PR #7, opened 2021-05-20
Ensure only users can change their own profile

FreeTrade:
It looks like there is a bug whereby a set of 6 private key holders can change the profile information of any user without the user's consent.

### DIFF
--- a/lib/block_view.go
+++ b/lib/block_view.go
@@ -4671,10 +4671,9 @@ func (bav *UtxoView) _connectUpdateProfile(
 		newProfileEntry = *existingProfileEntry
 
 		// Modifying a profile is only allowed if the transaction public key equals
-		// the profile public key or if the public key belongs to a paramUpdater.
+		// the profile public key
 		_, updaterIsParamUpdater := bav.Params.ParamUpdaterPublicKeys[MakePkMapKey(txn.PublicKey)]
-		if !reflect.DeepEqual(txn.PublicKey, existingProfileEntry.PublicKey) &&
-			!updaterIsParamUpdater {
+		if !reflect.DeepEqual(txn.PublicKey, existingProfileEntry.PublicKey) {
 
 			return 0, 0, nil, errors.Wrapf(
 				RuleErrorProfileModificationNotAuthorized,


jeansergegagnon:
Yeah, that looks a little fishy - how would it be possible for any other key that the original private key to be able to make that change (and I don't mean in the code - I mean from a blockchain perspective) - isn't the the whole premise of blockchain that only the private key holder can create new transactions for their pubic key?

I mean, if some "param updater" can change a profile - why wouldn't they be able to do anything, like withdraw?

Or am I missing something?

FreeTrade:
> Yeah, that looks a little fishy - how would it be possible for any other key that the original private key to be able to make that change (and I don't mean in the code - I mean from a blockchain perspective) - isn't the the whole premise of blockchain that only the private key holder can create new transactions for their pubic key?
> 
> I mean, if some "param updater" can change a profile - why wouldn't they be able to do anything, like withdraw?
> 
> Or am I missing something?

well this condition only exists in the update profile and swap identity actions - so i don't think there is any possibility of funds being affected, just identity

Mentors4EDU (approved):

Mentors4EDU:
ACK, I think the `ParamUpdater` parameter does more harm than good in this instance. Please review @maebeam

tijno:
I understand why this function is here and clearly its limited to just 5 admin keys i think.

ParamUpdates needed to standardise some aspects like bitclout to btc rate or usd rate - which is a good thing as we would not want nodes to adjust those.

Switch Keys and Update Profile also essential for onboarding vips, customer support, and long term growth.

Maybe implementation is not best one - but i can see the need.   

The thing that i think is important in these is what is the governance and control around this to ensure the key access is secure & staff operating those actions have some level of signoff before changes are made.

FreeTrade:
> The thing that i think is important in these is what is the governance and control around this to ensure the key access is secure & staff operating those actions have some level of signoff before changes are made.



> I understand why this function is here and clearly its limited to just 5 admin keys i think.
> 
> ParamUpdates needed to standardise some aspects like bitclout to btc rate or usd rate - which is a good thing as we would not want nodes to adjust those.
> 
> Switch Keys and Update Profile also essential for onboarding vips, customer support, and long term growth.
> 
> Maybe implementation is not best one - but i can see the need.
> 
> The thing that i think is important in these is what is the governance and control around this to ensure the key access is secure & staff operating those actions have some level of signoff before changes are made.

There are two other ways to go about this - only allow the private key holder to transfer an account, or only allow the 15,000 accounts to be transferrable by the superuser. It's just antithetical for a decentralized platform to have superusers that can reassign any account (and perhaps balance of the account too). It's one of the things that has got to be fixed before the platform can be called decentralized.

Mentors4EDU:
> It's just antithetical for a decentralized platform to have superusers that can reassign any account (and perhaps balance of the account too). It's one of the things that has got to be fixed before the platform can be called decentralized.


I fully agree, and it is definitely a huge red flag in regards to the premise of what this network should be.

tijno:
Ok so how would it work practically?

Lets take elon musk - he signs up to check it out with an account.

Happy, he want to claim his elonmusk account.

How can this be done for him if admins can only make changes to the original reserved elon musk key, and elon can only make changes to the new account he created ? 

Its all good flagging up problems but if we want this to become truly opensource proposing a solution and associated code is going to be helpful.

FreeTrade:
> Like how would a proper account swap feature work? One that maybe uses 2 transactions A->B and B accepts? It would be handy for other purposes i am sure.

You're right - a complete solution would require consent/signatures of both A and B.

For an emergency patch, and I think one is warranted for a problem as large as this, is that SWAP_IDENTITY should only be applicable to the 15,000 reserved accounts until a full solution is implemented.

tijno:
> For an emergency patch, and I think one is warranted for a problem as large as this, is that SWAP_IDENTITY should only be applicable to the 15,000 reserved accounts until a full solution is implemented.

This practically wont work though as it needs to swap with a different account that is not on the 15k list. So it needs to be able to take actions on both sides. Not just the 15k.

Also have there been instances where this was abused?

FreeTrade:
Yes, it would only be an improvement on the current situation rather than a full solution, hence 'emergency patch'.
I doubt it has been abused, but you simply can't have such an arbitrary power and trust that it won't be abused. That's not how people want to secure their identity and money.

tijno:
How would it work though if it was locked to the 15k ? 

Maybe you can show it in code ?

PR #13, opened 2021-05-27
Add support for block user transaction

lazynina:


FreeTrade:
Can you describe what this transaction does? Does it cover blocking/unblocking of users? Is it more like a twitter block (where blocked users cannot reply to your posts), or a mute(where muted users' content is simply not shown)?

lazynina:
> Can you describe what this transaction does? Does it cover blocking/unblocking of users? Is it more like a twitter block (where blocked users cannot reply to your posts), or a mute(where muted users' content is simply not shown)?

Currently, this will behave like a mute -- however the block transaction could later be expanded to act more like  a twitter block.

tijno:
im not too familiar with go so hesitant to submit pr for this

but looking at proposed changes, this suggests 1 transaction per blocked key?

would it be an idea to expand this to allow 1 transaction to block multiple keys in one go?

that way it would be possible to improve user experience if they wanted to block for example all pm spam bots, or diamonded bots or follow bots.

PR #15, opened 2021-05-29
Rph/like-reclout-diamond-indexes

redpartyhat:


lib/block_view.go
```diff
@@ -543,6 +546,13 @@ func (pe *PostEntry) IsDeleted() bool {
 	return pe.isDeleted
 }
 
+func IsQuotedReclout(postEntry *PostEntry) bool {
+	if postEntry.IsQuotedReclout && postEntry.RecloutedPostHash != nil {
```
lazynina, line 550: can't we just return `postEntry.IsQuotedReclout && postEntry.RecloutedPostHash != nil`?

lib/block_view.go
```diff
@@ -6575,6 +6614,164 @@ func (bav *UtxoView) GetPostsPaginatedForPublicKeyOrderedByTimestamp(publicKey [
...
+		copy(senderPKID[:], key[1+HashSizeBytes:])
+
+		diamondKey := &DiamondKey{
+			SenderPKID:      *senderPKID,
+			ReceiverPKID:    *receiverPKIDEntry.PKID,
+			DiamondPostHash: *postHash,
+		}
+
+		loadedEntry := bav.GetDiamondEntryForDiamondKey(diamondKey)
+		_ = loadedEntry
```
lazynina, line 6646: don't need to set the variable loadedEntry since we don't use it. Just call `bav.GetDiamondEntryForDiamondKey(diamondKey)`

lib/block_view.go
```diff
@@ -6575,6 +6614,164 @@ func (bav *UtxoView) GetPostsPaginatedForPublicKeyOrderedByTimestamp(publicKey [
...
+	for _, key := range keysFound {
+		// Sanity check that this is a reasonable key.
+		if len(key) != expectedKeyLength {
+			return nil, fmt.Errorf("UtxoView.GetDiamondsForPostHash: Invalid key length found: %d", len(key))
+		}
+
+		diamondPostHash := &BlockHash{}
+		copy(diamondPostHash[:], key[1:1+HashSizeBytes])
+		diamondPostEntry := bav.GetPostEntryForPostHash(diamondPostHash)
+		receiverPKIDEntry := bav.GetPKIDForPublicKey(diamondPostEntry.PosterPublicKey)
```
lazynina, line 6634: this should happen before the for loop is created.  This information is the same for all keys, so we can get a slight bump in performance by only computing these once. we know the post hash prior to the loop creation, so we don't need to get it from the key.

lib/block_view.go
```diff
@@ -6575,6 +6614,164 @@ func (bav *UtxoView) GetPostsPaginatedForPublicKeyOrderedByTimestamp(publicKey [
...
+	// Iterate over all the db keys & values and load them into the view.
+	expectedKeyLength := 1 + HashSizeBytes + btcec.PubKeyBytesLenCompressed
+	for _, key := range keysFound {
+		// Sanity check that this is a reasonable key.
+		if len(key) != expectedKeyLength {
+			return nil, fmt.Errorf("UtxoView.GetLikesForPostHash: Invalid key length found: %d", len(key))
+		}
+
+		likedPostHash := &BlockHash{}
+		copy(likedPostHash[:], key[1:1+HashSizeBytes])
```
lazynina, line 6675: we know the likedPostHash as it is the input to the function, so we can just copy this slice once before the foor loop

lib/block_view.go
```diff
@@ -6575,6 +6614,164 @@ func (bav *UtxoView) GetPostsPaginatedForPublicKeyOrderedByTimestamp(publicKey [
...
+	// Iterate over all the db keys & values and load them into the view.
+	expectedKeyLength := 1 + HashSizeBytes + btcec.PubKeyBytesLenCompressed
+	for _, key := range keysFound {
+		// Sanity check that this is a reasonable key.
+		if len(key) != expectedKeyLength {
+			return nil, fmt.Errorf("UtxoView.GetRecloutersForPostHash: Invalid key length found: %d", len(key))
+		}
+
+		recloutedPostHash := &BlockHash{}
+		copy(recloutedPostHash[:], key[1:1+HashSizeBytes])
```
lazynina, line 6713: we know the post hash before the for loop is created -- don't need to cut it out from the key.

lazynina (changes requested):

lib/block_view.go
lazynina: resolved with #16

lib/block_view.go
lazynina: resolved with #16

lib/block_view.go
lazynina: resolved with #16

lazynina (approved):
:+1:

PR #20, opened 2021-05-31
add statsd to server and expose it via getter method

lazynina:


maebeam (approved):

maebeam:
Maybe also refactor mempool to use this statsd instead of storing it?

lazynina:
> Maybe also refactor mempool to use this statsd instead of storing it?

done! Moved the mempool stats reporter to server.go

PR #25, opened 2021-06-06
Add HasMedia function and mediaRequired flag for only getting posts w…

bshin94:
…ith media

Related PRs:
https://github.com/bitclout/backend/pull/35
https://github.com/bitclout/frontend/pull/96

### DIFF
--- a/lib/block_view.go
+++ b/lib/block_view.go
@@ -2,6 +2,7 @@ package lib
 
 import (
 	"encoding/hex"
+	"encoding/json"
 	"fmt"
 	"math"
 	"math/big"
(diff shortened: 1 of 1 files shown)


lib/block_view.go
```diff
@@ -543,6 +544,16 @@ func (pe *PostEntry) IsDeleted() bool {
 	return pe.isDeleted
 }
 
+func (pe *PostEntry) HasMedia() bool {
+	bodyJSONObj := BitCloutBodySchema{}
+	err := json.Unmarshal(pe.Body, &bodyJSONObj)
+	//Only return true if json can be parsed and ImageUrls is not nil/non-empty
+	if err != nil || bodyJSONObj.ImageURLs == nil || len(bodyJSONObj.ImageURLs) == 0 {
```
maebeam, line 551: do you want posts with embedded youtube, vimeo, tiktok, etc links? if so, you'll need to check the ExtraData for `EmbedVideoURL`

bshin94: Ah yea we do, thanks for the call out. Will add.

PR #43, opened 2021-06-22
Fix @mention notifications

maebeam:


2000c5 (commented):
Want to make sure the fix is not to only catch commas and other standard "terminators" other than space?

All the cases I have seen were followed by a space.

maebeam:
I'm not sure I understand your comment. The terminators are relevant both before and after the @mention. Prior to this change a mention not followed or preceded by a space would fail.

2000c5:
> I'm not sure I understand your comment. The terminators are relevant both before and after the @mention. Prior to this change a mention not followed or preceded by a space would fail.

My apologies... Just wanted to make sure.

Thank you for your efforts!

lazynina (approved):
:rocket:

PR #44, opened 2021-06-22
Create new balance entry if previous balance is deleted when buying creator coin

lazynina:


lib/block_view.go
```diff
@@ -5315,7 +5315,10 @@ func (bav *UtxoView) HelpConnectCreatorCoinBuy(
 	buyerBalanceEntry, hodlerPKID, creatorPKID :=
 		bav._getBalanceEntryForHODLerPubKeyAndCreatorPubKey(
 			txn.PublicKey, existingProfileEntry.PublicKey)
-	if buyerBalanceEntry == nil {
+	// If the user does not have a balance entry or the user's balance entry is deleted and we have passed the
+	// BuyCreatorCoinAfterDeletedBalanceEntryFixBlockHeight, we create a new balance entry.
+	if buyerBalanceEntry == nil ||
+			(buyerBalanceEntry.isDeleted && blockHeight > BuyCreatorCoinAfterDeletedBalanceEntryFixBlockHeight){
```
maebeam, line 5321: This is an annoying nit but can we give this `(a && b)` condition its own line / variable name?

redpartyhat (approved):

lib/block_view.go
lazynina: I need to check for nil before checking isDeleted so I'm going to keep it as is.

maebeam: Ahhh you're right. Ignore me!

maebeam (approved):

PR #48, opened 2021-06-29
omit TxIndexMetadata in json if empty for TransactionMetadata

lazynina:
This will reduce the payload for `/api/v0/get-notifications` and `/api/v1/transaction-info` by excluding the keys for transaction metadata attributes that don't exist. E.g. we only expect to see basic transfer and follow metadata in a follow transaction, so we shouldn't include the keys for other transaction metadata types, such as `LikeTxindexMetadata`


maebeam (approved):
Awesome!

PR #50, opened 2021-06-30
Add consensus level decentralized post pinning mechanism to core. 

bluepartyhat:


lib/block_view.go
```diff
@@ -2759,6 +2801,42 @@ func (bav *UtxoView) _deleteRecloutEntryMappings(recloutEntry *RecloutEntry) {
 	bav._setRecloutEntryMappings(&tombstoneRecloutEntry)
 }
 
+func (bav *UtxoView) _addPostEntryPinning(postEntry *PostEntry) {
```
lazynina, line 2804: can we make this `_setPinEntryMappings(publicKey []byte, postHash *BlockHash, pinEntry *PinEntry) {`

lib/block_view.go
```diff
@@ -2759,6 +2801,42 @@ func (bav *UtxoView) _deleteRecloutEntryMappings(recloutEntry *RecloutEntry) {
...
+	pinEntry, prevEntryExists := bav.PublicKeyToPostHashToPinEntry[pkMapKey][*postEntry.PostHash]
+	if prevEntryExists {
+		pinEntry.pinned = true
+	} else {
+		pinEntry = &PinEntry{ tstampNanos: postEntry.TimestampNanos, pinned: true, }
+	}
+	bav.PublicKeyToPostHashToPinEntry[pkMapKey][*postEntry.PostHash] = pinEntry
+}
+
+func (bav *UtxoView) _deletePostEntryPinning(postEntry *PostEntry) {
```
lazynina, line 2822: `_deletePinEntryMappings(publicKey []byte, postHash *BlockHash, pinEntry *PinEntry)`

lib/block_view.go
```diff
@@ -4495,6 +4604,12 @@ func (bav *UtxoView) _connectSubmitPost(
 				bav._updateQuoteRecloutCount(newRecloutedPostEntry, 1)
 			}
 		}
+
+		// If the new post is pinned, we add the pin entry. We don't consider if the post is hidden
+		// as we do not allow new posts to be hidden.
+		if isPinned {
```
lazynina, line 4610: What happens if we are hiding and pinning a post at the same time?

lazynina (changes requested):
see feedback

lib/block_view.go
```diff
@@ -6568,7 +6683,78 @@ func (bav *UtxoView) GetAllPosts() (_corePosts []*PostEntry, _commentsByPostHash
...
 }
 
-func (bav *UtxoView) GetPostsPaginatedForPublicKeyOrderedByTimestamp(publicKey []byte, startPostHash *BlockHash, limit uint64, mediaRequired bool) (_posts []*PostEntry, _err error) {
+func (bav *UtxoView) GetPinnedPostsForPublicKeyOrderedByTimestamp(publicKey []byte, limit uint64,
+	mediaRequired bool) (_posts []*PostEntry, _postsIncluded map[BlockHash]struct{}, _err error) {
+	dbPrefix := _dbReverseSeekPrefixForPostsPinned(publicKey)
+
+	// Get the posts associated from the database. We get double the limit incase a post was recently hidden
+	// and had not been reflected on the keys in the database yet.
+	keysFound, _ := _enumerateLimitedKeysReversedForPrefix(bav.Handle, dbPrefix, 2 * limit)
```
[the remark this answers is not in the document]
lazynina: see this for dynamically getting enough keys:
https://github.com/bitclout/backend/blob/main/routes/user.go#L1998

lib/block_view.go
```diff
@@ -6568,7 +6683,78 @@ func (bav *UtxoView) GetAllPosts() (_corePosts []*PostEntry, _commentsByPostHash
 	return allCorePosts, commentsByPostHash, nil
 }
 
-func (bav *UtxoView) GetPostsPaginatedForPublicKeyOrderedByTimestamp(publicKey []byte, startPostHash *BlockHash, limit uint64, mediaRequired bool) (_posts []*PostEntry, _err error) {
+func (bav *UtxoView) GetPinnedPostsForPublicKeyOrderedByTimestamp(publicKey []byte, limit uint64,
```
lazynina, line 6686: this function may be better suited to live in post.go in backend

lib/block_view.go
```diff
@@ -4297,6 +4381,7 @@ func (bav *UtxoView) _connectSubmitPost(
 		// which seems like undesired behavior if a paramUpdater is trying to reduce
 		// spam
 		newPostEntry.IsHidden = txMeta.IsHidden
+		newPostEntry.IsPinned = isPinned
```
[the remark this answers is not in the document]
bluepartyhat: I think it should be left. Depending on the frontend, the previous state could be conveyed with this such that a pinned post being unhidden brings it back to your pinned posts collection. It feels unnecessary to add more nuanced consensus rules around this, but I get the argument against this.

KrassensteinBitClout:
Is this a global feature or will each node be able to have their own pinned posts?  I see this is written to Core so I assume it's global.

lazynina:
> Is this a global feature or will each node be able to have their own pinned posts?  I see this is written to Core so I assume it's global.

This feature allows users to pin posts to their own profile and storing the the fact that the post is pinned will live on the blockchain. So this is a "global" or what I generally think of as a "protocol-level" feature.

This is distinct from posts that are pinned to the global feed. Those will only ever be at the node-level. Each node operator will have the ability to pin posts to their global feed.

Hope that answers your question.

lib/block_view.go
```diff
@@ -543,9 +558,12 @@ type PostEntry struct {
 	// How many comments this post has
 	CommentCount uint64
 
-	// Indicator if a post is pinned or not.
+	// Indicator if a post is pinned or not to the user's profile.
 	IsPinned bool
 
+	// Indicator if a post is pinned to the global feed.
+	IsGlobalPinned bool
```
diamondhands0, line 565: Hmm.. I don't think we want anyone to have the ability to pin to the global feed do we?

lib/block_view.go
```diff
@@ -4199,6 +4269,12 @@ func (bav *UtxoView) _connectSubmitPost(
 		copy(recloutedPostHash[:], recloutedPostHashBytes)
 		delete(extraData, RecloutedPostHash)
 	}
+	// Set the IsPinned attribute based on extra data and check that we've passed the PinnedPostTrackingUpdateBlockHeight.
+	isPinned := false
+	if _, hasPinned := extraData[IsPinnedPostKey]; hasPinned && blockHeight > PinnedPostTrackingUpdateBlockHeight {
+		isPinned = true
+		delete(extraData, IsPinnedPostKey)
```
diamondhands0, line 4276: Why do we modify the extraData here?

lib/block_view.go
```diff
@@ -4341,6 +4435,17 @@ func (bav *UtxoView) _connectSubmitPost(
 			}
 		}
 
+		// We set/delete the pin entry mappings depending on the previous and current state of the post.
+		if hidingPostEntry {
```
diamondhands0, line 4439: Oof... complicated...

lib/block_view.go
```diff
@@ -276,6 +276,21 @@ type RecloutEntry struct {
 	isDeleted bool
 }
 
+// PinEntry stores the timestamp and state of a PostEntry.
+// It's used in the UtxoView to keep track of pinned posts not yet in the database.
+type PinEntry struct {
```
diamondhands0, line 281: High-level thoughts after reading through.

It would be good to understand what the value of a PinEntry is. My view on pinned posts is we'd implement them the exact same way as we implement IsHidden on posts, which is brain-dead simple and doesn't even require any change to the disconnect logic.

Then, to get an index of the pinned posts, we can modify PutPostEntryMappings and DeletePostEntryMappings in db_utils.go to simply update a separate little index of <PublicKey, Timestamp> for the pinned posts. We could even include *another* field in addition to IsPinned on the PostEntry, which is PinTimestamp, which can allow the user to order the pinned posts on their profile by pinning them in a different order. This would also require no changes to the disconnect logic I think.

I may be missing the reason why PinEntry is required, and if that's the case please let me know!

lib/block_view.go
lazynina: This is just the admin ability to pin to global feed that already exists. It was renamed in this PR.

lib/block_view.go
lazynina: Similar to why we delete the reclout keys from the extra data map - it saves space in the DB for each PostEntry as we don't need the keys in the PostEntry's PostExtraData map since we explicitly extract it to the IsPinned attribute.

lib/block_view.go
lazynina: I think the reason @bluepartyhat added the PinEntry due to the complications that would arise in `DBPutPostEntryMappingsWithTxn`. Unlike `isHidden`, `isPinned` results in different indexing. So when we flush to the DB, we need to know the previous pin status of the PostEntry (which is not retained on the PostEntry) so we can delete it in `DBDeletePostEntryMappingsWithTxn` and can then be correctly re-added in `DBPutPostEntryMappingsWithTxn`.  I think adding the complication there would probably be preferable to adding the PinEntry, but let's chat about this since the indexing is odd.

PR #63, opened 2021-07-10
Add postgres

maebeam:


cmd/run.go
```diff
@@ -147,7 +148,7 @@ func SetupRunFlags(cmd *cobra.Command) {
 		"When set, all blocks produced by the block producer will be signed by this " +
 			"seed.")
 	cmd.PersistentFlags().StringSlice("trusted-block-producer-public-keys", []string{
-		"BC1YLhULV1UBa7gYxEttkLdtzzkvXBwmgmyPKgJaGs35g2tiPafA2hv",
+		"BC1YLgS1zDJQqywFpsty4fFheUrZxVQNKEsrttppvUESFZCq6Nfoypm",
```
redpartyhat, line 151: Should this be changing?

lib/block_view.go
```diff
@@ -7558,43 +7573,38 @@ func (bav *UtxoView) _flushBalanceEntriesToDbWithTxn(txn *badger.Txn) error {
 }
 
 func (bav *UtxoView) FlushToDbWithTxn(txn *badger.Txn) error {
-	// Flush the utxos to the db.
-	if err := bav._flushUtxosToDbWithTxn(txn); err != nil {
-		return err
+	// Only flush to BadgerDB if Postgres is disabled
+	if bav.Postgres == nil {
```
diamondhands0, line 7577: Hmm... so we only flush to db if Postgres is set?

lib/block_view.go
```diff
@@ -1068,22 +1070,32 @@ func (bav *UtxoView) CopyUtxoView() (*UtxoView, error) {
...
+	// _ResetViewMappingsAfterFlush because it is not something that is affected by a
+	// flush operation. Moreover, its value is consistent with the view regardless of
+	// whether or not the view is flushed or not. Additionally the utxo view does
+	// not concern itself with the header chain (see comment on GetBestHash for more
+	// info on that).
+	if view.Postgres != nil {
+		view.TipHash = view.Postgres.GetChain("main").TipHash
+	} else {
+		view.TipHash = DbGetBestHash(view.Handle, ChainTypeBitCloutBlock /* don't get the header chain */)
+	}
```
redpartyhat, line 1097: Shouldn't the node depend on the badger tip hash more than the postgres tiphash?  I don't think we want to depend on things in the postgres instance, even if it is available.

lib/block_view.go
diamondhands0: I think it's possibly better to have Postgres *in addition* to Badger initially, rather than trying to replace it. It looks like doing it this way makes you forced into using Postgres, whereas it may be better to leave it as optional until *everything* can be done in Postgres.

lib/postgres.go
```diff
@@ -0,0 +1,741 @@
...
+		}
+
+		if txn.Signature != nil {
+			transaction.R = BigintToHash(txn.Signature.R)
+			transaction.S = BigintToHash(txn.Signature.S)
+		}
+
+		transactions = append(transactions, transaction)
+
+		for i, input := range txn.TxInputs {
```
diamondhands0, line 357: ii

lib/block_view.go
donhardman: Absolutely agree that Postgres is a totally bad idea to force to use it. Especially in the way we talk about the core project. I thought that by architecture backend project should operate data like indexes and API queries. But core stays in independent and lightweight mode and is easy to run by everyone who wants to get raw nodes and support network consensus (I hope we will see it soon, kind of POS or whatever).

Adding such a monster as Postgres in core I guess is a bad idea at all. It should be in the backend project. But it's just my opinion. Cheers.

lib/postgres.go
redpartyhat: To clarify, Mr. Hands does not like using single letter variables in code because they are hard to find and replace.  We should replace all instances of "for i".

cmd/run.go
```diff
@@ -59,6 +59,9 @@ func SetupRunFlags(cmd *cobra.Command) {
 	cmd.PersistentFlags().Bool("regtest", false,
 		"Can only be used in conjunction with --testnet. Creates a private testnet node with fast block times"+
 			"and instantly spendable block rewards.")
+	cmd.PersistentFlags().String("postgres-uri", "", "URI for syncing the node with postgres")
+	cmd.PersistentFlags().String("postgres-data-dir", "", "Start a local embedded postgres instance"+
+		"and persist its data in this directory")
```
redpartyhat, line 64: I would clarify that it is starting a local embedded postgres instance *instead* of a remote postgres instance and that it will be ignored if a URI is provided.  I think these info strings make sense but making it a touch more explicit would be nice.

Side note: it would be nice to clarify our node's current working relationship with postgres somewhere.  That is, if you spin up a postgres node, it will use postgres for most things and badger DB for some things that have not yet been migrated (with examples).  I'm not sure the best place to make that clear though.  Maybe we could put something here and in the docs?

lib/block_view.go
```diff
@@ -371,104 +371,6 @@ func (bav *UtxoView) GetRecloutPostEntryStateForReader(readerPK []byte, postHash
...
-func (ss StakeIDType) String() string {
-	if ss == StakeIDTypePost {
-		return "post"
-	} else if ss == StakeIDTypeProfile {
-		return "profile"
-	} else {
-		return "unknown"
-	}
-}
-
```
redpartyhat, line 471: Love it.  Thanks for the clean up!

lib/block_view.go
```diff
@@ -742,6 +609,9 @@ type UtxoView struct {
 	// Messages data
 	MessageKeyToMessageEntry map[MessageKey]*MessageEntry
 
+	// Postgres stores message data slightly differently
+	MessageMap map[BlockHash]*Message
```
redpartyhat, line 613: Not using overloaded terms like "message" as variable names is preferred (same reasoning as not using single letter variable names).  I'd suggest we call this new type of message entry a, "PGMessageEntry" instead. And "PGMessageEntryMap".  If we ever rip out badger, we can then easily find and replace and kill all the PG tags.

redpartyhat: I also think this naming convention will help to clarify the distinction between the two message types for new developers working on the code base.

lib/block_view.go
```diff
@@ -6367,6 +6571,65 @@ func (bav *UtxoView) ConnectBlock(
 	return utxoOps, nil
 }
 
+func (bav *UtxoView) Preload(bitcloutBlock *MsgBitCloutBlock) error {
```
redpartyhat, line 6574: This seems cool-- just to clarify, the idea here is that rather than doing a call to postgres every time we need something in the block that isn't in the view, we try and do a few batch queries up front, correct?

lib/block_view_test.go
```diff
@@ -7659,37 +7653,37 @@ func TestSpendOffOfUnminedTxnsBitcoinExchange(t *testing.T) {
...
+	//		mempoolTxsAdded, err := mempool.processTransaction(
+	//			burnTxn, true /*allowUnconnectedTxn*/, true /*rateLimit*/, 0, /*peerID*/
+	//			true /*verifySignatures*/)
+	//		require.NoErrorf(err, "on index: %v", reverseIndex)
+	//		require.Equal(1, len(mempoolTxsAdded))
+	//		require.NotEqual(
+	//			BlockHash{},
+	//			mempoolTxsAdded[0].Tx.TxnMeta.(*BitcoinExchangeMetadata).BitcoinMerkleRoot)
+	//	}
+	//}
```
redpartyhat, line 7686: We probably shouldn't merge this with these tests commented out.  Are these bitcoin exchange tests incompatible with the postgres integration for some reason?  Naively, it seems to me that postgres should not effect bitcoin exchange txns.

lib/server.go
```diff
@@ -329,12 +332,12 @@ func NewServer(
...
-		_bitcoinConnectPeer)
-	if err != nil {
-		return nil, errors.Wrapf(err, "NewServer: Problem initializing BitcoinManager")
-	}
+	//_bitcoinManager, err := NewBitcoinManager(
+	//	_db, _params, timesource, _bitcoinDataDir, _incomingMessages,
+	//	_bitcoinConnectPeer)
+	//if err != nil {
+	//	return nil, errors.Wrapf(err, "NewServer: Problem initializing BitcoinManager")
+	//}
```
redpartyhat, line 340: As per my earlier comment, not sure why we need to kill the bitcoin manager in order to make postgres work.

lib/utils.go
```diff
@@ -8,7 +8,7 @@ import (
 	"github.com/unrolled/secure"
 )
 
-const SECURE_MIDDLEWARE_RESTRICTIVE_CONTENT_SECURITY_POLICY = "default-src 'self'"
+const SECURE_MIDDLEWARE_RESTRICTIVE_CONTENT_SECURITY_POLICY = ""
```
redpartyhat, line 11: I'm not a CSP expert, could you elaborate on why do we need to kill this for Postgres?

migrate/20210623152412_create_tables.go
```diff
@@ -0,0 +1,366 @@
...
+				post_hash        BYTEA,
+				timestamp        BIGINT NOT NULL
+			);
+		`)
+		if err != nil {
+			return err
+		}
+
+		_, err = db.Exec(`
+			CREATE TABLE profiles (
```
redpartyhat, line 223: It would be great if we could turn off pluralization so that it is easier to find the struct corresponding to a specific postgres table.  For example, this table would be "profile" since that maps directly to the "profile" struct.

lib/postgres.go
```diff
@@ -0,0 +1,1368 @@
...
+}
+
+func (follow *Follow) NewFollowEntry() *FollowEntry {
+	return &FollowEntry{
+		FollowerPKID: follow.FollowerPKID,
+		FollowedPKID: follow.FollowedPKID,
+	}
+}
+
+type Diamond struct {
```
redpartyhat, line 311: You're probably going to hate this but it'd be ideal if we could change all these structs to be "PG[name]Entry".  So in this case, it would be "PGDiamondEntry".  The reason being, if you wanted to search for this specific struct in the codebase, searching for the word "diamond" would return a ton of erroneous matches.  Please discuss with @diamondhands0, if you think this is unnecessary... Sorry for the annoying nit!

lib/db_utils.go
```diff
@@ -3313,30 +3294,6 @@ func DBGetPostEntryByPostHash(db *badger.DB, postHash *BlockHash) *PostEntry {
...
-
-	return stakeID
-}
-
-func StakeIDToHash(stakeID []byte) *BlockHash {
-	hash := &BlockHash{}
-	copy(hash[:], stakeID[:HashSizeBytes])
-	return hash
-}
-
```
redpartyhat, line 3339: Woo! Cleanup!!

lib/block_view.go
maebeam: Correct!

lib/utils.go
maebeam: I was having local dev issues for some unknown reason. I'll put this back.

migrate/20210623152412_create_tables.go
maebeam: I resolved this by manually defining all of the table names in the corresponding structs

cmd/config.go
```diff
@@ -140,6 +142,10 @@ func (config *Config) Print() {
 		glog.Infof("Mempool Dump Directory: %s", config.MempoolDumpDirectory)
 	}
 
+	if config.PostgresURI != "" {
```
diamondhands0, line 145: Please make sure n0 and n0_test are working after this change with proper default flags and whatnot.

cmd/node.go
```diff
@@ -118,13 +122,36 @@ func (node *Node) Start() {
...
+		}
+
+		db = pg.Connect(options)
+	}
+
+	if db != nil {
+		node.Postgres = lib.NewPostgres(db)
+
+		// Make sure we're migrated
+		migrate.LoadMigrations()
```
diamondhands0, line 140: Add comments explaining, in detail, how the migrations work here. For example, what is LoadMigrations? Why do we need to call that before calling Run()? Where are the migrations stored and how are they being loaded? Make sure you give the devs all the info they need to trace everything here.

cmd/run.go
```diff
@@ -59,6 +59,9 @@ func SetupRunFlags(cmd *cobra.Command) {
 	cmd.PersistentFlags().Bool("regtest", false,
 		"Can only be used in conjunction with --testnet. Creates a private testnet node with fast block times"+
 			"and instantly spendable block rewards.")
+	cmd.PersistentFlags().String("postgres-uri", "", "BETA: Use Postgres as the backing store for chain data."+
+		"When enabled, most data is stored in postgres although badger is still currently used for some state. Run your"+
```
diamondhands0, line 63: Typo: Run your<space>"

lib/db_utils.go
```diff
@@ -3514,30 +3528,6 @@ func DBGetPostEntryByPostHash(db *badger.DB, postHash *BlockHash) *PostEntry {
 	return ret
 }
 
-func _dbGetStakeIDPostDBKey(postHash *BlockHash, totalAmountStakedNanos uint64) []byte {
```
diamondhands0, line 3517: Glad  this is killed.

lib/postgres.go
```diff
@@ -0,0 +1,1414 @@
...
+	return &Postgres{
+		db: db,
+	}
+}
+
+func LogSelect(query *orm.Query) error {
+	selectQuery := orm.NewSelectQuery(query)
+	fmter := orm.NewFormatter().WithModel(selectQuery)
+	queryStr, _ := selectQuery.AppendQuery(fmter, nil)
+	glog.Info(string(queryStr))
```
diamondhands0, line 31: Is it possible we'd ever log something sensitive here? Pretty sure the answer is no, but just flagging. DM me to discuss if so since the logging is definitely useful.

lib/postgres.go
```diff
@@ -0,0 +1,1414 @@
...
+// Tables
+//
+// When we can, we use unique fields (or combinations of unique fields) as the primary keys on the models.
+// This lets us use the WherePK() query while also minimizing columns and indicies on disk.
+//
+
+type PGChain struct {
+	tableName struct{} `pg:"pg_chains"`
+
+	Name    string `pg:",pk"`
```
diamondhands0, line 45: Just being lazy: Why do the columns have commas?

diamondhands0: I don't really get the annotation scheme here. Maybe add a comment under the Tables explaining it and give examples.

lib/postgres.go
```diff
@@ -0,0 +1,1414 @@
...
+	TipHash *BlockHash
```
diamondhands0, line 46: Why are some fields not annotated? Explain in your comment at the top.

diamondhands0: Ok I think I get it. You only annotate when you want to add things to the column definition like pk or use_zero. Still good to maybe explain it at the top. It may be redundant, but if it helps the lazy then it's worth it.

diamondhands0: So I assume the column names are interpolated into Postgres via magic...? I don't love it, but it's fine for now...

diamondhands0: Hate magic...

diamondhands0: Also in your comment, explain in full how something in the migration maps to something in this file, and how it all fits together. Make it as easy as possible for a dev to understand how to fully trace everything.

lib/postgres.go
```diff
@@ -0,0 +1,1414 @@
...
+	LikeCount         uint64
+	RecloutCount      uint64
+	QuoteRecloutCount uint64
+	DiamondCount      uint64
+	CommentCount      uint64
+	Pinned            bool
+	ExtraData         map[string][]byte
+}
+
+func (post *PGPost) NewPostEntry() *PostEntry {
```
diamondhands0, line 291: Why isn't this defined as static like:
```
func NewPostEntry() *PostEntry {
```

lib/postgres.go
```diff
@@ -0,0 +1,1414 @@
...
+
+	if post.ParentPostHash != nil {
+		postEntry.ParentStakeID = post.ParentPostHash.ToBytes()
+	}
+
+	return postEntry
+}
+
+func (post *PGPost) HasMedia() bool {
+	bodyJSONObj := BitCloutBodySchema{}
```
diamondhands0, line 317: Seems inefficient. Maybe leave a comment documenting that it's inefficient.

lib/postgres.go
```diff
@@ -0,0 +1,1414 @@
...
+		if parentHash != nil {
+			blockNode.Parent = blockMap[*parentHash]
+		}
+	}
+
+	return blockMap, nil
+}
+
+func (postgres *Postgres) GetChain(name string) *PGChain {
+	chain := &PGChain{
```
diamondhands0, line 469: Comment this function please.

lib/postgres.go
```diff
@@ -0,0 +1,1414 @@
...
+	if blockNode.Parent != nil {
+		block.ParentHash = blockNode.Parent.Hash
+	}
+
+	_, err := tx.Model(block).WherePK().OnConflict("(hash) DO UPDATE").Insert()
+	return err
+}
+
+func (postgres *Postgres) GetBlockIndex() (map[BlockHash]*BlockNode, error) {
+	var blocks []PGBlock
```
diamondhands0, line 430: Comment all of these functions.

lib/postgres.go
```diff
@@ -0,0 +1,1414 @@
...
+	bestChain := &PGChain{
+		TipHash: tipHash,
+		Name:    name,
+	}
+
+	_, err := tx.Model(bestChain).WherePK().OnConflict("(name) DO UPDATE").Insert()
+	return err
+}
+
+func (postgres *Postgres) InsertTransactionsTx(tx *pg.Tx, bitCloutTxns []*MsgBitCloutTxn, blockHash *BlockHash) error {
```
diamondhands0, line 497: Comment all of these functions.

lib/postgres.go
```diff
@@ -0,0 +1,1414 @@
...
+
+func (postgres *Postgres) UpsertBlockAndTransactions(blockNode *BlockNode, bitcloutBlock *MsgBitCloutBlock) error {
+	return postgres.db.RunInTransaction(postgres.db.Context(), func(tx *pg.Tx) error {
+		err := postgres.UpsertBlockTx(tx, blockNode)
+		if err != nil {
+			return err
+		}
+
+		blockHash := blockNode.Hash
+		err = postgres.UpsertChainTx(tx, "main", blockHash)
```
diamondhands0, line 746: Don't use a raw string here?

lib/postgres.go
```diff
@@ -0,0 +1,1414 @@
...
+
+		return nil
+	})
+}
+
+func (postgres *Postgres) GetUtxoEntryForUtxoKey(utxoKey *UtxoKey) *UtxoEntry {
+	utxo := &PGTransactionOutput{
+		OutputHash:  &utxoKey.TxID,
+		OutputIndex: utxoKey.Index,
+		Spent:       false,
```
diamondhands0, line 764: Curious: Are we keeping spent UTXOs around? Makes sense if we are, but this would be different than what we were doing before.

lib/postgres.go
```diff
@@ -0,0 +1,1414 @@
...
+		// TODO: Block height?
+		UtxoType: utxo.OutputType,
+		isSpent:  utxo.Spent,
+		UtxoKey:  utxoKey,
+	}
+}
+
+func (postgres *Postgres) GetUtxoEntriesForPublicKey(publicKey []byte) []*UtxoEntry {
+	var transactionOutputs []*PGTransactionOutput
+	err := postgres.db.Model(&transactionOutputs).Where("public_key = ?", publicKey).Select()
```
diamondhands0, line 784: I would maybe use fmt.Sprintf("%v = ?", PublicKeyString) in cases like these to avoid getting fucked when you change something. But it does make it pretty painful to read...

lib/postgres.go
```diff
@@ -0,0 +1,1414 @@
...
+			ProfilePic:              profileEntry.ProfilePic,
+			CreatorBasisPoints:      profileEntry.CreatorBasisPoints,
+			BitCloutLockedNanos:     profileEntry.BitCloutLockedNanos,
+			NumberOfHolders:         profileEntry.NumberOfHolders,
+			CoinsInCirculationNanos: profileEntry.CoinsInCirculationNanos,
+			CoinWatermarkNanos:      profileEntry.CoinWatermarkNanos,
+		}
+
+		if profileEntry.isDeleted {
+			deleteProfiles = append(deleteProfiles, profile)
```
diamondhands0, line 892: I don't think this code would ever hit because you're continueing above

lib/postgres.go
```diff
@@ -0,0 +1,1414 @@
...
+		_, err := tx.Model(&deleteProfiles).Returning("NULL").Delete()
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (postgres *Postgres) flushPosts(tx *pg.Tx, view *UtxoView) error {
```
diamondhands0, line 915: Pausing here.

lib/block_view.go
```diff
@@ -114,6 +114,7 @@ func MakeUsernameMapKey(nonLowercaseUsername []byte) UsernameMapKey {
 	return usernameMapKey
 }
 
+// DEPRECATED: Replace all instances with lib.PublicKey
```
diamondhands0, line 117: Nice change I think.

lib/postgres.go
```diff
@@ -0,0 +1,1414 @@
...
+
+//
+// Tables
+//
+// When we can, we use unique fields (or combinations of unique fields) as the primary keys on the models.
+// This lets us use the WherePK() query while also minimizing columns and indicies on disk.
+//
+
+type PGChain struct {
+	tableName struct{} `pg:"pg_chains"`
```
diamondhands0, line 43: I would unpluralize the name unless it's very difficult to tell the library to do it. It's easier to go back and forth from the migration file if pg_chain maps 1:1 to PGChain in the code. Pluralization is magic and magic is bad.

lib/postgres.go
diamondhands0: Sigh.. on second thought it is a lot harder to read... I think instead maybe we'll just never ever change column names...

lib/postgres.go
diamondhands0: Back!

lib/postgres.go
```diff
@@ -0,0 +1,1414 @@
...
+	}
+
+	return nil
+}
+
+func (postgres *Postgres) flushPosts(tx *pg.Tx, view *UtxoView) error {
+	var insertPosts []*PGPost
+	var deletePosts []*PGPost
+	for _, postEntry := range view.PostHashToPostEntry {
+		post := &PGPost{
```
diamondhands0, line 919: Question: Is there anything stopping us from replacing PostEntry with PGPost, and similarly for the other structs? I think some fields are irrelevant, like isDeleted. But I think we can define those irrelevant fields as lowercase camelcase in the PGPost. 

I think we should merge as-is, but just a thought on something we could do later to minimize duplication.

lib/postgres.go
```diff
@@ -0,0 +1,1414 @@
...
+		return nil
+	}
+	return posts
+}
+
+//
+// Comments
+//
+
+// TODO: Pagination
```
diamondhands0, line 1218: Sigh.. pagination seems hard... We should discuss how this is going to work.

migrate.go
```diff
@@ -0,0 +1,28 @@
+package main
```
diamondhands0, line 1: This needs detailed explanation. What is this main() doing? Why does it exist? How does one use it?

lib/server.go
```diff
@@ -498,7 +505,7 @@ func (srv *Server) GetBlocks(pp *Peer, maxHeight int) {
 	// Note this check is not strictly necessary since messageHandler will ultimately
 	// not allow blocks to be processed if the BitcoinManager is not synced, but checking
 	// this here allows for the optimization of not requesting them in the first place.
-	if !srv.bitcoinManager.IsCurrent(false /*considerCumWork*/) {
+	if srv.bitcoinManager.IsCurrent(false /*considerCumWork*/) {
```
diamondhands0, line 508: Was removing the ! intentional here?

migrate/20210623152412_create_tables.go
```diff
@@ -0,0 +1,366 @@
+package migrate
+
+import (
+	"github.com/go-pg/pg/v10/orm"
+	migrations "github.com/robinjoseph08/go-pg-migrations/v3"
+)
+
+func init() {
```
diamondhands0, line 8: Comments that explain in plain english how all this works.

migrate/20210623152412_create_tables.go
```diff
@@ -0,0 +1,366 @@
+package migrate
```
diamondhands0, line 1: Do we need so many numbers in the migration filename schema? How the following to make it more readable:
```
YYYY-MM-DD_HH-MM-SS_<migration name>
```

migrate/20210623152412_create_tables.go
```diff
@@ -0,0 +1,366 @@
...
+		`)
+		if err != nil {
+			return err
+		}
+
+		_, err = db.Exec(`
+			CREATE TABLE pg_blocks (
+				hash              BYTEA PRIMARY KEY,
+				parent_hash       BYTEA,
+				height            BIGINT NOT NULL,
```
diamondhands0, line 24: Hmm... so it looks like a BIGINT is equivalent to an int64. Which means that we're potentially storing uint64 values in something that's one bit less. This should be fine assuming that we never store math.MaxUint64, but it would break if we did try to store that value. Maybe add a check for it before storing somehow?

migrate/20210623152412_create_tables.go
```diff
@@ -0,0 +1,366 @@
...
+				difficulty_target BYTEA  NOT NULL,
+				cum_work          BYTEA  NOT NULL,
+				status            TEXT   NOT NULL,
+				tx_merkle_root    BYTEA  NOT NULL,
+				timestamp         BIGINT NOT NULL,
+				nonce             BIGINT NOT NULL,
```
diamondhands0, line 30: Hmmm... nonce and  extranonce can definitely exceed the width of an int64. But maybe it's OK because they're casted when fetched? A bit surprising that this works.

migrate/20210623152412_create_tables.go
```diff
@@ -0,0 +1,366 @@
...
+		if err != nil {
+			return err
+		}
+
+		_, err = db.Exec(`
+			CREATE TABLE pg_transaction_outputs (
+				output_hash  BYTEA    NOT NULL,
+				output_index INT      NOT NULL,
+				output_type  SMALLINT NOT NULL,
+				public_key   BYTEA    NOT NULL,
```
diamondhands0, line 60: Have you considered making public_key a string to make querying easier?

diamondhands0: Or is there some way to query it that's easy when you store it in bytes?

migrate/20210623152412_create_tables.go
```diff
@@ -0,0 +1,366 @@
...
+				creator_coin_to_transfer_nanos BIGINT NOT NULL,
+				receiver_public_key            BYTEA NOT NULL
+			);
+		`)
+		if err != nil {
+			return err
+		}
+
+		_, err = db.Exec(`
+			CREATE TABLE pg_metadata_swap_identities (
```
diamondhands0, line 195: How does a metadata link back to its txn if it's in another table? Is there a foreign key constraint or something?

diamondhands0: Nevermind dumb question.

lib/block_view.go
```diff
@@ -3807,17 +3754,30 @@ func (bav *UtxoView) GetDiamondEntryForDiamondKey(diamondKey *DiamondKey) *Diamo
 	// If we get here it means no value exists in our in-memory map. In this case,
 	// defer to the db. If a mapping exists in the db, return it. If not, return
 	// nil.
-	dbDiamondEntry := DbGetDiamondMappings(
-		bav.Handle, &diamondKey.ReceiverPKID, &diamondKey.SenderPKID, &diamondKey.DiamondPostHash)
-	if dbDiamondEntry != nil {
-		bav._setDiamondEntryMappings(dbDiamondEntry)
+	var diamondEntry *DiamondEntry
+	if bav.Postgres != nil {
+		diamond := bav.Postgres.GetDiamond(&diamondKey.SenderPKID, &diamondKey.ReceiverPKID, &diamondKey.DiamondPostHash)
```
diamondhands0, line 3759: Nit: It feels like it would be kinda nice if the raw Postgres querying, and the raw manipulation of PGDiamond and other PG* structs was encapsulated into postgres.go rather than bubbled up here. That would give a clean interface that would allow us to potentially swap out the storage layer in the future. That said, if there are many instances of this to the point where wrapping these things in their own functions is annoying then I would skip it.

diamondhands0: On second thought just don't worry about this. It's too annoying and not needed.

lib/block_view.go
```diff
@@ -3826,11 +3786,19 @@ func (bav *UtxoView) GetPostEntryForPostHash(postHash *BlockHash) *PostEntry {
 	// If we get here it means no value exists in our in-memory map. In this case,
 	// defer to the db. If a mapping exists in the db, return it. If not, return
 	// nil.
-	dbPostEntry := DBGetPostEntryByPostHash(bav.Handle, postHash)
-	if dbPostEntry != nil {
-		bav._setPostEntryMappings(dbPostEntry)
+	if bav.Postgres != nil {
+		post := bav.Postgres.GetPost(postHash)
+		if post != nil {
+			return bav.setPostMappings(post)
```
diamondhands0, line 3792: Nit: We should be consistent about what layer of the stack PG* structs are allowed. I think you shouldn't write set*Mappings functions to take PG* structs. Instead I would cast it first to keep taht interface clean and consistent.

lib/block_view.go
```diff
@@ -4031,11 +4110,22 @@ func (bav *UtxoView) GetProfileEntryForUsername(nonLowercaseUsername []byte) *Pr
 	// defer to the db. If a mapping exists in the db, return it. If not, return
 	// nil.
 	// Note that the DB username lookup is case-insensitive.
-	dbProfileEntry := DBGetProfileEntryForUsername(bav.Handle, nonLowercaseUsername)
-	if dbProfileEntry != nil {
-		bav._setProfileEntryMappings(dbProfileEntry)
+	if bav.Postgres != nil {
+		profile := bav.Postgres.GetProfileForUsername(string(nonLowercaseUsername))
```
diamondhands0, line 4114: Make sure that GetProfileForUsername does the same lowercasing as the original. It probably does, but just double check.

lib/block_view.go
```diff
@@ -4165,6 +4290,73 @@ func (bav *UtxoView) _deleteProfileEntryMappings(profileEntry *ProfileEntry) {
 	bav._setProfileEntryMappings(&tombstoneProfileEntry)
 }
 
+// Takes a Postgres Profile, sets all the mappings on the view, returns the equivalent ProfileEntry and PKIDEntry
+func (bav *UtxoView) setProfileMappings(profile *PGProfile) (*ProfileEntry, *PKIDEntry) {
```
diamondhands0, line 4294: Again, I think the casting from PGEntry to ProfileEntry shoudl happen before we call set*Mappings in order to maintain a clean interface and keep Postgres as low in teh stack as possible.

lib/block_view.go
```diff
@@ -6624,6 +6818,22 @@ func (bav *UtxoView) _connectSwapIdentity(
 	bav._setPKIDMappings(&newFromPKIDEntry)
 	bav._setPKIDMappings(&newToPKIDEntry)
 
+	// Postgres doesn't have a concept of PKID Mappings. Instead, we need to save an empty
+	// profile with the correct PKID and public key
+	if bav.Postgres != nil {
```
diamondhands0, line 6823: Hmm... why change the semantics here vs having a PKID<->public_key table?

lib/block_view.go
```diff
@@ -8048,6 +8257,65 @@ func (bav *UtxoView) ConnectBlock(
...
+func (bav *UtxoView) Preload(bitcloutBlock *MsgBitCloutBlock) error {
+	// We can only preload if we're using postgres
+	if bav.Postgres == nil {
+		return nil
+	}
+
+	// One iteration for all the PKIDs
+	var publicKeys []*PublicKey
+	for _, txn := range bitcloutBlock.Txns {
+		if txn.TxnMeta.GetTxnType() == TxnTypeFollow {
```
diamondhands0, line 8269: Why only follow? I assume because it was slow for some reason? Again, comments here would be helpful...

lib/block_view.go
```diff
@@ -9473,78 +9769,77 @@ func (bav *UtxoView) _flushBalanceEntriesToDbWithTxn(txn *badger.Txn) error {
...
+		}
+		if err := bav._flushMessageEntriesToDbWithTxn(txn); err != nil {
+			return err
+		}
+		if err := bav._flushBalanceEntriesToDbWithTxn(txn); err != nil {
+			return err
+		}
 	}
 
+	// Always flush to BadgerDB.
```
diamondhands0, line 9803: So the stuff below this line is what's yet to be moved right?

diamondhands0: Is there anything hard about moving the stuff below? Also, separately, is there anything more to do other than migrating these over to postgres.go?

lib/block_view_test.go
```diff
@@ -125,7 +125,7 @@ func TestBasicTransfer(t *testing.T) {
 		txn.PublicKey = recipientPkBytes
 
 		_signTxn(t, txn, recipientPrivString)
-		utxoView, _ := NewUtxoView(db, params, nil)
+		utxoView, _ := NewUtxoView(db, params, nil, nil)
```
diamondhands0, line 128: It looks like all we need to do to run all of our tests off of Postgres is to do something like:

```
func NewUtxoViewFromEnv(db, pgDB, params) {
  if <env> {
    return NewUtxoView(db, params, nil, nil)
  } else {
    return NewUtxoView(nil, params, nil, pgDB)
  }
}
```

Then in all the places where we initialize the view we can use this function, and we can test PG by running all the tests with a different env. Then eventually we can get rid of all reliance on badger.

diamondhands0: Probably best to merge what you have now if all tests are passing, but just flagging this as one way we could work the kinks out of PG by utilizing all of our old tests.

lib/postgres.go
maebeam: Nope!

lib/postgres.go
maebeam: The `PGPost` needs to either be an argument or the method needs to be defined on `PGPost`. IMO, defining it on `PGPost` makes the most sense.

lib/postgres.go
maebeam: I was going to combine Transaction Outputs and UTXOs but this will actually make warp sync more difficult. I'm going to separate these out.

lib/postgres.go
maebeam: Thank you! The continue code needed to be deleted.

lib/postgres.go
maebeam: The table name is explicitly defined and I'm following SQL conventions. I think it's more confusing to not use plural table names

lib/postgres.go
maebeam: The goal is to eventually migrate from `PostEntry` to `PGPost` and delete PostEntry entirely once we're fully moved off of badger. Baby steps are needed as `**Entry` objects are everywhere, especially in backend, and doing everything at once is much more error prone.

lib/server.go
maebeam: Bad merge resolution, thanks

migrate/20210623152412_create_tables.go
maebeam: We would need to fork the library which we will likely do in the future as I'd like to rewrite it. This migration name is generated by "go run migrate.go create"

lib/block_view.go
maebeam: This is WIP. Follows are one of the slower parts of the sync.

lib/block_view.go
maebeam: Nothing hard, just saving the easy stuff for last

lib/block_view_test.go
maebeam: Yes and yes!

lib/block_view.go
maebeam: There's no need to store PKID<->Public Key until a swap identity occurs or a profile is created. It's much more space efficient this way.

PR #69, opened 2021-07-15
Pn/derived keys

AeonSw4n:
Core idea is we accept txns signed by non-owner (derived) keys in [`_verifySignatureDerived()`](https://github.com/AeonSw4n/core/blob/4b0523f4f45840ac38cb4a797718ba0c5f5ac74d/lib/block_view.go#L2530) which is called in [`_connectBasicTransfer()`](https://github.com/AeonSw4n/core/blob/4b0523f4f45840ac38cb4a797718ba0c5f5ac74d/lib/block_view.go#L2787) on signature verification. Client looks for all derived keys associated with the txn public key and checks if any of them matches the signature. With this approach, I'm not making any changes to how txns are sent. A derived key must first be authorized by sending a new `TxnTypeAuthorizeDerivedKey` transaction constructed in [`CreateAuthorizeDerivedKeyTxn()`](https://github.com/AeonSw4n/core/blob/4b0523f4f45840ac38cb4a797718ba0c5f5ac74d/lib/blockchain.go#L3001). When clients receive the Authorize txn, they [`connect`](https://github.com/AeonSw4n/core/blob/4b0523f4f45840ac38cb4a797718ba0c5f5ac74d/lib/block_view.go#L5153) it and modify the new [`PKIDToDerivedKeyEntry`](https://github.com/AeonSw4n/core/blob/4b0523f4f45840ac38cb4a797718ba0c5f5ac74d/lib/block_view.go#L809) map in `UtxoView`. I also added a new prefix in the DB for [Authorize entries](https://github.com/AeonSw4n/core/blob/4b0523f4f45840ac38cb4a797718ba0c5f5ac74d/lib/db_utils.go#L3789). When we flush to the DB, we call [`_flushDerivedKeyEntryToDbWithTxn`](https://github.com/AeonSw4n/core/blob/4b0523f4f45840ac38cb4a797718ba0c5f5ac74d/lib/block_view.go#L7944). Let me know if this makes sense!

Note: I haven't written tests yet, and last time I checked, there was no YT tutorial on how to add a new txn to BitClout. So this is probably buggy :)


BogdanDidenko:
I'm not a big expert in this area, but why not use encryption way like
https://ledger.readthedocs.io/en/latest/background/hd_keys.html#child-key-derivation-function

As far as I understand, it is possible to check that the child's public key is derived from a parent public key if child key isn't hardened

This kind of solution would allow us to avoid the need to have a special db and send transactions to create a derived keys.

AeonSw4n:
Good direction! This was one of the ideas. ;)
However, in our setting, you could have multiple third-party apps holding your derived keys. BIP32's CKD would result in key leakage. Check out [this paper](https://eprint.iacr.org/2014/998.pdf) for more info. While this solution would be difficult to implement in our setting, _theoretically_, we could use [this](https://link.springer.com/article/10.1007/s10207-019-00476-5). But this hasn't been implemented to my knowledge.

BogdanDidenko:
Thanks for the links, I read and understood the problem better.

I think that this problem should have some graceful math solution, but I have not studied this direction)

What I searched on Google led me to BIP43 / BIP44
It would be nice to invite more experts in modern encryption to discuss

I have found this article but I don't know how useful it is:

https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2021-February/018381.html


> Identity revocations
> --------------------
> 
> It will be possible to revoke identities using single-use-seals mechanics ([originally proposed by Peter Todd] [1]) based on the bitcoin blockchain. What's important here is that this will not lead to the increase in transaction size and may be simply combined with normally happening transactions using special signature design procedure (see below).
> ...
>

andyboyd:
I have one question about this, but first some assumptions based on what I understand of what's here:

1. This approach requires writing to the blockchain every time a key is derived
2. Txns signed with derived keys will fail if the authorisation for the derived key cannot be found on the chain, or if the expiry of the derived key has passed
3. Writes to the chain take non-zero time to sync across to nodes after they are committed (this can be minutes, rather than seconds)

So, if I have an app that is using derived keys to sign transactions, and I need to sign in to a new account, or my derived key has expired and I need a new one, my client will need to authorise a new derived key. This involves a blockchain write. Does this mean that there is potentially a delay of a few minutes before the new derived key will actually be able to sign transactions, or am I misunderstanding something? Perhaps it is available on the node where it was created sooner than that?

AeonSw4n:
Yup the assumptions are correct. The derived key allows you to sign transactions via your node without identity. A node validates a txn signed with a derived key if:
1. corresponding auth txn is already in the mempool || auth txn is in a committed block
2. derived key hasn't expired

When it comes to external consistency issues that you're referring to, it's true that with this approach it can take a couple of minutes to guarantee that a txn commits globally. Any blockchain transaction suffers from this. However, it'll be instantly available in your mempool (and should be in other nodes' mempools as well).

Also note:
1. This is a "lazy" authorize txn - identity returns txn metadata and it's up to the third party to broadcast the txn.
2. The expiration date could potentially be variable (e.g. 1 week, 1 month, 1 year, etc.)

Ideally, we would ditch the authorize txn in the future, hence allowing for external consistency. It's a non-trivial and cool crypto problem to solve. 

Let me know what you think. Happy to discuss this further!

andyboyd:
I think that's fine. I suspect there may be issues for any clients that point at multiple nodes, where a derived key may only be valid for a couple of minutes on one of the nodes they are pointing at, but I can't think of a use case for a client that would want to write to multiple nodes, so it's probably not an issue. As long as the auth txn is immediately available on the mempool of the node you're pointing at, and that makes txns on the same node work immediately, I think that's workable. Sounds good to me 👍

lib/block_view.go
```diff
@@ -4962,6 +5150,86 @@ func (bav *UtxoView) _connectSwapIdentity(
...
+	// call _connectBasicTransfer to verify signatures
+	totalInput, totalOutput, utxoOpsForTxn, err := bav._connectBasicTransfer(
+		txn, txHash, blockHeight, verifySignatures)
+	if err != nil {
+		return 0, 0, nil, errors.Wrapf(err, "_connectAuthorizeDerivedKey: ")
+	}
+
+	// Force the input to be non-zero so that we can prevent replay attacks.
+	if totalInput == 0 {
+		return 0, 0, nil, RuleErrorProfileUpdateRequiresNonZeroInput
```
maebeam, line 5207: Let's make our own rule error for this and not reuse the update profile one

lib/block_view.go
```diff
@@ -725,6 +725,44 @@ func (pe *ProfileEntry) IsDeleted() bool {
 	return pe.isDeleted
 }
 
+// Obsolete?
```
maebeam, line 728: Looks obsolete to me. Let's delete them all

lib/block_view.go
```diff
@@ -2441,6 +2527,79 @@ func _isEntryImmatureBlockReward(utxoEntry *UtxoEntry, blockHeight uint32, param
...
+	}
+	txHash := Sha256DoubleHash(txBytes)
+
+	// Iterate through all derived key mappings to see if one matches the signature
+	// We're making this map to prevent a situation in which we have an outdated
+	// derived key entry in the DB, and an updated entry is in the mempool
+	derivedKeyMap := make(map[PKID]bool)
+	// Get owner PKID for map index
+	ownerPKID := *PublicKeyToPKID(txn.PublicKey)
+	for _, derivedKeyEntry := range bav.PKIDToDerivedKeyEntry[ownerPKID] {
```
maebeam, line 2544: We need to figure out a smart and efficient way to avoid this loop.

maebeam (commented):
Looks very solid overall. Great work Peter

BogdanDidenko:
> it'll be instantly available in your mempool (and should be in other nodes' mempools as well).

This also needs to be careful, I suppose.
If the transaction for creating a derived key is not added to the block for some reason, then any transactions signed with this key also cannot be added to the blocks.

Maybe we should move this discussion to this CIP https://github.com/bitclout/core/issues/58 ?

maebeam:
> If the transaction for creating a derived key is not added to the block for some reason, then any transactions signed with this key also cannot be added to the blocks.

This is correct, and after much deliberation on Discord we've decided this is a trade off we're ok with.

BogdanDidenko:
> after much deliberation on Discord
It was interesting for me to take a look at these deliberations on Discord. I looked into the "-technical-discussion" but did not find
I think it would be nice to duplicate ( just copy / paste) these deliberations to CIP https://github.com/bitclout/core/issues/58

tijno:
@AeonSw4n cant wait for this.

I did note you mentioned:

> The expiration date could potentially be variable (e.g. 1 week, 1 month, 1 year, etc.)

That would be amazing.

Will it also support different authentication levels? Eg - the key can post, but it can not transfer funds? I dont think it does currently but not sure how hard this would be to add.

carlosgnet:
> Will it also support different authentication levels? Eg - the key can post, but it can not transfer funds? I dont think it does currently but not sure how hard this would be to add.

I have just added a new discussion for this: https://github.com/bitclout/cips/discussions/115

lib/block_view.go
```diff
@@ -2268,8 +2313,6 @@ func (bav *UtxoView) _disconnectCreatorCoin(
 		} else {
 			// We do a simliar sanity-check as above, but in this case we don't need to
 			// reset the creator mappings.
-			deltaBuyerNanos := transactorBalanceEntry.BalanceNanos - operationData.PrevTransactorBalanceEntry.BalanceNanos
```
maebeam, line 2271: Looks like a merge mistake

lib/constants.go
```diff
@@ -1056,3 +1056,9 @@ const (
 	MinMaxCopiesPerNFT = 1
 	MaxMaxCopiesPerNFT = 10000
 )
+
+// Signature control byte constants.
+const (
+	DERControlByte = 48
```
maebeam, line 1062: Some additional comments here would be great

lib/block_view.go
```diff
@@ -2936,24 +3023,95 @@ func _isEntryImmatureBlockReward(utxoEntry *UtxoEntry, blockHeight uint32, param
...
 	}
-	// Verify that the transaction is signed by the specified key.
-	if txn.Signature == nil || !txn.Signature.Verify(txHash[:], txnPk) {
-		return RuleErrorInvalidTransactionSignature
+
+	// We check for public key solution iter for this transaction. This
+	// allows us to recover public key from the signature.
+	// If the transaction was signed with the owner key, iter is equal to 0.
+	// If the transaction was signed with a derived key, iter is between 1-4.
+	iter := sigByte - DERControlByte
```
maebeam, line 3056: Nit: To me `iter` keyword means `iterator` and I'm expecting a loop. Can we pick a different name for this variable?

lib/block_view.go
```diff
@@ -2936,24 +3023,95 @@ func _isEntryImmatureBlockReward(utxoEntry *UtxoEntry, blockHeight uint32, param
...
+
+		// Recover public key from the signature
+		sigCompact := SignatureSerializeCompactWithIter(sig, int(iter), false)
+		derivedPublicKey, _, err := btcec.RecoverCompact(btcec.S256(), sigCompact, txHash[:])
+		if err != nil {
+			return errors.Wrapf(err, "_verifySignature: Problem recovering public key ")
+		}
+
+		// Get owner PKID
+		ownerPKID := PublicKeyToPKID(txn.PublicKey)
```
maebeam, line 3079: @diamondhands0 Should derived keys be PKID based or Public Key based? PKID feels correct to me but want to confirm

lib/block_view.go
```diff
@@ -2936,24 +3023,95 @@ func _isEntryImmatureBlockReward(utxoEntry *UtxoEntry, blockHeight uint32, param
...
+
+		// We check if the derived key exists in DB and hasn't expired.
+		authEntry := DBGetOwnerToDerivedKeyMapping(bav.Handle, ownerPKID, derivedPKID)
+		if authEntry.OperationType == AuthorizeDerivedKeyOperationValid && authEntry.ExpirationBlock > uint64(blockHeight){
+			return nil
+			// I'm almost certain this is redundant since we already recovered the public key.
+			//if sig.Verify(txHash[:], derivedPublicKey) {
+			//	return nil
+			//}
+		}
```
maebeam, line 3111: This should be a helper on `UtxoView` and the result from the database should be cached in the `PKIDToDerivedKeyEntry` map

lib/block_view.go
```diff
@@ -2936,24 +3023,95 @@ func _isEntryImmatureBlockReward(utxoEntry *UtxoEntry, blockHeight uint32, param
...
+					}
+				}
+			}
+		}
+
+		// We check if the derived key exists in DB and hasn't expired.
+		authEntry := DBGetOwnerToDerivedKeyMapping(bav.Handle, ownerPKID, derivedPKID)
+		if authEntry.OperationType == AuthorizeDerivedKeyOperationValid && authEntry.ExpirationBlock > uint64(blockHeight){
+			return nil
+			// I'm almost certain this is redundant since we already recovered the public key.
```
maebeam, line 3107: Let's confirm this

lib/block_view.go
```diff
@@ -6675,6 +6864,150 @@ func (bav *UtxoView) _connectSwapIdentity(
 	return totalInput, totalOutput, utxoOpsForTxn, nil
 }
 
+// _verifyAccessSignature verifies if the accessSignature is valid.
```
maebeam, line 6867: Explain what an `accessSignature` is

lib/block_view.go
```diff
@@ -6675,6 +6864,150 @@ func (bav *UtxoView) _connectSwapIdentity(
...
+	}
+
+	// We make sure that the key hasn't been deleted in the past.
+	// If the derivedKey has been deleted in the past we return.
+	// Get owner PKID
+	ownerPKID := PublicKeyToPKID(ownerPublicKey)
+	// Get derived PKID
+	derivedPKID := PublicKeyToPKID(derivedPublicKey)
+	// Get previous authorize entry from the Db. This is helpful when disconnecting.
+	prevAuthEntry := DBGetOwnerToDerivedKeyMapping(bav.Handle, ownerPKID, derivedPKID)
```
maebeam, line 6948: This should use the helper that I mentioned above

lib/block_view.go
```diff
@@ -6675,6 +6864,150 @@ func (bav *UtxoView) _connectSwapIdentity(
...
+		AuthorizeDerivedKeyOperationValid,
+	}
+	bav._setDerivedKeyMappings(ownerPublicKey, derivedPublicKey, &derivedKeyEntry)
+
+	// call _connectBasicTransfer to verify txn signature.
+	totalInput, totalOutput, utxoOpsForTxn, err := bav._connectBasicTransfer(
+		txn, txHash, blockHeight, verifySignatures)
+	if err != nil {
+		// Since we've failed, we revert the utxoView mapping.
+		bav._setDerivedKeyMappings(ownerPublicKey, derivedPublicKey, prevDerivedKeyEntryCopy)
```
maebeam, line 6980: Tricky code worth reviewing @diamondhands0

lib/block_view.go
```diff
@@ -2774,6 +2817,47 @@ func (bav *UtxoView) _disconnectNFTBid(
...
+			"OperationTypeAuthorizeDerivedKey but found type %v",
+			utxoOpsForTxn[operationIndex].Type)
+	}
+	txMeta := currentTxn.TxnMeta.(*AuthorizeDerivedKeyMetadata)
+
+	// Revert the Db to the previous authorize entry from utxoOps
+	authEntry := utxoOpsForTxn[operationIndex].AuthEntry
+	senderPKID := PublicKeyToPKID(currentTxn.PublicKey)
+	derivedPKID := PublicKeyToPKID(txMeta.DerivedPublicKey)
+	err := DBPutDerivedKeyMapping(bav.Handle, senderPKID, derivedPKID, authEntry)
```
maebeam, line 2840: DB puts / gets should not be used in connect/disconnect methods. Keep them in setMappings helpers and let the flush do all DB writing

lib/network.go
```diff
@@ -2586,7 +2602,7 @@ func (msg *MsgBitCloutTxn) Copy() (*MsgBitCloutTxn, error) {
 	return newTxn, nil
 }
 
-func (msg *MsgBitCloutTxn) Sign(privKey *btcec.PrivateKey) (*btcec.Signature, error) {
+func (msg *MsgBitCloutTxn) Sign(privKey *btcec.PrivateKey, isDerivedKey bool) (*btcec.Signature, error) {
```
maebeam, line 2605: Is this extra param needed?

lib/network.go
```diff
@@ -2599,9 +2615,47 @@ func (msg *MsgBitCloutTxn) Sign(privKey *btcec.PrivateKey) (*btcec.Signature, er
...
+		}
+
+		// Get the public key solution based on btcsuite/btcd RecoverCompact method.
+		// Iteration is between 1-4.
+		iteration := 1 + int((txnSignatureCompact[0] - 27) & ^byte(4))
+
+		// Encode the public key solution in the first byte of the signature.
+		// Normally DER signatures start with 0x30 or 48 in base-10. We set
+		// the first byte to 0x30 + 0x1-4 depending on the solution.
+		txnSignatureBytes[0] = byte(48 + iteration)
```
maebeam, line 2653: Should 48 be `DERControlByte`?

lib/utils.go
```diff
@@ -122,3 +122,36 @@ func ComputeKeysFromSeedWithNet(seedBytes []byte, index uint32, isTestnet bool)
...
 	return pubKey, privKey, btcDepositAddress, nil
 }
+
+
+// SignatureSerializeCompactWithIter turns a btcec.Signature on S256 into a compact encoding. The isCompressed
+// parameter should be used if the given signature should reference a compressed public key or not.
+// If successful, the bytes of the compact signature will be returned in the format:
+// <(byte of 27+public key solution)+4 if compressed >< padded bytes for signature R><padded bytes for signature S>
+// where the R and S parameters are padded up to the bitlengh. Based on btcsuite/btcd implementation.
+func SignatureSerializeCompactWithIter(sig *btcec.Signature, iter int, isCompressedKey bool) []byte {
```
maebeam, line 132: Can you link to the referenced implementation?

maebeam:
Overall it looks great. I am slightly concerned about the performance of recovering a key from a signature but don't think it's worth changing the implementation.

lib/block_view.go
```diff
@@ -2936,24 +3023,95 @@ func _isEntryImmatureBlockReward(utxoEntry *UtxoEntry, blockHeight uint32, param
...
+		// Get owner PKID
+		ownerPKID := PublicKeyToPKID(txn.PublicKey)
+		// Get derived PKID
+		derivedPKID := PublicKeyToPKID(derivedPublicKey.SerializeCompressed())
+
+		// We check if there's an entry in UtxoView that authorizes this derived key
+		if entryMap, ok := bav.PKIDToDerivedKeyEntry[*ownerPKID]; ok {
+			if entry, ok := entryMap[*derivedPKID]; ok {
+				// Check if the key hasn't been revoked
+				if entry != nil {
```
maebeam, line 3087: Need to check `isDeleted`

lib/block_view.go
```diff
@@ -2936,24 +3023,95 @@ func _isEntryImmatureBlockReward(utxoEntry *UtxoEntry, blockHeight uint32, param
...
+						return nil
+						// I'm almost certain this is redundant since we've already recovered the public key.
+						//if sig.Verify(txHash[:], derivedPublicKey) {
+						//	return nil
+						//}
+					} else if entry.OperationType == AuthorizeDerivedKeyOperationInValid {
+						// If the owner de-authorized a key in this block, we want to immediately prevent any
+						// transactions signed with the derived key. So we return and skip Db check.
+						return RuleErrorInvalidTransactionSignature
+					}
```
maebeam, line 3098: Else, error, invalid operation type

lib/block_view.go
```diff
@@ -2936,24 +3023,95 @@ func _isEntryImmatureBlockReward(utxoEntry *UtxoEntry, blockHeight uint32, param
...
+				}
+			}
+		}
+
+		// We check if the derived key exists in DB and hasn't expired.
+		authEntry := DBGetOwnerToDerivedKeyMapping(bav.Handle, ownerPKID, derivedPKID)
```
maebeam, line 3104: Don't access the DB directly. Always check the view

lib/block_view.go
```diff
@@ -2936,24 +3023,95 @@ func _isEntryImmatureBlockReward(utxoEntry *UtxoEntry, blockHeight uint32, param
...
+			return errors.Wrapf(err, "_verifySignature: Problem recovering public key ")
+		}
+
+		// Get owner PKID
+		ownerPKID := PublicKeyToPKID(txn.PublicKey)
+		// Get derived PKID
+		derivedPKID := PublicKeyToPKID(derivedPublicKey.SerializeCompressed())
+
+		// We check if there's an entry in UtxoView that authorizes this derived key
+		if entryMap, ok := bav.PKIDToDerivedKeyEntry[*ownerPKID]; ok {
```
maebeam, line 3084: Why not just use getOwner here?

lib/block_view.go
```diff
@@ -1004,6 +1033,9 @@ type UtxoOperation struct {
 	NFTSpentUtxoEntries       []*UtxoEntry
 	PrevAcceptedNFTBidEntries *[]*NFTBidEntry
 
+	// For disconnecting authorize derived key.
+	AuthEntry *AuthorizeEntry
```
maebeam, line 1037: Why not just use DerivedKeyEntry here?

lib/block_view.go
AeonSw4n: isDeleted becomes OperationType, and we're already checking for AuthorizeDerivedKeyOperationValid on line 3088. But good point, I'll change the naming so it's not confusing.

lib/network.go
AeonSw4n: oops

lib/blockchain.go
```diff
@@ -3214,6 +3214,73 @@ func (bc *Blockchain) CreateCreatorCoinTransferTxnWithDiamonds(
 	return txn, totalInput, changeAmount, fees, nil
 }
 
+func (bc *Blockchain) CreateAuthorizeDerivedKeyTxn(
+	ownerPublicKey []byte,
+	derivedPublicKey []byte,
```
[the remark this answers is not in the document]
AeonSw4n: You could have a single derived key authorized on multiple owner key pairs.

lib/block_view.go
```diff
@@ -2987,24 +3067,74 @@ func _isEntryImmatureBlockReward(utxoEntry *UtxoEntry, blockHeight uint32, param
...
-	// Convert the txn public key into a *btcec.PublicKey
-	txnPk, err := btcec.ParsePubKey(txn.PublicKey, btcec.S256())
+
+	// We make a local copy of the signature as it should be read-only.
+	sigCopy := append([]byte{}, txn.Signature...)
+	// Verify that the control byte is within the allowed range.
+	// DERControlByte means txn was signed with the owner's private key.
+	// Otherwise, txn was signed with a derived key and we will attempt
+	// to recover the public key from the signature.
+	sigByte := sigCopy[0]
```
diamondhands0, line 3084: You have good comments but I'm still very confused by what's going on here. It looks like you're trying to encode a boolean isDerivedKey inside of the signature somehow. If this is the case, I think it's error-prone and a bit too clever. Instead, I would create a simple addition to extraData that contains this information, and modify this function to support an isDerivedKey boolean.

lib/block_view.go
```diff
@@ -2987,24 +3067,74 @@ func _isEntryImmatureBlockReward(utxoEntry *UtxoEntry, blockHeight uint32, param
...
 	}
-	// Verify that the transaction is signed by the specified key.
-	if txn.Signature == nil || !txn.Signature.Verify(txHash[:], txnPk) {
-		return RuleErrorInvalidTransactionSignature
+
+	// We check for public key solution iteration for this transaction. This
+	// allows us to recover public key from the signature.
+	// If the transaction was signed with the owner key, iteration is equal to 0.
+	// If the transaction was signed with a derived key, iteration is between 1-4.
+	iteration := sigByte - DERControlByte
```
diamondhands0, line 3100: I'm sure you've tested this, and your comments are pretty good, but I'm unfortunately still very confused about what this iteration stuff is doing. Please add in-depth commentary or a link to somewhere where I can get smarter about this so that I can verify that there aren't any holes. The fear I have is that this could be exploited in some way to fool us into signing on behalf of someone else.

lib/block_view.go
```diff
@@ -2987,24 +3067,74 @@ func _isEntryImmatureBlockReward(utxoEntry *UtxoEntry, blockHeight uint32, param
...
+	if iteration == 0 {
```
diamondhands0, line 3101: I think this logic would be a lot nicer if it was more like:

```
if isDerivedKey {
// something
} else {
// something
}
```

This sigByte stuff is too clever and makes the code a lot harder to read. I'd prefer to take an efficiency hit on an extraData bool rather than embed this in the signature itself.

lib/block_view.go
```diff
@@ -2987,24 +3067,74 @@ func _isEntryImmatureBlockReward(utxoEntry *UtxoEntry, blockHeight uint32, param
...
+		derivedPk, _, err := btcec.RecoverCompact(btcec.S256(), sigCompact, txHash[:])
+		if err != nil {
+			return errors.Wrapf(err, "_verifySignature: Problem recovering public key: ")
+		}
+
+		ownerPublicKey := txn.PublicKey
+		derivedPublicKey := derivedPk.SerializeCompressed()
+
+		// Look for a derived key entry and check if it's valid.
+		derivedKeyEntry := bav._getDerivedKeyMappingForOwner(ownerPublicKey, derivedPublicKey)
```
diamondhands0, line 3126: This if statement is a little tricky, and I think it's missing the isDeleted case. I think we should do something a little simpler like:

```
if derivedKeyEntry == nil || derivedKeyEntry.isDeleted == true {
return RuleErrorDerivedKeyInvalid
}
```

lib/block_view.go
```diff
@@ -844,6 +860,9 @@ type UtxoView struct {
 	// Coin balance entries
 	HODLerPKIDCreatorPKIDToBalanceEntry map[BalanceEntryMapKey]*BalanceEntry
 
+	// Derived Key entries. First index is owner PKID, then derived PKID
+	PKIDToDerivedKeyEntry          map[PKID]map[PKID]*DerivedKeyEntry
```
diamondhands0, line 864: A couple of separate thoughts here:

First, it feels like everything dealing with derived keys should be public key based. It should be PublicKey1 -> <PublicKey2, PublicKey3>. And we should leave PKID as a 1:1 mapping with each public key that's separate from all this DerivedKey stuff.

Second, this map of maps is a bit too complex. Our convention is to keep everything as <key> -> Entry. This being said, it seems like we should track everything as:
```
type DerivedKeyMapKey struct {
OwnerPublicKey []byte
DerivedPublicKey []byte
}

map[DerivedKeyMapKey]*DerivedKeyEntry
```

I know this will make the GetAllDerivedKeysForPublicKey function slower, since you will have to iterate over the whole map. But I think it's worth the simplification. I also don't see that function used anywhere.

diamondhands0: I'm going to pause my review here while I discuss public keys vs PKIDs for this with @maebeam.

lib/block_view.go
AeonSw4n: Hah, I can imagine how confusing this code must look without first reading some background explanation. I've added a longer comment higher in the function, and here's some more info:

Problem: Imagine a user has more than one authorized derived key. In such a case, when we sign a txn with a derived key, how does `_verifySignature()` know which particular derived key was used?

Slow solution: Check all authorized derived keys, user-wise.
Expensive solution: Pass the derived public key in extraData.

Cool solution: Use the signature itself to find out which derived public key was used. This way we don't need to pass any more data to a transaction, and it "just works."
Question: Is this secure? There are two attack vectors: theory-based and implementation-based. For the former, turns out breaking this scheme is equivalent to breaking ecdsa. I made a more detailed, developer-friendly explanation of the theory here: https://github.com/AeonSw4n/PublicKeyRecoveryECDSA
In regards to implementation errors, I tried writing a lot of tests in block_view_test.go.

lib/block_view.go
AeonSw4n: Please let me know if you still think we'd want the additional field in extraData after my comment on the public key iteration.

lib/block_view.go
AeonSw4n: I'm using `OperationType` to determine if the key is authorized or not. I was thinking that having `isDeleted` in the `derivedKeyEntry` might cause confusion since in other txns we're using it for tombstone entries. When we remove a `derivedKeyEntry` from UtxoView in `_disconnectAuthorizeDerivedKey()` it'll revert to the previous entry, or if none exists, it'll set `OperationType` to `OperationValid` and `expirationBlock` to `0` which by default fails the expiration check. This design choice allows us to treat `OperationType` as a one-time switch, where if you de-authorize a key, you can't ever authorize it again. Let me know if you still would like the `isDeleted` field.

AeonSw4n (commented):
Thanks for the detailed review! in regards to your comments on PKID, I changed everything to be public key based including the utxo map key. Not a huge fan that it causes `GetAllDerivedKeyMappingsForOwner` function to be `O(n)` but I agree it makes sense consistency-wise. Also this function is used in the backend repo.

lib/block_view.go
AeonSw4n: I changed everything to be public key based including the utxo map key. Not a huge fan that it causes `GetAllDerivedKeyMappingsForOwner` function to be `O(n)` but I agree it makes sense consistency-wise. Also, this function is used in the backend repo.

migrate/20210623152412_create_tables.go
```diff
@@ -44,8 +44,7 @@ func init() {
 				type       SMALLINT NOT NULL,
 				public_key BYTEA,
 				extra_data JSONB,
-				r          BYTEA,
-				s          BYTEA
+				signature  BYTEA,
```
maebeam: Please create a new migration file using `go run scripts/migrate.go create` to modify this column and add the new table

AeonSw4n: Good point, created!

AeonSw4n: Actually, do you think I should also revert to btcec.Signature in MsgBitCloutTxn (previously changed it to []byte) and r, s in Postgres?

migrate/20210907001756_derived_keys_update.go
```diff
@@ -0,0 +1,74 @@
...
+			);
+		`)
+		if err != nil {
+			return err
+		}
+
+		_, err = db.Exec(`
+			ALTER TABLE IF EXISTS pg_transactions 
+				DROP COLUMN IF EXISTS r, 
+				DROP COLUMN IF EXISTS s
```
maebeam, line 40: Can drops and adds both happen in this exec?

migrate/20210907001756_derived_keys_update.go
```diff
@@ -0,0 +1,74 @@
...
+			return err
+		}
+
+		return nil
+	}
+
+	down := func(db orm.DB) error {
+		_, err := db.Exec(`
+			DROP TABLE pg_metadata_derived_keys;
+			DROP TABLE pg_derived_keys;
```
maebeam, line 62: Please revert the changes to pg_transactions

AeonSw4n: Sounds good.

lib/block_view.go
```diff
@@ -2919,24 +3053,72 @@ func _isEntryImmatureBlockReward(utxoEntry *UtxoEntry, blockHeight uint32, param
...
-	txnPk, err := btcec.ParsePubKey(txn.PublicKey, btcec.S256())
+	sig, err := btcec.ParseDERSignature(txn.Signature, btcec.S256())
+
+	// Look for the derived key in transaction ExtraData and validate it. For transactions
+	// signed using a derived key, the derived public key is passed to ExtraData.
+	var derivedPk *btcec.PublicKey
+	var derivedPkBytes []byte
+	isDerived := false
+	if txn.ExtraData != nil {
+		if derivedPkBytes, isDerived = txn.ExtraData[DerivedPublicKey]; isDerived {
```
diamondhands0, line 3071: Nit: Avoid assignment in if statements since this can be error-prone. Instead prefer:

```
derivedPkBytes, isDerived = txn.ExtraData[DerivedPublicKey];
if isDerived {
...
}
```

lib/block_view.go
```diff
@@ -2919,24 +3053,72 @@ func _isEntryImmatureBlockReward(utxoEntry *UtxoEntry, blockHeight uint32, param
...
-		return errors.Wrapf(err, "_verifySignature: Problem parsing public key: ")
+		return errors.Wrapf(err, "_verifySignature: Problem parsing owner public key: ")
 	}
-	// Verify that the transaction is signed by the specified key.
-	if txn.Signature == nil || !txn.Signature.Verify(txHash[:], txnPk) {
-		return RuleErrorInvalidTransactionSignature
+
+	// If no derived key is present in ExtraData, we check if transaction was signed by the owner.
+	// If derived key is present in ExtraData, we check if transaction was signed by the derived key.
+	if isDerived == false {
```
diamondhands0, line 3088: Nit: isDerived is redundant and error-prone. I would instead do the following:

1) Define isDerived as a local variable in the `if txn.ExtraData != nil` block

2) Instead of checking `isDerived == false` do `derivedPk == nil`

lib/block_view.go
```diff
@@ -4557,6 +4743,102 @@ func (bav *UtxoView) _deleteProfileEntryMappings(profileEntry *ProfileEntry) {
...
+func (bav *UtxoView) _getDerivedKeyMappingForOwner(ownerPublicKey []byte, derivedPublicKey []byte) *DerivedKeyEntry {
+	// Check if the entry exists in utxoView
+	derivedKeyMapKey := MakeDerivedKeyMapKey(*NewPublicKey(ownerPublicKey), *NewPublicKey(derivedPublicKey))
+	entry, exists := bav.DerivedKeyToDerivedEntry[derivedKeyMapKey]
+	if exists {
+		return entry
+	}
+
+	// Check if the entry exists in the DB.
+	ownerPk := NewPublicKey(ownerPublicKey)
```
diamondhands0, line 4756: Nit: Define these at the top and use them in MakeDerivedKeyMapKey

lib/block_view.go
```diff
@@ -2919,24 +3053,72 @@ func _isEntryImmatureBlockReward(utxoEntry *UtxoEntry, blockHeight uint32, param
...
+			return nil
+		}
+	} else {
+		// Look for a derived key entry in UtxoView and DB, check if it exists nor is deleted.
+		derivedKeyEntry := bav._getDerivedKeyMappingForOwner(ownerPkBytes, derivedPkBytes)
+		if derivedKeyEntry == nil || derivedKeyEntry.isDeleted {
+			return RuleErrorDerivedKeyNotAuthorized
+		}
+
+		// Return an error if transaction public keys don't match derivedKeyEntry public keys.
```
diamondhands0, line 3100: Nit: This is a sanity check so update comment as follows:

// Sanity-check that transaction public keys line up with looked-up derivedKeyEntry public keys.

lib/block_view.go
```diff
@@ -2919,24 +3053,72 @@ func _isEntryImmatureBlockReward(utxoEntry *UtxoEntry, blockHeight uint32, param
...
+			derivedKeyEntry.ExpirationBlock <= uint64(blockHeight) {
+			return RuleErrorDerivedKeyNotAuthorized
+		}
+
+		// All checks passed so we try to verify the signature.
+		if sig.Verify(txHash[:], derivedPk) {
+			return nil
+		}
+
+		return RuleErrorDerivedKeyNotAuthorized
```
diamondhands0, line 3118: Wow man! I think this works, and it's braindead simple. Great work! The hard part is over. Now onto the rest :)

lib/block_view.go
```diff
@@ -7060,6 +7342,153 @@ func (bav *UtxoView) _connectSwapIdentity(
...
 }
 
+// _verifyAccessSignature verifies if the accessSignature is correct. Valid
+// accessSignature is the signed hash of (derivedPublicKey + expirationBlock)
+// in DER format, made with the ownerPublicKey.
+func _verifyAccessSignature(ownerPublicKey []byte, derivedPublicKey []byte,
+	expirationBlock uint64, accessSignature []byte) error {
+
+	// Compute a hash of derivedPublicKey+expirationBlock
+	expirationBlockByte := UintToBuf(expirationBlock)
```
diamondhands0, line 7352: Nit: Did you mean for this to be singular "byte" rather than plural "bytes"? UintToBuf returns a slice not a single byte.

lib/block_view.go
```diff
@@ -7060,6 +7342,153 @@ func (bav *UtxoView) _connectSwapIdentity(
...
+			txn.TxnMeta.GetTxnType().String())
+	}
+
+	// Make sure transaction hasn't expired.
+	if txMeta.ExpirationBlock <= uint64(blockHeight) {
+		return 0, 0, nil, RuleErrorAuthorizeDerivedKeyExpiredDerivedPublicKey
+	}
+
+	// Validate the owner public key.
+	ownerPublicKey := txn.PublicKey
```
diamondhands0, line 7404: Fun question for you. I don't think it's harmful, but can you have a derived key off of a derived key? E.g:

`original_owner_pubkey -> derived_key_1 -> derived_key_2` ?

lib/block_view.go
```diff
@@ -7060,6 +7342,153 @@ func (bav *UtxoView) _connectSwapIdentity(
...
+		if prevDerivedKeyEntry.OperationType == AuthorizeDerivedKeyOperationNotValid {
+			return 0, 0, nil, RuleErrorAuthorizeDerivedKeyDeletedDerivedPublicKey
+		}
+	}
+
+	// At this point we've verified the access signature, which means the derived key is authorized
+	// to sign on behalf of the owner. We now add a temporary derived key to utxoView with
+	// OperationType set to Valid so we pass txn signature verification in _connectBasicTransfer.
+	derivedKeyEntry := DerivedKeyEntry {
+		*NewPublicKey(ownerPublicKey),
```
diamondhands0, line 7443: Nit: Always label the struct fields. E.g.:

```
OwnerPublicKey: *NewPublicKey(ownerPublicKey),
...
```

lib/block_view.go
```diff
@@ -7060,6 +7342,153 @@ func (bav *UtxoView) _connectSwapIdentity(
...
+	}
+	bav._setDerivedKeyMapping(&derivedKeyEntry)
+
+	// Call _connectBasicTransfer to verify txn signature.
+	totalInput, totalOutput, utxoOpsForTxn, err := bav._connectBasicTransfer(
+		txn, txHash, blockHeight, verifySignatures)
+	if err != nil {
+		// Since we've failed, we revert the utxoView mapping to what it was previously.
+		// We're doing this manually because we haven't fully connected the txn yet.
+		bav._deleteDerivedKeyMapping(&derivedKeyEntry)
```
diamondhands0, line 7457: Heh. Right now we don't clean up if an error occurs. Rather, the caller is supposed to assume the view is completely broken when this happens. Just to keep things consistent, let's revert these two lines and just return the error. Same for below.

diamondhands0: Oh I see. This is a little hackey what you're doing here. Hmm...

diamondhands0: I'm confused. Why do you need to setDerivedKeyMappings above before calling _connectBasicTransfer? It seems like it should work without you having to do this. Are you trying to support the case where a derived key can create a sub-derived-key?

AeonSw4n: I'm allowing the Authorize transaction to be signed by the derived key (since the access signature is already signed by the owner). For first-time keys there's no existing entry in the DB so `_verifySignature()` would fail. If I add the key before calling the `_connectBasicTransfer` then it's loaded in utxoview and `_verifySignature()` passes.

lib/block_view.go
```diff
@@ -2757,6 +2812,82 @@ func (bav *UtxoView) _disconnectNFTBid(
...
+		return fmt.Errorf("_disconnectAuthorizeDerivedKey: DerivedKeyEntry for "+
+			"public key %v, derived key %v was found to be nil or deleted: %v",
+			PkToString(ownerPublicKey, bav.Params), PkToString(derivedPublicKey, bav.Params),
+			derivedKeyEntry)
+	}
+
+	// If we had a previous derivedKeyEntry set then compare it with the current entry.
+	if prevDerivedKeyEntry != nil {
+		// Sanity check public keys. This should never fail.
+		if !reflect.DeepEqual(ownerPublicKey, prevDerivedKeyEntry.OwnerPublicKey[:]) {
```
diamondhands0, line 2867: Great sanity check. We love sanity checks like this.

lib/block_view.go
```diff
@@ -7060,6 +7342,153 @@ func (bav *UtxoView) _connectSwapIdentity(
...
+		txMeta.ExpirationBlock, txMeta.AccessSignature)
+	if err != nil {
+		return 0, 0, nil, errors.Wrap(RuleErrorAuthorizeDerivedKeyAccessSignatureNotValid, err.Error())
+	}
+
+	// Get current (previous) derived key entry. We might revert to it later so we copy it.
+	prevDerivedKeyEntry := bav._getDerivedKeyMappingForOwner(ownerPublicKey, derivedPublicKey)
+
+	// We make sure that the key hasn't been invalidated in the past. If that's the case,
+	// we return. A derived key can't be re-authorized after being revoked in the past.
```
diamondhands0, line 7432: Nit: Please explain the reasoning behidn this.

lib/block_view.go
```diff
@@ -7060,6 +7342,153 @@ func (bav *UtxoView) _connectSwapIdentity(
...
+	// to sign on behalf of the owner. We now add a temporary derived key to utxoView with
+	// OperationType set to Valid so we pass txn signature verification in _connectBasicTransfer.
+	derivedKeyEntry := DerivedKeyEntry {
+		*NewPublicKey(ownerPublicKey),
+		*NewPublicKey(derivedPublicKey),
+		txMeta.ExpirationBlock,
+		AuthorizeDerivedKeyOperationValid,
+		false,
+	}
+	bav._setDerivedKeyMapping(&derivedKeyEntry)
```
diamondhands0, line 7449: Let the people know this is a dirty hack and explain the why in detail.

lib/block_view.go
diamondhands0: I think you can still revert the delete+set. If it errors then the view is assumed to be compromised and we'll generate a new one.

lib/block_view.go
```diff
@@ -7060,6 +7342,153 @@ func (bav *UtxoView) _connectSwapIdentity(
...
+		// Since we've failed, we revert the utxoView mapping to what it was previously.
+		// We're doing this manually because we haven't fully connected the txn yet.
+		bav._deleteDerivedKeyMapping(&derivedKeyEntry)
+		bav._setDerivedKeyMapping(prevDerivedKeyEntry)
+		return 0, 0, nil, RuleErrorAuthorizeDerivedKeyRequiresNonZeroInput
+	}
+
+	// Now that we've verified this txn is good to go, we update the utxoView entry
+	// OperationType to what it's supposed to be.
+	derivedKeyEntry.OperationType = txMeta.OperationType
```
diamondhands0, line 7473: Just to be safe, we like to _delete then _set whenever we're modifying a field like we're doing here. That future-proofs us in the event we add a new index on a field or something.

lib/block_view.go
diamondhands0: Eh.. on second thought just keep the _delete+_set. Doesn't hurt.

lib/block_view.go
```diff
@@ -7060,6 +7342,153 @@ func (bav *UtxoView) _connectSwapIdentity(
...
+	bav._setDerivedKeyMapping(&derivedKeyEntry)
+
+	if verifySignatures {
+		// _connectBasicTransfer has already checked that the transaction is
+		// signed by the top-level public key, which we take to be the poster's
```
diamondhands0, line 7478: Nit: It's signed either by the top-level key OR the derived key. Let's make that distinction here.

lib/postgres.go
```diff
@@ -1903,6 +1987,31 @@ func (postgres *Postgres) GetNFTBid(nftPostHash *BlockHash, bidderPKID *PKID, se
...
+
+func (postgres *Postgres) GetAllDerivedKeysForOwner(ownerPublicKey *PublicKey) []*PGDerivedKey {
+	var keys []*PGDerivedKey
+	err := postgres.db.Model(&keys).Where("owner_public_key = ?", *ownerPublicKey).Select()
+	if err != nil {
+		return nil
+	}
+	return keys
+}
+
```
diamondhands0, line 2014: All the Postgres stuff looks fine. Were you able to test it?

diamondhands0: I think @maebeam knows an environment variable you can set to run the test. Or maybe our continuous integration does it for you.

lib/blockchain_test.go
```diff
@@ -741,10 +741,27 @@ func _signTxn(t *testing.T, txn *MsgBitCloutTxn, privKeyStrArg string) {
...
 	require.NoError(err)
 	privKey, _ := btcec.PrivKeyFromBytes(btcec.S256(), privKeyBytes)
-	require.NoError(err)
 	txnSignature, err := txn.Sign(privKey)
 	require.NoError(err)
-	txn.Signature = txnSignature
+	txn.Signature = txnSignature.Serialize()
+}
+
+// Signs a txn with derived key. Signature contains solution iteration,
```
diamondhands0, line 749: I think we can delete this now. Sad; it was a fun function.

lib/constants.go
```diff
@@ -828,3 +831,22 @@ const (
...
+// We use a custom signature encoding which allows us to support derived keys.
+// Standard BitClout transaction signatures are stored in a DER format:
+// 0x30 <length of whole message> <0x02> <length of R> <R> 0x2 <length of S> <S>.
+// The first byte, 0x30, is called the DER control byte.
+// For transactions signed with a derived key, we use a different control byte ranging
+// from 0x31 to 0x34. This encodes a public key iteration. The iteration allows us to
+// recover the derived public key from the signature. This way we don't need to transmit
+// the derived key every time we want to sign a transaction on behalf of the owner.
+// More information can be found in section 4.1.6 of SEC 1 Ver 2.0, page 47-48 (53 and 54 in the pdf).
+const (
```
diamondhands0, line 845: I think we don't need this anymore as well.

lib/network.go
```diff
@@ -2245,7 +2250,7 @@ type MsgBitCloutTxn struct {
 	// inputs to the transaction. The exception to this rule is that
 	// BLOCK_REWARD and CREATE_bitclout transactions do not require a signature
 	// since they have no inputs.
-	Signature *btcec.Signature
+	Signature []byte
```
diamondhands0, line 2253: Huh. This seems like an interesting change. What prompted this?

diamondhands0: Ah, it was the custom signature format right? Given that we removed that I would revert this to how it was.

lib/network.go
```diff
@@ -2604,6 +2612,36 @@ func (msg *MsgBitCloutTxn) Sign(privKey *btcec.PrivateKey) (*btcec.Signature, er
 	return txnSignature, nil
 }
 
+// SignTransactionWithDerivedKey the signature contains solution iteration,
+// which allows us to recover signer public key from the signature.
+// Returns (new txn bytes, txn signature, error)
+func SignTransactionWithDerivedKey(txnBytes []byte, privateKey *btcec.PrivateKey) ([]byte, []byte, error){
```
diamondhands0, line 2618: Sad, but I think we should delete this now.

lib/network.go
```diff
@@ -2663,7 +2701,7 @@ func (msg *MsgBitCloutTxn) UnmarshalJSON(data []byte) error {
 		TxOutputs []*BitCloutOutput
 		TxnMeta   BitCloutTxnMetadata
 		PublicKey []byte
-		Signature *btcec.Signature
+		Signature []byte
```
diamondhands0, line 2704: Revert if we can.

lib/postgres.go
```diff
@@ -100,8 +100,7 @@ type PGTransaction struct {
 	Type      TxnType    `pg:",use_zero"`
 	PublicKey []byte     `pg:",type:bytea"`
 	ExtraData map[string][]byte
-	R         *BlockHash `pg:",type:bytea"`
-	S         *BlockHash `pg:",type:bytea"`
+	Signature []byte     `pg:",type:bytea"`
```
diamondhands0, line 103: Is this still needed?

lib/block_view.go
AeonSw4n: wow... so obvious, how have I missed this one!

lib/block_view.go
AeonSw4n: Hah nice. You can definitely have a derived key of a derived key. Though `key_2` won't automatically sign on behalf of `owner_pubkey`; you need an access signature and a new authorize transaction for that.

lib/blockchain_test.go
AeonSw4n: This one actually comes in handy in `block_view_test.go`, I updated it now to reflect the new `ExtraData` approach.

lib/constants.go
AeonSw4n: rip

lib/network.go
AeonSw4n: Ha! It is actually useful in the backend, https://github.com/AeonSw4n/backend/blob/fc59f/routes/admin_transaction.go#L351 , but I've updated the comment and method to reflect the new `ExtraData` approach.

lib/network.go
AeonSw4n: Makes sense.

lib/postgres.go
AeonSw4n: Nope, reverted.

lib/network.go
AeonSw4n: Yup, exactly. Reverted.

lib/postgres.go
AeonSw4n: Yeah, tested it manually. It would be ideal to add another unit test to `block_view_test.go` for Postgres, but to do this I'd need a way to generate a test Postgres Db, similarly to `GetTestBadgerDb()` from `blockchain_test.go`. Idk, is there a function like this already?

lib/block_view.go
```diff
@@ -7620,6 +7911,166 @@ func (bav *UtxoView) _connectSwapIdentity(
...
+	}
+
+	// Get current (previous) derived key entry. We might revert to it later so we copy it.
+	prevDerivedKeyEntry := bav._getDerivedKeyMappingForOwner(ownerPublicKey, derivedPublicKey)
+
+	// Authorize transactions can be signed by both owner and derived keys. However, this
+	// poses a risk in a situation where a malicious derived key, which has previously been
+	// de-authorized by the owner, were to attempt to re-authorize itself.
+	// To prevent this, the following check completely blocks a derived key once it has been
+	// de-authorized. This makes the lifecycle of a derived key more controllable.
```
diamondhands0, line 8004: Great comment.

lib/block_view.go
```diff
@@ -7620,6 +7911,166 @@ func (bav *UtxoView) _connectSwapIdentity(
...
+	// key entry to UtxoView, to support first-time derived keys (they don't exist in the DB yet).
+	// As a result, and if the derived key is present in transaction's ExtraData, we will
+	// pass signature verification in _connectBasicTransfer() -> _verifySignature().
+	//
+	// NOTE: Setting a mapping in UtxoView prior to fully validating a transaction shouldn't be
+	// reproduced elsewhere. It's error-prone, controversial, some even call it "a dirty hack!"
+	// All considered, this feature greatly simplifies the flow in identity - from the moment you
+	// generate a derived key, you can use it to sign any transaction offline, including authorize
+	// transactions. It also resolves issues in situations where the owner account has insufficient
+	// balance to submit an authorize transaction.
```
diamondhands0, line 8023: Awesome comment. Always tend toward over-commenting like this!

lib/block_view.go
```diff
@@ -7620,6 +7911,166 @@ func (bav *UtxoView) _connectSwapIdentity(
...
 }
 
+// _verifyAccessSignature verifies if the accessSignature is correct. Valid
+// accessSignature is the signed hash of (derivedPublicKey + expirationBlock)
+// in DER format, made with the ownerPublicKey.
+func _verifyAccessSignature(ownerPublicKey []byte, derivedPublicKey []byte,
+	expirationBlock uint64, accessSignature []byte) error {
+
+	// Compute a hash of derivedPublicKey+expirationBlock.
+	expirationBlockBytes := UintToBuf(expirationBlock)
```
diamondhands0, line 7921: Do two last things for me here:
1. Validate that the derivedPublicKey length is btcec.PubKeyLenCompressed
2. Use _EncodeUint64() or whatever that function is both here and when you produce the access signature. UintToBuf is a variable-length encoding scheme, whereas _EncodeUint64() is fixed-width.

This will guarantee that the entire accessBytes has a consistent width, and an attacker can't play games by e.g. providing a shorter public key to get a different height approved...

lib/block_view.go
```diff
@@ -7620,6 +7911,166 @@ func (bav *UtxoView) _connectSwapIdentity(
...
+		return fmt.Errorf("_verifyAccessSignature: Invalid signature")
+	}
+
+	return nil
+}
+
+func (bav *UtxoView) _connectAuthorizeDerivedKey(
+	txn *MsgBitCloutTxn, txHash *BlockHash, blockHeight uint32, verifySignatures bool) (
+	_totalInput uint64, _totalOutput uint64, _utxoOps []*UtxoOperation, _err error) {
+
```
diamondhands0, line 7951: You need to do one more thing here, which is you need to gate this function on a block height to avoid causing a fork. We have a block height that is triggering Wednesday called NFTTransferOrBurnBlockHeight. You need to do the following:

1. Rename NFTTransferOrBurnBlockHeight to NFTTransferOrBurnAndDerivedKeysBlockHeight
2. Add a block like this to the beginning of this function: https://github.com/bitclout/core/blob/main/lib/block_view.go#L7192
3. Fix your tests by setting the value to zero like line 19225 here: https://github.com/bitclout/core/blob/main/lib/block_view_test.go

lib/block_view.go
diamondhands0: A little sus but seems harmless so resolving.

lib/network.go
diamondhands0: Ah ok looks good.

lib/postgres.go
diamondhands0: No, but the tests that run via the continuous integration exercise Postgres so no need to do anything beyond a single manual sanity check.

lib/block_view.go
AeonSw4n: Yes indeed, we don't want a fork! Thanks for the detailed steps :)

lib/block_view.go
AeonSw4n: Lol, so much h*zing. Had to write `EncodeUint64()` in identity from scratch, but should be good now. Super unlikely anything would have happened after we verify `derivedPublicKey` length; though for sure, better safe than sorry.

lib/block_view.go
AeonSw4n: I wonder whose quote is it? 🧐

diamondhands0 (approved):

PR #74, opened 2021-07-24
Fix TXIndex shutdown crashes

gfodor:
When SIGTERM'ing the node, memory access segfaults were occurring in `txindex.go`.

If TXIndex was disabled, the reference was being accessed regardless, causing a crash.

If TXIndex was enabled, the Update goroutine was not being stopped before the database was closed.

This fixes both issues. I'm new to Go so apologize if this patch is doing something non-idiomatic or wrong. SIGTERM now seems to shut down properly.

### DIFF
--- a/lib/txindex.go
+++ b/lib/txindex.go
@@ -6,6 +6,7 @@ import (
 	"github.com/dgraph-io/badger/v3"
 	"path/filepath"
 	"reflect"
+	"sync"
 	"time"
 
 	chainlib "github.com/btcsuite/btcd/blockchain"
@@ -28,6 +29,12 @@ type TXIndex struct {
 
 	// Core params object
 	Params *BitCloutParams
+
+	// Update wait group
+	updateWaitGroup sync.WaitGroup
+
+	// Shutdown channel
+	stopUpdateChannel chan struct{}
 }
 
 func NewTXIndex(coreChain *Blockchain, bitcoinManager *BitcoinManager, params *BitCloutParams, dataDirectory string) (*TXIndex, error) {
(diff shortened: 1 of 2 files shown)


gfodor:
It looks like BlockProducer has this same issue wrt graceful shutdown of occasional segfaults. I will try to open another PR for that problem.

tijno:
Awesome! Welcome. Beyond my ability to review but great to see nee contributors to backend/core.

AeonSw4n (commented):
Looks good. Glad to see more devs helping improve BitClout. I've made some small stylistic suggestions. Could we also merge with #75 into a single PR?

lib/txindex.go
```diff
@@ -28,6 +29,12 @@ type TXIndex struct {
 
 	// Core params object
 	Params *BitCloutParams
+
+	// Update wait group
+	updateWaitGroup sync.WaitGroup
```
AeonSw4n, line 34: Could we make this `UpdateWaitGroup`

lib/txindex.go
```diff
@@ -28,6 +29,12 @@ type TXIndex struct {
...
+
+	// Shutdown channel
+	stopUpdateChannel chan struct{}
```
AeonSw4n, line 37: How about `StopUpdateChannel`

lib/txindex.go
gfodor: Sure! Got a tl;dr on core heuristics for public vs private? As a third party the prevalence of public fields and methods has been appreciated, but wasn’t sure if there is any obvious cut points being used to decide.

maebeam (approved):
Thank you so much!

lib/txindex.go
AeonSw4n: Was just thinking making it public is more future-proof. And you mentioned you're new to Go so wasn't sure if that was intentional or not. Nonetheless, congrats on the double-merge! :)

PR #75, opened 2021-07-24
Fix shutdown crash with block producer

gfodor:
Fixes the same issue as https://github.com/bitclout/core/pull/74 with the block producer.


maebeam (approved):
Thanks for fixing this! Been meaning to handle this for a while.

PR #77, opened 2021-07-27
Fix profile vulnerability

maebeam:
Add check to ensure that either (1) profilePublicKey is owned by the txn signer or (2) the signer is a param updater.


bluepartyhat (approved):
Looks good to me!

PR #86, opened 2021-07-31
Properly shutdown block producer

gfodor:
I'm not sure if this is a regression, but it looks like the block producer is now not being shut down on Stop(). (Was this always broken? I had a previous PR which dealt with graceful shutdowns for it, so maybe I was fooling myself that it was being shut down?)


maebeam (approved):

maebeam:
Thanks!

PR #95, opened 2021-08-18
Add a special case for diamondLevel = 0

diamondhands0:


tijno (commented):
seems fine to me but may want go expert to review

out of interest what is the special case for using level 0?

maebeam (approved):

PR #108, opened 2021-08-27
add reserved profile

lazynina:


dgsus:
@diamondhands0 @lazynina why is the dev team reserving profiles?

PR #113, opened 2021-09-08
NFT Transfers & Burns

redpartyhat:


Lux1n:
Thank you for this, this will remove a few major pain points with the current NFT experience!

The hard work is very much appreciated ❤️

lib/block_view.go
```diff
@@ -6290,6 +6503,11 @@ func (bav *UtxoView) _connectUpdateNFT(
 		return 0, 0, nil, RuleErrorCannotUpdateNonExistentNFT
 	}
 
+	// Verify the NFT is not a pending transfer.
+	if prevNFTEntry.IsPending {
```
diamondhands0, line 6507: Do we want to do this check in _connectNFTBid() and _acceptNFTBit() as well or nah?

lib/block_view.go
```diff
@@ -6930,6 +7148,314 @@ func (bav *UtxoView) _connectNFTBid(
...
+	}
+
+	// Force the input to be non-zero so that we can prevent replay attacks.
+	if totalInput == 0 {
+		return 0, 0, nil, RuleErrorNFTTransferRequiresNonZeroInput
+	}
+
+	if verifySignatures {
+		// _connectBasicTransfer has already checked that the transaction is
+		// signed by the top-level public key, which we take to be the poster's
```
diamondhands0, line 7249: s/the poster's public key/the NFT owner's public key

lib/block_view.go
```diff
@@ -6930,6 +7148,314 @@ func (bav *UtxoView) _connectNFTBid(
...
+		// public key.
+	}
+
+	// Now we are ready to transfer the NFT.
+
+	// Create the updated NFTEntry.
+	newNFTEntry := &NFTEntry{
```
diamondhands0, line 7256: Hmm... this seems error-prone compared to setting only the fields that are changing. For example, if we add a new field to NFTEntry then we could forget to initialize it here. Is there a reason you did it this way vs the following?

```
// Make a copy of the previous NFT
newNFTEntry := *prevNFTEntry
// Update the fields that were set during this transfer.
newNFTEntry.OwnerPKID = receiverPKID.PKID
newNFTEntry.UnlockableText = txMeta.UnlockableTex
newNFTEntry.IsPencing = true
```

lib/block_view.go
```diff
@@ -6930,6 +7148,314 @@ func (bav *UtxoView) _connectNFTBid(
...
+	}
+
+	// Force the input to be non-zero so that we can prevent replay attacks.
+	if totalInput == 0 {
+		return 0, 0, nil, RuleErrorAcceptNFTTransferRequiresNonZeroInput
+	}
+
+	if verifySignatures {
+		// _connectBasicTransfer has already checked that the transaction is
+		// signed by the top-level public key, which we take to be the poster's
```
diamondhands0, line 7344: s/poster/NFT owner

lib/block_view.go
```diff
@@ -6930,6 +7148,314 @@ func (bav *UtxoView) _connectNFTBid(
...
+		// public key.
+	}
+
+	// Now we are ready to transfer the NFT.
+
+	// Create the updated NFTEntry (everything the same except for IsPending) and set it.
+	newNFTEntry := &NFTEntry{}
```
diamondhands0, line 7351: I think this works but prefer:

```
// Make a copy of the previous NFT entry
newNFTEntry := *prevNFTEntry
// Set the NFT to no longer be pending
newNFTEntry.IsPending = false
// Grab the pointer here.
bav._setNFTEntryMappings(&newNFTEntry)
```

lib/block_view.go
diamondhands0, line 7344: s/poster/NFT owner

lib/block_view.go
diamondhands0, line 7351: I think this works but prefer:

```
// Make a copy of the previous NFT entry
newNFTEntry := *prevNFTEntry
// Set the NFT to no longer be pending
newNFTEntry.IsPending = false
// Grab the pointer here.
bav._setNFTEntryMappings(&newNFTEntry)
```

diamondhands0, line 7351: I think this works but prefer:

```
// Make a copy of the previous NFT entry
newNFTEntry := *prevNFTEntry
// Set the NFT to no longer be pending
newNFTEntry.IsPending = false
// Grab the pointer here.
bav._setNFTEntryMappings(&newNFTEntry)
```

lib/block_view.go
```diff
@@ -6930,6 +7148,314 @@ func (bav *UtxoView) _connectNFTBid(
...
+	}
+
+	// Force the input to be non-zero so that we can prevent replay attacks.
+	if totalInput == 0 {
+		return 0, 0, nil, RuleErrorBurnNFTRequiresNonZeroInput
+	}
+
+	if verifySignatures {
+		// _connectBasicTransfer has already checked that the transaction is
+		// signed by the top-level public key, which we take to be the poster's
```
diamondhands0, line 7432: s/poster/NFT owner

lib/block_view.go
```diff
@@ -6930,6 +7148,314 @@ func (bav *UtxoView) _connectNFTBid(
...
+		// public key.
+	}
+
+	// Create a backup before we burn the NFT.
+	prevNFTEntry := &NFTEntry{}
+	*prevNFTEntry = *nftEntry
```
diamondhands0, line 7438: Prefer prevNFTEntry := *nftEntry and then use &prevNFTEntry.

lib/block_view.go
```diff
@@ -6930,6 +7148,314 @@ func (bav *UtxoView) _connectNFTBid(
...
+
+	// Delete the NFT.
+	bav._deleteNFTEntryMappings(nftEntry)
+
+	// Save a copy of the previous postEntry and then increment NumNFTCopiesBurned.
+	prevPostEntry := &PostEntry{}
+	*prevPostEntry = *nftPostEntry
```
diamondhands0, line 7445: Same format here. prevPostEntry := *nftPostEntry and then use &prevPostEntry

lib/block_view.go
```diff
@@ -6930,6 +7148,314 @@ func (bav *UtxoView) _connectNFTBid(
...
+	nftPostEntry.NumNFTCopiesBurned++
```
diamondhands0, line 7446: Heh. Nice that this can only be incremented. Makes life easy.

lib/block_view_test.go
```diff
@@ -18967,3 +19210,503 @@ func TestBitCloutDiamondErrorCases(t *testing.T) {
...
+		post2 := DBGetPostEntryByPostHash(db, post2Hash)
+		require.Equal(uint64(1), post2.NumNFTCopiesBurned)
+	}
+
+	// Roll all successful txns through connect and disconnect loops to make sure nothing breaks.
+	_rollBackTestMetaTxnsAndFlush(testMeta)
+	_applyTestMetaTxnsToMempool(testMeta)
+	_applyTestMetaTxnsToViewAndFlush(testMeta)
+	_disconnectTestMetaTxnsFromViewAndFlush(testMeta)
+	_connectBlockThenDisconnectBlockAndFlush(testMeta)
```
diamondhands0, line 19711: Man, what a nice battery of tests we're doing here.

lib/constants.go
```diff
@@ -118,6 +118,10 @@ var (
 	// rather than in creator coin.
 	// Triggers: 3pm PT on 8/16/2021
 	BitCloutDiamondsBlockHeight = uint32(52112)
+
+	// NFTTransfersBlockHeight defines the height at which NFT transfer txns, accept NFT
+	// transfer txns, and NFT burn txns will be accepted.
+	NFTTransferOrBurnBlockHeight = uint32(100000)
```
diamondhands0, line 124: Let's merge with this set to the really high number then do a follow-up PR with it set to whenever we think people can upgrade by. I'm thinking Monday or Tuesday of next week could be good if we merge it quickly.

lib/block_view.go
```diff
@@ -6930,6 +7148,314 @@ func (bav *UtxoView) _connectNFTBid(
...
+		OwnerPKID:                  receiverPKID.PKID,
+		NFTPostHash:                prevNFTEntry.NFTPostHash,
+		SerialNumber:               prevNFTEntry.SerialNumber,
+		IsForSale:                  prevNFTEntry.IsForSale,
+		MinBidAmountNanos:          prevNFTEntry.MinBidAmountNanos,
+		UnlockableText:             txMeta.UnlockableText,
+		LastAcceptedBidAmountNanos: prevNFTEntry.LastAcceptedBidAmountNanos,
+		IsPending:                  true,
+	}
+	bav._setNFTEntryMappings(newNFTEntry)
```
diamondhands0, line 7267: So I think our convention in general is to call _deleteNFTEntryMappings(prevNFTEntry) *and then* call _setNFTEntryMappings(newNFTEntry). Like so:

```
bav._deleteNFTEntryMappings(prevNFTEntry)
bav._setNFTEntryMappings(newNFTEntry)
```

I think it works as-is, but doing it this way future-proofs us in the event we create additional indexes in the view. For example, if we at some point created a mapping that used OwnerPKID as a key, then we would need to call _delete first to remove the old mapping.

I would do this in all the _connect *and* _disconnect functions btw.

diamondhands0, line 7267: So I think our convention in general is to call _deleteNFTEntryMappings(prevNFTEntry) *and then* call _setNFTEntryMappings(newNFTEntry). Like so:

```
bav._deleteNFTEntryMappings(prevNFTEntry)
bav._setNFTEntryMappings(newNFTEntry)
```

I think it works as-is, but doing it this way future-proofs us in the event we create additional indexes in the view. For example, if we at some point created a mapping that used OwnerPKID as a key, then we would need to call _delete first to remove the old mapping.

I would do this in all the _connect *and* _disconnect functions btw.

lib/block_view.go
redpartyhat: I don't see any harm in making things more strict.  Added.

lib/block_view.go
redpartyhat: No reason, just copied it from _connectCreateNFT.  Changed.

lib/block_view.go
redpartyhat: I like it. Cleaner.

redpartyhat: Appears to be a duplicate comment.  Resolving.

redpartyhat:
All comments have been addressed.  I also added an "rph/nft-transfers" branch for backend, frontend, and identity. These branches were used to test the changes locally and validate the new transaction types worked as expected.

diamondhands0 (approved):

superzordon (approved):

PR #120, opened 2021-09-22
Ln/native video

lazynina:
Updates the DeSoBodySchema to add VideoURLs.

Also omits attributes of the DesoBodySchema if they are empty - this will save a few KBs if a post doesn't have an image or video.


lib/network.go
```diff
@@ -857,11 +857,12 @@ func (msg *MsgDeSoGetBlocks) String() string {
...
 
-// Within a post, the body typically has a particular
+// DeSoBodySchema Within a post, the body typically has a particular
 // schema defined below.
 type DeSoBodySchema struct {
-	Body      string
-	ImageURLs []string
+	Body      string `json:",omitempty"`
+	ImageURLs []string `json:",omitempty"`
+	VideoURLs []string `json:",omitempty"`
```
lazynina, line 865: I think this change is acceptable, but would like other core devs (@redpartyhat, @diamondhands0 , @maebeam ) to weigh in. Omitting VideoURLs when they are empty allow us to avoid a fork as existing posts can still be marshaled and unmarshaled in the same fashion since it won't expect the VideoURLs.

I think adding omitempty to the other attributes is okay as well and will save us a little space in the long run as we won't have keys in body if there is no value.

lib/block_view_test.go
```diff
@@ -15521,7 +15521,7 @@ func TestNFTBasic(t *testing.T) {
 	{
 		// Balance before.
 		m0BalBeforeNFT := _getBalance(t, chain, nil, m0Pub)
-		require.Equal(uint64(27), m0BalBeforeNFT)
+		require.Equal(uint64(28), m0BalBeforeNFT)
```
redpartyhat, line 15524: @lazynina, why did all these balance values change? I can't figure it out...

lazynina: See this comment: https://github.com/deso-protocol/core/pull/120#discussion_r714284576

I added a new attribute to the body schema and was going to only add omitempty for the new VideoURLs attribute to maintain backwards compatibility. However, if we add omitempty to the Body and ImageURLs as well, we can save  a KB per post

redpartyhat: Oooooh, of course! I was misreading this change.  I thought the posts were costing *more* but they are actually costing *less*, which makes sense.  Awesome.

lib/network.go
redpartyhat: So, just to clarify my understanding... I don't think adding VideoURLs to the DeSoBodySchema would actually cause a fork with existing forks.  If Golang tries to unmarshal a JSON string that is missing a key, it just initializes the struct with the default value.  Therefore, old posts wouldn't actually break.  See here for an example of unmarshalling a JSON string that is missing a key: https://play.golang.org/p/e4QnZgGpsh5. Let me know if I am missing something and it would in fact cause a fork.  

That being said, I see no reason not to omit empty values when storing these JSON blobs on the chain! This change LGTM.

maebeam (approved):
lgtm!

lib/network.go
lazynina: Okay that makes sense - my understanding of how it would marshal when missing a key was incorrect. If you're good with this change, then let's do it! I synced a whole node from scratch with this update, so it won't mess up the existing chain.

redpartyhat (approved):

PR #123, opened 2021-09-28
Fix column names in psql

kanshi:
The recent renames from BitClout to DeSo missed some '_'s in the migration scripts. Here is a fix I used for that on love4src.


maebeam:
Thanks!

PR #131, opened 2021-10-22
Allow pending NFT sales (FORKING CHANGE)

redpartyhat:
As it stands, users that want to put transferred NFTs up for sale must first "accept" the NFT transfer.  This is cumbersome and does not add any obvious utility. This PR removes that restriction, automatically accepting an NFT transfer when it is put up for sale.

In drafting this change, I took inventory of all NFT transactions and noted those that are affected by the "IsPending" status of an NFT.  This is the current state of NFT consensus:

Unaffected NFT Transaction:
  - CreateNFT
  - BurnNFT
  - NFTTransfer (one can already transfer a pending NFT)

Affected NFT Transactions:
  - NFTBid -> Can't bid on a pending NFT.
  - AcceptNFTBid -> Can't accept a bid on a pending NFT.
  - UpdateNFT -> You can't update an NFT that is a pending transfer.
  - AcceptNFTTransfer -> Sanity check's that an NFT is not for sale before accepting a transfer.  Therefore, if an NFT was for sale *and* pending, one could not accept it.  Since it is unreasonable to create a state where the user cannot accept a transfer, this must be avoided.

Given the current state of consensus, the best solution is to force a user to accept a transfer if they update an NFT.  This prevents us from being in a situation where an NFT is for sale and pending while keeping all other logic working as expected.

### DIFF
--- a/lib/block_view.go
+++ b/lib/block_view.go
@@ -6820,9 +6820,13 @@ func (bav *UtxoView) _connectUpdateNFT(
 		return 0, 0, nil, RuleErrorCannotUpdateNonExistentNFT
 	}
 
-	// Verify the NFT is not a pending transfer.
-	if prevNFTEntry.IsPending {
-		return 0, 0, nil, RuleErrorCannotUpdatePendingNFTTransfer
+	// Prior to the "AllowPendingNFTSalesBlockHeight", one could not update (ie. put for sale)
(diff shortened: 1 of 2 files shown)


maebeam (approved):
I like it!

lazynina (approved):
LGTM

Issue #136, opened 2021-11-01
Invalid transaction type in block 74069 - SWAP_IDENTIY (missing second T)


starykolarz:
Introduced here:
https://github.com/deso-protocol/core/commit/a4bbfa5c83aaa2585ca8f91a8487618828e489fc#diff-a5a5f778c343d89a36db05b482345198caeb9d6aa30bd34685c76e73e934f22e

Tx data:
        {
     "TransactionIDBase58Check": "3JuESvYxAcPXMoCRN3StmLPHKp2hkdSwnDQHgkK1CdKhdsq8LNYfFy",
     "RawTransactionHex": "01af7e5c2406ac513fbda473e5c03eb995ca8d4178db22ba0c03b190a855f7ff720001021134b0ccce10a8316ffb868fa86418fdd52c051dbf73bd722a8fb9a375cfebb58da0d8050c442103576d48b30347170e00b784aa1bf1c0312405c0fc0b0280549ff365a41c73145421035808ea17bddcad4915e2b6ad021bb07a22fa08247a433e0ce311ca6931baf64421021134b0ccce10a8316ffb868fa86418fdd52c051dbf73bd722a8fb9a375cfebb5004630440220776baef173baaca82a8b04bfef6f93ef84ddfed8c02ce9322ae5ba6e6a335e52022021ccf380fbdf1fd31e6248ec6032a00f9b1c02f611d9163b74e767d0186b8917",
     "Inputs": [         {
        "TransactionIDBase58Check": "3JuEU54U7jehFcQp5UDvuYcARbUKHwjKNUR4yU7pKdLLfLw7dSu7sB",
        "Index": 0
     }],
     "Outputs": [         {
        "PublicKeyBase58Check": "BC1YLfoSyJWKjHGnj5ZqbSokC3LPDNBMDwHX3ehZDCA3HVkFNiPY5cQ",
        "AmountNanos": 11931661
     }],
     "SignatureHex": "30440220776baef173baaca82a8b04bfef6f93ef84ddfed8c02ce9322ae5ba6e6a335e52022021ccf380fbdf1fd31e6248ec6032a00f9b1c02f611d9163b74e767d0186b8917",
     "TransactionType": "SWAP_IDENTIY",
     "BlockHashHex": "0000000000006be449a09dbc93675c597934c4be3bf0e762185e4a11183dd738",
     "TransactionMetadata":          {
[...]

PR #137, opened 2021-11-01
Typo in Swap Identity Txt String

tijno:
Note this invalid transaction type is now on chain - not sure whether that will cause issues syncing - my nodes seem to be in sync. 

See #136

@maebeam


tijno:
see #138

closing this

maebeam:
Thanks for the report + PR

PR #141, opened 2021-11-06
Historical balance lookup support

maebeam:


lib/blockchain.go
```diff
@@ -2306,14 +2302,8 @@ func (bc *Blockchain) ProcessBlock(desoBlock *MsgDeSoBlock, verifySignatures boo
...
-	if bc.server != nil {
-		go func() {
-			bc.server.incomingMessages <- &ServerMessage{
-				Msg: &MsgDeSoBlockAccepted{
-					block: desoBlock,
-				},
-			}
-		}()
+	if bc.eventManager != nil {
+		bc.eventManager.blockAccepted(&BlockEvent{Block: desoBlock})
```
diamondhands0, line 2306: I think this will cause the ProcessBlock to hang and/or run slower because it's not running it in a goroutine.

PR #142, opened 2021-11-08
Buy Now NFTs (FORKING CHANGE)

lazynina:
Add support for selling an NFT at a fixed price instead of an auction style.

What's been done:
1. Move the logic of connecting/disconnecting NFT sales out of `_connectAcceptNFTBid` and `_disconnectAcceptNFTBid` in block_view.go into a helper function that can be used by `_connectNFTBid` and `_disconnectNFTBid` when an NFT is a buy now NFT.
2. Update `_connectNFTBid` and `_disconnectNFTBid` to automatically sell an NFT when a bid is submitted on an NFT that is a Buy Now NFT.
3. Prevent making an NFT a `Buy Now` NFT if the NFT has unlockable content.
4. Prevent putting an NFT on sale as a `Buy Now` NFT with a Purchase price of 0.
5. Add rosetta fields to UtxoOperation struct to support the Output to the coin that is generated by this transaction.
6.  Create a utility function that gathers enough inputs to cover a certain amount of spend. This helps gather enough inputs for an NFT bid on a Buy Now NFT - similar to how we gather additional inputs for Creator Coin purchases.
7. Add some new Errors
8. Add test for Buy Now functionality.


lib/block_view.go
```diff
@@ -2715,8 +2736,7 @@ func (bav *UtxoView) _disconnectAcceptNFTBid(
 
 	// (1) Set the old NFT entry.
 	if operationData.PrevNFTEntry == nil || operationData.PrevNFTEntry.isDeleted {
-		return fmt.Errorf("_disconnectAcceptNFTBid: prev NFT entry doesn't exist; " +
-			"this should never happen")
+		return fmt.Errorf("prev NFT entry doesn't exist; this should never happen")
```
redpartyhat, line 2739: In core, errors are generally prefixed with the name of the function returning the error in order to make it easier to track them down (since you basically always have a pseudo-stack trace).  I would add "_helpDisconnectNFTSold: " to the error messages here for consistency.

lib/block_view.go
```diff
@@ -2737,27 +2756,28 @@ func (bav *UtxoView) _disconnectAcceptNFTBid(
...
-	if operationData.NFTSpentUtxoEntries == nil || len(operationData.NFTSpentUtxoEntries) == 0 {
-		return fmt.Errorf("_disconnectAcceptNFTBid: NFTSpentUtxoEntries was nil; " +
-			"this should never happen")
-	}
-	// Note: these UTXOs need to be unspent in reverse order.
-	for ii := len(operationData.NFTSpentUtxoEntries) - 1; ii >= 0; ii-- {
-		spentUtxoEntry := operationData.NFTSpentUtxoEntries[ii]
-		if err := bav._unSpendUtxo(spentUtxoEntry); err != nil {
-			return errors.Wrapf(err, "_disconnectAcceptNFTBid: Problem unSpending utxo %v: ", spentUtxoEntry)
+	if !isNFTBid {
```
redpartyhat, line 2769: You can actually just check the UtxoOperation type here.  I would either kill the "isNFTBid" function parameter or add a sanity check to ensure that the UtxoOperation matches the function caller's expectation (ie. make sure that they didn't set isNFTBid to false when the UtxoOperation has OperationTypeNFTBid).  It would also be good to add a nice comment above this if statement to explain why bidder UTXOs don't need to be reverted in the buy-it-now NFT bid case.

redpartyhat: It also seems good to add an else statement with some sanity checks: (1) sanity check that the operation doesn't have any NFTSpentUtxoEntries if it is an NFT bid, (2) sanity check that the NFT entry has "IsBuyNow" set to true (and you might want to sanity check the opposite case as well).

lib/block_view.go
```diff
@@ -7039,11 +7131,250 @@ func (bav *UtxoView) _connectAcceptNFTBid(
...
+	nftKey := MakeNFTKey(txMeta.NFTPostHash, txMeta.SerialNumber)
+	prevNFTEntry := bav.GetNFTEntryForNFTKey(&nftKey)
+	// Get the post entry, verify it exists.
+	nftPostEntry := bav.GetPostEntryForPostHash(txMeta.NFTPostHash)
+
 	existingProfileEntry := bav.GetProfileEntryForPublicKey(nftPostEntry.PosterPublicKey)
 	if existingProfileEntry == nil || existingProfileEntry.isDeleted {
 		return 0, 0, nil, fmt.Errorf(
-			"_connectAcceptNFTBid: Profile missing for NFT pub key: %v %v",
+			"Profile missing for NFT pub key: %v %v",
```
redpartyhat, line 7377: Same comment as with _helpDisconnectNFTSold.  Seems no lose to add the function name to the error strings here.

lib/block_view.go
```diff
@@ -6881,6 +6956,22 @@ func (bav *UtxoView) _connectUpdateNFT(
 			txMeta.NFTPostHash.String())
 	}
 
+	if isBuyNow && blockHeight < BuyNowNFTBlockHeight {
```
redpartyhat, line 6959: The current way the code is written, we accept any "IsBuyNowKey" as being legit even if we have not passed the BuyNowNFTBlockHeight.  This means that someone could see this PR and add the IsBuyNowKey to extra data to a transaction right now in order to break nodes that sync from scratch when this update is merged in the future. Instead, it seems better to wrap the code above that checks for the IsBuyNowKey in extra data in a blockheight check.  That way, "isBuyNow" is forced to be false until the BuyNowNFTBlockHeight hits.

lib/block_view.go
```diff
@@ -6881,6 +6956,22 @@ func (bav *UtxoView) _connectUpdateNFT(
...
+		return 0, 0, nil, errors.Wrapf(RuleErrorBuyNowNFTBeforeBlockHeight, "_connectUpdateNFT: ")
+	}
+
+	// We only need to make these checks if we are past the Buy Now NFT Block Height
+	if blockHeight >= BuyNowNFTBlockHeight {
```
redpartyhat, line 6964: Nit: It seems better to reduce complexity and just run these checks regardless of the block height.

lib/block_view.go
```diff
@@ -7349,53 +7688,73 @@ func (bav *UtxoView) _connectAcceptNFTBid(
...
+			// Rosetta fields.
+			AcceptNFTBidCreatorPublicKey: nftPostEntry.PosterPublicKey,
+			AcceptNFTBidBidderPublicKey: bidderPublicKey,
+			AcceptNFTBidCreatorRoyaltyNanos: creatorCoinRoyaltyNanos,
+		})
+	}
+
+	if txMeta.IsNFTBid {
+		// Add an operation to the list at the end indicating we've connected an NFT bid.
+		utxoOpsForTxn = append(utxoOpsForTxn, &UtxoOperation{
```
redpartyhat, line 7711: It seems error prone to duplicate so much code here.  We should build the part of the UtxoOperation that is consistent regardless of IsAcceptNFTBid vs. IsNFTBid and then selectively add the Rosetta fields with if/else if.

redpartyhat: You also might want to sanity check that IsAcceptNFTBid and IsNFTBid are never true at the same time.

lib/mempool.go
```diff
@@ -1451,18 +1452,28 @@ func ComputeTransactionMetadata(txn *MsgDeSoTxn, utxoView *UtxoView, blockHash *
 		realTxMeta := txn.TxnMeta.(*NFTBidMetadata)
 		_ = realTxMeta
 
+		isBuyNow := false
+		// Check if the Bid is on a Buy Now NFT.
+		if val, exists := txn.ExtraData[IsBuyNowKey]; exists && reflect.DeepEqual(val, []byte{BoolToByte(true)}) {
```
redpartyhat, line 1457: Hmmmm, since Rosetta depends on this, maybe we should block height restrict this extradata lookup? Just to be extra safe? This would align with the change I proposed in _connectUpdateNFT().

lib/block_view.go
```diff
@@ -6768,6 +6821,21 @@ func (bav *UtxoView) _connectCreateNFT(
 	if postEntry.IsNFT {
 		return 0, 0, nil, RuleErrorCreateNFTOnPostThatAlreadyIsNFT
 	}
+	if isBuyNow && blockHeight < BuyNowNFTBlockHeight {
```
redpartyhat, line 6824: We should remove this check and instead blockHeight restrict the chunk of code that gets the IsBuyNowKey from extraData.

lib/block_view.go
```diff
@@ -6768,6 +6821,21 @@ func (bav *UtxoView) _connectCreateNFT(
...
+		return 0, 0, nil, errors.Wrapf(RuleErrorBuyNowNFTBeforeBlockHeight, "_connectCreateNFT: ")
+	}
+	// We only need to make these checks if we are past the BuyNowNFTBlockHeight
+	if blockHeight >= BuyNowNFTBlockHeight {
```
redpartyhat, line 6828: Similar to the _connectUpdateNFT() comment, it seems better to reduce complexity and remove this if statement / always run these checks.

lib/block_view.go
```diff
@@ -7039,11 +7131,250 @@ func (bav *UtxoView) _connectAcceptNFTBid(
...
+	VerifySignatures bool
+
+	// These both shouldn't be true
+	IsAcceptNFTBid   bool
+	IsNFTBid         bool
+}
+
+func (bav *UtxoView) _helpConnectNFTSold(txMeta HelpConnectNFTSoldStruct) (
+	_totalInput uint64, _totalOutput uint64, _utxoOps []*UtxoOperation, _err error) {
+	if txMeta.IsAcceptNFTBid && txMeta.IsNFTBid || !txMeta.IsAcceptNFTBid && !txMeta.IsNFTBid {
```
redpartyhat, line 7366: A simpler way to express this check is: "if txMeta.IsAcceptNFTBid == txMeta.IsNFTBid".

lib/block_view.go
```diff
@@ -7205,12 +7543,14 @@ func (bav *UtxoView) _connectAcceptNFTBid(
...
 
 	newNFTEntry := &NFTEntry{
-		LastOwnerPKID:  updaterPKID.PKID,
+		LastOwnerPKID:  prevNFTEntry.OwnerPKID,
 		OwnerPKID:      txMeta.BidderPKID,
 		NFTPostHash:    txMeta.NFTPostHash,
 		SerialNumber:   txMeta.SerialNumber,
 		IsForSale:      false,
 		UnlockableText: txMeta.UnlockableText,
+		// TODO: We automatically flip IsBuyNow to false. Otherwise, someone could buy this NFT from them?
```
redpartyhat, line 7552: Makes sense to me!

lib/mempool.go
redpartyhat, line 1457: It seems a bit fragile that TxIndex relies on the IsBuyNowKey being in the ExtraData of NFTBid's in order to determine if they are buy now bids while _connectNFTBid() relies on the IsBuyNow param on the NFT Entry.  I suggest we get rid of the extradata on the NFTBid and instead check to see if the NFTEntry that was bid on is now owned by the bidder.  Wdyt?

lib/blockchain.go
```diff
@@ -3787,6 +3807,38 @@ func (bc *Blockchain) AddInputsAndChangeToTransactionWithSubsidy(
...
+		nftKey := MakeNFTKey(txMeta.NFTPostHash, txMeta.SerialNumber)
+		nftEntry := utxoView.GetNFTEntryForNFTKey(&nftKey)
+
+		if nftEntry != nil && nftEntry.IsBuyNow {
+			spendAmount += txMeta.BidAmountNanos
+			if txArg.ExtraData == nil {
+				txArg.ExtraData = make(map[string][]byte)
+			}
+			// We also set IsBuyNow to true in extra data
+			txArg.ExtraData[IsBuyNowKey] = []byte{BoolToByte(true)}
```
redpartyhat, line 3837: I suggest that we get ride of this extra data on the NFT bid.  See my reasoning in the comment on ComputeTransactionMetadata() for NFTBids.

lib/block_view.go
```diff
@@ -7039,11 +7131,250 @@ func (bav *UtxoView) _connectAcceptNFTBid(
...
+		bav._setNFTBidEntryMappings(newBidEntry)
+		// Okay here's where the fun happens. We are submitting a bid on a Buy Now enabled NFT. We create the bid then we call the
+		// _helpConnectNFTSold to handle the royalty payout logic and such.
+		totalInput, totalOutput, utxoOpsForTxn, err := bav._helpConnectNFTSold(HelpConnectNFTSoldStruct{
+			NFTPostHash: txMeta.NFTPostHash,
+			SerialNumber: txMeta.SerialNumber,
+			BidderPKID: bidderPKID.PKID,
+			BidAmountNanos: txMeta.BidAmountNanos,
+
+			BidderInputs: []*DeSoInput{},//txMeta.BidderInputs,
```
redpartyhat, line 7325: Nit: kill comment.

lib/block_view.go
```diff
@@ -7039,11 +7131,250 @@ func (bav *UtxoView) _connectAcceptNFTBid(
...
+
+		// Add an operation to the list at the end indicating we've connected an NFT bid.
+		utxoOpsForTxn = append(utxoOpsForTxn, &UtxoOperation{
+			Type:            OperationTypeNFTBid,
+			PrevNFTBidEntry: prevNFTBidEntry,
+		})
+
+		return totalInput, totalOutput, utxoOpsForTxn, nil
+	} else {
+		newBidEntry := &NFTBidEntry{
```
redpartyhat, line 7310: It's sort of weird that we create an NFT bid entry only to immediately delete it later in _helpConnectNFTSold().  I  assumed we would skip the nftBidEntry and just immediately update the nftEntry for the buyer.  I think this is fine though, and certainly works / streamlines the code well.  Maybe just add a quick comment that explains that the bid entry created here is a temporary placeholder and immediately deleted when we call _helpConnectNFTSold()?

lib/block_view.go
```diff
@@ -7166,18 +7504,18 @@ func (bav *UtxoView) _connectAcceptNFTBid(
...
+			"sum of royalties (%d, %d) is less than bid amount (%d)",
 			creatorRoyaltyNanos, creatorCoinRoyaltyNanos, txMeta.BidAmountNanos)
 	}
 
 	bidAmountMinusRoyalties := txMeta.BidAmountNanos - creatorRoyaltyNanos - creatorCoinRoyaltyNanos
-
+	txn, txHash, verifySignatures := txMeta.Txn, txMeta.TxHash, txMeta.VerifySignatures
 	// Connect basic txn to get the total input and the total output without
 	// considering the transaction metadata.
 	totalInput, totalOutput, utxoOpsFromBasicTransfer, err := bav._connectBasicTransfer(
```
redpartyhat, line 7515: BUG ALERT: We are not currently checking to see if there was sufficient output to cover a "buy it now" NFT sale.  Right now, one could submit just enough inputs to cover the cost of the txn bytes but not enough inputs to cover the cost of their BuyItNow bid and still have the transaction succeed (except for the fact that the "hardcore balance check" safety net code would probably catch it).  If this is a buy it now NFT, we need to check that totalInput is greater than totalOutput + bid amount.  See the code around "RuleErrorCreatorCoinTxnOutputExceedsInput" for an example.

lib/mempool.go
lazynina: seems reasonable to me!

lib/block_view.go
lazynina: yep - will add a comment. Basically creating this bid here allows us to reuse the code that is now in _helpConnectNFTBid

lib/block_view.go
```diff
@@ -2737,27 +2756,49 @@ func (bav *UtxoView) _disconnectAcceptNFTBid(
...
-	if operationData.NFTSpentUtxoEntries == nil || len(operationData.NFTSpentUtxoEntries) == 0 {
-		return fmt.Errorf("_disconnectAcceptNFTBid: NFTSpentUtxoEntries was nil; " +
-			"this should never happen")
-	}
-	// Note: these UTXOs need to be unspent in reverse order.
-	for ii := len(operationData.NFTSpentUtxoEntries) - 1; ii >= 0; ii-- {
-		spentUtxoEntry := operationData.NFTSpentUtxoEntries[ii]
-		if err := bav._unSpendUtxo(spentUtxoEntry); err != nil {
-			return errors.Wrapf(err, "_disconnectAcceptNFTBid: Problem unSpending utxo %v: ", spentUtxoEntry)
+	// We do not need to revert bidder UTXOs if this is an NFT Bid on a Buy Now NFT, because the bidder inputs are specified
```
diamondhands0, line 2769: Good comment.

lib/block_view.go
```diff
@@ -2737,27 +2756,49 @@ func (bav *UtxoView) _disconnectAcceptNFTBid(
...
+	// as transaction inputs as opposed to bidder inputs that are specified in the transaction metadata since the transactor
+	// and the bidder are the same in this scenario.
+	if operationData.Type != OperationTypeNFTBid {
```
diamondhands0, line 2772: Nit: I think it's safer to write this as:
```
if operationData.Type == OperationTypeAcceptNFTBid {
...
} else if operationData.Type == OperationTypeNFTBic {
...
} else {
// error because we got a weird type
}
```

lib/block_view.go
```diff
@@ -2737,27 +2756,49 @@ func (bav *UtxoView) _disconnectAcceptNFTBid(
...
+		// Note: these UTXOs need to be unspent in reverse order.
+		for ii := len(operationData.NFTSpentUtxoEntries) - 1; ii >= 0; ii-- {
+			spentUtxoEntry := operationData.NFTSpentUtxoEntries[ii]
+			if err := bav._unSpendUtxo(spentUtxoEntry); err != nil {
+				return errors.Wrapf(err, "_helpDisconnectNFTSold: Problem unSpending utxo %v: ", spentUtxoEntry)
+			}
+		}
+	} else {
+		// Check that there are no NFTSpentUtxoEntries.
+		if operationData.NFTSpentUtxoEntries != nil && len(operationData.NFTSpentUtxoEntries) > 0 {
```
diamondhands0, line 2793: Nit: `if len(operationData.NFTSpentUtxoEntries) > 0 {` is sufficient because `len(nil)` returns zero.

lib/block_view.go
```diff
@@ -2815,25 +2851,58 @@ func (bav *UtxoView) _disconnectNFTBid(
...
 	}
-	nftBidKey := MakeNFTBidKey(bidderPKID.PKID, txMeta.NFTPostHash, txMeta.SerialNumber)
-	nftBidEntry := bav.GetNFTBidEntryForNFTBidKey(&nftBidKey)
-	// Sanity-check that it exists.
-	if nftBidEntry == nil || nftBidEntry.isDeleted {
-		return fmt.Errorf("_disconnectNFTBid: Bid entry for "+
-			"nftBidKey %v doesn't exist; this should never happen", nftBidKey)
-	}
+	// If there isn't a previous NFT entry, we're dealing with a bid on a non-BuyNow NFT.
+	if operationData.PrevNFTEntry == nil {
```
diamondhands0, line 2872: Nit: Feels like we should have a more robust/explicit way of knowing whether this bid was on a BuyNow NFT vs on a normal NFT. I could see us setting PrevNFTEntry for other reasons in the future, which would cause an unexpected bug to occur here if we don't make this if statement a little more narrow.

diamondhands0: Maybe you reverse the check and do:
```
if operationData.PrevNFTEntry != nil && operationData.PrevNFTEntry.IsBuyNow {
  // In this case, the bid was on a buy-now NFT, which means that we need to ...
} else {
  // In this case, we know that we're dealing with a non-buy-now NFT...
}
```

lib/block_view.go
```diff
@@ -2815,25 +2851,58 @@ func (bav *UtxoView) _disconnectNFTBid(
...
+			return fmt.Errorf("_disconnectNFTBid: Bid entry for "+
+				"nftBidKey %v doesn't exist; this should never happen", nftBidKey)
+		}
 
-	// If a previous entry exists, set it.
-	if operationData.PrevNFTBidEntry != nil {
-		bav._setNFTBidEntryMappings(operationData.PrevNFTBidEntry)
+		// Delete the existing NFT bid entry.
+		bav._deleteNFTBidEntryMappings(nftBidEntry)
 	}
```
diamondhands0, line 2906: Feels like we can consolidate this block a bit:

```
if operationData.PrevNFTEntry != nil {
  if !operationData.PrevNFTEntry.IsBuyNow {
    // Error since this should never happen
  }
  // If we get here, we know this bid was on a buy-now NFT, and that the underlying NFT
  // was sold outright to the bidder.

  // If there is a previous NFT entry, that means the NFT entry was modified and that only happens for Buy Now NFTs when bidding.
  if err := bav._helpDisconnectNFTSold(operationData, txMeta.NFTPostHash); err != nil {
    return errors.Wrapf(err, "_disconnectNFTBid: ")
  }
}
// The steps after this point apply irrespective of whether the bid was on a buy-now NFT.

nftBidKey := MakeNFTBidKey(bidderPKID.PKID, txMeta.NFTPostHash, txMeta.SerialNumber)
nftBidEntry := bav.GetNFTBidEntryForNFTBidKey(&nftBidKey)
// Sanity-check that it exists.
if nftBidEntry == nil || nftBidEntry.isDeleted {
  return fmt.Errorf("_disconnectNFTBid: Bid entry for "+
    "nftBidKey %v doesn't exist; this should never happen", nftBidKey)
}
// Delete the existing NFT bid entry.
bav._deleteNFTBidEntryMappings(nftBidEntry)

// If a previous entry exists, set it. This will only be the case for non-buy-now NFTs.
if operationData.PrevNFTBidEntry != nil {
  // Sanity-check that there is no PrevNFTEntry
  if operationData.PrevNFTEntry != nil {
      // error
  }

  bav._setNFTBidEntryMappings(operationData.PrevNFTBidEntry)
}
```

lib/block_view.go
```diff
@@ -6768,6 +6843,16 @@ func (bav *UtxoView) _connectCreateNFT(
...
+
+	// We can't encrypt unlockable content if Buy Now is enabled.
+	if txMeta.HasUnlockable && isBuyNow {
+		return 0, 0, nil, errors.Wrapf(RuleErrorCannotHaveUnlockableAndBuyNowNFT, "_connectCreateNFT: ")
+	}
+	// We can't sell an NFT for 0.
+	if isBuyNow && txMeta.MinBidAmountNanos == 0 {
+		return 0, 0, nil, errors.Wrapf(RuleErrorBuyNowMustHaveMinBidAmountNanos, "_connectCreateNFT: ")
+	}
```
diamondhands0, line 6854: Quick question: What does this break? It seems fine to set an NFT up for sale to the first person who bids. I would let the frontends handle the validation around this.

lib/block_view.go
```diff
@@ -6881,6 +6974,15 @@ func (bav *UtxoView) _connectUpdateNFT(
 			txMeta.NFTPostHash.String())
 	}
 
+	// We only need to make these checks if we are past the Buy Now NFT Block Height
+	if postEntry.HasUnlockable && isBuyNow {
+		return 0, 0, nil, errors.Wrapf(RuleErrorCannotHaveUnlockableAndBuyNowNFT, "_connectUpdateNFT: ")
+	}
+
+	if isBuyNow && txMeta.MinBidAmountNanos == 0 {
```
diamondhands0, line 6982: Again, unless there's a compelling reason to enforce this check in consensus, I would let the frontends do this validation.

lib/block_view.go
```diff
@@ -7039,11 +7142,253 @@ func (bav *UtxoView) _connectAcceptNFTBid(
...
+	// that are used by future transactions.
+	BidderInputs []*DeSoInput
+
+	BlockHeight      uint32
+	Txn              *MsgDeSoTxn
+	TxHash           *BlockHash
+	VerifySignatures bool
+
+	// These both shouldn't be true
+	IsAcceptNFTBid   bool
```
diamondhands0, line 7374: The IsNFTBid and IsAcceptNFTBid fields are super confusing. It took me a while to understand the if statement right below this. Can we just use the TxnType to switch on this or something and remove these fields?

lib/block_view.go
```diff
@@ -7039,11 +7142,253 @@ func (bav *UtxoView) _connectAcceptNFTBid(
...
+	IsNFTBid         bool
+}
+
+func (bav *UtxoView) _helpConnectNFTSold(txMeta HelpConnectNFTSoldStruct) (
+	_totalInput uint64, _totalOutput uint64, _utxoOps []*UtxoOperation, _err error) {
+	if txMeta.IsAcceptNFTBid == txMeta.IsNFTBid {
```
diamondhands0, line 7380: Nit: I wouldn't call this txMeta. I would call it args since we use txMeta to mean something different in all other contexts.

I would also make args.Type be the thing that you use instead of having two mutually exclusive bools, as mentioned in the comment above.

lib/block_view.go
```diff
@@ -7039,11 +7142,253 @@ func (bav *UtxoView) _connectAcceptNFTBid(
...
+	}
+
+	// Validate the nftEntry.  Note that there is a special case where a bidder can submit a bid
+	// on SerialNumber zero.  This acts as a blanket bid on any serial number version of this NFT
+	// As a result, the nftEntry will be nil and should not be validated.
+	nftKey := MakeNFTKey(txMeta.NFTPostHash, txMeta.SerialNumber)
+	nftEntry := bav.GetNFTEntryForNFTKey(&nftKey)
+	bidderPKID := bav.GetPKIDForPublicKey(txn.PublicKey)
+	if bidderPKID == nil || bidderPKID.isDeleted {
+		return 0, 0, nil, fmt.Errorf("_connectNFTBid: PKID for bidder public key %v doesn't exist; this should never happen", string(txn.PublicKey))
```
diamondhands0, line 7217: Nit: Keep lines under 80 characters long. GoLand has a guide built into it by default that helps you know when you've hit this limit. Makes things a lot easier to read in IDE or terminal.

lib/block_view.go
```diff
@@ -7039,11 +7142,253 @@ func (bav *UtxoView) _connectAcceptNFTBid(
 			len(txMeta.UnlockableText), bav.Params.MaxPrivateMessageLengthBytes)
 	}
 
-	// Get the poster's profile.
+	// Verify the NFT bid entry being accepted exists and has a bid consistent with the metadata.
+	// If we did not require an AcceptNFTBid txn to have a bid amount, it would leave the door
+	// open for an attack where someone replaces a high bid with a low bid after the owner accepts.
+	nftBidKey := MakeNFTBidKey(txMeta.BidderPKID, txMeta.NFTPostHash, txMeta.SerialNumber)
```
diamondhands0, line 7148: Am I missing something or is this redundant? It seems like you're doing this exact same check in the _helpConnectNFTSold function. So maybe you can delete this from here?

lib/block_view.go
```diff
@@ -7063,89 +7408,96 @@ func (bav *UtxoView) _connectAcceptNFTBid(
...
-	for _, bidderInput := range txMeta.BidderInputs {
-		bidderUtxoKey := UtxoKey(*bidderInput)
-		bidderUtxoEntry := bav.GetUtxoEntryForUtxoKey(&bidderUtxoKey)
-		if bidderUtxoEntry == nil || bidderUtxoEntry.isSpent {
-			return 0, 0, nil, RuleErrorBidderInputForAcceptedNFTBidNoLongerExists
-		}
+	// We only need to validate the bidder UTXOs when connecting an AcceptNFTBid transaction since the transactor and
+	// the bidder are different users.  For NFTBid transactions on Buy Now NFTs, there are additional inputs to cover
+	// the bid amount. We do not need to make explicitly make change for the bidder in that situation either.
+	if txMeta.IsAcceptNFTBid {
```
diamondhands0, line 7455: This would be `if args.TxnType == TxnTypeAcceptNFTBid`

lib/block_view.go
diamondhands0: +1. I would put this check in a secon dif statement after the IsAcceptNFT one:
```
if args.TxnType == TxnTypeAcceptNFTBid {
  // As you have it now
} else if args.TxnType == TxnTypeNFTBid {
  // Do the checks for the buy-now NFT
} else {
  // error unexpected txn type
}
```

diamondhands0: Also don't forget about overflows :)

lib/block_view.go
```diff
@@ -7166,25 +7518,54 @@ func (bav *UtxoView) _connectAcceptNFTBid(
...
 	if txMeta.BidAmountNanos < (creatorRoyaltyNanos + creatorCoinRoyaltyNanos) {
 		return 0, 0, nil, fmt.Errorf(
-			"_connectAcceptNFTBid: sum of royalties (%d, %d) is less than bid amount (%d)",
+			"_helpConnectNFTSold: sum of royalties (%d, %d) is less than bid amount (%d)",
 			creatorRoyaltyNanos, creatorCoinRoyaltyNanos, txMeta.BidAmountNanos)
 	}
 
 	bidAmountMinusRoyalties := txMeta.BidAmountNanos - creatorRoyaltyNanos - creatorCoinRoyaltyNanos
-
+	txn, txHash, verifySignatures := txMeta.Txn, txMeta.TxHash, txMeta.VerifySignatures
```
diamondhands0, line 7526: Nit: I don't like relabels like this. They typically make the code harder to read. I would instead just rename these vars in all the places below.

lib/block_view.go
diamondhands0: Oh ok I see you added it below. Seems fine, but I would prefer if it was consolidated into the if statement above just to make it so that the checks are in the same place logically.

lib/block_view.go
```diff
@@ -7166,25 +7518,54 @@ func (bav *UtxoView) _connectAcceptNFTBid(
...
 	// Force the input to be non-zero so that we can prevent replay attacks.
 	if totalInput == 0 {
-		return 0, 0, nil, RuleErrorAcceptNFTBidRequiresNonZeroInput
+		return 0, 0, nil, errors.Wrapf(RuleErrorAcceptNFTBidRequiresNonZeroInput, "_helpConnectNFTSold: ")
+	}
+
+	if txMeta.IsNFTBid {
+		bidAmountNanos := txMeta.BidAmountNanos
+		// Check that the bid amount is non-zero.
+		if bidAmountNanos == 0 {
```
diamondhands0, line 7545: Is this check needed? Seems like `bidAmountNanos < prevNFTEntry.MinBidAmountNanos` has you covered.

diamondhands0: Going with a previous comment, it seems like bidding zero should be allowed if the person puts an NFT up for sale with a min bid of zero. It's kindof like a "race" for who can bid on the NFT first.

lib/block_view.go
```diff
@@ -7349,53 +7731,62 @@ func (bav *UtxoView) _connectAcceptNFTBid(
...
 		NFTSpentUtxoEntries:       spentUtxoEntries,
 		PrevAcceptedNFTBidEntries: prevAcceptedBidHistory,
-
 		// Rosetta fields.
 		AcceptNFTBidCreatorPublicKey: nftPostEntry.PosterPublicKey,
 		AcceptNFTBidBidderPublicKey: bidderPublicKey,
 		AcceptNFTBidCreatorRoyaltyNanos: creatorCoinRoyaltyNanos,
-	})
+	}
+	if txMeta.IsAcceptNFTBid {
```
diamondhands0, line 7749: You could actually put `Type: args.Type` if you put the OperationType in the helper struct instead of the bool.

lib/block_view.go
```diff
@@ -7651,6 +7918,7 @@ func (bav *UtxoView) _connectNFTTransfer(
 	newNFTEntry.OwnerPKID = receiverPKID.PKID
 	newNFTEntry.UnlockableText = txMeta.UnlockableText
 	newNFTEntry.IsPending = true
+	newNFTEntry.IsBuyNow = false
```
diamondhands0, line 7921: Should we make it so IsBuyNow can only be set after IsPending = false? I don't think it's necessary but just a thought I had while reading.

lib/block_view.go
```diff
@@ -7039,11 +7142,253 @@ func (bav *UtxoView) _connectAcceptNFTBid(
...
+		}
+
+		// Verify that the bidder is not the current owner of the NFT.
+		if reflect.DeepEqual(nftEntry.OwnerPKID, bidderPKID.PKID) {
+			return 0, 0, nil, RuleErrorNFTOwnerCannotBidOnOwnedNFT
+		}
+
+		// Verify that the bid amount is greater than the min bid amount for this NFT.
+		// We allow BidAmountNanos to be 0 if there exists a previous bid entry. A value of 0 indicates that we should delete the entry.
+		if txMeta.BidAmountNanos < nftEntry.MinBidAmountNanos && !(txMeta.BidAmountNanos == 0 && prevNFTBidEntry != nil) {
```
diamondhands0, line 7247: Ah, going through everything again I realize now why you don't allow a bid of zero-- it's because a bid of zero is being used to indicate that a bid should be deleted. So I guess it's fine to enforce that a bid is non-zero in your checks in general. But I would still allow MinBidAmount to be zero since that seems like it should be harmless.

lib/block_view.go
diamondhands0: Let's add a test to catch this case also.

lib/block_view.go
```diff
@@ -7039,11 +7142,253 @@ func (bav *UtxoView) _connectAcceptNFTBid(
...
+		if nftEntry.IsBuyNow && txMeta.BidAmountNanos == 0 {
+			return 0, 0, nil, RuleErrorZeroBidOnBuyNowNFT
+		}
+		// Verify that we are not bidding on a Buy Now NFT before the Buy Now NFT Block Height. This should never happen.
+		if nftEntry.IsBuyNow && blockHeight < BuyNowNFTBlockHeight {
+			return 0, 0, nil, errors.Wrapf(RuleErrorBuyNowNFTBeforeBlockHeight, "_connectNFTBid: ")
+		}
+		if nftEntry.IsBuyNow && txMeta.BidAmountNanos >= nftEntry.MinBidAmountNanos {
+			isBuyNowBid = true
+		}
```
diamondhands0, line 7259: The logic around the MinBidAmountNanos is a bit confusing. I would add a comment as to why you need to check MinBidAmountNanos again in the IsBuyNow check. I think it's because you want zero bids to be handled in the !isBuyNowBid block below. But if not, then this extra check seems a bit weird.

lib/block_view.go
diamondhands0: Update: Nevermind. Clearly a bid of zero means you want to delete the bid, which doesn't make sense for an IsNFTBid I think...

lib/block_view.go
```diff
@@ -7408,134 +7799,10 @@ func (bav *UtxoView) _connectAcceptNFTBid(
...
 		return 0, 0, nil, fmt.Errorf(
-			"_connectAcceptNFTBid: Sum of participant diffs is >0 (%d, %d, %d, %d)",
+			"_helpConnectNFTSold: Sum of participant diffs is >0 (%d, %d, %d, %d)",
 			sellerDiff, bidderDiff, creatorDiff, coinDiff)
 	}
 
-	return totalInput, totalOutput, utxoOpsForTxn, nil
-}
-
-func (bav *UtxoView) _connectNFTBid(
```
diamondhands0, line 7418: Nit: I think the code would have been easier to review if you hadn't moved the _connectNFTBid() function, which I think messed up the diff.

lib/block_view_test.go
```diff
@@ -18895,6 +19001,500 @@ func TestNFTPreviousOwnersCantAcceptBids(t *testing.T) {
...
+		// Make sure royalties to creator and to coin are paid out correctly.
+		expectedCreatorRoyalty := bidAmountNanos / 10
+		require.Equal(uint64(2), expectedCreatorRoyalty)
+		expectedCoinRoyalty := 2 * bidAmountNanos / 10
+		require.Equal(uint64(4), expectedCoinRoyalty)
+		desoLocked, _ := _getCreatorCoinInfo(t, db, params, m0Pub)
+		require.Equal(m0DeSoLockedBefore+expectedCoinRoyalty, desoLocked)
+	}
+
+	// Roll all successful txns through connect and disconnect loops to make sure nothing breaks.
```
diamondhands0, line 19491: I'd add something with a transfer + buy now if you can. The IsPending is an interesting edge case.

Other things to consider that you might have done that I might have missed:
* I think you might have done this and I missed it, but bidding with an amount that's below the min for IsBuyNow=true is interesting.
* Also bidding zero when you have IsBuyNow is interesting.

lib/block_view.go
lazynina: It wasn't moved - just an annoying way in which GitHub displays the diff. I had to move out all the logic that "sells" the NFT so we didn't have duplicated code to handle royalties and NFT ownership.

lib/blockchain.go
```diff
@@ -3794,6 +3814,33 @@ func (bc *Blockchain) AddInputsAndChangeToTransactionWithSubsidy(
...
+			if err != nil {
+				return 0, 0, 0, 0, errors.Wrapf(err,
+					"_computeInputsForTxn: Problem getting augmented UtxoView from mempool: ")
+			}
+		}
+
+		nftKey := MakeNFTKey(txMeta.NFTPostHash, txMeta.SerialNumber)
+		nftEntry := utxoView.GetNFTEntryForNFTKey(&nftKey)
+
+		if nftEntry != nil && nftEntry.IsBuyNow {
```
diamondhands0, line 3838: Check isDeleted too and error if it is.

lib/constants.go
```diff
@@ -123,6 +123,11 @@ var (
 	// transfer txns, NFT burn txns, and AuthorizeDerivedKey txns will be accepted.
 	// Triggers: 12PM PT on 9/15/2021
 	NFTTransferOrBurnAndDerivedKeysBlockHeight = uint32(60743)
+
+	// BuyNowNFTBlockHeight defines the height at which NFTs can be sold at a fixed price instead of an auction style.
+	// Note: this will be adjusted, 90000 is a placeholder until a block height is decided.
+	// Triggers: TBD on TBD
+	BuyNowNFTBlockHeight = uint32(90000)
```
diamondhands0, line 130: Just to make it foolproof, I'd put a really high value here for now with a FIXME:

```
	// BuyNowNFTBlockHeight defines the height at which NFTs can be sold at a fixed price instead of an auction style.
	// FIXME: Currently set to a really high value until we decide when we want this to trigger.
	BuyNowNFTBlockHeight = math.MaxUint32
```

lib/mempool.go
```diff
@@ -1463,21 +1464,33 @@ func ComputeTransactionMetadata(txn *MsgDeSoTxn, utxoView *UtxoView, blockHash *
...
-		}
+		isBuyNow := false
 
 		// We don't send notifications for standing offers.
 		if realTxMeta.SerialNumber != 0 {
 			nftKey := MakeNFTKey(realTxMeta.NFTPostHash, realTxMeta.SerialNumber)
 			nftEntry := utxoView.GetNFTEntryForNFTKey(&nftKey)
+
+			bidderPKID := utxoView.GetPKIDForPublicKey(txn.PublicKey)
+			if reflect.DeepEqual(nftEntry.OwnerPKID, bidderPKID) {
```
diamondhands0, line 1475: Hmm.. I think this is wrong in the case where someone bids on their own non-buy-now NFT. I think an easy way to do this is to maybe use the UtxoOps of the connected txn to determine this, since that's the most fool-proof.

diamondhands0:
So overall this looks pretty great. I love how the helper function consolidates the code and makes buy now super clean.

I just had one super annoying thought that I figured I'd voice. So, with this change it looks like you can do Buy Now *or* auction. But I'm wondering if want the protocol to be more flexible, and support bids below the Buy Now threshold even when an NFT has isBuyNow = true. Like for me, I might just want to set a really high buy now price in case a whale comes in, but then still entertain lower offers. This is how eBay does it as well.

I think to support this we'd need to have a BuyNowPriceNanos that's distinct from MinBidAmountNanos. We'd need to do some checks around that too. But I wonder if it's easier than it seems. For example, we could simply replace the IsBuyNow bool with `BuyNowPriceNanos > 0` and I think everything would mooostly work?

Anyway, @lazynina let's discuss offline. If it's not too much work, it seems like the extra flexibility at the protocol layer could be nice, even if most frontends don't implement it. But it could also be not worth it since you can always just message someone to ask if they're willing to take a lower price, in which case they can switch it into auction mode and accept your bid.

lib/block_view.go
lazynina: I'm a little confused about the comment that begins with `I see you added it below`. What is it? the check for overflow or the checks if the Txn is a NFT Bid

lib/block_view.go
lazynina: yeah this is smooth. thx!

lib/block_view.go
lazynina: 0 means cancel my bid

lib/block_view.go
lazynina: 0 means cancel my bid.

lib/block_view.go
lazynina: I totally forgot we had TxnType. Duh. yeah it's easier that way.

lib/block_view.go
lazynina: yes it is redundant! deleting.

lib/block_view.go
lazynina: yep yep. must smoother. I just fetch it from the Txn instead so that there can't be any trickery.

lazynina: so `Txn.TxnMeta.GetTxnType() == TxnTypeAcceptNFTBid`

lib/block_view.go
lazynina: correct! yeah 0 is cancel my bid. unless we change this at a later date to be a -1 means cancel my bid. but that introduces a lot of changes. so I'm inclined to not allow 0 at this point in time.

lib/block_view.go
lazynina: `OperationType` and `TxnType` are two different types unfortunately.

lib/block_view.go
lazynina: hmm I think we should? I guess it's a little strange to sell a pending NFT lol

lazynina:
> So overall this looks pretty great. I love how the helper function consolidates the code and makes buy now super clean.
> 
> I just had one super annoying thought that I figured I'd voice. So, with this change it looks like you can do Buy Now _or_ auction. But I'm wondering if want the protocol to be more flexible, and support bids below the Buy Now threshold even when an NFT has isBuyNow = true. Like for me, I might just want to set a really high buy now price in case a whale comes in, but then still entertain lower offers. This is how eBay does it as well.
> 
> I think to support this we'd need to have a BuyNowPriceNanos that's distinct from MinBidAmountNanos. We'd need to do some checks around that too. But I wonder if it's easier than it seems. For example, we could simply replace the IsBuyNow bool with `BuyNowPriceNanos > 0` and I think everything would mooostly work?
> 
> Anyway, @lazynina let's discuss offline. If it's not too much work, it seems like the extra flexibility at the protocol layer could be nice, even if most frontends don't implement it. But it could also be not worth it since you can always just message someone to ask if they're willing to take a lower price, in which case they can switch it into auction mode and accept your bid.

I think you wanted to allow Buy Nows to have a 0 Buy Now price (but require users to pay at least 1 nanos). So I think if this key exists, then we will treat it as a buy now even if the value is 0. If there is no key, then it is not a buy now.

lib/blockchain.go
lazynina: This is checked in the line above.

lazynina:
replaced with #166

PR #146, opened 2021-11-20
Added DeSocialWorld

ews99:
Our node is running for months now and also serving https://desocial.nl and https://desocountries.com


tijno:
I support this.

PR #147, opened 2021-11-20
Add NFTz Node

mvanhalen:
Add NFTz Node. We started early October and have Minting and Post options. nftz.zone is our frontend. DeSo transactions are processed by node.nftz.zone or api.nftz.zone. Bumped to id 12 since I saw there is a pr for 11


tijno:
I support this.

PR #149, opened 2021-11-29
Added additional nodes

HPaulson:
Added Nodes 13 - 18 to the registry list. 

Each of these machines is owned and operated by SeismicCore, LLC on behalf of our clients listed in the "Owner" field therein. Each node meets all necessary listing criteria, and has been operating for at least 30 days.

### DIFF
--- a/lib/nodes.go
+++ b/lib/nodes.go
@@ -85,4 +85,34 @@ var NODES = map[uint64]DeSoNode{
 		URL:   "https://nftz.zone",
 		Owner: "mvanhalen",
 	},
+	13: {
+		Name:  "Cloutible",
+		URL:   "https://cloutible.club",
+		Owner: "DawaynePerza",
+	},
+	14: {
+		Name:  "Agbegbe",
+		URL:   "https://agbegbe.org",
+		Owner: "TheParkerazzi",
+	},
+	15: {
+		Name:  "CloutingAround",
+		URL:   "https://cloutingaround.dev",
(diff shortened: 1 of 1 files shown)


lib/nodes.go
```diff
@@ -85,4 +85,34 @@ var NODES = map[uint64]DeSoNode{
...
+		Owner: "paulobrien",
+	},
+	17: {
+		Name:  "Mousai",
+		URL:   "https://deso.mousai.stream",
+		Owner: "marlonjm2k",
+	},
+	18: {
+		Name:  "KoalaTBooks",
+		URL:   "https://koalatbooks.com",
```
maebeam, line 115: This domain was created two weeks ago, please resubmit it later.

maebeam:
The others look good though!

lib/nodes.go
HPaulson: @maebeam The node itself has been online for more than 30d, as per the requirement. The domain is newer, but the node was still live and processing tx's for weeks prior to the domain registration (It was used for API calls and other interactions using IP)

Not quite sure why domain registry would be used to calculate node uptime, as the metrics are entirely unrelated.

maebeam: Thanks for the clarification on uptime

HPaulson: @maebeam No problem -- thanks for the merge. As per the ineligible node, I'll go head and submit another PR to remove it.

PR #150, opened 2021-11-30
Fix Prefix Overlap: Requires resync

lazynina:
bump _PrefixPostHashSerialNumberToAcceptedBidEntries from 54 to 55, bump _PrefixAuthorizeDerivedKey from 54 to 56. resolve overlapping prefix issue.

Note: Requires resync


maebeam (approved):
Nice, let's add the test to this PR

AeonSw4n (approved):

tijno:
@lazynina Does this affect 3rd party nodes syncing with diamond/deso if they have not done full resync?

lazynina:
> @lazynina Does this affect 3rd party nodes syncing with diamond/deso if they have not done full resync?

@tijno it will affect them if they upgrade. The core team is working to determine the best course of action for performing this upgrade hence why it has not been merged. 

There isn't an imminent threat from the overlapping prefixes here since the construction of the keys for each index are of different lengths and there is no code that seeks over those prefixes yet. Seeking over prefixes with different length keys can cause some issues.

cmd/node.go
```diff
@@ -110,6 +110,12 @@ func (node *Node) Start() {
 		panic(err)
 	}
 
+	// Make sure there is no prefix conflict in BadgerDb
+	err = lib.CheckForPrefixOverlap()
```
diamondhands0, line 114: Right now this test only works if someone remembers to add their new prefix to the tester function. This isn't robust because people will forget to do this, and the value of the test will be lost.

Instead, what you want is to make it so that a compile error occurs somehow if you forget to add the prefix to the test.

Here's one idea:
```
// Define NEXT_TAG as a variable
NEXT_TAG = 57

// Increment through to NEXT_TAG in the test
for ii := 0; ii < NEXT_TAG; ii++ {
  _, inPrefixMap := prefixMap[ii]
  if !inPrefixMap {
    // Allow for any exceptions
    // return an error if not covered by exceptions
  }
}
```

diamondhands0: Also nitpick: All the testing code should live in a test file not in the live code.

AeonSw4n: Okay, updated the checks. I feel like it's harder to miss this check if it's in db_utils vs db_utils_test

AeonSw4n: Actually, I thought about this and you're right it's cleaner when it's in db_utils_test

lib/db_utils_test.go
```diff
@@ -14,6 +18,128 @@ import (
...
+		key := hex.EncodeToString(prefix)
+		if _, exists := prefixMap[key]; exists {
+			err = errors.Errorf("Found DB prefix conflict at %v", prefix)
+		}
+		prefixMap[key] = true
+	}
+
+	// Make sure we didn't forget to add the new prefix to prefixList.
+	// We iterate through all prefixes from 0 to NEXT_PREFIX to see if we missed something.
+	for i := uint32(0); i <= NEXT_PREFIX; i++ {
```
diamondhands0, line 103: Nit: Always prefer ii or vv. These are easier to find and replace or hop to if you're using shortcuts like * or #.

PR #153, opened 2021-12-02
Split block view

maebeam:
Why:
- GoLand and GitHub struggle with files this large. GitHub won't even render it and anytime you make a mistake in GoLand it takes forever to reprocess the entire file
- Files this large make it easy for random code to get stuck somewhere. During this split I found a number of methods that don't belong.
- Files this large make it impossible to read the entire file from top to bottom and understand everything it's doing
- This split sets us up to eventually move block view into its own package if we decide it makes sense to do so.


maebeam:
Closing in favor of #164

PR #163, opened 2021-12-11
Adding node

AdonousTech:


lib/nodes.go
```diff
@@ -125,4 +125,9 @@ var NODES = map[uint64]DeSoNode{
 		URL:   "https://deverso.io/",
 		Owner: "Nordian",
 	},
+	21: {
+		Name:  "NinjaNode",
+		URL:   "https://node.bitcloutapps.ninja",
+		Owner: "mattpitts"
```
maebeam, line 131: Trailing comma please

lib/nodes.go
```diff
@@ -125,4 +125,9 @@ var NODES = map[uint64]DeSoNode{
...
+	}
```
maebeam, line 132: Trailing comma please

maebeam (changes requested):

PR #167, opened 2021-12-15
return boolean indicating if error is from double spend detection or not

lazynina:


maebeam (approved):

lib/bitcoin_burner.go
```diff
@@ -717,19 +718,19 @@ func CheckBitcoinDoubleSpend(txnHash *chainhash.Hash, blockCypherAPIKey string,
 }
 
 func BlockCypherPushAndWaitForTxn(txnHex string, txnHash *chainhash.Hash,
-	blockCypherAPIKey string, doubleSpendWaitSeconds float64, params *DeSoParams) error {
+	blockCypherAPIKey string, doubleSpendWaitSeconds float64, params *DeSoParams) (_err error, _isDoubleSpend bool) {
```
diamondhands0, line 721: Nit: Convention is generally to make the last argument the error. Would quick fix it here.

diamondhands0 (approved):

PR #170, opened 2021-12-20
DeSo V3 Messages

AeonSw4n:
All works, for reference:
[backend PR](https://github.com/deso-protocol/backend/pull/253)
[identity PR](https://github.com/deso-protocol/identity/pull/107)
[frontend PR](https://github.com/deso-protocol/frontend/pull/543)

reference DIP: https://github.com/deso-protocol/dips/blob/p/messages-v3/dips/dip-5.md


lib/blockchain.go
```diff
@@ -3565,6 +3589,55 @@ func (bc *Blockchain) CreateAuthorizeDerivedKeyTxn(
...
 	}
 
+	if messagingPublicKey != "" && messagingKeyName != "" && messagingKeySignature != "" {
+		messagingPk, err := hex.DecodeString(messagingPublicKey)
+		if err != nil {
+			return nil, 0, 0, 0, errors.Wrapf(err, "CreateAuthorizeDerivedKeyTxn: " +
+				"Problem decoding messaging public key")
+		}
+
+		if err = IsByteArrayValidPublicKey([]byte(messagingPublicKey), "CreateAuthorizeDerivedKeyTxn: " +
```
AeonSw4n, line 3599: messagingPk, use validPublicKeyAndName

lib/block_view_message.go
```diff
@@ -79,86 +178,172 @@ func (bav *UtxoView) deleteMessageMappings(message *PGMessage) {
...
+	for _, messagingKeyEntry := range dbMessagingKeys {
+		key := *messagingKeyEntry.MessagingPublicKey
+		// Check if we have seen the messaging key before.
+		if _, exists := messagingKeysMap[key]; !exists {
+			messagingKeysMap[key] = messagingKeyEntry
+		} else {
+			// If the messaging key was deleted in the UtxoView, we can now delete it from our map.
+			if messagingKeysMap[key].isDeleted {
+				delete(messagingKeysMap, key)
+			}
```
lazynina, line 228: Do we want to delete this here? Seems unnecessary and could result in a bug where we've deleted the key from the map before flushing to the db.

lib/block_view_message.go
```diff
@@ -79,86 +178,172 @@ func (bav *UtxoView) deleteMessageMappings(message *PGMessage) {
...
-		messageKey := MakeMessageKey(publicKey, dbMessageEntry.TstampNanos)
-		bav._getMessageEntryForMessageKey(&messageKey)
+	// We define an auxiliary map to keep track of messages in UtxoView and DB.
+	messagesMap := make(map[MessageKey]*MessageEntry)
+
+	// First look for messages in the UtxoView. We don't skip deleted entries for now as we will do it later.
+	for messageKey, messageEntry := range bav.MessageKeyToMessageEntry {
+		for _, messagingKeyEntry := range _messagingKeyEntries {
+			if reflect.DeepEqual(messageKey.PublicKey[:], messagingKeyEntry.MessagingPublicKey[:]) {
+				// We will add the messages by sender so that we have no overlaps in the DB in some weird edge cases.
```
lazynina, line 264: What are these edge cases referenced here?

lib/block_view_message.go
```diff
@@ -79,86 +178,172 @@ func (bav *UtxoView) deleteMessageMappings(message *PGMessage) {
...
+	}
+	for _, messageEntry := range dbMessageEntries {
+		mapKey := MakeMessageKey(messageEntry.SenderMessagingPublicKey[:], messageEntry.TstampNanos)
+		if _, exists := messagesMap[mapKey]; !exists {
+			messagesMap[mapKey] = messageEntry
+		} else {
+			// If the message was deleted in the UtxoView, then we remove it from our map struct.
+			if messagesMap[mapKey].isDeleted {
+				delete(messagesMap, mapKey)
+			}
```
lazynina, line 285: Once again, this is a little concerning. We typically don't delete entries that are marked with `isDeleted=true` when we're reading from the view

lib/block_view_message.go
```diff
@@ -218,19 +396,124 @@ func (bav *UtxoView) _connectPrivateMessage(
...
+	if extraV, hasExtraV := txn.ExtraData["V"]; hasExtraV {
+		rr := bytes.NewReader(extraV)
+		version, err = ReadUvarint(rr)
+		if err != nil {
+			return 0, 0, nil, errors.Wrapf(RuleErrorPrivateMessageInvalidVersion,
+				"_connectPrivateMessage: Problem reading version from ExtraData, error: (%v)", err)
+		}
+		if version > 3 {
+			return 0, 0, nil, errors.Wrapf(RuleErrorPrivateMessageInvalidVersion,
+				"_connectPrivateMessage: Problem reading version from ExtraData")
```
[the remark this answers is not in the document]
lazynina: also, can't I specify `V=0` in the extra data map? We probably want to make sure that version is between 1 and 3 to be valid, correct?

lib/block_view_message.go
```diff
@@ -218,19 +396,124 @@ func (bav *UtxoView) _connectPrivateMessage(
...
+	// to these messaging keys, as opposed to encrypting to user's main keys.
+	if version == 3 {
+		// Make sure DeSo V3 messages are live.
+		if blockHeight < DeSoV3MessagesBlockHeight {
+			return 0, 0, nil, errors.Wrapf(
+				RuleErrorPrivateMessageMessagingPartyBeforeBlockHeight,
+				"_connectPrivateMessage: messaging party used before block height")
+		}
+		// Look for messaging keys in transaction ExtraData
+		// TODO: Do we want to make the ExtraData keys shorter to save space and transaction cost?
```
lazynina, line 438: Probably a good idea to keep the key length short if we can help it while maintaining some level of significance in the key name

lib/block_view_message.go
```diff
@@ -218,19 +396,124 @@ func (bav *UtxoView) _connectPrivateMessage(
...
+		}
+		// We do an analogous validation for the recipient's messaging key.
+		recipientMessagingKeyName, existsRecipientName := txn.ExtraData[RecipientMessagingKeyName]
+		if existsRecipient && existsRecipientName {
+			// This check exists because of an annoying edge-case where group owner wanted to send a message
+			// to the group. In this scenario, the recipient public key is the same as the user's public key,
+			// which would make us fail one of the previous checks. To circumvent this, the transaction
+			// metadata in this case would have recipient public key set to messaging public key.
+			// There is also an edge case where recipient public key can equal messaging public key if it's
+			// a base key, so we make an exception for that.
```
lazynina, line 474: this is a very helpful comment.

lib/block_view_message.go
```diff
@@ -308,52 +577,396 @@ func (bav *UtxoView) _disconnectPrivateMessage(
...
+		rr := bytes.NewReader(extraV)
+		version, err = ReadUvarint(rr)
+		if err != nil {
+			return errors.Wrapf(RuleErrorPrivateMessageInvalidVersion,
+				"_disconnectPrivateMessage: Problem reading version from ExtraData, error: (%v)", err)
+		}
+		if version > 3 {
+			return errors.Wrapf(RuleErrorPrivateMessageInvalidVersion,
+				"_disconnectPrivateMessage: Problem reading version from ExtraData")
+		}
```
lazynina, line 593: same as mentioned above. we should validate that the version is between 1 and 3. or better yet, define a function called `isMessageVersionValid` or something. 

Also this code is duplicated from above - can we refactor?

lib/block_view_types.go
```diff
@@ -98,8 +102,10 @@ const (
 	OperationTypeAcceptNFTTransfer            OperationType = 21
 	OperationTypeBurnNFT                      OperationType = 22
 	OperationTypeAuthorizeDerivedKey          OperationType = 23
+	OperationTypeMessagingKey                 OperationType = 24
+	OperationTypeMessageParty                 OperationType = 25
 
-	// NEXT_TAG = 24
+	// NEXT_TAG = 26
```
lazynina, line 108: merge main into your branch please - want to be sure we don't have overlap here.

lib/db_utils.go
```diff
@@ -233,10 +233,21 @@ var (
...
+	_PrefixMessageRecipient = []byte{58}
+
+	// Prefix for Message to Message Party. Prefix works similarly to regular private messages.
+	_PrefixMessageToMessageParty = []byte{59}
+
 	// TODO: This process is a bit error-prone. We should come up with a test or
 	// something to at least catch cases where people have two prefixes with the
 	// same ID.
-	// NEXT_TAG: 55
+	// NEXT_TAG: 58
```
lazynina, line 250: this should be 60 based on your values - but let's merge in main so we're confident that we're not overlapping.

lib/block_view_message.go
AeonSw4n: We're deleting from a local map that only makes reference to UtxoView's map, so it seems like it should be fine to delete here.

lib/block_view_message.go
AeonSw4n: This is check is for fetching messages from UtxoView and DB and it's potentially possible that the same message is present in both places. To make sure we don't fetch the same message twice, we will index messages in the local `messagesMap` by sender public key (we could alternatively index it by recipient public key, so we just choose sender).

lib/block_view_message.go
AeonSw4n: Should be fineee, per above comment.

lib/block_view_message.go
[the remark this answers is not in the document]
AeonSw4n: Good catch!

lib/block_view_message.go
AeonSw4n: Yeaaah, I wonder what we could rename them to. @diamondhands0 what do you think about this one?

lib/block_view_message.go
AeonSw4n: good, cleaner code! I like

PR #172, opened 2021-12-27
remove check that length of outputs is non-zero

lazynina:
Although highly unlikely, It is possible to have 0 outputs if inputs EXACTLY cover the fees for a transaction.

### DIFF
--- a/lib/block_view.go
+++ b/lib/block_view.go
@@ -2,7 +2,6 @@ package lib
 
 import (
 	"fmt"
-	"github.com/davecgh/go-spew/spew"
 	"math"
 	"reflect"
 	"strings"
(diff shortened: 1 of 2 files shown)


maebeam (approved):

maebeam:
Is this a forking change? Seems like a node running this change would accept a transaction when another one wouldn't?

lazynina:
> Is this a forking change? Seems like a node running this change would accept a transaction when another one wouldn't?

yes you're right. Let's hold off on this and incorporate it with the upcoming NFT V2 fork

maebeam:
Merged in a41b2fa02c3a4024b174f203e2e51196a0c5ddbc

PR #186, opened 2022-01-17
fix error with pointer when generating DESO royalties for other creators

lazynina:


maebeam (commented):
Wait, why doesn't the old code work? I'm confused 🤔

PR #191, opened 2022-01-18
Add new constant to constants file to represent the NodeSource map key in Extra Data

superzordon:
Enables https://github.com/deso-protocol/backend/pull/275


lazynina (approved):
🔥

Issue #192, opened 2022-01-18
Postgres migration error for latest and stable builds 2.01


mvanhalen:
When setting up a fresh node (latest and stable build 2.01) and Postgres Database it starts, creates the first 41 tables and then gives the following error:

` Finished running "20210623152412_create_tables"
backend     | panic: 20210623152412_create_tables: ERROR #42601 syntax error at or near ")"
backend     |
backend     | goroutine 1 [running]:
backend     | github.com/deso-protocol/core/cmd.(*Node).Start(0xc0001602d0)
backend     |   /deso/src/core/cmd/node.go:137 +0xae6
backend     | github.com/deso-protocol/backend/cmd.Run(0x1e84e00, {0x11e0357, 0x0, 0x0})
backend     |   /deso/src/backend/cmd/run.go:29 +0x98
backend     | github.com/spf13/cobra.(*Command).execute(0x1e84e00, {0x1f5be78, 0x0, 0x0})
backend     |   /root/go/pkg/mod/github.com/spf13/cobra@v1.1.3/command.go:856 +0x5f8
backend     | github.com/spf13/cobra.(*Command).ExecuteC(0x1e84b80)
backend     |   /root/go/pkg/mod/github.com/spf13/cobra@v1.1.3/command.go:960 +0x3ad
backend     | github.com/spf13/cobra.(*Command).Execute(...)
backend     |   /root/go/pkg/mod/github.com/spf13/cobra@v1.1.3/command.go:897
backend     | github.com/deso-protocol/backend/cmd.Execute()
backend     |   /deso/src/backend/cmd/root.go:24 +0x25
backend     | main.main()
backend     |   /deso/src/backend/main.go:15 +0x17`

Tried it a couple of times for stable and latest build , every time same result. Will setup go to see if I can spot the error.

tijno:
This line should not have `,`

https://github.com/deso-protocol/core/blob/main/migrate/20220106162320_create_dao_coin_tables.go#L18

tijno:
Looks like this is fixed with latest image
https://hub.docker.com/layers/desoprotocol/backend/latest/images/sha256-5cc9e687cefd5426fc8f9e977ab8a3e768425d25b7de72d4bd16c699d630c1a3?context=explore

PR #193, opened 2022-01-18
Fix sql statements for DAOCoin/Nft

tijno:
See #192


lazynina (approved):

lazynina:
@tijno it should be automatically released. If I recall correctly, we publish an image on every commit to main.

PR #197, opened 2022-01-23
Fix column typo Postgres

mvanhalen:


mvanhalen:
Should be applied after Lazy Nina branch Postgres fixes merge

lazynina:
@mvanhalen - I've made this change in my branch so closing it out here

PR #206, opened 2022-02-02
[fix] Disable possibility to transfer zero dao coins

donhardman:
This transaction sent zero dao coins – 3JuEUJQUPTYk4zRBQ9v7aRfYcs1dj8Sg7TUkdAuV1tyopd6ezn1SCX

And looks like its useless action at all and should be forbidden.

This fix disables zero dao coin transfers and throws out error on connect.


lazynina (approved):

lazynina:
will need to set block heights correctly, but otherwise this looks like a good change. @diamondhands0 - thoughts on this PR?

PR #213, opened 2022-02-09
Extra Data on Posts, NFTs, Messages, Messaging Groups, and Derived Keys

lazynina:
Note: this doesn't HAVE to be a forking change since this only parses the extra data and doesn't change consensus. But one would have different ProfileEntry's if they synced from scratch without it.


lib/block_view_profile_test.go
```diff
@@ -4115,5 +4167,5 @@ func TestAuthorizeDerivedKeyBasic(t *testing.T) {
 
 	// After we rolled back the blocks, db should reset
 	_verifyTest(authTxnMeta.DerivedPublicKey, 0, 0, AuthorizeDerivedKeyOperationValid, nil)
-	fmt.Println("Successfuly run TestAuthorizeDerivedKeyBasic()")
+	fmt.Println("Successfuly run TestAuthorizelDerivedKeyBasic()")
```
lazynina, line 4170:
```suggestion
	fmt.Println("Successfuly run TestAuthorizeDerivedKeyBasic()")
```

lib/blockchain.go
```diff
@@ -3691,11 +3717,18 @@ func (bc *Blockchain) CreateAuthorizeDerivedKeyTxn(
...
 	}
 
-	extraData := make(map[string][]byte)
+	derivedKeyExtraData := make(map[string][]byte)
 	if derivedKeySignature {
-		extraData[DerivedPublicKey] = derivedPublicKey
+		derivedKeyExtraData[DerivedPublicKey] = derivedPublicKey
+	}
+
+	//
```
lazynina, line 3725:
```suggestion
	// Delete protected keys
```

PR #222, opened 2022-02-23
Fix Issue with ParentStakeID set when modifying Post

lazynina:


diamondhands0:
Wow, cool find! Great work guys!

PR #224, opened 2022-02-23
add AcceptedBlockHeight to NFTBidEntry (REQUIRES RESYNC)

lazynina:


lib/postgres.go
```diff
@@ -729,19 +727,21 @@ func (nft *PGNFT) NewNFTEntry() *NFTEntry {
...
-	NFTPostHash    *BlockHash `pg:",pk,type:bytea"`
-	SerialNumber   uint64     `pg:",pk,use_zero"`
-	BidAmountNanos uint64     `pg:",use_zero"`
-	Accepted       bool       `pg:",use_zero"`
+	BidderPKID          *PKID      `pg:",pk,type:bytea"`
+	NFTPostHash         *BlockHash `pg:",pk,type:bytea"`
+	SerialNumber        uint64     `pg:",pk,use_zero"`
+	BidAmountNanos      uint64     `pg:",use_zero"`
+	Accepted            bool       `pg:",use_zero"`
+	AcceptedBlockHeight *uint32    `pg:",use_zero"`
```
AeonSw4n, line 735: Why do we use a pointer to a uint32?

AeonSw4n (approved):
Looks good, just wonder why we use the pointer for the uint32.

lib/postgres.go
lazynina: We use a pointer because the PGNFTBid and NFTBidEntry are used to represent both active bid (ones that can be accepted) as well as accepted bids. If the AcceptedBlockHeight is nil, then the bid HAS NOT been accepted.

AeonSw4n: All good then

diamondhands0: Fine to leave this, but for future reference I would avoid pointers to ints in Go. Better to just have an extra bool that's set to false by default when you create a struct because it's easier to read, and less likely to trigger errors (e.g. by dereferencing a nil pointer accidentally or something).

In this case, could you also just have `AcceptedBlockHeight = 0` to indicate bid hasn't been accepted yet? That seems like the most elegant solution unless there's an edge case I'm missing.

PR #241, opened 2022-03-09
Add DAO Coin Limit Order block view skeleton.

mattfoley8:


lib/block_view.go
```diff
@@ -9,7 +9,6 @@ import (
 	"time"
 
 	"github.com/btcsuite/btcd/btcec"
-	"github.com/dgraph-io/badger/v3"
```
mattfoley8, line 12: `go fmt` removed unused dependency

lazynina (approved):

PR #253, opened 2022-03-14
Add DAO coin limit order connect txn validations.

mattfoley8:


lib/block_view_dao_coin_limit_order.go
```diff
@@ -37,37 +38,135 @@ func (bav *UtxoView) _connectDAOCoinLimitOrder(
...
+		break
+	case DAOCoinLimitOrderEntryDenominatedCoinTypeDAOCoin:
+		return 0, 0, nil, RuleErrorDAOCoinLimitOrderUnsupportedDenominatedCoinType
+	default:
+		return 0, 0, nil, RuleErrorDAOCoinLimitOrderUnsupportedDenominatedCoinType
+	}
+
+	// If denominated in $DESO, validate DenominatedCoinCreatorPKID is all zeroes.
+	if txMeta.DenominatedCoinType == DAOCoinLimitOrderEntryDenominatedCoinTypeDESO {
+		if !reflect.DeepEqual(ZeroPKID, txMeta.DenominatedCoinCreatorPKID) {
```
lazynina, line 62: This will always be false since `ZeroPKID` is a direct reference to a `PKID` and `txMeta.DenominatedCoinCreatorPKID` is a pointer.
```suggestion
		if !reflect.DeepEqual(ZeroPKID, *txMeta.DenominatedCoinCreatorPKID) {
```

lib/block_view_dao_coin_limit_order.go
```diff
@@ -37,37 +38,135 @@ func (bav *UtxoView) _connectDAOCoinLimitOrder(
...
+
+	// Validate price > 0.
+	if !txMeta.PriceNanos.Gt(uint256.NewInt()) {
+		return 0, 0, nil, RuleErrorDAOCoinLimitOrderInvalidPrice
+	}
+
+	// If denominated in $DESO, confirm PriceNanos is uint64.
+	if !txMeta.PriceNanos.IsUint64() {
+		return 0, 0, nil, RuleErrorDAOCoinLimitOrderInvalidPrice
+	}
```
lazynina, line 94:
```suggestion
	if txMeta.DenominatedCoinType == DAOCoinLimitOrderEntryDenominatedCoinTypeDESO && !txMeta.PriceNanos.IsUint64() {
		return 0, 0, nil, RuleErrorDAOCoinLimitOrderInvalidPrice
	}
```

lib/block_view_dao_coin_limit_order.go
```diff
@@ -37,37 +38,135 @@ func (bav *UtxoView) _connectDAOCoinLimitOrder(
...
-	// Validate Quantity > 0
-	// If DESO buy, validate quantity < max uint64, order.Quantity.IsUint64()
-	// If DESO buy, validate price * quantity < max uint64
+	// Validate quantity > 0.
+	if !txMeta.Quantity.Gt(uint256.NewInt()) {
+		return 0, 0, nil, RuleErrorDAOCoinLimitOrderInvalidQuantity
+	}
+
+	// Validate quantity is uint64.
+	if !txMeta.Quantity.IsUint64() {
```
lazynina, line 105:
```suggestion
	if txMeta.DenominatedCoinType == DAOCoinLimitOrderEntryDenominatedCoinTypeDESO && !txMeta.Quantity.IsUint64() {
```

lazynina: I think we only want it to be less than max uint64 if it's denominated in DESO. Although this makes me a little concerned now since I expect DAO coin to be valued far less than DESO.  let's chat about this offline.

lib/block_view_dao_coin_limit_order.go
```diff
@@ -37,37 +38,135 @@ func (bav *UtxoView) _connectDAOCoinLimitOrder(
...
+		return 0, 0, nil, RuleErrorDAOCoinLimitOrderInvalidQuantity
+	}
+
+	// If $DESO buy, validate price * quantity < max uint64.
+	if txMeta.DenominatedCoinType == DAOCoinLimitOrderEntryDenominatedCoinTypeDESO {
+		if txMeta.PriceNanos.Uint64()*txMeta.Quantity.Uint64() >= math.MaxUint64 {
```
lazynina, line 111: this will just overflow. I think you should do the multiplication in uint256 and then use `IsUint64`

lib/block_view_dao_coin_limit_order.go
```diff
@@ -37,37 +38,135 @@ func (bav *UtxoView) _connectDAOCoinLimitOrder(
...
+	if txMeta.OperationType == DAOCoinLimitOrderEntryOrderTypeBid {
+		if txMeta.DenominatedCoinType == DAOCoinLimitOrderEntryDenominatedCoinTypeDESO {
+			desoBalanceNanos, err := bav.GetDeSoBalanceNanosForPublicKey(bav.GetPublicKeyForPKID(transactorPKID))
+
+			if err != nil {
+				return 0, 0, nil, err
+			}
+
+			// User is trying to open a bid order but doesn't have enough $DESO.
+			if desoBalanceNanos < (txMeta.PriceNanos.Uint64() * txMeta.Quantity.Uint64()) {
```
lazynina, line 151: I think you'll want to make a new Uint256 out of the desoBalanceNanos and compare as uint256s to avoid overflow issues

lib/block_view_dao_coin_limit_order.go
mattfoley8: Price refers to $DESO or DAO coin denomination.
Quantity refers solely to the number of DAO coins for sale.
I'm not sure we have to require that Quantity is a uint64.
It would just have to be less than or equal to the max supply of DAO coins.
What is the global max of DAO coins? Or is there a way to check the max number of DAO coins by PKID?

lib/block_view_dao_coin_limit_order.go
```diff
@@ -37,37 +37,140 @@ func (bav *UtxoView) _connectDAOCoinLimitOrder(
...
+	default:
+		return 0, 0, nil, RuleErrorDAOCoinLimitOrderUnsupportedOperationType
+	}
+
+	// Validate price > 0.
+	if txMeta.PriceNanos.Cmp(NewFloat()) > 0 {
+		return 0, 0, nil, RuleErrorDAOCoinLimitOrderInvalidPrice
+	}
+
+	// If denominated in $DESO, confirm PriceNanos is uint64.
```
lazynina, line 94:
```suggestion
	// If denominated in $DESO, confirm PriceNanos is less than max uint64.
```

lib/block_view_dao_coin_limit_order.go
```diff
@@ -37,37 +37,140 @@ func (bav *UtxoView) _connectDAOCoinLimitOrder(
...
+	if txMeta.DenominatedCoinType == DAOCoinLimitOrderEntryDenominatedCoinTypeDESO && !IsUint64(&txMeta.PriceNanos) {
+		return 0, 0, nil, RuleErrorDAOCoinLimitOrderInvalidPrice
+	}
+
+	// If denominated in DAO coins, confirm PriceNanos is uint256.
```
lazynina, line 99:
```suggestion
	// If denominated in DAO coins, confirm PriceNanos is less than max uint256.
```

lib/block_view_dao_coin_limit_order.go
```diff
@@ -37,37 +37,140 @@ func (bav *UtxoView) _connectDAOCoinLimitOrder(
...
+		}
+	}
+
+	// Validate transfer restriction status, if DAO coin can only be transferred to whitelisted members.
 	// TODO
 
-	// ----- Begin validations
-	// Create rule errors for each of these validations below
-	// lib/errors.go
+	// Validate if ask order, that the seller has enough of the DAO coin they're trying to sell.
```
lazynina, line 128:
```suggestion
	// If ask order, validate that the seller has enough of the DAO coin they're trying to sell.
```

lazynina (approved):

lazynina (approved):

PR #256, opened 2022-03-14
Convert DAO coin PriceNanos to big float.

mattfoley8:


lib/db_utils.go
```diff
@@ -5802,7 +5802,7 @@ func DBGetPaginatedPostsOrderedByTime(
 	postIndexKeys, _, err := DBGetPaginatedKeysAndValuesForPrefix(
 		db, startPostPrefix, _PrefixTstampNanosPostHash, /*validForPrefix*/
 		len(_PrefixTstampNanosPostHash)+len(maxUint64Tstamp)+HashSizeBytes, /*keyLen*/
-		numToFetch, reverse /*reverse*/, false /*fetchValues*/)
+		numToFetch, reverse                                                 /*reverse*/, false /*fetchValues*/)
```
lazynina, line 5805: this is ugly, but go fmt is dumb. let's just break this onto new lines so go fmt doesn't make it look so dumb

lazynina (approved):

PR #26, opened 2021-06-07
SOFT FORK: Today, BitClout joins the environmental movement.

diamondhands0:
I am proud to announce that the dev community has committed to moving to
a zero-waste consensus mechanism, and that it is taking a major step in
that direction with a new update.

Cryptocurrencies have gotten a lot of heat recently for being harmful to
the environment.

This is due in large part to "proof of work," a consensus mechanism that
burns CPU power to secure the network.

BitClout has relied on proof of work to get off the ground, but since
launching the dev community has already invested in consensus mechanisms
that allow BitClout to operate on orders of magnitude less power than
most existing proof of work cryptocurrencies.

Now, in a show of solidarity with the environmental movement, and to
take the first big step toward a zero-waste consensus mechanism,
BitClout is deploying a new update that reduces its power consumption by
a full order of magnitude.

BitClout is reducing the block reward, and thus the amount of
environmental waste its consensus mechanism incurs, by a factor of ten.

This was a difficult change to make, and it took the input of several
major node operators to deploy it safely.

We understand that miners in particular may not like this change;
however, there is now consensus around the decision to push the limits
of the BitClout chain toward conservation rather than allowing the
depletion of what are ultimately finite environmental resources.

There is still more work to do to move to a zero-waste consensus
mechanism without compromising on decentralization (e.g. full proof of
stake).

But the dev community is committed to getting there sooner rather than
later, and we hope this change shows just how committed we are.

PR #260, opened 2022-03-15
SOFT FORK: Today, Deso joins the decentralized movement

HPaulson:
I am proud to announce that the dev community has committed to moving to
a decentralized consensus mechanism, and that it is taking a major step in
that direction with a new update.

Deso has gotten a lot of heat recently for lacking true
decentralization.

This is due in large part to "trusted producers," a consensus mechanism that
enables five keys to control the network.

Deso has relied on trusted producers to get off the ground, but since
launching the dev community has already invested in consensus mechanisms
that allow Deso to operate on orders of magnitude less control than
the existing trusted producers model.

Now, in a show of solidarity with the web 3.0 movement, and to
take the first big step toward a decentralized consensus mechanism,
Deso is deploying a new update that reduces its centralized control by
a full order of magnitude.

Deso is removing trusted producers, and thus the amount of
control its consensus mechanism incurs, by a factor of ten.

This was a difficult change to make, and it took the input of several
major node operators to deploy it safely.

We understand that the Deso Foundation in particular may not like this change;
however, there is now consensus around the decision to push the limits
of the Deso chain toward decentralization rather than allowing the
depletion of what are ultimately finite community resources.

There is still more work to do to move to a zero-control consensus
mechanism without compromising on centralization (e.g. distribution of supply).

But the dev community is committed to getting there sooner rather than
later, and we hope this change shows just how committed we are.

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -144,22 +144,6 @@ func SetupRunFlags(cmd *cobra.Command) {
 	cmd.PersistentFlags().String("block-producer-seed", "",
 		"When set, all blocks produced by the block producer will be signed by this "+
 			"seed.")
-	cmd.PersistentFlags().StringSlice("trusted-block-producer-public-keys", []string{
-		"BC1YLgS1zDJQqywFpsty4fFheUrZxVQNKEsrttppvUESFZCq6Nfoypm",
-		"BC1YLh768bVj2R3QpSiduxcvn7ipxF3L3XHsabZYtCGtsinUnNrZvNN",
-		"BC1YLgsiUgM1Vr35YwbkSfZB3NC9tyrMXBPuJ2SEBf8naDf6PRpNit9",
-		"BC1YLgW5jWudzSUvrvNkD4GReN3kvGvsTuqLLttKfsCbXb7vLSCjwTk",
-		"BC1YLi8X7U9DZc2UqPE4s5PjrNJJUa6PKygD7VF4u8vy96srm18YvEX",
-	},
-		"When set, this node will only accept new blocks that are signed by the trusted block "+
-			"producers. This setting, is pretty novel. It allows a network of full nodes who "+
-			"trust each other to create their own network that can't be easily taken over by a 51% "+
-			"attack. In some sense, it uses trust in order to lower the amount of work needed to "+
-			"protect the network, making it highly eco-friendly. Then, if full nodes ever want to "+
-			"allow open mining, all they need to do is unset these public keys (or one of the owners "+
-			"of the public keys can release her key material, pulling a metaphorical 'ripcord'). "+
-			"Importantly, until this point, the network will be completely protected from a 51% attack, "+
-			"giving it time to accumulate the necessary hash power.")
 	cmd.PersistentFlags().Uint64("trusted-block-producer-start-height", 37000,
 		"If --trusted-block-producer-public-keys is set, then all blocks after this height must "+
 			"be signed by one of these keys in order to be considered valid. Setting this value to zero "+


sungkhum:
Let's get decentralized!

AdityaChaudhary0005:
image

diamondhands0:
Hey all, looks like there's some confusion here. The --trusted_block_producer flag is optional, which means that anybody who wants to run their node without relying on signatures from trusted nodes can easily do so. The only reason we have this flag as an option is because removing it would result in mining costs being significantly higher, as the comment explains.

Moreover, this whole mechanism is slated for removal once we move to PoS, which is right around the corner.

Lastly, I will add for reference that Coinbase imposes some extremely rigorous checks on decentralization prior to listing, which attempt to ensure both phsyical decentralization *and* economic decentralization of the network. If DeSo were not sufficiently decentralized then we could not have listed, and this is also why so-called "permissioned" networks almost never make it through.

HPaulson:
@diamondhands0  I'm very sorry this was sprung on you all. That was not the intent, and it's meant to be a positive change that will significantly increase how seriously Deso is taken by the broader crypto ecosystem in light of all of the controversy over centralization. It does also significantly decrease the Foundation's footprint, which is a good thing both for decentralization but also for the ability of securing the network.

In the future, changes like these will be submitted as DIPs (Deso Improvement Proposals) similar to BIPs in Bitcoin. The only reason we did not go through that process for this change is because we didn't feel like using that repository. That said, it will be used shortly, and it will allow the devs to get a lot more input before merging changes like this.

dgsus:
> Hey all, looks like there's some confusion here. The --trusted_block_producer flag is optional, which means that anybody who wants to run their node without relying on signatures from trusted nodes can easily do so. The only reason we have this flag as an option is because removing it would result in mining costs being significantly higher, as the comment explains.
> 
> 
> 
> Moreover, this whole mechanism is slated for removal once we move to PoS, which is right around the corner.
> 
> 
> 
> Lastly, I will add for reference that Coinbase imposes some extremely rigorous checks on decentralization prior to listing, which attempt to ensure both phsyical decentralization *and* economic decentralization of the network. If DeSo were not sufficiently decentralized then we could not have listed, and this is also why so-called "permissioned" networks almost never make it through.

looking forward to reading about the pos implementation plans ser.

also about decentralisation, i personally find it hard to believe... how many nodes are running reliably? 20, 30? only what the trusted block producers say go into this version of the chain that deso is currently running. how is that not centralized? 

anyway, if going away with pos its a least some good news. how about swap_identity? also going away with pos? 

cheers!

iPaulPro:
> Hey all, looks like there's some confusion here. The --trusted_block_producer flag is optional, which means that anybody who wants to run their node without relying on signatures from trusted nodes can easily do so. The only reason we have this flag as an option is because removing it would result in mining costs being significantly higher, as the comment explains.

Perhaps I _am_ misunderstanding. If I run a node and remove the trusted block producers flag, then a block is produced by a miner not on the core list and my node accepts it (but core nodes do not), don't we end up with 2 different chains?

HPaulson:
In all seriousness-

Of course, I submitted this PR as a satyrical parody of #26(0), but I do think it's an opportunity to have a solid conversation about this topic (especially after Nader's comment above, which I do greatly appreciate).

@diamondhands0 your position re: removing signature requirements would simply cause "3rd party nodes" (this isn't a great term, but I'm using it here to describe nodes not owned by core) to be out of sync & forked. Let me explain:

Many of the major DeSo products are core-owned, and thus submit tx's to a node also owned by the core (ex. BitClout, Diamond, DaoDao, etc). This isn't great, but I realize it's a problem that gets solved with time & more products. The issue, though, arises when we take a look at trusted producers.

Your argument that any node can just remove the signature requirement, since it's optional, has a major fallacy. Since core-run nodes are effectively the only nodes that people use to submit txs, anyone running a node that doesn't align with their chain would be considered out of sync, creating a hard fork. Removing this optional flag does exactly that: If I were to remove this flag & enable mining on all of the nodes in my network, eventually one would find a new block first & share it with peers accordingly. As soon as this happens, my network would accept the block, but yours wouldn't; that creates a hard fork.

Since people mainly utilize core-owned nodes right now, that's MY problem, not yours, even if every single node on the network (besides yours) had this flag disabled.

This creates a situation where your 5 nodes (which are trusted by default, I will note) effectively own the network. Anyone else who produces a block or attempts to collect mining reward would effectively fork the chain (even if their block is valid), and their node is now entirely useless as it's on a new chain. The general public, exchanges, and your major products all have this flag set to only trust your nodes, so even if in theory you could remove the flag, this wouldn't be effective in practice.

Imagine if I changed any other validation condition, even something major like signatures, on my network. Is it possible to change this? Of course; in theory it's optional . Does it work to contribute to the chain? Obviously not- this creates a fork, and my nodes would be considered invalid (since they approved invalid txs). This is a simple blockchain concept -- it's not about what MY nodes accept, it's about what the network at large accepts; since every exchange & your major products only trust your nodes, that's the network at large, and anything outside of these trusted blocks is considered invalid, causing DeSo to be centralized.

Per Coinbase-

- We've heard this claim mentioned by core a whole lot, but with no substantive evidence. As mentioned above, DeSo (as listed on coinbase, since anyone who doesn't use the flag created a new chain & new asset) is entirely centralized from a consensus standpoint, so I'm not too sure how vigorous this check is. I'm sure these checks do exist, but considering Coinbase Ventures backed DeSo & we know for a fact that consensus is centralized, I'm not so sure this check was very "rigorous" as you claim. Saying DeSo is decentralized and then using your own investors as evidence seems a little off to me, instead let's focus on the technicals; code doesn't lie.

Per economic centralization -- I'm really unsure how Coinbase would've done any kind of effective check on this. Of course your team created the wallet ("rich") endpoint, but anyone can create multiple wallets. At the first TownHall, you actually claimed to own "most of" the 2m DeSo given to founders, then in the next breath said that no entity owns more than 5%. 

Not to mention, ~80% of DeSo was created on a chain that nobody actually has access to, and was compiled into the genesis block. You've provided us with some useless data packet that we can't even use without the proper software, and on this topic you've said things like "It's personal wealth" and "That's not really important" (1st TownHall). With this in mind, especially your own claim about the 2m to founders, I'm not too sure it's fair to say DeSo is economically decentralized.

I'd love some additional information on these topics that includes actual evidence instead of waffling the same unsubstantiated claims we've heard for months.

PR #261, opened 2022-03-15
Add go test to GitHub Actions CI.

mattfoley8:
Examples of caching Go dependencies on Linux GitHub Actions:
* https://github.com/actions/cache/blob/main/examples.md#linux-1
* https://markphelps.me/posts/speed-up-your-go-builds-with-actions-cache/


mattfoley8:
GitHub Actions is currently down. :cry: 
```
Update - GitHub Actions is experiencing degraded performance. We are still investigating and will provide an update when we have one.
Mar 16, 14:29 UTC
```

PR #263, opened 2022-03-15
Mf/add dao coin limit order txn testing

mattfoley8:


lib/block_view_dao_coin_limit_order.go
```diff
@@ -271,19 +271,20 @@ func (bav *UtxoView) _connectDAOCoinLimitOrder(
 
 			// Update order quantities.
 			var daoCoinsToTransfer uint256.Int
+			orderIsComplete := false
 
 			if requestedOrder.Quantity.Lt(&order.Quantity) {
```
lazynina, line 276: I think we should do `requestedOrder.Quantity.Gt(&order.Quantity)` here instead of `Lt` and flip the logic. This way we don't have to duplicate the check for whether or not the order is complete.

lazynina (changes requested):
one smallll change - but then good to go into the feature branch

lib/block_view_dao_coin_limit_order.go
mattfoley8: It's not that simple because there's really 3 conditions, each with two that overlap. We also need to delete the order if they're equivalent: `bav._deleteDAOCoinLimitOrderEntryMappings(order)`, so the updated logic would be something like: 
```
if requestedOrder.Quantity == order.Quantity {
  // close order ***
  // set requestedOrder.Quantity = 0
  // set orderIsComplete to true +++
} else if requestedOrder.Quantity > order.Quantity {
  // close order ***
  // set requestedOrder.Quantity -= order.Quantity
} else {
  // requestedOrder.Quantity < order.Quantity
  // set order.Quantity -= requestedOrder.Quantity
  // set requestedOrder.Quantity = 0
  // set OrderIsComplete = true +++
}
```
So with your change from `if < elsif >=` to `if > elsif <=`, we would be consolidating the `orderIsComplete = true` logic (denoted +++ above), but would need a separate check for the `close order` logic (denoted *** above) in the `else`. Something like the following which I don't know is better.
```
			if requestedOrder.Quantity.Gt(&order.Quantity) {
				daoCoinsToTransfer = order.Quantity
				requestedOrder.Quantity = *uint256.NewInt().Sub(&requestedOrder.Quantity, &order.Quantity)
				bav._deleteDAOCoinLimitOrderEntryMappings(order)
			} else {
				daoCoinsToTransfer = requestedOrder.Quantity
				order.Quantity = *uint256.NewInt().Sub(&order.Quantity, &requestedOrder.Quantity)
				requestedOrder.Quantity = *uint256.NewInt()
				orderIsComplete = true

				if order.Quantity.IsZero() {
					bav._deleteDAOCoinLimitOrderEntryMappings(order)
				}
			}
```

lazynina (approved):

PR #268, opened 2022-03-17
add function to calculate total cost of an order

lazynina:
add function to calculate total cost of an order with new Price representation
Fix `IsBetterAskThan` and `IsBetterBidThan`
Remove erroneous conditional in `DBGetMatchingDAOCoinBidOrders`

### DIFF
--- a/lib/db_utils.go
+++ b/lib/db_utils.go
@@ -6249,6 +6249,7 @@ func DBGetMatchingDAOCoinAskOrders(txn *badger.Txn, inputOrder *DAOCoinLimitOrde
 	queryOrder.Quantity = MaxUint256.Clone()
 
 	key := DBKeyForDAOCoinLimitOrder(queryOrder, false)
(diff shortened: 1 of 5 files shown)


lib/db_utils.go
```diff
@@ -6370,14 +6370,6 @@ func DBGetMatchingDAOCoinBidOrders(txn *badger.Txn, inputOrder *DAOCoinLimitOrde
 			break
 		}
 
-		// Break if ask price is greater than requested bid price. In practice, this means we should break if the
-		// ask price per denominated coin is less than the requested bid price per denomiated coin.
-		// order.PriceNanos > inputOrder.PriceNanos
-		// order.PriceNanosPerDenominatedCoin < inputOrder.PriceNanosPerDenominatedCoin
-		if order.PriceNanosPerDenominatedCoin.Lt(inputOrder.PriceNanosPerDenominatedCoin) {
```
mattfoley8, line 6377: I think we still want this check. We have a corresponding one in the function to match requested bid orders.

mattfoley8 (approved):

PR #273, opened 2022-03-17
Add DAO coin limit order postgres tests.

mattfoley8:


lib/blockchain_test.go
```diff
@@ -170,6 +172,19 @@ func NewLowDifficultyBlockchainWithParams(params *DeSoParams) (
...
+
+	// TODO: convert this from env variable to cmd line flag.
+	if len(os.Getenv("TEST_POSTGRES")) > 0 {
+		postgresDb = NewPostgres(pg.Connect(&pg.Options{
+			Addr:     "localhost:5432",
+			User:     "admin",
+			Database: "admin",
+			Password: "",
+		}))
+	} else {
```
lazynina, line 185: We can mix this else actually. It's initialized to nil

lazynina (approved):

PR #286, opened 2022-03-29
Update DAOCoinLimitOrderMetadata type so it can be serialized to json

iamsofonias:
PKID is a type alias for byte array, which does not serialize directly to json. As a result the `DAOCoinLimitOrderMetadata` doesn't convert to json so it can be passed to the frontend. This PR does two things:
1) Changes the type of `MatchingBidsInputsMap` field from a map to a slice to address the above issue
2) Renames the field to `MatchedBidsTransactors` so its clear that the utxo inputs are being grouped by transactor when forming the other side of the trade

### DIFF
--- a/lib/block_view_dao_coin_limit_order.go
+++ b/lib/block_view_dao_coin_limit_order.go
@@ -620,18 +620,14 @@ func (bav *UtxoView) _connectDAOCoinLimitOrder(
 	} else {
 		desoAllowedToSpendByPublicKey[*NewPublicKey(txn.PublicKey)] = 0
 	}
+
 	// Iterate over all the UTXOs in the txn and spend them. As we spend each one,
 	// add the amount each account is allowed to spend to our map.
 	spentUtxoEntries := []*UtxoEntry{}
-	publicKeys := []PublicKey{}
-	for publicKey := range txMeta.MatchingBidsInputsMap {
-		publicKeys = append(publicKeys, publicKey)
-	}
-	sort.Slice(publicKeys, func(ii, jj int) bool {
-		return bytes.Compare(publicKeys[ii].ToBytes(), publicKeys[jj].ToBytes()) > 0
-	})
-	for _, publicKey := range publicKeys {
-		matchingBidsInputs := txMeta.MatchingBidsInputsMap[publicKey]
+	for _, transactor := range txMeta.MatchedBidsTransactors {
+		publicKey := *transactor.TransactorPublicKey
+		matchingBidsInputs := transactor.Inputs
+
 		// If no balance recorded so far, initialize to zero.
 		if _, exists := desoAllowedToSpendByPublicKey[publicKey]; !exists {
 			desoAllowedToSpendByPublicKey[publicKey] = 0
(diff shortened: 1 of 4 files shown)


lib/network.go
```diff
@@ -5515,40 +5522,12 @@ func (txnData *DAOCoinLimitOrderMetadata) ToBytes(preSignature bool) ([]byte, er
...
-		inputs, exists := txnData.MatchingBidsInputsMap[*bidderPk]
-		if !exists {
-			return nil, fmt.Errorf("DAOCoinLimitOrderMetadata: Error "+
-				"serializing inputs for pubkey %v at index %v. Map value "+
-				"did not exist. This should never happen", pkStr, ii)
-		}
-		data = append(data, bidderPk[:]...)
-		data = append(data, UintToBuf(uint64(len(inputs)))...)
-		for _, input := range inputs {
+	for _, transactor := range txnData.MatchedBidsTransactors {
```
iamsofonias, line 5527: No need to worry about sorting by public key since since we're dealing with a slice instead of a map

mattfoley8: Couldn't the ordering of the slice be different resulting in a different txn hash?

iamsofonias: Hmm, given an existing slice, serializing and deserializing it to/from bytes won't change the order of elements. Same thing with serializing/deserializing to/from json. So we're safe in this scenario.

But if we ever need to reconstruct the same transaction from scratch (ex: based on input params from the API), then we could end up with a different ordering of elements. The place to address that is in the `CreateDAOCoinLimitOrderTxn` function which is where we construct this metadata type. I can make the change there, but will we ever hit this scenario?

mattfoley8: I'm not sure, but this ominous comment makes me think we should just double-check some consistent sorted ordering in the slice.
```
	// Maps MUST be iterated in sorted order or else the serialization of the txn
	// will be non-deterministic. This is very bad because it results in the txn
	// hash being different each time.
```

lib/block_view_dao_coin_limit_order.go
```diff
@@ -620,18 +620,14 @@ func (bav *UtxoView) _connectDAOCoinLimitOrder(
...
-	publicKeys := []PublicKey{}
-	for publicKey := range txMeta.MatchingBidsInputsMap {
-		publicKeys = append(publicKeys, publicKey)
-	}
-	sort.Slice(publicKeys, func(ii, jj int) bool {
-		return bytes.Compare(publicKeys[ii].ToBytes(), publicKeys[jj].ToBytes()) > 0
-	})
-	for _, publicKey := range publicKeys {
-		matchingBidsInputs := txMeta.MatchingBidsInputsMap[publicKey]
+	for _, transactor := range txMeta.MatchedBidsTransactors {
```
lazynina, line 627: I think we do still need to sort this slice somehow to make it deterministic (hypersync requirement). Let's sort by public key

lib/network.go
lazynina, line 5527: let's sort these here inputs here as well please.

lazynina (changes requested):

lazynina (approved):

PR #287, opened 2022-03-29
Mf/add safe math uint64 utils

mattfoley8:


mattfoley8:
This was branched off of the SafeMathUint256 branch which got merged so it includes those extra commits, but you can see from the modified files that it only adds new SafeMathUint64 utils and implements in a few places.

lazynina (approved):

PR #290, opened 2022-03-30
Add UTXO view API DAO coin limit order getters.

mattfoley8:


lib/block_view_dao_coin_limit_order.go
```diff
@@ -1062,16 +1066,147 @@ func (bav *UtxoView) _getDAOCoinLimitOrderEntry(inputEntry *DAOCoinLimitOrderEnt
...
+				// This should never happen as DAO coin limit
+				// order ToBytes() never returns an error.
+				return nil, err
+			}
+			orderEntryString := string(orderEntryBytes)
+
+			outputEntries = append(outputEntries, orderEntry)
+			seenOrderEntries[orderEntryString] = true
+		}
+	}
```
lazynina, line 1167: Generally the process we follow for these functions is: 
1. Get everything from the DB first
2. Merge into view
3. Iterate over view and get the stuff you need
As an example, see `GetNFTEntriesForPostHash` on line 63 of `block_view_nft.go`

lazynina (changes requested):

lib/block_view_dao_coin_limit_order.go
```diff
@@ -1062,16 +1066,147 @@ func (bav *UtxoView) _getDAOCoinLimitOrderEntry(inputEntry *DAOCoinLimitOrderEnt
...
+
+	// Get orders from the database.
+	dbAdapter := DbAdapter{
```
lazynina, line 1107: should we add an function on UtxoView called `GetDBAdapter` so we don't have this construction going on everywhere?

lazynina (approved):

PR #298, opened 2022-04-04
Add postgres CI whitelisted testing.

mattfoley8:


lib/postgres_test.go
```diff
@@ -0,0 +1,27 @@
+package lib
+
+import (
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+func TestNewPostgresFromURI(t *testing.T) {
+	require := require.New(t)
+	_ = require
```
lazynina, line 10:
```suggestion
```

scripts/migrate.go
```diff
@@ -21,13 +22,20 @@ const directory = "migrate"
...
+	if len(os.Getenv("POSTGRES_URI")) > 0 {
+		pgOptions = lib.ParsePostgresURI(os.Getenv("POSTGRES_URI"))
+	} else {
+		pgOptions = &pg.Options{
+			Addr:     "localhost:5432",
+			User:     "admin",
+			Database: "admin",
+			Password: "",
+		}
+	}
```
lazynina, line 36: I generally prefer initializing this variable with the default value and just having a IF statement, as opposed to an if-else.

lazynina (changes requested):

PR #311, opened 2022-04-07
Mf/add operation type column

mattfoley8:
There is an off-by-one nano bug in the ComputeBaseUnitsToBuyUint256 function which I am still debugging. Otherwise this PR should be g2g.


lib/block_view_dao_coin_limit_order.go
```diff
@@ -1229,6 +1389,12 @@ func (bav *UtxoView) IsValidDAOCoinLimitOrder(order *DAOCoinLimitOrderEntry, isC
 		return RuleErrorDAOCoinLimitOrderCannotBuyAndSellSameCoin
 	}
 
+	// Validate operation type.
+	if order.OperationType != DAOCoinLimitOrderOperationTypeASK &&
```
lazynina, line 1393: This check seems off? an order can only have one operation type, no?

lib/block_view_dao_coin_limit_order.go
```diff
@@ -347,10 +354,14 @@ func (bav *UtxoView) _connectDAOCoinLimitOrder(
...
-			// This is the total amount of sellCoin the current order is willing to sell. Note that
-			// the asset the matching order is SELLING is the same asset that the
-			// transactor's order is BUYING.
-			matchingOrderTotalBuyCoinToSell, err := matchingOrder.BaseUnitsToSellUint256()
+			// Calculate leftover transactor and matching order quantities
+			// as well as the number of coins exchanged.
+			updatedTransactorOrderQuantityToFill,
+				updatedMatchingOrderQuantityToFill,
+				daoCoinBaseUnitsBoughtByTransactor,
+				daoCoinBaseUnitsSoldByTransactor,
```
lazynina, line 362: should these variables be renamed since they COULD be DESO?

lib/block_view_dao_coin_limit_order.go
```diff
@@ -366,183 +377,79 @@ func (bav *UtxoView) _connectDAOCoinLimitOrder(
...
+			filledOrders = append(filledOrders, transactorOrderFilledOrder)
 
-				// Mark matching order for deletion.
+			// Update quantity for matching order.
+			matchingOrderFilledOrder := &FilledDAOCoinLimitOrder{
+				TransactorPKID:                            matchingOrder.TransactorPKID,
+				BuyingDAOCoinCreatorPKID:                  matchingOrder.BuyingDAOCoinCreatorPKID,
+				SellingDAOCoinCreatorPKID:                 matchingOrder.SellingDAOCoinCreatorPKID,
+				BuyingDAOCoinQuantityInBaseUnitsPurchased: daoCoinBaseUnitsSoldByTransactor,
+				SellingDAOCoinQuantityInBaseUnitsSold:     daoCoinBaseUnitsBoughtByTransactor,
```
lazynina, line 420: I think I incorrectly named these two struct attributes, should these really be `BuyingCoinQuantityInBaseUnitsPurchased` and `SellingCoinQuantityInBaseUnitsPurchased`?

lib/block_view_dao_coin_limit_order.go
```diff
@@ -1029,6 +936,259 @@ func (bav *UtxoView) _deleteDAOCoinLimitOrderEntryMappings(entry *DAOCoinLimitOr
...
 
+func _calculateDAOCoinsTransferredInLimitOrderMatch(
+	transactorOrder *DAOCoinLimitOrderEntry,
+	matchingOrder *DAOCoinLimitOrderEntry,
+	transactorQuantityToFillInBaseUnits *uint256.Int) (
+	updatedTransactorQuantityToFillInBaseUnits *uint256.Int,
+	updatedMatchingQuantityToFillInBaseUnits *uint256.Int,
+	transactorBuyingCoinBaseUnitsTransferred *uint256.Int,
+	transactorSellingCoinBaseUnitsTransferred *uint256.Int,
+	err error) {
```
lazynina, line 947:
```suggestion
	_updatedTransactorQuantityToFillInBaseUnits *uint256.Int,
	_updatedMatchingQuantityToFillInBaseUnits *uint256.Int,
	_transactorBuyingCoinBaseUnitsTransferred *uint256.Int,
	_transactorSellingCoinBaseUnitsTransferred *uint256.Int,
	_err error) {
```
so we explicitly return all values

lib/block_view_dao_coin_limit_order.go
```diff
@@ -1029,6 +936,259 @@ func (bav *UtxoView) _deleteDAOCoinLimitOrderEntryMappings(entry *DAOCoinLimitOr
...
+				matchingOrder.ScaledExchangeRateCoinsToSellPerCoinToBuy,
+				transactorQuantityToFillInBaseUnits)
+			if err != nil {
+				return nil, nil, nil, nil, err
+			}
+
+			// We use the transactor's quantity as-is as the number
+			// of coins that were sold by the transactor.
+			transactorSellingCoinBaseUnitsTransferred = transactorQuantityToFillInBaseUnits
+			return
```
lazynina, line 1008:
```suggestion
			return
```
Is this return here correct? If so, please supply the return values

lib/block_view_dao_coin_limit_order.go
```diff
@@ -1029,6 +936,259 @@ func (bav *UtxoView) _deleteDAOCoinLimitOrderEntryMappings(entry *DAOCoinLimitOr
...
+		// the quantity bought by the transactor.
+		transactorBuyingCoinBaseUnitsTransferred, err = ComputeBaseUnitsToSellUint256(
+			matchingOrder.ScaledExchangeRateCoinsToSellPerCoinToBuy, matchingOrderQuantityToBuy)
+		if err != nil {
+			return nil, nil, nil, nil, err
+		}
+
+		// And its quantity to buy is equal to the quantity sold by the transactor.
+		transactorSellingCoinBaseUnitsTransferred = matchingOrderQuantityToBuy
+		return
```
lazynina, line 1031: supply return values please.

lib/block_view_dao_coin_limit_order.go
```diff
@@ -1029,6 +936,259 @@ func (bav *UtxoView) _deleteDAOCoinLimitOrderEntryMappings(entry *DAOCoinLimitOr
...
+			transactorBuyingCoinBaseUnitsTransferred = transactorQuantityToFillInBaseUnits
+
+			transactorSellingCoinBaseUnitsTransferred, err = ComputeBaseUnitsToBuyUint256(
+				matchingOrder.ScaledExchangeRateCoinsToSellPerCoinToBuy,
+				transactorQuantityToFillInBaseUnits)
+			if err != nil {
+				return nil, nil, nil, nil, err
+			}
+
+			return
```
lazynina, line 1087: please add return values explicitly

lib/block_view_dao_coin_limit_order.go
```diff
@@ -1029,6 +936,259 @@ func (bav *UtxoView) _deleteDAOCoinLimitOrderEntryMappings(entry *DAOCoinLimitOr
...
+		transactorBuyingCoinBaseUnitsTransferred = matchingOrderQuantityToSell
+
+		// And its quantity bought is equal to the quantity sold by the transactor.
+		transactorSellingCoinBaseUnitsTransferred, err = ComputeBaseUnitsToBuyUint256(
+			matchingOrder.ScaledExchangeRateCoinsToSellPerCoinToBuy, matchingOrderQuantityToSell)
+		if err != nil {
+			return nil, nil, nil, nil, err
+		}
+
+		return
```
lazynina, line 1111: return values please

lib/block_view_dao_coin_limit_order.go
```diff
@@ -1029,6 +936,259 @@ func (bav *UtxoView) _deleteDAOCoinLimitOrderEntryMappings(entry *DAOCoinLimitOr
...
+			// The transactor's quantity represents their selling coin.
+			transactorBuyingCoinBaseUnitsTransferred, err = ComputeBaseUnitsToSellUint256(
+				matchingOrder.ScaledExchangeRateCoinsToSellPerCoinToBuy,
+				transactorQuantityToFillInBaseUnits)
+			if err != nil {
+				return nil, nil, nil, nil, err
+			}
+
+			transactorSellingCoinBaseUnitsTransferred = transactorQuantityToFillInBaseUnits
+			return
```
lazynina, line 1137: return values please

lib/block_view_dao_coin_limit_order.go
```diff
@@ -1029,6 +936,259 @@ func (bav *UtxoView) _deleteDAOCoinLimitOrderEntryMappings(entry *DAOCoinLimitOr
...
+		transactorBuyingCoinBaseUnitsTransferred = transactorQuantityToFillInBaseUnits
+
+		transactorSellingCoinBaseUnitsTransferred, err = ComputeBaseUnitsToBuyUint256(
+			matchingOrder.ScaledExchangeRateCoinsToSellPerCoinToBuy,
+			transactorQuantityToFillInBaseUnits)
+		if err != nil {
+			return nil, nil, nil, nil, err
+		}
+
+		return
```
lazynina, line 1150: return values please

lib/block_view_dao_coin_limit_order.go
```diff
@@ -1029,6 +936,259 @@ func (bav *UtxoView) _deleteDAOCoinLimitOrderEntryMappings(entry *DAOCoinLimitOr
...
+		transactorBuyingCoinBaseUnitsTransferred = matchingOrder.QuantityToFillInBaseUnits
+
+		transactorSellingCoinBaseUnitsTransferred, err = ComputeBaseUnitsToBuyUint256(
+			matchingOrder.ScaledExchangeRateCoinsToSellPerCoinToBuy,
+			matchingOrder.QuantityToFillInBaseUnits)
+		if err != nil {
+			return nil, nil, nil, nil, err
+		}
+
+		return
```
lazynina, line 1176: return values please

lib/block_view_dao_coin_limit_order.go
```diff
@@ -1029,6 +936,259 @@ func (bav *UtxoView) _deleteDAOCoinLimitOrderEntryMappings(entry *DAOCoinLimitOr
...
+	// Which is equivalent to the transactor's selling coin.
+	transactorBuyingCoinBaseUnitsTransferred, err = ComputeBaseUnitsToBuyUint256(
+		matchingOrder.ScaledExchangeRateCoinsToSellPerCoinToBuy,
+		matchingOrder.QuantityToFillInBaseUnits)
+	if err != nil {
+		return nil, nil, nil, nil, err
+	}
+
+	transactorSellingCoinBaseUnitsTransferred = matchingOrder.QuantityToFillInBaseUnits
+	return
```
lazynina, line 1189: return value please

lib/block_view_dao_coin_limit_order.go
```diff
@@ -1257,7 +1423,7 @@ func (bav *UtxoView) IsValidDAOCoinLimitOrder(order *DAOCoinLimitOrderEntry, isC
 	}
 
 	// Validate quantity > 0.
-	if !order.QuantityToBuyInBaseUnits.Gt(uint256.NewInt()) {
+	if !order.QuantityToFillInBaseUnits.Gt(uint256.NewInt()) {
```
lazynina, line 1426:
```suggestion
	if order.QuantityToFillInBaseUnits.IsZero() {
```

lazynina (changes requested):

lib/block_view_dao_coin_limit_order.go
mattfoley8: I think it's right. If the operation type != ASK and the operation type != BID, then we error out: invalid operation type. It's a double negative. Equivalent to `if !(operation type == ASK || operation type == BID)`.

lib/block_view_dao_coin_limit_order.go
```diff
@@ -268,15 +269,21 @@ func (bav *UtxoView) _connectDAOCoinLimitOrder(
 	// Check to see if we have an existing order. If we do, then update our
 	// transactor order and delete the old one.
 	if prevTransactorOrder != nil && !prevTransactorOrder.isDeleted {
+		// If the operation types are equal, then we can just add the quantities
+		// as expected. If the operation types differ, then we throw an error.
+		if transactorOrder.OperationType != prevTransactorOrder.OperationType {
```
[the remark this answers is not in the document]
mattfoley8: We could actually convert the quantity and update the new order with the previous order's converted quantity. But I changed it to this. IMO this is safer. Feels more like a place for someone to try to exploit than a real feature someone would use. It's like saying, "I want to buy 10 DAO coin XYZ for $DESO and I also want to sell 5 $DESO for DAO coin XYZ." As opposed to just saying the amount of DAO coins you want to buy. IMO would be weird for someone to specify it this way in the same block height but I can revert it back to the updating the second order's quantity if you think that's preferable. IMO safer to error out as is done here though.

lib/block_view_dao_coin_limit_order.go
mattfoley8: I think this is right. It shows for the given order how much DAO coin base units they bought and sold. It's inverted depending on whether you're looking at the transactor or the matching order. I.e. the # DAO coins bought by the transactor == # DAO coins sold by the matching order and vice versa. Which is why you see them flipped here. We're looking at the matching order.

lib/block_view_dao_coin_limit_order.go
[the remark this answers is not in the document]
lazynina: Agree this is safer - keep it as is!

lib/block_view_dao_coin_limit_order.go
lazynina: Yep - I read it wrong. Good stuff

lib/block_view_dao_coin_limit_order.go
lazynina: These can't be amounts of DESO?

mattfoley8: You're right. Updating now.

lib/block_view_types.go
```diff
@@ -1487,21 +1487,22 @@ type DAOCoinLimitOrderEntry struct {
 	// 2, then we would need to convert 2 SellingDAOCoinCreatorPKID
 	// coins to get 1 BuyingDAOCoinCreatorPKID. Note, however, that to represent
 	// 2, we would actually have to set ScaledExchangeRateCoinsToSellPerCoinToBuy to
-	// be equal to 2*2^128 because of the representation format we describe
+	// be equal to 2*1e38 because of the representation format we describe
```
mattfoley8, line 1490: nit: 2*1e138 --> 1e38

PR #317, opened 2022-04-11
TotalInput and TotalOutput should consider inputs and outputs to/from bidders and difference should equal fee

lazynina:


lib/block_view_dao_coin_limit_order.go
```diff
@@ -787,6 +794,13 @@ func (bav *UtxoView) _connectDAOCoinLimitOrder(
 					if err = createUTXO(newDESOSurplus.Uint64(), pubKey, UtxoTypeDAOCoinLimitOrderPayout); err != nil {
 						return 0, 0, nil, err
 					}
+
+					// Increase totalOutput to accoutn for outputs generated by matching orders.
```
mattfoley8, line 798:
```suggestion
					// Increase totalOutput to account for outputs generated by matching orders.
```

lib/block_view_dao_coin_limit_order.go
```diff
@@ -846,6 +860,17 @@ func (bav *UtxoView) _connectDAOCoinLimitOrder(
 		FilledDAOCoinLimitOrders:             filledOrders,
 	})
 
+	// Just to be safe, we confirm that totalInput doesn't exceed totalOutput.
```
mattfoley8, line 863:
```suggestion
	// Just to be safe, we confirm that totalOutput doesn't exceed totalInput.
```

mattfoley8 (approved):

diamondhands0 (approved):
Great catch!

PR #327, opened 2022-04-13
Mf/add dao coin limit order id field

mattfoley8:


migrate/20220309224100_create_dao_coin_limit_order_tables.go
```diff
@@ -41,21 +41,14 @@ func init() {
...
-				block_height                                       BIGINT NOT NULL,
-
-				PRIMARY KEY (
-					transactor_pkid,
-					buying_dao_coin_creator_pkid,
-					selling_dao_coin_creator_pkid,
-					scaled_exchange_rate_coins_to_sell_per_coin_to_buy,
-					block_height
-				)
+				block_height                                       BIGINT NOT NULL
```
mattfoley8, line 51: It may be useful for us to include some additional indices here:
* `transactor_pkid`: This would help power the GetOrdersByTransactorPKID query.
* `buying_dao_coin_creator_pkid, selling_dao_coin_creator_pkid`: This would help power the GetMatchingOrders query.
* `order_id`: This is the primary key so we already have an index here. This helps power the GetOrdersByOrderID query. :heavy_check_mark: 
@lazynina LMK what you think.

mattfoley8: Added w/ https://github.com/deso-protocol/core/pull/327/commits/f6ede7c1cc02a3139c289f4abfbdcab4e2aa7d0e.

lib/block_view_dao_coin_limit_order.go
```diff
@@ -269,6 +269,34 @@ func (bav *UtxoView) _connectDAOCoinLimitOrder(
...
+			OrderID:        txMeta.CancelOrderID,
+			TransactorPKID: transactorPKIDEntry.PKID,
+		})
+		if err != nil {
+			return 0, 0, nil, err
+		}
+		if existingTransactorOrder == nil {
+			return 0, 0, nil, RuleErrorDAOCoinLimitOrderToCancelNotFound
+		}
+
```
lazynina, line 285: we should add a check that the order is created by the transactor.

lib/block_view_dao_coin_limit_order.go
```diff
@@ -1347,13 +1295,29 @@ func (bav *UtxoView) _getDAOCoinLimitOrderEntry(inputEntry *DAOCoinLimitOrderEnt
...
+		return nil, nil
+	}
+
+	// Confirm that the existing order entry in the database found by OrderID
+	// was created by this transactor. Otherwise, it would be possible for a
+	// user to delete an order created by another user! In this case, we don't
+	// throw an error but instead just return nil, i.e. no order found.
+	if !inputEntry.TransactorPKID.Eq(outputEntry.TransactorPKID) {
+		return nil, nil
+	}
```
lazynina, line 1318: I think this check belongs in the connect logic. There may be reasons in the future that we want this function to just return an order entry given an orderId

lib/block_view_types.go
```diff
@@ -1809,20 +1818,17 @@ func ComputeBaseUnitsToSellUint256(
 	return quantityToSellUint256, nil
 }
 
+// An OrderID uniquely identifies an order, but we also include
+// TransactorPKID as a safe-guard to ensure that the transactor
+// can only modify their own orders.
```
lazynina, line 1823: Can't we validate this from the entry?

lib/blockchain.go
```diff
@@ -3178,6 +3178,11 @@ func (bc *Blockchain) CreateDAOCoinLimitOrderTxn(
 	// We will set FeeNanos to it's true value after we add inputs and outputs.
 	metadata.FeeNanos = math.MaxUint64
 
+	// Initialize CancelOrderID to the ZeroBlockHash if nil.
+	if metadata.CancelOrderID == nil {
+		metadata.CancelOrderID = &ZeroBlockHash
+	}
```
lazynina, line 3184: Why do we want to initialize this to the ZeroBlockHash? Doing this increases the size of the transaction

lib/db_utils.go
```diff
@@ -6410,6 +6418,13 @@ func DBPutDAOCoinLimitOrderWithTxn(txn *badger.Txn, order *DAOCoinLimitOrderEntr
 		return errors.Wrapf(err, "DBPutDAOCoinLimitOrderWithTxn: problem storing limit order")
 	}
 
+	// Store in index: _PrefixDAOCoinLimitOrderByOrderID
+	key = DBKeyForDAOCoinLimitOrderByOrderID(order)
+
+	if err = txn.Set(key, orderBytes); err != nil {
+		return errors.Wrapf(err, "DBPutDAOCoinLimitOrderWithTxn: problem storing limit order")
```
lazynina, line 6425: we should probably update this error statement to indicate which prefix failed. Will help if we ever have to debug.

lib/db_utils.go
```diff
@@ -6432,5 +6447,12 @@ func DBDeleteDAOCoinLimitOrderWithTxn(txn *badger.Txn, order *DAOCoinLimitOrderE
 		return errors.Wrapf(err, "DBDeleteDAOCoinLimitOrderWithTxn: problem deleting limit order")
 	}
 
+	// Delete from index: _PrefixDAOCoinLimitOrderByOrderID
+	key = DBKeyForDAOCoinLimitOrderByOrderID(order)
+
+	if err := txn.Delete(key); err != nil {
+		return errors.Wrapf(err, "DBDeleteDAOCoinLimitOrderWithTxn: problem deleting limit order")
```
lazynina, line 6454: same as above

lib/network.go
```diff
@@ -5618,7 +5618,7 @@ func (txnData *DAOCoinLimitOrderMetadata) ToBytes(preSignature bool) ([]byte, er
 	data = append(data, EncodeUint256(txnData.ScaledExchangeRateCoinsToSellPerCoinToBuy)...)
 	data = append(data, EncodeUint256(txnData.QuantityToFillInBaseUnits)...)
 	data = append(data, UintToBuf(uint64(txnData.OperationType))...)
-	data = append(data, BoolToByte(txnData.CancelExistingOrder))
+	data = append(data, txnData.CancelOrderID.ToBytes()...)
```
lazynina, line 5621: If we wanted to support a nil value for `CancelOrderID` I think we can use the scheme that we use for the `PostHashToModify` in `SubmitPostMetadata.ToBytes()` function

lib/block_view_dao_coin_limit_order.go
```diff
@@ -289,102 +317,22 @@ func (bav *UtxoView) _connectDAOCoinLimitOrder(
 
 	// Create entry from txn metadata for the transactor.
 	transactorOrder := &DAOCoinLimitOrderEntry{
-		TransactorPKID:                            transactorPKIDEntry.PKID,
-		BuyingDAOCoinCreatorPKID:                  buyCoinPKIDEntry.PKID,
-		SellingDAOCoinCreatorPKID:                 sellCoinPKIDEntry.PKID,
+		OrderID:                   txHash,
```
diamondhands0, line 320: Make 100% sure you've modified every instantiation of DAOCoinLimitOrderEntry to now include the OrderID. I usually do this by changing the name of the struct and finding compile errors. E.g. I'll rename its definition to DAOCoinLimitOrderEntryy and then fix every instance manually, then Refactor>Rename it back. Guarantees that every instance is fixed.

lib/block_view_dao_coin_limit_order.go
```diff
@@ -289,102 +317,22 @@ func (bav *UtxoView) _connectDAOCoinLimitOrder(
...
-		bav._deleteDAOCoinLimitOrderEntryMappings(prevTransactorOrder)
-
-		// Update quantity to add that of the previous order at this level.
-		transactorOrder.QuantityToFillInBaseUnits, err = SafeUint256().Add(
-			transactorOrder.QuantityToFillInBaseUnits,
-			prevTransactorOrder.QuantityToFillInBaseUnits)
-		if err != nil {
-			return 0, 0, nil, errors.Wrapf(err, "Error updating order quantity: ")
-		}
-	}
```
diamondhands0, line 386: LOVE this deletion <3

lib/block_view_dao_coin_limit_order.go
```diff
@@ -1468,68 +1432,18 @@ func (bav *UtxoView) GetAllDAOCoinLimitOrdersForThisTransactor(transactorPKID *P
...
 // ###########################
 // ## VALIDATIONS
 // ###########################
 
-func (bav *UtxoView) IsValidDAOCoinLimitOrder(order *DAOCoinLimitOrderEntry, isCancelOrder bool) error {
+func (bav *UtxoView) IsValidDAOCoinLimitOrder(order *DAOCoinLimitOrderEntry) error {
 	// Returns an error if the input order is invalid. Otherwise returns nil.
 
+	// Validate OrderID.
+	if order.OrderID == nil || order.OrderID.IsZeroBlockHash() {
```
diamondhands0, line 1443: Do you need the nil check? If so, make sure to add it above as well.

lib/block_view_types.go
```diff
@@ -1809,20 +1818,17 @@ func ComputeBaseUnitsToSellUint256(
...
 type DAOCoinLimitOrderMapKey struct {
-	TransactorPKID                            PKID
-	BuyingDAOCoinCreatorPKID                  PKID
-	SellingDAOCoinCreatorPKID                 PKID
-	ScaledExchangeRateCoinsToSellPerCoinToBuy uint256.Int
-	BlockHeight                               uint32
+	OrderID        BlockHash
+	TransactorPKID PKID
```
diamondhands0, line 1826: Do you even need TransactorPKID?

lib/db_utils.go
```diff
@@ -304,14 +305,23 @@ var (
...
-	// > -> <DAOCoinLimitOrder>
+	//   OrderID [32]byte
+	// > -> <DAOCoinLimitOrderEntry>
+	//
+	// This index allows users to query for a single order by ID.
+	// This is useful in e.g. cancelling an order.
+	// <
+	//   _PrefixDAOCoinLimitOrderByOrderID
+	//   OrderID [32]byte
+	// > -> <DAOCoinLimitOrderEntry>
```
diamondhands0, line 316: Interesting. I was thinking you wouldn't need the entries here anymore, and could just "normalize" it to make it so you just look it up from the OrderID index when you need it. But actually I like what you did here because it keeps the lookups fast and requires minimal code changes. Just make sure the code that updates these three indexes always keeps things 100% consistent. That means updating all 3 indexes whenever you modify a single order.

lazynina (changes requested):

lib/db_utils.go
```diff
@@ -6261,12 +6279,14 @@ func DBGetMatchingDAOCoinLimitOrders(
...
 	//   * Set BlockHeight to 0 as this becomes math.MaxUint32 in the key.
-	//   * Set TransactorPKID to MaxPKID to make sure we seek from just beyond the last entry.
+	//   * Set TransactorPKID to MaxPKID.
+	//   * Set OrderID to MaxBlockHash.
 	queryOrder.SellingDAOCoinCreatorPKID = inputOrder.BuyingDAOCoinCreatorPKID
 	queryOrder.BuyingDAOCoinCreatorPKID = inputOrder.SellingDAOCoinCreatorPKID
 	queryOrder.ScaledExchangeRateCoinsToSellPerCoinToBuy = MaxUint256.Clone()
 	queryOrder.BlockHeight = uint32(0)
 	queryOrder.TransactorPKID = MaxPKID.NewPKID()
+	queryOrder.OrderID = NewBlockHash(MaxUint256.Bytes())
```
mattfoley8, line 6289: @diamondhands0 @lazynina What do we think about this as a semi-hack to get the MaxBlockHash? I noticed both uint256s and BlockHashs are 32 bytes. Genius? Or too hacky?

lazynina: A little hacky - I think it would be more consistent to use the `maxHash` defined in blockchain.go

lazynina (approved):
Just change the max blockhash please

lib/db_utils.go
mattfoley8: Updated. :+1:

PR #342, opened 2022-04-21
Mf/add limit order metadata order type field

mattfoley8:


lib/postgres.go
```diff
@@ -1308,6 +1309,8 @@ func (postgres *Postgres) InsertTransactionsTx(tx *pg.Tx, desoTxns []*MsgDeSoTxn
 				SellingDAOCoinCreatorPublicKey:            txMeta.SellingDAOCoinCreatorPublicKey,
 				ScaledExchangeRateCoinsToSellPerCoinToBuy: txMeta.ScaledExchangeRateCoinsToSellPerCoinToBuy.Hex(),
 				QuantityToFillInBaseUnits:                 txMeta.QuantityToFillInBaseUnits.Hex(),
+				OperationType:                             uint64(txMeta.OperationType),
```
lazynina, line 1312: good catch

migrate/20220309224100_create_dao_coin_limit_order_tables.go
```diff
@@ -19,6 +19,7 @@ func init() {
 				scaled_exchange_rate_coins_to_sell_per_coin_to_buy TEXT,
 				quantity_to_fill_in_base_units                     TEXT,
 				operation_type                                     BIGINT,
+				order_type                                         BIGINT,
```
lazynina, line 22: doesn't really have to be a big int? smallint is probably most efficient for operation_type and order_type

lib/network.go
```diff
@@ -5608,6 +5609,14 @@ type DAOCoinLimitOrderMetadata struct {
 	FeeNanos uint64
 }
 
+type DAOCoinLimitOrderType uint64
```
lazynina, line 5612: we can probably make this a uint8?

iamsofonias:
Looks good, leaving to Nina to stamp. 

One question though: should we also add validation in the transaction metadata construction for FillOrKill orders (https://github.com/deso-protocol/core/blob/main/lib/blockchain.go#L3169)? Ex: if the transaction we want to construct is a FillOrKill order, but there aren't enough bidder inputs to populate an executable order, then we know the order submission will fail. The API can return the appropriate error before the order is even submitted in this case. This avoids the roundtrip of signing and submitting a transaction that will fail, and it feels like this it'll help avoid confusing for devs later on.

lib/block_view_dao_coin_limit_order_test.go
```diff
@@ -1411,7 +1411,7 @@ func TestDAOCoinLimitOrder(t *testing.T) {
 		m0DESOBalanceNanosAfter = _getBalance(t, chain, mempool, m0Pub)
 		m1DESOBalanceNanosAfter := _getBalance(t, chain, mempool, m1Pub)
 		require.Equal(m0DESOBalanceNanosBefore, m0DESOBalanceNanosAfter)
-		require.Equal(m1DESOBalanceNanosBefore-uint64(34), m1DESOBalanceNanosAfter) // Hard-coded fee.
+		require.Equal(m1DESOBalanceNanosBefore-uint64(35), m1DESOBalanceNanosAfter) // Hard-coded fee.
```
lazynina, line 1414: I *think* we can get the fee from DAO coin limit order txns: 
```
testMeta.txns[len(testMeta.txns)-1].TxnMeta.(*DAOCoinLimitOrderMetadata).FeeNanos
```

maybe we write a little func to get this so we can get rid of hard coded fees?

lazynina: alternatively, we update all the helpers like `_doDAOCoinLimitOrderTxn` to return fees and then append them to a list of fees on TestMeta?

lazynina (changes requested):
Also add some validation in the txn construction function?

lib/block_view_dao_coin_limit_order_test.go
mattfoley8: I went w/ your first suggestion. See most recent commit.

lib/block_view_dao_coin_limit_order_test.go
```diff
@@ -1647,10 +1652,11 @@ func TestDAOCoinLimitOrder(t *testing.T) {
 		_, _, _, _, err = _connectDAOCoinLimitOrderTxn(
 			testMeta, m1Pub, m1Priv, currentTxn, totalInputMake)
 		require.Error(err)
-		require.Contains(err.Error(), RuleErrorDAOCoinLimitOrderFeeNanosOverflow)
+		require.Contains(err.Error(), RuleErrorDAOCoinLimitOrderFeeNanosBelowMinTxFee)
```
mattfoley8, line 1655: This is the better error message when someone tries to set FeeNanos to zero $DESO so this is a good change IMO.

lazynina (approved):

lazynina (approved):

PR #347, opened 2022-04-29
Support DAOCoinLimitOrder cancels for derived keys

iamsofonias:
Derived keys can't cancel DAO coin limit orders since the spending limit lookup requires the buying & selling coin public keys to always be specified in the transaction metadata. We side-step this by first using order id to look up the order entry, and using the buying & selling pkids from there.

It doesn't appear we need any changes in the disconnect logic. The affected transactions disconnect as expected in the tests.

### DIFF
--- a/lib/block_view.go
+++ b/lib/block_view.go
@@ -1599,9 +1599,24 @@ func (bav *UtxoView) _checkDerivedKeySpendingLimit(
 		}
 	case TxnTypeDAOCoinLimitOrder:
 		txnMeta := txn.TxnMeta.(*DAOCoinLimitOrderMetadata)
+		var buyingCoinPublicKey []byte
+		var sellingCoinPublicKey []byte
+		if txnMeta.CancelOrderID != nil {
+			orderEntry, err := bav._getDAOCoinLimitOrderEntry(txnMeta.CancelOrderID)
+			if err != nil || orderEntry == nil {
+				return utxoOpsForTxn, errors.Wrapf(
+					RuleErrorDerivedKeyInvalidDAOCoinLimitOrderOrderID,
+					"_checkDerivedKeySpendingLimit: Invalid DAO coin limit order ID %v",
+					txnMeta.CancelOrderID)
+			}
+			buyingCoinPublicKey = bav.GetPublicKeyForPKID(orderEntry.BuyingDAOCoinCreatorPKID)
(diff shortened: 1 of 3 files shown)


lib/block_view.go
```diff
@@ -1599,9 +1599,24 @@ func (bav *UtxoView) _checkDerivedKeySpendingLimit(
...
+		var sellingCoinPublicKey []byte
+		if txnMeta.CancelOrderID != nil {
+			orderEntry, err := bav._getDAOCoinLimitOrderEntry(txnMeta.CancelOrderID)
+			if err != nil || orderEntry == nil {
+				return utxoOpsForTxn, errors.Wrapf(
+					RuleErrorDerivedKeyInvalidDAOCoinLimitOrderOrderID,
+					"_checkDerivedKeySpendingLimit: Invalid DAO coin limit order ID %v",
+					txnMeta.CancelOrderID)
+			}
+			buyingCoinPublicKey = bav.GetPublicKeyForPKID(orderEntry.BuyingDAOCoinCreatorPKID)
```
mattfoley8, line 1612: There's a small edge case here where someone provides a CancelOrderID and a BuyingDAOCoinCreatorPublicKey and/or SellingDAOCoinCreatorPublicKey in their metadata. We ignore the Buying and Selling DAOCoinCreatorPublicKeys in that case. IMO that's fine. One could argue that that's confusing for the user, but maybe, maybe not that's true. Alternatively, we could throw a new error like cannot specify buying public key when cancelling an order. And the same for selling public key.

iamsofonias: Good point, throwing an error clearly signals to the user that the buying and selling coin public keys will never be used. However, this should defer to whatever the behavior is in [_connectDAOCoinLimitOrder](https://github.com/deso-protocol/core/blob/main/lib/block_view_dao_coin_limit_order.go#L282). There, we quietly ignore the buying and selling coin fields if the CancelOrderID field is provided.

IMO, any transaction that goes through when signed by the user's private key should go through when signed by a valid derived key (and vice versa). Similarly, any transaction that's rejected when signed by one should be rejected when signed by the other. Inconsistency here would be a bigger point of confusion.

Adding the check and throwing the error in _connectDAOCoinLimitOrder would keep the two paths consistent, but do we really want to introduce new behavior there?

diamondhands0 (approved):

PR #357, opened 2022-05-17
Fix RawDecodeWithoutMetadata for NFTBidEntryBundle

lazynina:
Currently, we initialize a slice of BidEntries with the expected length and then append to this slice. This results in a slice that is twice the length expected with the first half of the slice as nil BidEntries. 

To resolve this, we set the index in the slice to the decoded bid entry directly.


AeonSw4n:
Damn, how could I miss this!? Great find @lazynina

PR #359, opened 2022-05-20
Update NinjaNode => VoSocial App

AdonousTech:


AdonousTech:
Should I submit as a new node?

lazynina:
> Should I submit as a new node?

sorry for the delay here - been swamped with DAODAO the past week. I just looked at the commit history and saw you added this node originally, so it's okay to update the name.

PR #377, opened 2022-06-28
Better error message for DAO coin limit order transfer restriction.

mattfoley8:


lib/blockchain.go
```diff
@@ -3539,15 +3539,15 @@ func (bc *Blockchain) CreateDAOCoinLimitOrderTxn(
 	}
 
 	// Validate txn metadata.
-	err = utxoView.IsValidDAOCoinLimitOrderMetadata(txn.PublicKey, metadata)
+	blockHeight := bc.blockTip().Height + 1
+	err = utxoView.IsValidDAOCoinLimitOrderMetadata(txn.PublicKey, blockHeight, metadata)
```
iamsofonias, line 3543: Surfacing the new error message here doesn't require a hard fork because it's happening during transaction construction. We don't want to put this operation behind the fork height

mattfoley8: Good point. It's true. But IMO the benefit of reusing our existing validation functions (which run during txn construction and connection) outweighs the benefit of validating separately (without considering fork height) during txn construction in the (hopefully short) interim time between now and when we hit our fork heights.

lazynina (approved):

PR #379, opened 2022-06-30
Metamask-compatible access bytes encoding

AeonSw4n:


lib/block_view_derived_key.go
```diff
@@ -49,12 +50,42 @@ func _verifyAccessSignatureWithTransactionSpendingLimit(ownerPublicKey []byte, d
...
+}
+
+func AssembleAccessBytesWithMetamaskStrings(derivedPublicKey []byte, expirationBlock uint64,
+	transactionSpendingLimit *TransactionSpendingLimit, params *DeSoParams) []byte {
+
+	encodingString := "DECENTRALIZED SOCIAL\n\n"
+	encodingString += "Your derived public key: " + Base58CheckEncode(derivedPublicKey, false, params) + "\n\n"
+	encodingString += "The expiration block of your key: " + strconv.FormatUint(expirationBlock, 10) + "\n\n"
+	encodingString += transactionSpendingLimit.ToMetamaskString(params)
+	return []byte(encodingString)
```
lazynina, line 88: @AeonSw4n - can you explain how this new access sig works? I think some additional comments here would be helpful for future maintenance. I know some of this lives in identity, but core should have the requisite knowledge documented here as well.

lib/block_view_profile.go
```diff
@@ -1005,10 +1005,8 @@ func _verifyEthPersonalSignature(signer, data, signature []byte) error {
 	}
 
 	// Change the data bytes into Ethereum's personal_sign message standard. This will prepend the message prefix and hash
-	// the prepended message using keccak256. We turn data into a hex string and treat it as a character sequence which is
-	// how MetaMask treats it.
-	dataHex := hex.EncodeToString(data)
-	hash, _ := TextAndHash([]byte(dataHex))
+	// the prepended message using keccak256.
+	hash, _ := TextAndHash(data)
```
lazynina, line 1009: I'm assuming this change is safe since this is a new feature, but basically we don't need to hex encode the data?

lib/network.go
```diff
@@ -4774,6 +4775,161 @@ type TransactionSpendingLimit struct {
...
+	// TransactionCountLimitMap
+	if len(tsl.TransactionCountLimitMap) > 0 {
+		var txnCountStr []string
+		str += _indt(indentationCounter) + "Transaction Count Limit: \n"
+		indentationCounter++
+		for txnType, limit := range tsl.TransactionCountLimitMap {
+			txnCountStr = append(txnCountStr, _indt(indentationCounter)+txnType.String()+": "+
+				strconv.FormatUint(limit, 10)+"\n")
+		}
+		// Ensure deterministic ordering of the transaction count limit strings by doing a lexicographical sort.
```
lazynina, line 4801: :+1:

lib/network.go
```diff
@@ -4774,6 +4775,161 @@ type TransactionSpendingLimit struct {
 	DAOCoinLimitOrderLimitMap map[DAOCoinLimitOrderLimitKey]uint64
 }
 
+func (tsl *TransactionSpendingLimit) ToMetamaskString(params *DeSoParams) string {
```
lazynina, line 4778: This function is kinda gnarly, but I get why we have to do it. Can we add a comment above this function to explain the "why" here?

lib/network.go
```diff
@@ -4774,6 +4775,161 @@ type TransactionSpendingLimit struct {
...
+			indentationCounter--
+
+			opString += _indt(indentationCounter) + "]\n"
+			creatorCoinLimitStr = append(creatorCoinLimitStr, opString)
+		}
+		// Ensure deterministic ordering of the transaction count limit strings by doing a lexicographical sort.
+		sort.Strings(creatorCoinLimitStr)
+		for _, limitStr := range creatorCoinLimitStr {
+			str += limitStr
+		}
```
lazynina, line 4833: we could probably refactor this little snippet so we don't repeat it in each loop

lib/network.go
```diff
@@ -4774,6 +4775,161 @@ type TransactionSpendingLimit struct {
...
+	// CreatorCoinOperationLimitMap
+	if len(tsl.CreatorCoinOperationLimitMap) > 0 {
+		var creatorCoinLimitStr []string
+		str += _indt(indentationCounter) + "Creator Coin Operation Limits:\n"
+		indentationCounter++
+		for limitKey, limit := range tsl.CreatorCoinOperationLimitMap {
+			opString := _indt(indentationCounter) + "[\n"
+
+			indentationCounter++
+			opString += _indt(indentationCounter) + "Creator Public Key: " +
```
lazynina, line 4818: technically this isn't the creator pub key, but rather the PKID. MOST of the time it will be the same, but should the message reflect that it's a PKID?

lazynina: applies to all uses of PKID below as well.

lib/network_test.go
```diff
@@ -1294,3 +1297,198 @@ func TestDecodeBlockVersion0(t *testing.T) {
...
+	// Generate a random GlobalDESOLimit field.
+	_populateTotalDESOLimit := func() uint64 {
+		return rand.Uint64()
+	}
+	// Generate a random TransactionCountLimitMap field.
+	_populateTransactionCountLimitMap := func(operationCount int) map[TxnType]uint64 {
+		operationMap := make(map[TxnType]uint64)
+
+		for ; operationCount > 0; operationCount-- {
+			txnTyp := TxnType(uint8(rand.Int()%maxTxnType + 1))
```
lazynina, line 1326: you're not guaranteed to have 26 txn types with this since you may overwrite existing ones. Is this okay?

lib/network_test.go
```diff
@@ -1294,3 +1297,198 @@ func TestDecodeBlockVersion0(t *testing.T) {
...
+				SellingDAOCoinCreatorPKID: *NewPKID(RandomBytes(int32(PublicKeyLenCompressed))),
+			}
+			operationMap[randomDAOLimitOperation] = rand.Uint64()
+		}
+		return operationMap
+	}
+
+	// Test encoding of all possible combinations of TransactionSpendingLimit fields.
+	_runTestOnSpendingLimit := func(spendingLimit *TransactionSpendingLimit, params *DeSoParams) bool {
+		fmt.Println(spendingLimit.ToMetamaskString(params))
```
lazynina, line 1387: we should probably remove this print statement before this goes into main

lazynina (commented):
Overall looks good, just think some additional comments to clarify these changes would be helpful for future maintenance

lib/network.go
```diff
@@ -4774,6 +4775,161 @@ type TransactionSpendingLimit struct {
 	DAOCoinLimitOrderLimitMap map[DAOCoinLimitOrderLimitKey]uint64
 }
 
+func (tsl *TransactionSpendingLimit) ToMetamaskString(params *DeSoParams) string {
+	var str string
+	var indentationCounter int
+	_indt := _computeIndentation
```
lazynina, line 4781: can't we just rename the function define below? I don't love the idea of renaming a function inside another

lib/network_test.go
lazynina, line 1387:
```suggestion
```

lazynina (changes requested):

lib/network_test.go
AeonSw4n: haha okay, good find. This was me being lazy, but gotchu, it's fixed now.

lib/network.go
AeonSw4n: yup, makes sense

lib/block_view_profile.go
AeonSw4n: This was there before because metamask signs data as a string instead of a byte array. Previously metamask would sign the hex encoding of the data, while now we just sign the metamask string.

lib/block_view_message.go
```diff
@@ -814,6 +928,23 @@ func (bav *UtxoView) _connectMessagingGroup(
 	// It's worth noting that we index messaging keys by the owner public key and messaging key name.
 	existingEntry := bav.GetMessagingGroupKeyToMessagingGroupEntryMapping(messagingGroupKey)
 
+	// We will update the existing entry if it exists, or otherwise create a new utxoView entry. The new entry can currently
```
AeonSw4n, line 931: Think about base key group chats (above)

lib/network.go
```diff
@@ -5977,6 +6372,16 @@ func DeserializePubKeyToUint64Map(data []byte) (map[PublicKey]uint64, error) {
 // MessagingGroupMetadata
 // ==================================================================
 
+// MessagingGroupOperation represents V3 Group Chat Messages ExtraData["MessagingGroupOperationType"] values
+type MessagingGroupOperation byte
+
+const (
```
AeonSw4n, line 6378: add some comments

lazynina (approved):

PR #380, opened 2022-06-30
Public key recovery in signature verification

AeonSw4n:
Metamask thing. Needs Postgres tests! (in #384)

### DIFF
--- a/lib/block_view.go
+++ b/lib/block_view.go
@@ -1123,7 +1123,9 @@ func (bav *UtxoView) _verifySignature(txn *MsgDeSoTxn, blockHeight uint32) (_der
 	txHash := Sha256DoubleHash(txBytes)
 
 	// Look for the derived key in transaction ExtraData and validate it. For transactions
-	// signed using a derived key, the derived public key is passed to ExtraData.
+	// signed using a derived key, the derived public key is passed in ExtraData. Alternatively,
+	// the signature uses DeSo-DER encoding, meaning we can recover the derived public key from
+	// the signature.
 	var derivedPk *btcec.PublicKey
 	derivedPkBytes, isDerived, err := IsDerivedSignature(txn)
 	if err != nil {
@@ -1149,7 +1151,7 @@ func (bav *UtxoView) _verifySignature(txn *MsgDeSoTxn, blockHeight uint32) (_der
 	// If derived key is present in ExtraData, we check if transaction was signed by the derived key.
 	if derivedPk == nil {
 		// Verify that the transaction is signed by the specified key.
-		if txn.Signature.Sign.Verify(txHash[:], ownerPk) {
+		if txn.Signature.Verify(txHash[:], ownerPk) {
 			return nil, nil
 		}
 	} else {
@@ -1187,9 +1189,10 @@ func (bav *UtxoView) _verifySignature(txn *MsgDeSoTxn, blockHeight uint32) (_der
 				blockHeight)
 		}
 
-		// All checks passed so we try to verify the signature.
-		// FIXME: This step can be avoided for DeSo-DER signatures. At this point, we've already recovered and verified the signature.
-		if txn.Signature.Sign.Verify(txHash[:], derivedPk) {
+		// All checks passed so we try to verify the signature. This step can be avoided for DeSo-DER signatures.
+		// At this point, we've already recovered the derived public key and verified the signature.
+		// IT'S SUPER IMPORTANT TO QUADRUPLE-CHECK THIS LOGIC WHEN MODIFYING ANY CODE RELATED TO RECOVERABLE SIGNATURES.
+		if txn.Signature.IsRecoverable || txn.Signature.Verify(txHash[:], derivedPk) {
 			return derivedPk.SerializeCompressed(), nil
 		}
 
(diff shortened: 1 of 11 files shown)


lazynina (approved):
overall looks good - love the amount of comments!

lib/block_view.go
```diff
@@ -1192,12 +1199,45 @@ func (bav *UtxoView) _verifySignature(txn *MsgDeSoTxn, blockHeight uint32) (_der
...
+		derivedPkBytes, isDerived := txn.ExtraData[DerivedPublicKey]
+		if isDerived {
+			// Make sure both encodings aren't used concurrently.
+			if txn.Signature.IsRecoverable {
+				return nil, false, errors.Wrapf(RuleErrorDerivedKeyHasBothExtraDataAndRecoveryId,
+					"IsDerivedSignature: transaction signed with a derived key can either store public key in "+
+						"ExtraData or use the DeSo-DER recoverable signature encoding")
+			}
+			return derivedPkBytes, isDerived, nil
+		}
```
lazynina, line 1218: I would prefer not to nest If statements this deeply, it increase cognitive complexiity.
```suggestion
		// Make sure both encodings aren't used concurrently.
		if isDerived  && txn.Signature.IsRecoverable {
			return nil, false, errors.Wrapf(RuleErrorDerivedKeyHasBothExtraDataAndRecoveryId,
				"IsDerivedSignature: transaction signed with a derived key can either store public key in "+
					"ExtraData or use the DeSo-DER recoverable signature encoding")
		}
		if isDerived {
			return derivedPkBytes, isDerived, nil
		}
```

lib/block_view.go
```diff
@@ -1143,7 +1149,7 @@ func (bav *UtxoView) _verifySignature(txn *MsgDeSoTxn, blockHeight uint32) (_der
 	// If derived key is present in ExtraData, we check if transaction was signed by the derived key.
 	if derivedPk == nil {
 		// Verify that the transaction is signed by the specified key.
-		if txn.Signature.Verify(txHash[:], ownerPk) {
+		if txn.Signature.Sign.Verify(txHash[:], ownerPk) {
```
lazynina, line 1152: instead of updating all the usages of `txn.Signature.Verify` to `txn.Signature.Sign.Verify`, let's create a function on `DeSoSignature` called `Verify`

```
func (desoSignature DeSoSignature) Verify(hash []byte, pubKey *PublicKey) bool {
	return desoSignature.Sign.Verify(hash, pubKey)
}

**Review by lazynina — CHANGES_REQUESTED** [review]

File: lib/block_view.go [hunk 3]
```diff
@@ -1182,7 +1188,8 @@ func (bav *UtxoView) _verifySignature(txn *MsgDeSoTxn, blockHeight uint32) (_der
 		}
 
 		// All checks passed so we try to verify the signature.
-		if txn.Signature.Verify(txHash[:], derivedPk) {
+		// FIXME: This step can be avoided for DeSo-DER signatures. At this point, we've already recovered and verified the signature.
+		if txn.Signature.Sign.Verify(txHash[:], derivedPk) {
```
lazynina, line 1192: Let's resolve the FIXME. I think skipping this is as simple as checking if `IsRecoverable` is false before this and returning if so

lib/network.go
```diff
@@ -2513,6 +2517,148 @@ func (desoOutput *DeSoOutput) GetEncoderType() EncoderType {
...
+	derSigMagicOffset = 0x30
+
+	// derSigMagicMaxRecoveryOffset is the maximal value of the DeSo-DER signature format. We enable public key recovery
+	// from ECDSA signatures. To facilitate this, we add the recovery id to the DER magic 0x30 first byte. The recovery id
+	// is in the range of [0, 3] and corresponds to the compact signature header magic. Adding recovery id to signature
+	// encoding is totally optional and leaving the first byte 0x30 is acceptable. Specifically, the DeSo-DER signatures
+	// have the following format:
+	// <0x30 + optionally (0x01 + recoveryId)> <length of whole message> <0x02> <length of R> <R> 0x2 <length of S> <S>.
+	// At this point, a familiar reader might arrive at some malleability concerns. After all that's why bip-62 enforced
+	// DER signatures. ECDSA malleability is prevented by allowing public key recovery iff it was produced with a derived key.
```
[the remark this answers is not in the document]
AeonSw4n: it's actually "if and only if" so I did iff.

lib/blockchain_test.go
```diff
@@ -762,14 +763,32 @@ func _signTxnWithDerivedKey(t *testing.T, txn *MsgDeSoTxn, privKeyStrArg string)
...
-		txn.ExtraData = make(map[string][]byte)
-	}
-	txn.ExtraData[DerivedPublicKey] = publicKey.SerializeCompressed()
-	txnSignature, err := txn.Sign(privateKey)
-	require.NoError(err)
 
-	txn.Signature = txnSignature
+	// We will randomly sign with the standard DER encoding + ExtraData, or with the DeSo-DER encoding.
+	signatureType := rand.Int() % 2
+	if signatureType == 0 {
```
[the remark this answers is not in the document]
AeonSw4n: It's tricky. It would require modifying many tests, and I don't think it'll increase the coverage by too much. I think the majority of the issues with the signatures would come up immediately, and going with option 3. by default should capture them.

lib/block_view.go
AeonSw4n: I like this suggestion a lot. Dons

lib/block_view.go
AeonSw4n: I'm always +1 on performance optimizations

lib/blockchain_test.go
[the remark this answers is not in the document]
lazynina: That's fine by me then

PR #383, opened 2022-07-05
K/add group chat muting

keshavm02:
DRAFT PR for group chat muting

### DIFF
--- a/lib/block_view_message.go
+++ b/lib/block_view_message.go
@@ -742,6 +742,12 @@ func (bav *UtxoView) _connectMessagingGroup(
 			RuleErrorMessagingKeyBeforeBlockHeight, "_connectMessagingGroup: "+
 				"Problem connecting messaging key, too early block height")
 	}
+
+	// Check that the transaction has the right TxnType.
+	if txn.TxnMeta.GetTxnType() != TxnTypeMessagingGroup {
+		return 0, 0, nil, fmt.Errorf("_connectMessagingGroup: called with bad TxnType %s",
+			txn.TxnMeta.GetTxnType().String())
+	}
 	txMeta := txn.TxnMeta.(*MessagingGroupMetadata)
 
 	// If the key name is just a list of 0s, then return because this name is reserved for the base key.
--- a/lib/block_view_message.go
+++ b/lib/block_view_message.go
@@ -929,6 +929,8 @@ func (bav *UtxoView) _connectMessagingGroup(
 		extraData = mergeExtraData(existingExtraData, txn.ExtraData)
 	}
 
+	muteList := existingEntry.MuteList
+
 	// TODO: Currently, it is technically possible for any user to add *any other* user to *any group* with
 	// a garbage EncryptedKey. This can be filtered out at the app layer, though, and for now it leaves the
 	// app layer with more flexibility compared to if we implemented an explicit permissioning model at the
@@ -940,6 +942,7 @@ func (bav *UtxoView) _connectMessagingGroup(
 		MessagingPublicKey:    messagingPublicKey,
 		MessagingGroupKeyName: NewGroupKeyName(txMeta.MessagingGroupKeyName),
 		MessagingGroupMembers: messagingMembers,
+		MuteList:              muteList,
 		ExtraData:             extraData,
 	}
 	// Create a utxoOps entry, we make a copy of the existing entry.
--- a/lib/block_view_message.go
+++ b/lib/block_view_message.go
@@ -831,6 +831,62 @@ func (bav *UtxoView) _connectMessagingGroup(
 		}
 	}
 
+	// V3 Messages: MUTING and UNMUTING members
+	if blockHeight >= bav.Params.ForkHeights.ExtraDataOnEntriesBlockHeight {
+		//var existingExtraData map[string][]byte
+		if existingEntry != nil && !existingEntry.isDeleted {
+			value, ok := existingEntry.ExtraData["OperationType"]
+			if ok {
+				muteList := existingEntry.MuteList
+
+				if string(value) == "MessagingGroupOperationMute" {
+					for _, s := range txMeta.MessagingGroupMembers {
+						// Add s to muteList
+						muteList = append(muteList, s)
+					}
+				} else if string(value) == "MessagingGroupOperationUnmute" {
+					for _, s := range txMeta.MessagingGroupMembers {
+						// Remove s from muteList
+						for i, toUnmute := range muteList {
+							if toUnmute == s {
+								muteList = append(muteList[:i], muteList[i+1:]...)
+								break
+							}
+						}
+					}
+				}
+				// Create a MessagingGroupEntry so we can add the entry to UtxoView.
+				messagingGroupEntry := MessagingGroupEntry{
+					GroupOwnerPublicKey:   &messagingGroupKey.OwnerPublicKey,
+					MessagingPublicKey:    messagingPublicKey,
+					MessagingGroupKeyName: NewGroupKeyName(txMeta.MessagingGroupKeyName),
+					MessagingGroupMembers: existingEntry.MessagingGroupMembers,
+					MuteList:              muteList,
+					ExtraData:             existingEntry.ExtraData,
+				}
+				// Create a utxoOps entry, we make a copy of the existing entry.
+				var prevMessagingKeyEntry *MessagingGroupEntry
+				if existingEntry != nil && !existingEntry.isDeleted {
+					prevMessagingKeyEntry = &MessagingGroupEntry{}
+					rr := bytes.NewReader(EncodeToBytes(uint64(blockHeight), existingEntry))
+					if exists, err := DecodeFromBytes(prevMessagingKeyEntry, rr); !exists || err != nil {
+						return 0, 0, nil, errors.Wrapf(err,
+							"_connectMessagingGroup: Error decoding previous entry")
+					}
+				}
+				bav._setMessagingGroupKeyToMessagingGroupEntryMapping(&messagingGroupKey.OwnerPublicKey, &messagingGroupEntry)
+
+				// Construct UtxoOperation.
+				utxoOpsForTxn = append(utxoOpsForTxn, &UtxoOperation{
+					Type:                  OperationTypeMessagingKey,
+					PrevMessagingKeyEntry: prevMessagingKeyEntry,
+				})
+
+				return totalInput, totalOutput, utxoOpsForTxn, nil
+			}
+		}
+	}
+
 	// In DeSo V3 Messages, a messaging key can initialize a group chat with more than two parties. In group chats, all
 	// messages are encrypted to the group messaging public key. The group members are provided with an encrypted
 	// private key of the group's messagingPublicKey so that each of them can read the messages. We refer to
--- a/lib/block_view_types.go
+++ b/lib/block_view_types.go
@@ -1982,6 +1982,11 @@ type MessagingGroupEntry struct {
 	// is given to all group members.
 	MessagingGroupMembers []*MessagingGroupMember
 
+	// MuteList is a list of members that have been currently muted.
+	// Being muted means the member cannot send any messages to the group
+	// but can still cryptographically read new on-chain messages
+	MuteList []*MessagingGroupMember
+
 	// ExtraData is an arbitrary key value map
 	ExtraData map[string][]byte
 


lib/block_view_message.go
```diff
@@ -474,6 +472,26 @@ func (bav *UtxoView) _connectPrivateMessage(
 				"_connectPrivateMessage: at least one messaging party must be present")
 		}
 
+		// Reject message if sender is muted
+		var messagingGroupKey *MessagingGroupKey
+		messagingGroupKey = NewMessagingGroupKey(NewPublicKey(txMeta.RecipientPublicKey), txn.ExtraData[RecipientMessagingGroupKeyName])
```
AeonSw4n, line 477: At this point in the function, we know that either `existsSender` or `existsRecipient` is true (L#469) but we don't know if it's the recipient, or the sender, or both. When you're referring to for example `txn.ExtraData[RecipientMessagingGroupKeyName]`, we're assuming it's not empty, which it might actually be. Check out the validation logic for the recipient and the sender on lines L#498 and L#510. I think the mutelist requires both the recipient and sender present in ExtraData. In this case you should move your code after these two conditions and gate it with a condition of `existsSender && existsSenderName && existsRecipient && existsRecipientName`.

lib/block_view_message.go
```diff
@@ -825,6 +849,71 @@ func (bav *UtxoView) _connectMessagingGroup(
 		}
 	}
 
+	// V3 Messages: MUTING and UNMUTING members
+	if blockHeight >= bav.Params.ForkHeights.DeSoV3MessagesBlockHeight {
```
AeonSw4n, line 853: Oh you added a blockheight :) We would actually need to add a new future blockheight for your change to work

lib/block_view_message.go
```diff
@@ -825,6 +849,71 @@ func (bav *UtxoView) _connectMessagingGroup(
...
+				contains := func(memberList []*MessagingGroupMember, member *MessagingGroupMember) bool {
+					for _, a := range memberList {
+						if a == member {
+							return true
+						}
+					}
+					return false
+				}
+
+				if string(value) == "MessagingGroupOperationMute" {
```
AeonSw4n, line 867: nit: can we make `MessagingGroupOperationMute` a constant? Same with unmute.

lib/block_view_message.go
```diff
@@ -825,6 +849,71 @@ func (bav *UtxoView) _connectMessagingGroup(
 		}
 	}
 
+	// V3 Messages: MUTING and UNMUTING members
+	if blockHeight >= bav.Params.ForkHeights.DeSoV3MessagesBlockHeight {
+		if existingEntry != nil && !existingEntry.isDeleted {
+			value, operationTypeExists := txn.ExtraData["OperationType"]
```
AeonSw4n, line 855: nit: Can we make "OperationType" a constant? We usually put ExtraData constants towards the bottom of constants.go

AeonSw4n (commented):
This looks very solid, great job dude! I think after this round we should be at the finish line.

lib/block_view_message.go
keshavm02: good catch, done

lib/block_view_message.go
keshavm02: done

lib/block_view_message.go
keshavm02: not sure what that means, like is there an ever higher blockHeight requirement needed for this because this technically happened after V3 started?

lib/block_view_message.go
```diff
@@ -1007,6 +1109,64 @@ func (bav *UtxoView) _disconnectMessagingGroup(
 		}
 	}
 
+	// At this point, we know that prevMessagingKeyEntry exists
+	// UNDO MuteList Operation if applicable
+	if blockHeight >= bav.Params.ForkHeights.ExtraDataOnEntriesBlockHeight {
```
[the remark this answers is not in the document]
keshavm02: right, that was my guess too but wrote this just in case

lib/block_view_message.go
```diff
@@ -474,6 +472,26 @@ func (bav *UtxoView) _connectPrivateMessage(
...
+		messagingGroupEntry := bav.GetMessagingGroupKeyToMessagingGroupEntryMapping(messagingGroupKey)
+		if messagingGroupEntry != nil {
+			muteList := messagingGroupEntry.MuteList
+			if err := IsByteArrayValidPublicKey(senderMessagingPublicKey); err != nil {
+				return 0, 0, nil, errors.Wrapf(
+					RuleErrorPrivateMessageParsePubKeyError, "_connectPrivateMessage: Parse error: %v", err)
+			}
+			// TODO: Make the following more efficient by retrieving MuteList from hacked MessagingGroupEntry to avoid fetching bulky MessagingGroupEntry
+			for _, mutedMember := range muteList {
+				if reflect.DeepEqual(mutedMember.GroupMemberPublicKey, NewPublicKey(senderMessagingPublicKey)) {
```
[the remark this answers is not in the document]
keshavm02: ahh makes sense, thanks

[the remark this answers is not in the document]
keshavm02: done

lib/block_view_message.go
keshavm02: agreed, done

lib/block_view_message.go
```diff
@@ -825,6 +856,76 @@ func (bav *UtxoView) _connectMessagingGroup(
 		}
 	}
 
+	// V3 Messages: MUTING and UNMUTING members
+	if blockHeight >= bav.Params.ForkHeights.DeSoV3MessagesMutingBlockHeight {
+		if existingEntry != nil && !existingEntry.isDeleted {
+			value, operationTypeExists := txn.ExtraData["OperationType"]
+			if operationTypeExists {
+				// inline helper method
+				contains := func(memberList []*MessagingGroupMember, member *MessagingGroupMember) bool {
```
AeonSw4n, line 865: Since we're using `contains` only once, maybe we just write that logic directly in place where we were supposed to use `contains`?

lib/block_view_message.go
```diff
@@ -825,6 +856,76 @@ func (bav *UtxoView) _connectMessagingGroup(
...
+							return true
+						}
+					}
+					return false
+				}
+				// make deep copy of existingEntry.MuteList to prevent mempool errors
+				muteListCopy := make([]*MessagingGroupMember, len(existingEntry.MuteList))
+				copy(muteListCopy, existingEntry.MuteList)
+				// check if mute or unmute
+				if string(value) == MessagingGroupOperationMute {
```
AeonSw4n, line 882: I wonder if we want to throw an error whenever the member was already in the mutelist or if they weren't muted, wyt?

lib/block_view_message.go
```diff
@@ -825,6 +856,76 @@ func (bav *UtxoView) _connectMessagingGroup(
...
+					GroupOwnerPublicKey:   &messagingGroupKey.OwnerPublicKey,
+					MessagingPublicKey:    messagingPublicKey,
+					MessagingGroupKeyName: NewGroupKeyName(txMeta.MessagingGroupKeyName),
+					MessagingGroupMembers: existingEntry.MessagingGroupMembers,
+					MuteList:              existingEntry.MuteList,
+					ExtraData:             existingEntry.ExtraData,
+				}
+				// Create a utxoOps entry, we make a copy of the existing entry.
+				var prevMessagingKeyEntry *MessagingGroupEntry
+				if existingEntry != nil && !existingEntry.isDeleted {
```
AeonSw4n: I think we already checked for this condition?

lib/block_view_message.go
```diff
@@ -825,6 +856,76 @@ func (bav *UtxoView) _connectMessagingGroup(
...
+				}
+				bav._setMessagingGroupKeyToMessagingGroupEntryMapping(&messagingGroupKey.OwnerPublicKey, &messagingGroupEntry)
+
+				// Construct UtxoOperation.
+				utxoOpsForTxn = append(utxoOpsForTxn, &UtxoOperation{
+					Type:                  OperationTypeMessagingKey,
+					PrevMessagingKeyEntry: prevMessagingKeyEntry,
+				})
+
+				return totalInput, totalOutput, utxoOpsForTxn, nil
```
AeonSw4n, line 930: Hmm I don't love how this is duplicated, maybe we can somehow refactor the code so that we only create a utxoop once? (Nit: don't forget to merge ExtraData)

lib/db_utils.go
```diff
@@ -1680,6 +1680,9 @@ func DBPutMessagingGroupMemberWithTxn(txn *badger.Txn, snap *Snapshot, blockHeig
 		MessagingGroupMembers: []*MessagingGroupMember{
 			messagingGroupMember,
 		},
+		//MuteList: []*MessagingGroupMember{
```
AeonSw4n, line 1683: Should this be removed?

AeonSw4n (commented):
Aaaalmost there!

lib/block_view_message.go
```diff
@@ -825,6 +856,76 @@ func (bav *UtxoView) _connectMessagingGroup(
...
+					for _, a := range memberList {
+						if a == member {
+							return true
+						}
+					}
+					return false
+				}
+				// make deep copy of existingEntry.MuteList to prevent mempool errors
+				muteListCopy := make([]*MessagingGroupMember, len(existingEntry.MuteList))
+				copy(muteListCopy, existingEntry.MuteList)
```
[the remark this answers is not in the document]
keshavm02: fixed

lib/block_view_message.go
keshavm02: yes, fixed.

lib/block_view_message.go
keshavm02: moved `contains` right above its usage, but keeping the function as it helps simplify logic

lib/block_view_message.go
keshavm02: yes, done

lib/block_view_message.go
keshavm02: fixing duplication with if else is causing high complexity and errors, so leaving it be for now

lib/db_utils.go
keshavm02: nope, added and fixed

lib/block_view_message.go
keshavm02: if it ain't broke don't fix it

lib/block_view_message.go
```diff
@@ -474,6 +472,33 @@ func (bav *UtxoView) _connectPrivateMessage(
...
 
+		// Reject message if sender is muted
+		if blockHeight >= bav.Params.ForkHeights.DeSoV3MessagesMutingBlockHeight {
+			_, existsSenderKeyName := txn.ExtraData[SenderMessagingGroupKeyName]
+			recipientMessagingGroupKeyName, existsRecipientKeyName := txn.ExtraData[RecipientMessagingGroupKeyName]
+			// Ensure sender and recipient are both present for a group chat
+			if existsSender && existsSenderKeyName && existsRecipient && existsRecipientKeyName {
+				var messagingGroupKey *MessagingGroupKey
+				messagingGroupKey = NewMessagingGroupKey(NewPublicKey(txMeta.RecipientPublicKey), recipientMessagingGroupKeyName)
+				messagingGroupEntry := bav.GetMessagingGroupKeyToMessagingGroupEntryMapping(messagingGroupKey)
```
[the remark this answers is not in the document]
keshavm02: optimization done!

lib/block_view_message.go
```diff
@@ -474,6 +472,33 @@ func (bav *UtxoView) _connectPrivateMessage(
...
+				if messagingGroupEntry != nil {
+					muteList := messagingGroupEntry.MuteList
+					if err := IsByteArrayValidPublicKey(senderMessagingPublicKey); err != nil {
+						return 0, 0, nil, errors.Wrapf(
+							RuleErrorPrivateMessageParsePubKeyError, "_connectPrivateMessage: Parse error: %v", err)
+					}
+					// TODO: Make the following more efficient by retrieving MuteList from hacked MessagingGroupEntry to avoid fetching bulky MessagingGroupEntry
+					for _, mutedMember := range muteList {
+						if reflect.DeepEqual(mutedMember.GroupMemberPublicKey[:], senderMessagingPublicKey) {
```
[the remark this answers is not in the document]
keshavm02: yes fixed

lib/block_view_message.go
```diff
@@ -474,6 +472,33 @@ func (bav *UtxoView) _connectPrivateMessage(
 				"_connectPrivateMessage: at least one messaging party must be present")
 		}
 
+		// Reject message if sender is muted
+		if blockHeight >= bav.Params.ForkHeights.DeSoV3MessagesMutingBlockHeight {
```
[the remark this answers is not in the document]
keshavm02: done and cleaned up

lib/block_view_message.go
```diff
@@ -825,6 +863,101 @@ func (bav *UtxoView) _connectMessagingGroup(
...
+							entryCopy.MuteList = append(entryCopy.MuteList, s)
+						} else {
+							return 0, 0, nil, errors.Wrapf(RuleErrorMessagingMemberAlreadyMuted,
+								"_connectMessagingGroup: Cannot mute member that is already muted (%v).", s.GroupMemberPublicKey[:])
+						}
+					}
+				} else if string(value) == MessagingGroupOperationUnmute {
+					for _, s := range txMeta.MessagingGroupMembers {
+						isUnmuteValid := false
+						for i, toUnmute := range entryCopy.MuteList {
```
AeonSw4n, line 913: This loop is tricky as discussed

lib/block_view_message.go
AeonSw4n: ah, it would be reaaally nice if we could deduplicate this

lib/block_view_message.go
```diff
@@ -825,6 +863,101 @@ func (bav *UtxoView) _connectMessagingGroup(
 		}
 	}
 
+	// V3 Messages: MUTING and UNMUTING members
+	// Check if blockHeight lower than threshold and still trying to mute
+	if blockHeight < bav.Params.ForkHeights.DeSoV3MessagesMutingAndPrefixOptimizationBlockHeight {
```
AeonSw4n, line 868: nit: how about we consolidate and refactor this `if existingEntry != nil && !existingEntry.isDeleted` -> `blockHeight < bav.Params...` || `blockHeight >= bav params....`?

AeonSw4n (commented):
Let's fix that loop thing and deduplicate connectMessagingGroup logic.

lib/block_view_message.go
```diff
@@ -498,6 +496,40 @@ func (bav *UtxoView) _connectPrivateMessage(
 			messageEntry.RecipientMessagingPublicKey = NewPublicKey(recipientMessagingPublicKey)
 			messageEntry.RecipientMessagingGroupKeyName = NewGroupKeyName(recipientMessagingKeyName)
 		}
+
+		// Reject message if sender is muted
+		if blockHeight >= bav.Params.ForkHeights.DeSoV3MessagesMutingAndPrefixOptimizationBlockHeight {
+			// Ensure sender and recipient are both present for a group chat
+			if existsSender && existsSenderName && existsRecipient && existsRecipientName {
```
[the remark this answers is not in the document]
keshavm02: makes sense, done

lib/block_view_message.go
```diff
@@ -825,6 +863,101 @@ func (bav *UtxoView) _connectMessagingGroup(
 		}
 	}
 
+	// V3 Messages: MUTING and UNMUTING members
+	// Check if blockHeight lower than threshold and still trying to mute
+	if blockHeight < bav.Params.ForkHeights.DeSoV3MessagesMutingAndPrefixOptimizationBlockHeight {
+		if existingEntry != nil && !existingEntry.isDeleted {
+			if value, operationTypeExists := txn.ExtraData[MessagingGroupOperationType]; operationTypeExists {
+				if string(value) == MessagingGroupOperationMute || string(value) == MessagingGroupOperationUnmute {
```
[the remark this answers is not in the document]
keshavm02: true, as long as no one else adds another OperationType before `DeSoV3MessagesMutingAndPrefixOptimizationBlockHeight` goes live

lib/block_view_message.go
```diff
@@ -825,6 +863,101 @@ func (bav *UtxoView) _connectMessagingGroup(
...
+						// Make sure does not already exist to ensure no dups
+						contains := func(memberList []*MessagingGroupMember, member *MessagingGroupMember) bool {
+							for _, a := range memberList {
+								if a == member {
+									return true
+								}
+							}
+							return false
+						}
+						if !contains(entryCopy.MuteList, s) {
```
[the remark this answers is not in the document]
keshavm02: done

lib/block_view_message.go
keshavm02: fixed

lib/block_view_message.go
keshavm02: done

lib/block_view_message.go
keshavm02: it was tedious but done :)

AeonSw4n:
Closed in favor of #394

PR #384, opened 2022-07-11
Unlimited spending limit

AeonSw4n:


lib/block_view_test.go
```diff
@@ -489,10 +498,17 @@ func TestBasicTransfer(t *testing.T) {
 	_ = assert
 	_ = require
 
-	chain, params, db := NewLowDifficultyBlockchain()
+	postgres := InitializeTestPostgresInstance(t)
```
mattfoley8, line 501: IMO this logic should be refactored into some test_utils. Individual tests shouldn't have to know the specifics of a postgres instance initialization. Also ideally we would have one test and not two sister tests like `_testBasicTransfer` and `_testBasicTransferWithPostgres`.

lib/db_adapter.go
```diff
@@ -10,6 +10,14 @@ type DbAdapter struct {
 	snapshot   *Snapshot
 }
 
+func NewDbAdapter(chain *Blockchain) *DbAdapter {
```
mattfoley8, line 13: To match below, maybe this makes more sense w/ a signature like `func (chain *Blockchain) GetDbAdapter *DbAdapter {`.

lib/postgres.go
```diff
@@ -2916,3 +2944,50 @@ func (postgres *Postgres) GetNotifications(publicKey string) ([]*PGNotification,
 
 	return notifications, nil
 }
+
+//
+// Postgres Test tooling
```
mattfoley8, line 2949: IMO this is better off in a test_utils or postgres_test file and not here.

lib/postgres.go
```diff
@@ -2916,3 +2944,50 @@ func (postgres *Postgres) GetNotifications(publicKey string) ([]*PGNotification,
...
+//
+
+// Drop all tables in the postgres Db.
+func (postgres *Postgres) resetDatabase() error {
```
mattfoley8, line 2953: We can get this functionality for free with embedded postgres. There's a lot of logic here that assumes a local docker postgres for easier testing. IMO no reason to spin our own solution.

lib/postgres.go
```diff
@@ -2916,3 +2944,50 @@ func (postgres *Postgres) GetNotifications(publicKey string) ([]*PGNotification,
...
+
+	removeOptions := types.ContainerRemoveOptions{
+		Force: true,
+	}
+	err = cli.ContainerRemove(context.Background(), postgres.containerId, removeOptions)
+	if err != nil {
+		return errors.Wrapf(err, "forceKillContainer: Problem removing postgres container")
+	}
+
+	if postgres.directory == "" {
```
mattfoley8, line 2986: why even symlink the postgres data at all. if it's just for testing, may as well not mount the volume IMO.

lib/postgres_test.go
```diff
@@ -1,10 +1,284 @@
...
+func RunTestWithBadgerAndPostgresOptimized(t *testing.T, testFunction func(t *testing.T, postgres *Postgres)) {
+	var postgres *Postgres
+	postgresChannel := make(chan struct{})
+	go func() {
+		postgres = InitializeTestPostgresInstance(t)
+		postgresChannel <- struct{}{}
+	}()
+	testFunction(t, nil)
+	<-postgresChannel
+	testFunction(t, postgres)
```
mattfoley8, line 56: IMO running badger + postgres in parallel like this is of limited value. More useful IMO is to run badger tests in a single CI runner and postgres tests in a parallel runner. I.e. you run the entire suite (instead of each test) over badger or postgres in parallel.

lib/postgres_test.go
```diff
@@ -1,10 +1,284 @@
...
 
+// Set to true if you don't want to use the postgres testing framework.
+const DISABLE_POSTGRES_DOCKER_API = false
+
+// Config variables for spawning the postgres container.
+const CONTAINER_NAME = "test_postgresql"
+const LOCAL_PORT = "5433"
+const REMOTE_PORT = "5432"
+const POSTGRES_USER = "postgres"
+const POSTGRES_PASSWORD = "postgres"
```
mattfoley8, line 30: I also think we lose a lot by defining the postgres params in consts like this and dropping support for the PG_URI env var which is a common way to pass in pg params. The idea may be to encapsulate all postgres logic away from the user so they don't even need to know which params are being used, but what if for example, I'm already running a secondary postgres docker container on port 5433 and need to be able to configure the port.

lib/block_view_derived_key_test.go
```diff
@@ -18,6 +20,108 @@ const (
...
+	require := require.New(t)
+	_ = require
+
+	senderPkBytes, _, err := Base58CheckDecode(senderPkString)
+	require.NoError(err)
+
+	// Verify that expiration block was persisted in the db or is in mempool utxoView
+	if mempool == nil {
+		var derivedKeyEntry *DerivedKeyEntry
+		if chain.postgres != nil {
```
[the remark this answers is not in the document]
AeonSw4n: Good call!

lib/db_adapter.go
lazynina: agreed with mf ^

lib/network.go
```diff
@@ -4772,9 +4916,173 @@ type TransactionSpendingLimit struct {
 	// BuyingCreatorPKID || SellingCreatorPKID to number of
 	// transactions
 	DAOCoinLimitOrderLimitMap map[DAOCoinLimitOrderLimitKey]uint64
+
+	// ===== ENCODER MIGRATION UnlimitedDerivedKeysMigration =====
+	// IsUnlimited field determines whether this derived key has no spending limit.
```
lazynina, line 4921:
```suggestion
	// IsUnlimited field determines whether this derived key can perform any transaction on behalf of its owner and does not require a no spending limit.
```
let's be more clear about what `IsUnlimited` means

lib/postgres.go
```diff
@@ -2,22 +2,31 @@ package lib
...
 	"regexp"
 	"strings"
 )
 
 type Postgres struct {
 	db *pg.DB
+
+	// Only applies to a docker postgres db.
+	directory   string
+	containerId string
```
lazynina, line 29: I don't think we really need this for a postgres db?

lib/postgres.go
lazynina: agreed with MF^

lib/postgres.go
lazynina: agreed with MF^

lib/postgres.go
lazynina: agreed^

lib/postgres_test.go
```diff
@@ -1,10 +1,284 @@
...
+	}
+
+	return &Postgres{
+		db: db,
+	}
+}
+
+// startTestPostgresContainerAndConnect will spawn a postgres container with a volume mapped to some random directory,
+// and port-forwarded 0.0.0.0 : LOCAL_PORT <-> remote : REMOTE_PORT where REMOTE_PORT coincides with the postgres daemon.
+func startTestPostgresContainerAndConnect(t *testing.T) *Postgres {
```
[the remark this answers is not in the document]
lazynina: ^^yes

lazynina (changes requested):
The unlimited spending limits updates here seem ok, but the postgres testing changes should be broken into a separate PR. It makes it far hard to review and I don't think we necessarily want all those the postgres testing changes and don't want to hold up this PR as we debate them.

lib/block_view_derived_key_test.go
```diff
@@ -18,6 +20,100 @@ const (
...
+		require.NoError(err)
+		derivedKeyEntry := utxoView._getDerivedKeyMappingForOwner(senderPkBytes, derivedPublicKey)
+		// If we removed the derivedKeyEntry from utxoView altogether, it'll be nil.
+		// To pass the tests, we initialize it to a default struct.
+		if derivedKeyEntry == nil || derivedKeyEntry.isDeleted {
+			derivedKeyEntry = &DerivedKeyEntry{*NewPublicKey(senderPkBytes), *NewPublicKey(derivedPublicKey), 0, AuthorizeDerivedKeyOperationValid, nil, transactionSpendingLimit, nil, false}
+		}
+		require.Equal(derivedKeyEntry.ExpirationBlock, expirationBlockExpected)
+		require.Equal(derivedKeyEntry.OperationType, operationTypeExpected)
+	}
```
lazynina, line 111:
```suggestion
	var derivedKeyEntry *DerivedKeyEntry
	// Verify that expiration block was persisted in the db or is in mempool utxoView
	if mempool == nil {
		derivedKeyEntry := NewDbAdapter(chain).GetOwnerToDerivedKeyMapping(*NewPublicKey(senderPkBytes), *NewPublicKey(derivedPublicKey))
	} else {
		utxoView, err := mempool.GetAugmentedUniversalView()
		require.NoError(err)
		derivedKeyEntry := utxoView._getDerivedKeyMappingForOwner(senderPkBytes, derivedPublicKey)
	}
	// If we removed the derivedKeyEntry from utxoView altogether, it'll be nil.
	// To pass the tests, we initialize it to a default struct.
	if derivedKeyEntry == nil || derivedKeyEntry.isDeleted {
		derivedKeyEntry = &DerivedKeyEntry{*NewPublicKey(senderPkBytes), *NewPublicKey(derivedPublicKey), 0, AuthorizeDerivedKeyOperationValid, nil, transactionSpendingLimit, nil, false}
	}
	require.Equal(derivedKeyEntry.ExpirationBlock, expirationBlockExpected)
	require.Equal(derivedKeyEntry.OperationType, operationTypeExpected)
```
we can de-dupe this code here.

lib/block_view_derived_key_test.go
```diff
@@ -508,7 +667,7 @@ func _getAuthorizeDerivedKeyMetadataWithTransactionSpendingLimitAndDerivedPrivat
 	expirationBlockByte := EncodeUint64(expirationBlock)
 	accessBytes := append(derivedPublicKey, expirationBlockByte[:]...)
 
-	transactionSpendingLimitBytes, err := transactionSpendingLimit.ToBytes()
+	transactionSpendingLimitBytes, err := transactionSpendingLimit.ToBytes(0)
```
lazynina, line 670: hmm we should probably be able to test before and after the block height. we can modify the block height at which the fork occurs by updating params.ForkHeights, so we may need to pass a height in here?

lib/block_view_derived_key_test.go
```diff
@@ -2844,6 +3080,86 @@ func TestAuthorizedDerivedKeyWithTransactionLimitsHardcore(t *testing.T) {
...
+
+		// Unlimited spending limit.
+		transactionSpendingLimit = &TransactionSpendingLimit{
+			GlobalDESOLimit:              0,
+			TransactionCountLimitMap:     make(map[TxnType]uint64),
+			CreatorCoinOperationLimitMap: make(map[CreatorCoinOperationLimitKey]uint64),
+			DAOCoinOperationLimitMap:     make(map[DAOCoinOperationLimitKey]uint64),
+			NFTOperationLimitMap:         make(map[NFTOperationLimitKey]uint64),
+			IsUnlimited:                  true,
+		}
```
lazynina, line 3110: define a var for this unlimited spending limit struct?

lib/network.go
```diff
@@ -5062,6 +5066,11 @@ func (tsl *TransactionSpendingLimit) ToMetamaskString(params *DeSoParams) string
 		indentationCounter--
 	}
 
+	// IsUnlimited
+	if tsl.IsUnlimited {
+		str += "FULL ACCESS"
+	}
```
lazynina, line 5072: If it's full access, should we even include the spending limit object? seems a little unnecessary to have both

lazynina (changes requested):

lib/network.go
AeonSw4n: Feels like if we don't indicate that it's full access, people might not realize how much permission this key has.

lazynina: It should certainly say full access, but we should leave out the other pieces of the spending limit object like the maps and the global DESO limit.

lib/block_view_derived_key_test.go
AeonSw4n: We only use this like twice in the entire file, I personally prefer it to be more descriptive.

lib/block_view_derived_key_test.go
AeonSw4n: I did a very hardcore before - at - after block height test, lmk what you think

PR #392, opened 2022-08-25
T/allow ens for username

DeSoDog:


lib/block_view_profile.go
```diff
@@ -574,10 +575,13 @@ func (bav *UtxoView) _connectUpdateProfile(
 		return 0, 0, nil, RuleErrorProfileStakeMultipleSize
 	}
 	// If a username is set then it must adhere to a particular regex.
+
 	if len(txMeta.NewUsername) != 0 && !UsernameRegex.Match(txMeta.NewUsername) {
-		return 0, 0, nil, errors.Wrapf(RuleErrorInvalidUsername, "Username: %v", string(txMeta.NewUsername))
+		isAfterMetamaskBlockHeight := ENSRegex.Match(txMeta.NewUsername) && blockHeight > bav.Params.ForkHeights.UnlimitedDerivedKeysBlockHeight
+		if !isAfterMetamaskBlockHeight {
+			return 0, 0, nil, errors.Wrapf(RuleErrorInvalidUsername, "Username: %v", string(txMeta.NewUsername))
+		}
```
lazynina, line 583:
```suggestion
		if blockHeight >= bav.Params.ForkHeights.UnlimitedDerivedKeysBlockHeight && !ENSRegex.Match(txMeta.NewUsername) {
			return 0, 0, nil, errors.Wrapf(RuleErrorInvalidUsername, "Username: %v", string(txMeta.NewUsername))
		}
```

lib/block_view_profile_test.go
```diff
@@ -794,6 +795,39 @@ func TestUpdateProfile(t *testing.T) {
...
+			"m5 created by paramUpdater", /*newDescription*/
+			otherShortPic,                /*newProfilePic*/
+			11*100,                       /*newCreatorBasisPoints*/
+			1.5*100*100,                  /*newStakeMultipleBasisPoints*/
+			false /*isHidden*/)
+
+		require.Error(err)
+		require.Contains(err.Error(), RuleErrorInvalidUsername)
+
+		params.ForkHeights.UnlimitedDerivedKeysBlockHeight = uint32(0)
```
lazynina, line 818: instead of modifying the fork height, please mine more blocks so we go past the fork height.

lazynina (changes requested):

lib/block_view_profile.go
DeSoDog: Removing the variable name, but keeping the logic. The suggested conditional doesn't throw an error if the block height is less than the fork height

lib/block_view_profile_test.go
DeSoDog: Resolving due to miner sync issues

lib/block_view_profile.go
lazynina: oh duh. good catch

PR #394, opened 2022-09-01
P/k group muting

AeonSw4n:
Open in favor of #383


lib/block_view_message.go
```diff
@@ -825,95 +949,189 @@ func (bav *UtxoView) _connectMessagingGroup(
...
+		// would be tricky to impose in consensus, considering that removed users can't *forget* the messaging private key.
+		// Removing users can be facilitated in the application-layer, where we can issue a new group key and share it with
+		// all valid members.
+
+		// Map all members so that it's easier to check for overlapping members.
+		existingMembers := make(map[PublicKey]bool)
+
+		// Sanity-check a group's members can't contain the messagingPublicKey.
+		existingMembers[*messagingPublicKey] = true
+
```
lazynina, line 975: if you want to be even more memory efficient, you can use an empty struct. It doesn't look like we're checking for the "true" value in this map, so existence itself implies a "truthy" value. Empty structs cost 0 bytes and effectively make a map a hash set
```suggestion
		// Map all members so that it's easier to check for overlapping members.
		existingMembers := make(map[PublicKey]struct{})

		// Sanity-check a group's members can't contain the messagingPublicKey.
		existingMembers[*messagingPublicKey] = struct{}{}

```

lib/block_view_message.go
```diff
@@ -825,95 +949,189 @@ func (bav *UtxoView) _connectMessagingGroup(
...
+			existingMembers[*messagingMember.GroupMemberPublicKey] = true
+			newMessagingMembers = append(newMessagingMembers, messagingMember)
 		}
 
-		// Make sure the messagingMember public key and messaging key name are valid.
-		if err := ValidateGroupPublicKeyAndName(messagingMember.GroupMemberPublicKey[:], messagingMember.GroupMemberKeyName[:]); err != nil {
-			return 0, 0, nil, errors.Wrapf(err, "_connectMessagingGroup: "+
-				"Problem validating public key or messaging key for messagingMember (%v)", messagingMember.GroupMemberPublicKey[:])
+	case MessagingGroupOperationMuteMembers:
+		// Muting members assumes the group was already created.
```
lazynina, line 1045: I think we need to check block height here before proceeding

lib/block_view_message.go
```diff
@@ -825,95 +949,189 @@ func (bav *UtxoView) _connectMessagingGroup(
...
-		if memberGroupEntry == nil || memberGroupEntry.isDeleted {
-			return 0, 0, nil, errors.Wrapf(
-				RuleErrorMessagingMemberKeyDoesntExist, "_connectMessagingGroup: "+
-					"Problem verifying messaing key for messagingMember (%v)", messagingMember.GroupMemberPublicKey[:])
+		for _, newlyMutedMember := range txMeta.MessagingGroupMembers {
+			// Make sure GroupOwner is not muting herself
+			if reflect.DeepEqual(newlyMutedMember.GroupMemberPublicKey[:], existingEntry.GroupOwnerPublicKey[:]) {
+				return 0, 0, nil, errors.Wrapf(RuleErrorMessagingGroupOwnerMutingSelf,
+					"_connectMessagingGroup: GroupOwner cannot mute herself (%v).", existingEntry.GroupOwnerPublicKey[:])
+			}
```
lazynina, line 1069: should we prevent a user from muting themselves? also can any member mute any other member or is it only a group owner? This is a potential avenue for abuse in a groupchat. If I don't like  you, I just mute you. I think it's okay if that's the situation, but want to make sure we're considering the UX of this feature.

lib/block_view_message.go
```diff
@@ -825,95 +949,189 @@ func (bav *UtxoView) _connectMessagingGroup(
...
+			// Add member to newMuteList. We iterate over newMuteList to prevent mute list entries with duplicate public keys.
+			for _, addedMutedMember := range newMuteList {
```
lazynina, line 1071: we should probably use a hash set to track the newMuteList within this function. I can imagine we may run into very long lists of muted members in long running group chats. This could end up being an n^2 situation, which would be pretty annoying. Use a map with the value as a struct{} and then at the end, we can convert the map back to a slice.

lib/block_view_message.go
```diff
@@ -825,95 +949,189 @@ func (bav *UtxoView) _connectMessagingGroup(
...
+			// 2. Member already unmuted
+			// 3. Member does not exist in group hence cannot unmute nonexistent member
+			if !isUnmuteValid {
+				// 1. GroupOwner unmuting herself is invalid because GroupOwner can never be muted in the first place
+				if reflect.DeepEqual(member.GroupMemberPublicKey[:], existingEntry.GroupOwnerPublicKey[:]) {
+					return 0, 0, nil, errors.Wrapf(RuleErrorMessagingGroupOwnerUnmutingSelf,
+						"_connectMessagingGroup: GroupOwner cannot mute herself (%v).", existingEntry.GroupOwnerPublicKey[:])
+				}
+				// 2. Member already unmuted
+				for _, entryMembers := range existingEntry.MessagingGroupMembers {
```
lazynina, line 1113: we should try to remove these nested loops as much as possible

lib/block_view_message.go
```diff
@@ -825,95 +949,189 @@ func (bav *UtxoView) _connectMessagingGroup(
...
 
-		// Make sure the messagingMember public key and messaging key name are valid.
-		if err := ValidateGroupPublicKeyAndName(messagingMember.GroupMemberPublicKey[:], messagingMember.GroupMemberKeyName[:]); err != nil {
-			return 0, 0, nil, errors.Wrapf(err, "_connectMessagingGroup: "+
-				"Problem validating public key or messaging key for messagingMember (%v)", messagingMember.GroupMemberPublicKey[:])
+	case MessagingGroupOperationMuteMembers:
+		// Muting members assumes the group was already created.
+		if existingEntry == nil || existingEntry.isDeleted {
+			return 0, 0, nil, errors.Wrapf(RuleErrorMessagingGroupDoesntExist,
+				"_connectMessagingGroup: Can't mute members for a non-existing group")
```
lazynina, line 1048:
```suggestion
				"_connectMessagingGroup: Can't mute members for a non-existent group")
```

lib/block_view_message.go
```diff
@@ -825,95 +949,189 @@ func (bav *UtxoView) _connectMessagingGroup(
...
-				RuleErrorMessagingMemberAlreadyExists, "_connectMessagingGroup: "+
-					"Error, messagingMember already exists (%v)", messagingMember.GroupMemberPublicKey[:])
+		// Set the messaging members to the existing members since we won't be changing them.
+		newMessagingMembers = existingEntry.MessagingGroupMembers
+
+	case MessagingGroupOperationUnmuteMembers:
+		// Unmuting members assumes the group was already created.
+		if existingEntry == nil || existingEntry.isDeleted {
+			return 0, 0, nil, errors.Wrapf(RuleErrorMessagingGroupDoesntExist,
+				"_connectMessagingGroup: Can't mute members for a non-existing group")
```
lazynina, line 1087:
```suggestion
				"_connectMessagingGroup: Can't mute members for a non-existent group")
```

lib/server.go
```diff
@@ -605,6 +605,12 @@ func (srv *Server) GetSnapshot(pp *Peer) {
 	if !syncingPrefix {
 		// We will assign the peer to a non-existing prefix.
 		for _, prefix = range StatePrefixes.StatePrefixesList {
+			// FIXME: This is a temporary hack that we have to employ until we are confident nodes have
+			// 	downloaded the latest code that sends an empty db chunk for a non-existing prefix.
```
lazynina, line 609:
```suggestion
			// 	downloaded the latest code that sends an empty db chunk for a non-existent prefix.
```

lib/server.go
```diff
@@ -605,6 +605,12 @@ func (srv *Server) GetSnapshot(pp *Peer) {
 	if !syncingPrefix {
 		// We will assign the peer to a non-existing prefix.
```
lazynina, line 606:
```suggestion
		// We will assign the peer to a non-existent prefix.
```

lib/server.go
```diff
@@ -643,6 +649,19 @@ func (srv *Server) GetSnapshot(pp *Peer) {
 		"with Prefix (%v) and SnapshotStartEntry (%v)", pp, prefix, lastReceivedKey)
 }
 
+// FIXME: This is a temporary hack that we have to employ until we are confident nodes have
+// 	downloaded the latest code that sends an empty db chunk for a non-existing prefix. We
```
lazynina, line 653:
```suggestion
// 	downloaded the latest code that sends an empty db chunk for a non-existent prefix. We
```

lib/server.go
```diff
@@ -1237,6 +1256,12 @@ func (srv *Server) _handleSnapshot(pp *Peer, msg *MsgDeSoSnapshotData) {
 
 	var completedPrefixes [][]byte
 	for _, prefix := range StatePrefixes.StatePrefixesList {
+		// FIXME: This is a temporary hack that we have to employ until we are confident nodes have
+		// 	downloaded the latest code that sends an empty db chunk for a non-existing prefix.
```
lazynina, line 1260:
```suggestion
		// 	downloaded the latest code that sends an empty db chunk for a non-existent prefix.
```

lib/block_view_message_test.go
```diff
@@ -2014,8 +2107,28 @@ func TestGroupMessages(t *testing.T) {
...
 		// 	sender: 4
 		//	recipient: 4
+		// Since we're passed the ExtraData migration, the entry will have the extra data field. We add it after
+		// transaction is processed as an extra sanity-check.
+		messageEntry.ExtraData = make(map[string][]byte)
+		messageEntry.ExtraData[MessagesVersionString] = UintToBuf(MessagesVersion3)
+		messageEntry.ExtraData[SenderMessagingPublicKey] = entry.MessagingPublicKey[:]
+		messageEntry.ExtraData[SenderMessagingGroupKeyName] = NewGroupKeyName(defaultKey).ToBytes()
+		messageEntry.ExtraData[RecipientMessagingPublicKey] = entryRecipient.MessagingPublicKey.ToBytes()
+		messageEntry.ExtraData[RecipientMessagingGroupKeyName] = NewGroupKeyName(defaultKey).ToBytes()
```
lazynina, line 2117: We keep repeating this chunk of code. Can we move it to a helper function?

lib/block_view_message_test.go
```diff
@@ -2392,13 +2558,410 @@ func TestGroupMessages(t *testing.T) {
...
+
+		// MUTING TESTS
+		// Let us now mute m0
+		var muteList []*MessagingGroupMember
+		muteList = append(muteList, &MessagingGroupMember{
+			m0PublicKey,
+			BaseGroupKeyName(),
+			encrypt(privBytes, m0PubKey),
+		})
+		//require.Equal(false, _verifyMessagingKey(testMeta, senderPublicKey, entry))
```
lazynina, line 2577: Remove commented out code please

lib/block_view_message_test.go
```diff
@@ -2392,13 +2558,410 @@ func TestGroupMessages(t *testing.T) {
...
+		{
+			// Let us now try to mute m0 AGAIN
+			// This should produce an error since m0 is already muted
+			var muteList []*MessagingGroupMember
+			muteList = append(muteList, &MessagingGroupMember{
+				m0PublicKey,
+				BaseGroupKeyName(),
+				encrypt(privBytes, m0PubKey),
+			})
+			//require.Equal(false, _verifyMessagingKey(testMeta, senderPublicKey, entry))
```
lazynina, line 2651: remove commented out code please

lib/block_view_message_test.go
```diff
@@ -2392,13 +2558,410 @@ func TestGroupMessages(t *testing.T) {
...
+
+		// UNMUTING TESTS
+		// Let us now unmute m0
+		var unmuteList []*MessagingGroupMember
+		unmuteList = append(unmuteList, &MessagingGroupMember{
+			m0PublicKey,
+			BaseGroupKeyName(),
+			encrypt(privBytes, m0PubKey),
+		})
+		//require.Equal(false, _verifyMessagingKey(testMeta, senderPublicKey, entry))
```
lazynina, line 2674: commented out code - please remove

lib/block_view_message_test.go
```diff
@@ -2392,13 +2558,410 @@ func TestGroupMessages(t *testing.T) {
...
+		{
+			// Let us now try to unmute m0 AGAIN
+			// This should produce an error since m0 is already unmuted
+			var unmuteList []*MessagingGroupMember
+			unmuteList = append(unmuteList, &MessagingGroupMember{
+				m0PublicKey,
+				BaseGroupKeyName(),
+				encrypt(privBytes, m0PubKey),
+			})
+			//require.Equal(false, _verifyMessagingKey(testMeta, senderPublicKey, entry))
```
lazynina, line 2754:
```suggestion
```
commented out code - please remove

lib/block_view_message_test.go
```diff
@@ -2392,13 +2558,410 @@ func TestGroupMessages(t *testing.T) {
...
+			extraData = make(map[string][]byte)
+			extraData[MessagingGroupOperationType] = []byte{byte(MessagingGroupOperationUnmuteMembers)}
```
lazynina, line 2756: Note that you can define a map in one command if you'd like:
```
		extraData = map[string][]byte{
			MessagingGroupOperationType: []byte(byte(MessagingGroupOperationUnmuteMembers)),
		}
```

lib/block_view_message_test.go
```diff
@@ -2392,13 +2558,410 @@ func TestGroupMessages(t *testing.T) {
...
+			var muteList []*MessagingGroupMember
+			muteList = append(muteList, &MessagingGroupMember{
+				senderPublicKey,
+				BaseGroupKeyName(),
+				encrypt(privBytes, senderPkBytes),
+			})
+			//require.Equal(false, _verifyMessagingKey(testMeta, senderPublicKey, entry))
+			extraData := make(map[string][]byte)
+			extraData[MessagingGroupOperationType] = []byte{byte(MessagingGroupOperationMuteMembers)}
+			_messagingKeyWithExtraDataWithTestMeta(
```
lazynina, line 2806: Can we add tests for muting/unmuting oneself? I think that's a different rule error than the group owner one

lib/db_utils.go
```diff
@@ -1645,28 +1659,223 @@ func DBGetAllUserGroupEntries(handle *badger.DB, ownerPublicKey []byte) ([]*Mess
...
+	// the entry is the member specified. This is a bit of a hack to allow us to store a "back-reference"
+	// to the GroupEntry inside the value of this field.
+	memberGroupEntry := &MessagingGroupEntry{
+		GroupOwnerPublicKey:   messagingGroupEntry.GroupOwnerPublicKey,
+		MessagingPublicKey:    messagingGroupEntry.MessagingPublicKey,
+		MessagingGroupKeyName: messagingGroupEntry.MessagingGroupKeyName,
+		MessagingGroupMembers: []*MessagingGroupMember{
+			messagingGroupMember,
+		},
+		MuteList:  muteList,
```
lazynina, line 1738: do we need to sort the mute list and group members for determinism's sake?

lib/network.go
```diff
@@ -6371,6 +6371,16 @@ func DeserializePubKeyToUint64Map(data []byte) (map[PublicKey]uint64, error) {
...
 
+// MessagingGroupOperation represents V3 Group Chat Messages ExtraData["MessagingGroupOperationType"] values
+type MessagingGroupOperation byte
+
+const (
+	MessagingGroupOperationAddMembers MessagingGroupOperation = iota
+	MessagingGroupOperationRemoveMembers
+	MessagingGroupOperationMuteMembers
+	MessagingGroupOperationUnmuteMembers
+)
```
lazynina, line 6382: Typically we don't use iota's, because if we accidentally change the order of declaration later, previous transactions won't work. For example, if we moved RemoveMembers to the end, it would confuse the connect logic. Let's explicitly set values to numbers

lazynina (changes requested):
overall looks great, nice work @AeonSw4n

lib/block_view_message.go
```diff
@@ -490,14 +560,45 @@ func (bav *UtxoView) _connectPrivateMessage(
...
+				txMeta.RecipientPublicKey, recipientMessagingPublicKey, recipientMessagingKeyName, blockHeight); err != nil {
 				return 0, 0, nil, errors.Wrapf(RuleErrorPrivateMessageFailedToValidateMessagingKey,
 					"_connectPrivateMessage: failed to validate public key and key name, error: (%v)", err)
 			}
 			// If everything worked, update the messaging key information in the message entry.
 			messageEntry.RecipientMessagingPublicKey = NewPublicKey(recipientMessagingPublicKey)
 			messageEntry.RecipientMessagingGroupKeyName = NewGroupKeyName(recipientMessagingKeyName)
 		}
+
+		// After the DeSoUnlimitedDerivedKeysAndMessagesMutingAndMembershipIndexBlockHeight block height we force the usage of V3 messages.
```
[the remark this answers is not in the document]
AeonSw4n: Feels like the simplest way to get rid of the endless sea of backwards-compatibility issues with messages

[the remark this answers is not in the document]
AeonSw4n: I was thinking we should just make backend be backwards-compatible and attempt to do v3 by default. All v2 messages are valid under v3, it's just that v3 private messages need to be formatted differently, and requiring a unified transaction schema makes things simpler moving forward

lib/block_view_flush.go
```diff
@@ -997,25 +1021,47 @@ func (bav *UtxoView) _flushMessagingGroupEntriesToDbWithTxn(txn *badger.Txn, blo
...
-					recipient, &ownerPublicKey, messagingGroupEntry); err != nil {
+				// After the muting and prefix optimization block height, we will store the owner's groups
+				// in the membership index by default. This way we can easily query for the messaging group
+				// simplified entry in the db, saving read time, e.g. when we want to send a message to the group.
+
+				if err := DBPutMessagingGroupOwnerInMembershipIndexWithTxn(txn, bav.Snapshot, blockHeight,
+					ownerPublicKey, messagingGroupEntry); err != nil {
 					return errors.Wrapf(err, "UtxoView._flushMessagingGroupEntriesToDbWithTxn: "+
-						"Problem putting MessagingGroupEntry recipient (%v) to db", recipient)
+						"Problem putting MessagingGroupEntry %v to db", *messagingGroupEntry)
```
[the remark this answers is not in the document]
AeonSw4n: Isn't this comment going to print the whole entry? I'm dereferencing the pointer.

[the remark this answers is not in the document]
AeonSw4n: I've added more info to distinguish between different ops in the flush.

lib/block_view_message.go
AeonSw4n: The call to GetMessagingGroupOperation() is gated by the blockheight, so the operation will never be Mute/Unmute unless we're past the block height.

lib/block_view_message.go
AeonSw4n: Only the owner can mute others. During the connect logic for the messaging group txn we fetch an entry based on the txn.PublicKey, so you can only modify your own groups.

lib/block_view_message.go
AeonSw4n: yeah that makes sense

lib/block_view_message.go
AeonSw4n: the O(nlogn) thingy gets rid of them. This one in particular is just to come up with a better validation error, and since we will error regardless if we get into this the `if !isUnmuteValid` it makes the whole thing O(nlogn) still

lib/block_view_message_test.go
AeonSw4n: makes sense

lib/block_view_message_test.go
AeonSw4n: I know but it's 3 lines vs 2 lines.

lib/network.go
AeonSw4n: I generally prefer using iota over setting all values manually, but since these go into ExtraData I agree with you that it's better to define them explicitly

lib/block_view_message_test.go
AeonSw4n: Only owner can mute/unmute people. There was an edge-case where the owner tried muting someone who wasn't in the group. Keshav found this ln and I just fixed it too rn.

PR #398, opened 2022-09-11
Refactor V3 changes

AeonSw4n:


lib/block_view_flush.go
```diff
@@ -1021,46 +997,24 @@ func (bav *UtxoView) _flushMessagingGroupEntriesToDbWithTxn(txn *badger.Txn, blo
...
-				// After the muting and prefix optimization block height, we will store the owner's groups
-				// in the membership index by default. This way we can easily query for the messaging group
-				// simplified entry in the db, saving read time, e.g. when we want to send a message to the group.
-
-				if err := DBPutMessagingGroupOwnerInMembershipIndexWithTxn(txn, bav.Snapshot, blockHeight,
-					ownerPublicKey, messagingGroupEntry); err != nil {
-					return errors.Wrapf(err, "UtxoView._flushMessagingGroupEntriesToDbWithTxn: "+
-						"Fail while putting owner membership index. Problem putting MessagingGroupEntry %v to db", *messagingGroupEntry)
+				if reflect.DeepEqual(recipient.GroupMemberPublicKey[:], ownerPublicKey[:]) {
+					continue
```
lazynina, line 1011: this looks like we exclude the group owner from the db, is this right?

lib/block_view_message.go
```diff
@@ -142,6 +78,9 @@ func (bav *UtxoView) GetMessagingGroupKeyToMessagingGroupEntryMapping(
 
 	// If an entry exists in the in-memory map, return the value of that mapping.
 	if mapValue, exists := bav.MessagingGroupKeyToMessagingGroupEntry[*messagingGroupKey]; exists {
+		if mapValue.isDeleted {
+			return nil
+		}
```
lazynina, line 83: I think we actually want to return the mapValue if it's deleted in case one of the downstream functions is going to merge results from badger. Should we expose an IsDeleted function on the MessagingGroupEntry and expect downstream functions to properly handle?

lib/block_view_message.go
```diff
@@ -887,19 +774,14 @@ func (bav *UtxoView) _connectMessagingGroup(
...
-		if EqualGroupKeyName(NewGroupKeyName(txMeta.MessagingGroupKeyName), DefaultGroupKeyName()) {
-			// Verify the GroupOwnerSignature. it should be signature( messagingPublicKey || messagingKeyName )
-			// We need to make sure the default messaging key was authorized by the master public key.
-			// All other keys can be registered by derived keys.
-			bytes := append(txMeta.MessagingPublicKey, txMeta.MessagingGroupKeyName...)
-			if err := _verifyBytesSignature(txn.PublicKey, bytes, txMeta.GroupOwnerSignature, blockHeight, bav.Params); err != nil {
-				return 0, 0, nil, errors.Wrapf(err, "_connectMessagingGroup: "+
-					"Problem verifying signature bytes, error: %v", RuleErrorMessagingSignatureInvalid)
-			}
+	if EqualGroupKeyName(NewGroupKeyName(txMeta.MessagingGroupKeyName), DefaultGroupKeyName()) {
```
lazynina, line 777: I think we want to only do this check if we're before `DeSoUnlimitedDerivedKeysBlockHeight`

lib/block_view_message.go
lazynina:
```suggestion
```
let's delete so this behaves exactly as this function does currently in production

lib/block_view_message.go
```diff
@@ -1227,12 +975,6 @@ func (bav *UtxoView) _disconnectMessagingGroup(
 			utxoOpsForTxn[operationIndex].Type)
 	}
 
-	// Check that the transaction has the right TxnType.
```
lazynina, line 1230: I actually like having this sanity check in there. I think it's okay to add it in.

lazynina (changes requested):

lib/block_view_flush.go
AeonSw4n: yes, I reverted to what we had in main before the membership index.

lazynina (approved):

PR #404, opened 2022-10-10
K/v3 new stuff

keshavm02:
Only for reading through. Not for running locally.


lib/block_view_types.go
```diff
@@ -1980,7 +2116,7 @@ type MessagingGroupEntry struct {
 	// MessagingGroupMembers is a list of recipients in a group chat. Messaging keys can have
 	// multiple recipients, where the encrypted private key of the messaging public key
 	// is given to all group members.
-	MessagingGroupMembers []*MessagingGroupMember
+	MessagingGroupMembers []*MessagingGroupMember // Deprecated
```
AeonSw4n, line 2119: Should we rename to DeprecatedMessagingGroupMembers? Or something like that

AeonSw4n: also if we drop access groups from message computation, we might just be able to delete this.

lib/block_view_types.go
```diff
@@ -1991,8 +2127,32 @@ type MessagingGroupEntry struct {
...
+
+	messagingGroupEntry := MessagingGroupEntry{
+		GroupOwnerPublicKey:   groupOwnerPublicKey,
+		MessagingPublicKey:    messagingPublicKey,
+		MessagingGroupKeyName: messagingGroupKeyName,
+		MessagingGroupMembers: messagingGroupMembers,
+		ExtraData:             extraData,
+	}
+
+	if MigrationTriggered(blockHeight, DeSoUnlimitedDerivedKeysAndMessageMutingAndMembershipIndex) {
```
AeonSw4n, line 2142: hmm, I don't love using `MigrationTriggered` to handle regular block height logic. Till now, we used `MigrationTriggered` only in DeSoEncoder logic

lib/block_view_types.go
```diff
@@ -2104,6 +2287,8 @@ type MessagingGroupMember struct {
 
 	// EncryptedKey is the encrypted messaging public key, addressed to the recipient.
 	EncryptedKey []byte
+
+	IsMuted bool
```
AeonSw4n, line 2291: With the new design, we should probably remove this from the `MessagingGroupEntry`

lib/block_view_types.go
```diff
@@ -2112,6 +2297,7 @@ func (rec *MessagingGroupMember) ToBytes() []byte {
 	data = append(data, EncodeByteArray(rec.GroupMemberPublicKey.ToBytes())...)
 	data = append(data, EncodeByteArray(rec.GroupMemberKeyName.ToBytes())...)
 	data = append(data, EncodeByteArray(rec.EncryptedKey)...)
+	data = append(data, BoolToByte(rec.IsMuted))
```
AeonSw4n, line 2300: Should be gated by migration height

lib/block_view_types.go
```diff
@@ -2135,15 +2321,22 @@ func (rec *MessagingGroupMember) FromBytes(rr *bytes.Reader) error {
...
 }
 
 func (rec *MessagingGroupMember) RawEncodeWithoutMetadata(blockHeight uint64, skipMetadata ...bool) []byte {
-	data := []byte{}
+	var data []byte
 
 	data = append(data, EncodeToBytes(blockHeight, rec.GroupMemberPublicKey, skipMetadata...)...)
 	data = append(data, EncodeToBytes(blockHeight, rec.GroupMemberKeyName, skipMetadata...)...)
 	data = append(data, EncodeByteArray(rec.EncryptedKey)...)
+	data = append(data, BoolToByte(rec.IsMuted))
```
AeonSw4n, line 2339: Should be gated by migration height

lib/db_utils.go
```diff
@@ -479,6 +512,9 @@ func StatePrefixToDeSoEncoder(prefix []byte) (_isEncoder bool, _encoder DeSoEnco
 	} else if bytes.Equal(prefix, Prefixes.PrefixDAOCoinLimitOrderByOrderID) {
 		// prefix_id:"[62]"
 		return true, &DAOCoinLimitOrderEntry{}
+	} else if bytes.Equal(prefix, Prefixes.PrefixGroupMembershipIndex) {
```
AeonSw4n, line 515: We should also add prefixes `64`, and `65` with `return false, nil`. You can check a similar situation with prefix `PrefixPublicKeyBlockHashToBlockReward` and others

AeonSw4n (changes requested):
Overall, it looks chill, and the most challenging part is behind us. We should be close to completion once we implement the high-level refactors we discussed.

PR #407, opened 2022-10-16
Access groups and group messages

AeonSw4n:


lib/db_utils.go
```diff
@@ -479,6 +511,9 @@ func StatePrefixToDeSoEncoder(prefix []byte) (_isEncoder bool, _encoder DeSoEnco
 	} else if bytes.Equal(prefix, Prefixes.PrefixDAOCoinLimitOrderByOrderID) {
 		// prefix_id:"[62]"
 		return true, &DAOCoinLimitOrderEntry{}
+	} else if bytes.Equal(prefix, Prefixes.PrefixGroupMembershipIndex) {
+		// prefix_id:"[63]"
+		return true, &AccessGroupEntry{}
```
AeonSw4n, line 516: make sure to add all new prefixes here

AeonSw4n (commented):
@keshavm02

lib/block_view_access_groups.go
```diff
@@ -0,0 +1,964 @@
...
+	// If utxoView mapping doesn't exist, check DB.
+	isMuted, err := DBIsMemberMutedInGroupMembersIndex(bav.Handle, bav.Snapshot, messagingGroupEntry.GroupOwnerPublicKey, messagingGroupEntry.AccessGroupKeyName, memberPublicKey)
+	if err != nil {
+		return false, errors.Wrapf(err, "isMemberMuted: Problem checking if member is muted in DB")
+	}
+	return isMuted, nil
+}
+
+// setMemberMuted sets the muted status of a member in the group.
+func (bav *UtxoView) setMutedMembersMapping(
```
AeonSw4n, line 421: can we refactor these functions to use `<groupOwnerPublicKey, groupKeyName>` as parameters instead of the `AccessGroupEntry`?

lib/block_view.go
```diff
@@ -253,10 +264,10 @@ func (bav *UtxoView) CopyUtxoView() (*UtxoView, error) {
 	}
 
 	// Copy messaging group data
-	newView.MessagingGroupKeyToMessagingGroupEntry = make(map[MessagingGroupKey]*MessagingGroupEntry, len(bav.MessagingGroupKeyToMessagingGroupEntry))
-	for pkid, entry := range bav.MessagingGroupKeyToMessagingGroupEntry {
+	newView.AccessGroupKeyToAccessGroupEntry = make(map[AccessGroupKey]*AccessGroupEntry, len(bav.AccessGroupKeyToAccessGroupEntry))
+	for pkid, entry := range bav.AccessGroupKeyToAccessGroupEntry {
 		newEntry := *entry
-		newView.MessagingGroupKeyToMessagingGroupEntry[pkid] = &newEntry
+		newView.AccessGroupKeyToAccessGroupEntry[pkid] = &newEntry
```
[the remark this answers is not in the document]
keshavm02: done

lib/block_view_access_groups.go
```diff
@@ -0,0 +1,964 @@
...
+						RuleErrorMessagingMemberKeyDoesntExist, "_connectMessagingGroup: "+
+							"Problem verifying messaging key for messagingMember (%v)", messagingMember.GroupMemberPublicKey[:])
+				}
+
+				// Add the messagingMember to our helper structs.
+				newMessagingMembers = append(newMessagingMembers, messagingMember)
+			}
+		}
+
+	case AccessGroupOperationMuteMembers:
```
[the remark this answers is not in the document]
lazynina: It's not a new transaction type though, correct? but moving that logic to a helper would be nice - we follow a similar style for operation types in `_connectDAOCoin` where we call functions like `HelpConnectDAOCoinMint` to break up the logic and make it more digestible.

lib/block_view_access_groups.go
```diff
@@ -0,0 +1,1026 @@
...
+	var extraData map[string][]byte
+	if blockHeight >= bav.Params.ForkHeights.ExtraDataOnEntriesBlockHeight {
+		var existingExtraData map[string][]byte
+		if existingEntry != nil && !existingEntry.isDeleted {
+			existingExtraData = existingEntry.ExtraData
+		}
+		extraData = mergeExtraData(existingExtraData, txn.ExtraData)
+	}
+
+	// TODO: Currently, it is technically possible for any user to add *any other* user to *any group* with
```
lazynina, line 907: How can the app level tell if it's a garbage encrypted key? We should probably have some guards in place to prevent this junk

lib/block_view_types.go
```diff
@@ -75,8 +75,11 @@ const (
 	EncoderTypeUtxoOperationBundle
 	EncoderTypeMessageEntry
 	EncoderTypeGroupKeyName
-	EncoderTypeMessagingGroupEntry
-	EncoderTypeMessagingGroupMember
+	EncoderTypeGroupMembershipKey
+	EncoderTypeGroupEnumerationKey
+	EncoderTypeGroupMemberAttributesKey
+	EncoderTypeAccessGroupEntry
+	EncoderTypeAccessGroupMember
```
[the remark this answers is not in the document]
lazynina: agreed^

lib/block_view_types.go
```diff
@@ -702,8 +711,8 @@ type UtxoOperation struct {
 	// For disconnecting AuthorizeDerivedKey transactions.
 	PrevDerivedKeyEntry *DerivedKeyEntry
 
-	// For disconnecting MessagingGroupKey transactions.
-	PrevMessagingKeyEntry *MessagingGroupEntry
+	// For disconnecting AccessGroupKey transactions.
+	PrevMessagingKeyEntry *AccessGroupEntry
```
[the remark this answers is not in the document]
lazynina: Agreed^

lib/db_utils.go
```diff
@@ -244,50 +244,89 @@ type DBPrefixes struct {
...
+	//   - To be able to fetch all groups that a user is a member of (including groups that
+	//     they own). This is especially useful for Backend API to fetch all groups for a user.
+	//
+	// New <GroupMembershipIndex> :
+	// <prefix, GroupMemberPublicKey [33]byte, GroupOwnerPublicKey [33]byte, GroupKeyName [32]byte> -> <AccessGroupMember>
+	PrefixGroupMembershipIndex []byte `prefix_id:"[63]" is_state:"true"`
+
+	// Prefix for Enumerating over all the members of a group:
+	// <prefix, GroupOwnerPublicKey [33]byte, GroupKeyName [32]byte, GroupMemberPublicKey [33]byte> -> <>
+	PrefixGroupEnumerationIndex []byte `prefix_id:"[64]" is_state:"true"`
```
lazynina, line 318: Can we call this `PrefixGroupMemberEnumerationIndex`?

lib/db_utils.go
```diff
@@ -1482,159 +1533,153 @@ func DBGetLimitedMessageForMessagingKeys(handle *badger.DB, messagingKeys []*Mes
...
-	if entry := DBGetMessagingGroupEntryWithTxn(txn, snap, messagingGroupKey); entry == nil {
+	if entry := DBGetAccessGroupEntryWithTxn(txn, snap, accessGroupKey); entry == nil {
 		return nil
 	}
 
-	// When a messaging key entry exists, delete it from the DB.
-	if err := DBDeleteWithTxn(txn, snap, _dbKeyForMessagingGroupEntry(messagingGroupKey)); err != nil {
-		return errors.Wrapf(err, "DBDeleteMessagingGroupEntryWithTxn: Deleting "+
-			"entry for MessagingGroupKey failed: %v", messagingGroupKey)
+	// When a access key entry exists, delete it from the DB.
```
lazynina, line 1608:
```suggestion
	// When an access key entry exists, delete it from the DB.
```

lib/db_utils.go
```diff
@@ -1482,159 +1533,153 @@ func DBGetLimitedMessageForMessagingKeys(handle *badger.DB, messagingKeys []*Mes
...
-	ownerGroupEntries, err := DBGetMessagingGroupEntriesForOwnerWithTxn(txn, NewPublicKey(ownerPublicKey))
-	if err != nil {
-		return nil, errors.Wrapf(err, "DBGetAllUserGroupEntiresWithTxn: problem getting messaging entries")
-	}
-	userGroupEntries = append(userGroupEntries, ownerGroupEntries...)
-
 	// And add the groups where the user is a member
-	memberGroupEntries, err := DBGetAllMessagingGroupEntriesForMemberWithTxn(txn, NewPublicKey(ownerPublicKey))
+	// Never use the deprecated function for backwards compatibility here because of no access to blockHeight
+	accessGroupEntries, err := DBGetAllEntriesForPublicKeyFromMembershipIndexWithTxn(txn, snap, NewPublicKey(memberPublicKey))
```
lazynina, line 1668: This won't work if we're before the fork height, correct? If so, we'll prob need branching logic to call a version of this function with the pre-fork logic unfortunately.

lib/db_utils.go
```diff
@@ -1644,111 +1689,840 @@ func DBGetAllUserGroupEntries(handle *badger.DB, ownerPublicKey []byte) ([]*Mess
...
+	groupMemberPublicKey *PublicKey, groupOwnerPublicKey *PublicKey, groupKeyName *GroupKeyName) error {
+
+	// First pull up the mapping that exists for the public key passed in.
+	// If one doesn't exist then there's nothing to do.
+	existingMember := DBGetMemberFromMembershipIndexWithTxn(txn, snap, groupMemberPublicKey, groupOwnerPublicKey, groupKeyName)
+	if existingMember == nil {
+		return nil
+	}
+
+	// When a message exists, delete the mapping for the sender and receiver.
```
lazynina, line 1848: this comment doesn't make sense?

lib/db_utils.go
```diff
@@ -1644,111 +1689,840 @@ func DBGetAllUserGroupEntries(handle *badger.DB, ownerPublicKey []byte) ([]*Mess
...
+		// The key is of the form <prefix, GroupOwnerPublicKey, GroupKeyName, GroupMemberPublicKey>
+		// So we extract the GroupEnumerationKey.
+		enumerationKey := &GroupEnumerationKey{}
+		rr := bytes.NewReader(keyBytes)
+		if exists, err := DecodeFromBytes(enumerationKey, rr); !exists || err != nil {
+			return nil, errors.Wrapf(err, "DBGetAllGroupMembersForMessagingGroupWithTxn: "+
+				"problem decoding key (%v)", keyBytes)
+		}
+		memberPublicKey := &enumerationKey.GroupMemberPublicKey
+		messagingGroupMember := DBGetMemberFromMembershipIndexWithTxn(txn, snap, memberPublicKey, groupOwnerPublicKey, groupKeyName)
```
lazynina, line 1954: could we store `AccessGroupMember` as the value in this index so we don't have to do a second DB lookup?

lib/network.go
```diff
@@ -201,66 +201,74 @@ type DeSoMessage interface {
...
+	TxnTypeBurnNFT                       TxnType = 21
+	TxnTypeAuthorizeDerivedKey           TxnType = 22
+	TxnTypeAccessGroupCreate             TxnType = 23
+	TxnTypeDAOCoin                       TxnType = 24
+	TxnTypeDAOCoinTransfer               TxnType = 25
+	TxnTypeDAOCoinLimitOrder             TxnType = 26
+	TxnTypeAccessGroupAddMember          TxnType = 27
+	TxnTypeAccessGroupRemoveMember       TxnType = 28
+	TxnTypeAccessGroupSetMemberAttribute TxnType = 29
+	TxnTypeAccessGroupSetGroupAttribute  TxnType = 30
```
lazynina, line 233: wait we're adding new transaction types for each of these? It would be great if we could minimize the number of new txn types. Also, isn't adding member to access groups already a part of an existing txn? some clarity on why we're adding new txn types would be helpful

lib/network.go
```diff
@@ -462,8 +480,8 @@ func NewTxnMetadata(txType TxnType) (DeSoTxnMetadata, error) {
 		return (&BurnNFTMetadata{}).New(), nil
 	case TxnTypeAuthorizeDerivedKey:
 		return (&AuthorizeDerivedKeyMetadata{}).New(), nil
-	case TxnTypeMessagingGroup:
-		return (&MessagingGroupMetadata{}).New(), nil
+	case TxnTypeAccessGroupCreate:
+		return (&AccessGroupMetadata{}).New(), nil
 	case TxnTypeDAOCoin:
 		return (&DAOCoinMetadata{}).New(), nil
 	case TxnTypeDAOCoinTransfer:
```
lazynina, line 487: we need to add the new txn types here as well if we are in fact adding new txn types

lib/network.go
```diff
@@ -6391,70 +6455,116 @@ type MessagingGroupMetadata struct {
...
+	}
+
+	// Make sure this is a messaging group transaction.
+	if txn.TxnMeta.GetTxnType() != TxnTypeAccessGroupCreate {
+		return 0, fmt.Errorf("GetAccessGroupOperation: called on txn with type %v", txn.TxnMeta.GetTxnType())
+	}
+
+	// Sanity-check cast the transaction metadata to a messaging group metadata.
+	_, ok := txn.TxnMeta.(*AccessGroupMetadata)
+	if !ok {
```
lazynina, line 6541:
```suggestion
	if _, ok := txn.TxnMeta.(*AccessGroupMetadata); !ok {
```

lib/network.go
```diff
@@ -6391,70 +6455,116 @@ type MessagingGroupMetadata struct {
...
+	}
+
+	// If the transaction's ExtraData is nil then we assume AccessGroupOperationAddMembers
+	messagingGroupOperation := AccessGroupOperationAddMembers
+	if txn.ExtraData == nil {
+		return messagingGroupOperation, nil
+	}
+
+	// Check if the transaction's ExtraData contains a AccessGroupOperationType.
+	if operationTypeBytes, operationTypeExists := txn.ExtraData[AccessGroupOperationType]; operationTypeExists && len(operationTypeBytes) == 1 {
```
lazynina, line 6552: we should probably through some sort of error if the length of opeationTypeBytes is greater than expected

lib/server.go
```diff
@@ -667,6 +673,20 @@ func (srv *Server) GetSnapshot(pp *Peer) {
...
+//	downloaded the latest code that sends an empty db chunk for a non-existent prefix. We
+//	check if the prefix is the newly-added PrefixGroupMembershipIndex and if so, filter it out.
+func (srv *Server) CheckIfStatePrefixExistsForBlockHeight(blockHeight uint64, prefix []byte) bool {
+	switch prefix[0] {
+	case Prefixes.PrefixGroupMembershipIndex[0]:
+		if uint32(blockHeight) < srv.blockchain.params.ForkHeights.DeSoAccessGroupsBlockHeight {
+			return false
+		}
+	}
+	return true
```
lazynina, line 687:
```suggestion
	return prefix[0] != Prefixes.PrefixGroupMembershipIndex[0] || uint32(blockHeight) >= srv.blockchain.params.ForkHeights.DeSoAccessGroupsBlockHeight 
```
feels like we don't really need a switch-case statement here

lazynina (changes requested):
initial review - need to go through it again, but wanted to drop my first round of feedback in

lib/block_view_access_groups.go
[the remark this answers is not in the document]
keshavm02: That code is a bit outdated. We decided to form new txns for each:

create group
Add member
Remove member
Set member attribute
Set group attribute

Makes things a lot cleaner long-term

lib/block_view.go
```diff
@@ -252,11 +270,37 @@ func (bav *UtxoView) CopyUtxoView() (*UtxoView, error) {
...
+	for key, member := range bav.GroupMembershipKeyToAccessGroupMember {
+		newMember := *member
+		newView.GroupMembershipKeyToAccessGroupMember[key] = &newMember
+	}
+
+	newView.GroupMemberAttributes = make(map[GroupEnumerationKey]map[AccessGroupMemberAttributeType]*AttributeEntry, len(bav.GroupMemberAttributes))
+	for key, attributes := range bav.GroupMemberAttributes {
+		newAttributes := make(map[AccessGroupMemberAttributeType]*AttributeEntry, len(attributes))
+		for attribute, value := range attributes {
+			newValue := *value
```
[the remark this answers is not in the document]
keshavm02: Could be nil for eg. IsMuted value would be nil as there is no associated data. 

Basically, if an attribute for a member is in DB that means it’s set to true. The Value is simply if that attribute has any associated value like description/imageBytes etc.

lib/block_view_access_groups.go
```diff
@@ -0,0 +1,964 @@
...
+		if messagingKeyEntry.isDeleted {
+			continue
+		}
+		retMessagingKeyEntries = append(retMessagingKeyEntries, messagingKeyEntry)
+	}
+	return retMessagingKeyEntries, nil
+}
+
+// TODO: Update for Postgres
+func (bav *UtxoView) GetMessagesForUser(publicKey []byte, blockHeight uint32) (
```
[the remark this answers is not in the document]
keshavm02: done

lib/block_view_access_groups.go
```diff
@@ -0,0 +1,964 @@
...
+	_messageEntries []*MessageEntry, _messagingKeyEntries []*AccessGroupEntry, _err error) {
+
+	return bav.GetLimitedMessagesForUser(publicKey, math.MaxUint64, blockHeight)
+}
+
+// TODO: Update for Postgres
+func (bav *UtxoView) GetLimitedMessagesForUser(ownerPublicKey []byte, limit uint64, blockHeight uint32) (
```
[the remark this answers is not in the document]
keshavm02: done

lib/block_view_access_groups.go
```diff
@@ -0,0 +1,964 @@
...
+// setMessageEntryRotatingVersion sets the version of the messaging group key.
+// The version is used to determine the key rotation period.
+func setMessageEntryRotatingVersion(messageEntry *MessageEntry, blockHeight uint32, rotatingVersion uint64) {
+	if MigrationTriggered(uint64(blockHeight), DeSoAccessGroupsMigration) {
+		// Set the ExtraData["MessageEntryRotatingVersion"] to the provided value.
+		messageEntry.ExtraData[MessageRotatingVersion] = EncodeUint64(rotatingVersion)
+	}
+}
+
+func (bav *UtxoView) _connectMessagingGroup(
```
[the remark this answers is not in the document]
keshavm02: done

lib/block_view_access_groups.go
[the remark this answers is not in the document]
keshavm02: Currently Piotr is in the trenches finishing up those txns

lib/block_view_access_groups.go
```diff
@@ -0,0 +1,964 @@
...
+
+	if !EqualGroupKeyName(messagingGroupEntry.AccessGroupKeyName, NewGroupKeyName(keyName)) {
+		return fmt.Errorf("ValidateKeyAndNameWithUtxo: key name don't match for "+
+			"ownerPublicKey: %s", PkToString(ownerPublicKey, bav.Params))
+	}
+	return nil
+}
+
+// isMemberMuted returns true if the member is muted in the group.
+func (bav *UtxoView) isMemberMuted(
```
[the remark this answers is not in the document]
keshavm02: done

lib/block_view_types.go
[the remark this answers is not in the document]
keshavm02: done

lib/db_utils.go
keshavm02: done

lib/block_view_access_groups.go
keshavm02: done

lib/block_view.go
```diff
@@ -252,11 +270,37 @@ func (bav *UtxoView) CopyUtxoView() (*UtxoView, error) {
 		newView.MessageMap[txnHash] = &newMessage
 	}
 
-	// Copy messaging group data
-	newView.MessagingGroupKeyToMessagingGroupEntry = make(map[MessagingGroupKey]*MessagingGroupEntry, len(bav.MessagingGroupKeyToMessagingGroupEntry))
-	for pkid, entry := range bav.MessagingGroupKeyToMessagingGroupEntry {
+	// Copy access group data
+	newView.AccessGroupKeyToAccessGroupEntry = make(map[AccessGroupKey]*AccessGroupEntry, len(bav.AccessGroupKeyToAccessGroupEntry))
+	for pkid, entry := range bav.AccessGroupKeyToAccessGroupEntry {
```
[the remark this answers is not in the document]
keshavm02: understood, thank you! fixed now

lib/block_view.go
```diff
@@ -2276,11 +2320,31 @@ func (bav *UtxoView) _connectTransaction(txn *MsgDeSoTxn, txHash *BlockHash,
...
+		//} else if txn.TxnMeta.GetTxnType() == TxnTypeAccessGroupSetMemberAttribute {
+		//	totalInput, totalOutput, utxoOpsForTxn, err =
+		//		bav._connectAccessGroupSetMemberAttribute(
+		//			txn, txHash, blockHeight, verifySignatures)
+		//
+		//} else if txn.TxnMeta.GetTxnType() == TxnTypeAccessGroupSetGroupAttribute {
+		//	totalInput, totalOutput, utxoOpsForTxn, err =
+		//		bav._connectAccessGroupSetGroupAttribute(
+		//			txn, txHash, blockHeight, verifySignatures)
+
```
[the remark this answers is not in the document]
keshavm02: Yes sorry Piotr is working on that code now

lib/block_view_access_groups.go
```diff
@@ -0,0 +1,1026 @@
...
+	"github.com/pkg/errors"
+	"math"
+	"reflect"
+)
+
+// GetMessagingMember will check the membership index for membership of memberPublicKey in the group
+// <groupOwnerPublicKey, groupKeyName>. Based on the blockheight, we fetch the full group or we fetch
+// the simplified message group entry from the membership index. forceFullEntry is an optional parameter that
+// will force us to always fetch the full group entry.
+func (bav *UtxoView) GetMessagingMember(memberPublicKey *PublicKey, groupOwnerPublicKey *PublicKey,
```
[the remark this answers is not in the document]
keshavm02: Yes thanks 100% items have been refactored now!

lib/block_view_access_groups.go
```diff
@@ -0,0 +1,1026 @@
...
+		return mapValue
+	}
+
+	// If we get here, it means that the group has not been fetched in this utxoView. We fetch it from the db.
+	messagingGroupMember := DBGetMemberFromMembershipIndex(bav.Handle, bav.Snapshot, memberPublicKey, groupOwnerPublicKey, groupKeyName)
+	return messagingGroupMember
+}
+
+// SetMessagingMember will set the membership mapping of AccessGroupMember.
+func (bav *UtxoView) SetMessagingMember(messagingGroupEntry *AccessGroupEntry,
```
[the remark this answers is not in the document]
keshavm02: noted, fixing all setters!

lib/block_view_access_groups.go
```diff
@@ -0,0 +1,1026 @@
...
+
+	// set utxoView mapping
+	bav._setGroupMembershipKeyToAccessGroupMemberMapping(messagingGroupEntry, messagingGroupMember)
+}
+
+// GetMessagingEntry will check the membership index for membership of memberPublicKey in the group
+// <groupOwnerPublicKey, groupKeyName>. Based on the blockheight, we fetch the full group or we fetch
+// the simplified message group entry from the membership index. forceFullEntry is an optional parameter that
+// will force us to always fetch the full group entry.
+func (bav *UtxoView) GetMessagingEntry(memberPublicKey *PublicKey, groupOwnerPublicKey *PublicKey,
```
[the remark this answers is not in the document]
keshavm02: done

lib/block_view_access_groups.go
```diff
@@ -0,0 +1,1026 @@
...
+	groupMembershipKey := NewGroupMembershipKey(memberPublicKey, groupOwnerPublicKey, groupKeyName[:])
+
+	// If the group has already been fetched in this utxoView, then we get it directly from there.
+	if mapValue, exists := bav.GroupMembershipKeyToAccessGroupMember[*groupMembershipKey]; exists {
+		return mapValue
+	}
+
+	// If we get here, it means that the group has not been fetched in this utxoView. We fetch it from the db.
+	messagingGroupMember := DBGetMemberFromMembershipIndex(bav.Handle, bav.Snapshot, memberPublicKey, groupOwnerPublicKey, groupKeyName)
+	return messagingGroupMember
```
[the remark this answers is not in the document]
keshavm02: done

lib/block_view_access_groups.go
```diff
@@ -0,0 +1,1026 @@
...
+		bav._setAccessGroupKeyToAccessGroupEntryMapping(&messagingGroupKey.OwnerPublicKey, accessGroupEntry)
+	}
+	return accessGroupEntry
+
+	//}
+}
+
+func (bav *UtxoView) _setGroupMembershipKeyToAccessGroupMemberMapping(entry *AccessGroupEntry, member *AccessGroupMember) {
+	if entry == nil || member == nil {
+		return
```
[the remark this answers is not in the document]
keshavm02: Yes definitely. It shouldn't be called with a nil member.

lib/block_view_access_groups.go
```diff
@@ -0,0 +1,1026 @@
...
+	}
+	groupMembershipKey := NewGroupMembershipKey(member.GroupMemberPublicKey, entry.GroupOwnerPublicKey, entry.AccessGroupKeyName[:])
+	bav.GroupMembershipKeyToAccessGroupMember[*groupMembershipKey] = member
+}
+
+func (bav *UtxoView) _setAccessGroupKeyToAccessGroupEntryMapping(ownerPublicKey *PublicKey,
+	messagingGroupEntry *AccessGroupEntry) {
```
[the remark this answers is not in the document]
keshavm02: Yes sorry it's all 100% refactored now. I was prioritizing other code as i didn't know you were reviewing this afternoon 😅

lib/block_view_access_groups.go
```diff
@@ -0,0 +1,1026 @@
...
+	// This is our helper map to keep track of all user messaging keys.
+	accessKeysMap := make(map[AccessGroupKey]*AccessGroupEntry)
+
+	// Start by fetching all the messaging keys that we have in the UtxoView.
+	for accessKey, accessKeyEntry := range bav.AccessGroupKeyToAccessGroupEntry {
+		// We don't check for deleted entries now, we will do that later once we add messaging keys
+		// from the DB. For now we also omit the base key, we will add it later when querying the DB.
+
+		// Check if the messaging key corresponds to our public key.
+		if reflect.DeepEqual(accessKey.OwnerPublicKey[:], ownerPublicKey) {
```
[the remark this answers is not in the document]
keshavm02: noted, thank you

lib/block_view_access_groups.go
```diff
@@ -0,0 +1,1026 @@
...
+			accessKeysMap[accessKey] = accessKeyEntry
+			continue
+		}
+		// Now we will look for messaging keys where the public key is a recipient of a group chat.
+		if blockHeight >= bav.Params.ForkHeights.DeSoAccessGroupsBlockHeight {
+			member, err := DBGetGroupMemberForMessagingGroup(bav.Handle, bav.Snapshot, accessKeyEntry.GroupOwnerPublicKey, accessKeyEntry.AccessGroupKeyName, NewPublicKey(ownerPublicKey))
```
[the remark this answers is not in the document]
keshavm02: True, fixed!

lib/db_utils.go
keshavm02: +1

lib/block_view_access_groups.go
keshavm02: So right now the simple solution could be having a member attribute called "AcceptedInvitation" which denotes whether a member of a group accepts to be a part of it or now. 

This prevent spam big time, otherwise random channels could keep adding people.

This would also prevent garbage EncryptedKeys because they would decrypt garbage messages. The member can reject group invitation at this point due to unreadability of messages.

lib/db_utils.go
keshavm02: all a -> an fixed now

AeonSw4n:
TODO: Add Postgres support

lib/network.go
keshavm02: Yes in the works!

lib/network.go
keshavm02: +1

lib/block_view_access_groups.go
```diff
@@ -0,0 +1,1026 @@
...
+	return nil
+}
+
+// getGroupMemberAttributeEntry returns the group member attribute entry for the given group member.
+func (bav *UtxoView) getGroupMemberAttributeEntry(groupOwnerPublicKey *PublicKey, groupKeyName *GroupKeyName,
+	memberPublicKey *PublicKey, attributeType AccessGroupMemberAttributeType) (*AttributeEntry, error) {
+	// Create enumeration key.
+	enumerationKey := NewGroupEnumerationKey(groupOwnerPublicKey, groupKeyName[:], memberPublicKey)
+	// Check if enumerationKey exists in GroupMemberAttributes mapping.
+	attributeEntry, exists := bav.GroupMemberAttributes[*enumerationKey][attributeType]
```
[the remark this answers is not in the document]
keshavm02: Then the AttributeType for that enumerationKey would also not exist, making the above logic skip over the utxoView and checking in the DB. Tested the nested map logic in go playground :)

Something to note: Even though the AttributeType won't exist for a non-existent enumerationKey, when asked for a value for the non-existent key in the map, it would return a valid zero-value. So we definitely need to check the "exists" bool instead of value content to do this correctly.

lib/block_view_access_groups.go
keshavm02: What do you think of this current solution @lazynina ? I'm not sure how the app layer would determine the correctness of an EncryptedKey. The burden currently lies on the user.

lib/db_utils.go
keshavm02: Yes, sorry it's not clear in this function but we do have branching logic in place. I'll put in some comments here. Basically this function is being called only if we are already after the fork height. If we are before the fork height, there is a separate function being called that begins with "DEPRECATED".

lib/db_utils.go
keshavm02: fixed

lib/db_utils.go
keshavm02: Yes, definitely. Just thought we could save on storage especially when groups or channels might have million+ members! We're also not even calling this function anywhere yet as there is rarely a need to get ALL members. 

Should we still go ahead and store AccessGroupMember as the value here @lazynina @AeonSw4n ?

lib/network.go
keshavm02: Thanks, done. Now throwing error if len(operationTypeBytes) != 1

lib/server.go
keshavm02: fixed

PR #412, opened 2022-11-04
Access Groups, Dm Messages, Group Chats

AeonSw4n:

PR #414, opened 2022-11-06
Bugs fix and improvements to Node.Start() and Node.Stop()

### LINKED ISSUE #413: Panic when `node.nodeMessageChan <- lib.NodeErase` message is sent for the first time
When the message is sent for the first time using [node.nodeMessageChan <- lib.NodeErase](https://github.com/deso-protocol/core/blob/main/cmd/node.go#L233), it panics. This is because `node.IsRunning` is false when the message was sent. The value is set  at a [later point](https://github.com/deso-protocol/core/blob/main/cmd/node.go#L254).

hackintoshrao:
Context: I love the project and what it stands for. It's very important for an impactful and complex project like this to have great coverage of tests. I started working on writing unit and integration tests for the entire codebase. While I work on tests I'm fixing things as I find them. Here is the summary of the PR

- Fixes https://github.com/deso-protocol/core/issues/413
- Making Node.IsRunning flag atomic.
- Renaming test utility functions for better readability.
- Utility functions for changing the status of the node.
- Unit tests for the utility functions.


diamondhands0:
Wow great work! We're reviewing and should get back shortly.

integration_testing/tools.go
```diff
@@ -41,18 +42,18 @@ const MaxSyncBlockHeight = 1500
...
 		require.NoError(err)
 	}
 	return dbDir
 }
 
-// generateConfig creates a default config for a node, with provided port, db directory, and number of max peers.
+// GenerateTestConfig creates a default config for a node, with provided port, db directory, and number of max peers.
 // It's usually the first step to starting a node.
-func generateConfig(t *testing.T, port uint32, dataDir string, maxPeers uint32) *cmd.Config {
+func GenerateTestConfig(t *testing.T, port uint32, dataDir string, maxPeers uint32) *cmd.Config {
```
AeonSw4n, line 56: Any way we can deduplicate this and keep only a single `GenerateTestConfig` in cmd?

AeonSw4n (commented):
This is awesome Karthic, thank you so much for this contribution! I've added some small comments during my review. It would be great if we could incorporate them into the PR before we merge. Let me know what you think 😀

integration_testing/tools.go
hackintoshrao: SGTM. I'll do this change

cmd/node_test.go
```diff
@@ -0,0 +1,155 @@
...
+
+	testConfig := GenerateTestConfig(t, 18000, testDir, 10)
+
+	testConfig.ConnectIPs = []string{"deso-seed-2.io:17000"}
+
+	testNode := NewNode(&testConfig)
+
+	// expected running status should be false before the node is started
+	expectedRunningStatus := false
+	actualRunningstatus := testNode.IsRunning()
```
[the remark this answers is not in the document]
hackintoshrao: I'm particular about these things too, missed it by mistake. Thanks for pointing out.

cmd/node.go
```diff
@@ -34,9 +50,11 @@ type Node struct {
...
 	Postgres *lib.Postgres
 
-	// IsRunning is false when a NewNode is created, set to true on Start(), set to false
-	// after Stop() is called. Mainly used in testing.
-	IsRunning bool
+	// running is false when a NewNode is created, it is set to true on node.Start(),
+	// set to false after Stop() is called. Mainly used in testing.
+	// Use the convinience methods changeRunningStatus/StatusStartToStop/StatusStopToStart to change node status
+	// Use *Node.IsRunning() to retreive the node running status.
+	running *atomic.Bool
```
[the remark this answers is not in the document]
hackintoshrao: Thank you for the review. Is the Go 1.18 Compatibility documented anywhere in the docs? If not I'll make a documentation/README update.

Hmm, Atomic makes code much cleaner, but yes, will wait for a future 1.19 bump! Your recommendation sounds perfect, I'll soon ship the changes.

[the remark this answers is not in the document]
hackintoshrao: @AeonSw4n I've incorporated the changes. Let me know what you think.

cmd/node.go
```diff
@@ -251,7 +296,43 @@ func (node *Node) Start(exitChannels ...*chan struct{}) {
 			}
 		}
 	}
-	node.IsRunning = true
+
+	node.statusMutex.Lock()
+	// Load the node status.
+	// This is identify whether the node is initialized for the first time or it's a restart.
```
AeonSw4n, line 302: nit: to identify

AeonSw4n (commented):
Hey man, just reviewed your changes. Really solid code overall! I think before we're ready to merge this boyo into our codebase, I have to be confident we won't get screwed by the additional complexity entailed by the status changes and locking/unlocking of the `statusMutex`. We gotta make sure everything is intuitively forward compatible, and no dev in the future can break things too easily. Left you some suggestions, let me know what you think! I'd say after we address these things, we'll be ready to sync some nodes and make sure everything runs smooth as butter.

cmd/node.go
```diff
@@ -284,14 +371,139 @@ func (node *Node) Start(exitChannels ...*chan struct{}) {
...
+		glog.Errorf("Error stopping Node -- %v", err)
+		return err
+	}
+	return nil
+}
+
+// Loads the status of Node.
+// Modifying LoadStatus() and the validateStatus() functions and their tests are hard requirements
+// to add new status codes for the node.
+func (node *Node) LoadStatus() (NodeStatus, error) {
```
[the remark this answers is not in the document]
AeonSw4n: also maybe we make this method private and wrap it with some other `GetStatus()` that does actually hold the status mutex?

cmd/node.go
```diff
@@ -26,6 +27,42 @@ import (
...
 )
 
+var (
+	// ErrAlreadyStarted is returned when somebody tries to start an already
+	// running service.
+	ErrAlreadyStarted = errors.New("already started")
+	// ErrAlreadyStopped is returned when somebody tries to stop an already
+	// stopped service (without resetting it).
+	ErrAlreadyStopped = errors.New("already stopped")
+	// ErrNodeNeverStarted is returned when somebody tries to find or changge the
```
lazynina, line 37:
```suggestion
	// ErrNodeNeverStarted is returned when somebody tries to find or change the
```

cmd/node.go
```diff
@@ -26,6 +27,42 @@ import (
...
+	// running status of the server without never starting it once.
+	ErrNodeNeverStarted = errors.New("never started the node instance")
+	// Cannot set node status to NEVERSTARTED.
+	// NEVERSTARTED is the default status, cannot set deliberately.
+	ErrCannotSetToNeverStarted = errors.New("cannot set the node status to neverstarted")
+	// Error if the atomic compare of swap of the node status fails
+	ErrFailedtoChangeNodeStatus = errors.New("failed to change the status of the node")
```
lazynina, line 44: this variable does not appear to be used - if so, can we please remove it? If we do keep it, let's fix the capitalization on the name
```suggestion
	ErrFailedToChangeNodeStatus = errors.New("failed to change the status of the node")
```

cmd/node.go
```diff
@@ -34,9 +71,16 @@ type Node struct {
 	Config   *Config
 	Postgres *lib.Postgres
 
-	// IsRunning is false when a NewNode is created, set to true on Start(), set to false
-	// after Stop() is called. Mainly used in testing.
-	IsRunning bool
+	// status is nil when a NewNode is created, it is initialized and set to RUNNING [byte(1)] on node.Start(),
+	// set to STOPPED [byte(2)] after Stop() is called.
+
+	// Use the convinience methods changeNodeStatus/UpdateStatusRunning/UpdateStatusStopped to change node status.
```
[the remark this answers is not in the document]
lazynina:
```suggestion
	// Use the convenience methods changeNodeStatus/UpdateStatusRunning/UpdateStatusStopped to change node status.
```

cmd/node.go
```diff
@@ -34,9 +71,16 @@ type Node struct {
...
+	// Use *Node.IsRunning() to check if the node is running.
+	// Use *Node.LoadStatus() to retrieve the status of the node.
+	status *NodeStatus
+	// Held whenver the status of the node is read or altered.
```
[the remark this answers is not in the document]
lazynina:
```suggestion
	// Held whenever the status of the node is read or altered.
```

cmd/node.go
lazynina:
```suggestion
	// This is to identify whether the node is initialized for the first time or it's a restart.
```

cmd/node.go
```diff
@@ -251,7 +296,43 @@ func (node *Node) Start(exitChannels ...*chan struct{}) {
...
+
+	node.statusMutex.Lock()
+	// Load the node status.
+	// This is identify whether the node is initialized for the first time or it's a restart.
+	status, err := node.LoadStatus()
+	// FIXME: Replace panics with return value,
+	// and leave the decision making to the client library of the core whether to crash the app.
+	if err != nil {
+		panic("Failed to load node status")
+	}
```
lazynina, line 308: this needs to be fixed before we can merge this in. don't like having panics lying around

cmd/node.go
```diff
@@ -273,7 +358,9 @@ func (node *Node) Start(exitChannels ...*chan struct{}) {
 		case <-syscallChannel:
 		}
 
+		node.statusMutex.Lock()
 		node.Stop()
+		node.statusMutex.Unlock()
```
lazynina, line 363: Generally I prefer to have the handling of the mutexes confined to the methods that actually interact with the field we're trying to lock on. Could we move the lock/unlock on statusMutex into the `node.Stop` call?

cmd/node.go
```diff
@@ -251,7 +296,43 @@ func (node *Node) Start(exitChannels ...*chan struct{}) {
 			}
 		}
 	}
-	node.IsRunning = true
+
+	node.statusMutex.Lock()
+	// Load the node status.
+	// This is identify whether the node is initialized for the first time or it's a restart.
+	status, err := node.LoadStatus()
```
lazynina, line 303: same as I mentioned below about moving mutex usage into these helper function. Can we move the lock/unlock logic into `node.LoadStatus` and `node.UpdateStatusRunning`?

cmd/node.go
```diff
@@ -284,14 +371,139 @@ func (node *Node) Start(exitChannels ...*chan struct{}) {
...
+		glog.Errorf("Error stopping Node -- %v", err)
+		return err
+	}
+	return nil
+}
+
+// Changes node status to RUNNING.
+func (node *Node) UpdateStatusRunning() error {
+	if err := node.changeNodeStatus(RUNNING); err != nil {
+		glog.Errorf("Error stopping Node -- %v", err)
```
lazynina, line 388: Shouldn't this be `Error running Node`?

cmd/node.go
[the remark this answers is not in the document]
lazynina: would be best if all the mutex logic lives within these functions instead of handled by the callers.

cmd/node.go
```diff
@@ -284,14 +371,139 @@ func (node *Node) Start(exitChannels ...*chan struct{}) {
...
+func (node *Node) IsRunning() bool {
+	status, err := node.LoadStatus()
+
+	if err != nil {
+		return false
+	}
+	if status == RUNNING {
+		return true
+	}
+	return false
```
lazynina, line 492:
```suggestion
	return status == RUNNING
```

cmd/node.go
```diff
@@ -284,14 +371,139 @@ func (node *Node) Start(exitChannels ...*chan struct{}) {
...
 func (node *Node) Stop() {
 	node.runningMutex.Lock()
 	defer node.runningMutex.Unlock()
 
-	if !node.IsRunning {
+	// Change nodes running status to stop
+	// Node has to be in RUNNING status to be able to STOP!
+	// FIXME: Return error in cases where the the node fails to stop. This makes the function testable.
+	if err := node.UpdateStatusStopped(); err != nil {
+		glog.Errorf("Error stopping Node -- %v", err)
```
lazynina, line 503: seems like we glog this same statement twice. we should either do it in the `UpdateStatusStopped` or here. My preference is that the caller handles how to glog so we don't blow up our logs, although I don't believe we'd hit this error statement in production often

cmd/node.go
```diff
@@ -353,19 +565,36 @@ func (node *Node) listenToNodeMessages(exitChannels ...*chan struct{}) {
...
-		if !node.IsRunning {
-			panic("Node.listenToNodeMessages: Node is currently not running, nodeMessageChan should've not been called!")
-		}
-		glog.Infof("Node.listenToNodeMessages: Stopping node")
-		node.Stop()
-		glog.Infof("Node.listenToNodeMessages: Finished stopping node")
 		switch operation {
+		case lib.NodeRestart:
+			// Using Mutex while accessing the Node status to avoid any race conditions
+			node.statusMutex.Lock()
```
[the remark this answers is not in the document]
lazynina: yes^ - big piece from my review is that we don't want to have these mutexes flying around all the calling functions, but should instead handle the lock/unlock in the helper functions as much as we can.

cmd/node.go
```diff
@@ -353,19 +565,36 @@ func (node *Node) listenToNodeMessages(exitChannels ...*chan struct{}) {
...
+				node.Stop()
+			}
+			node.statusMutex.Unlock()
+			glog.Infof("Node.listenToNodeMessages: Finished stopping node")
+
 			if err := os.RemoveAll(node.Config.DataDirectory); err != nil {
 				glog.Fatal(lib.CLog(lib.Red, fmt.Sprintf("IMPORTANT: Problem removing the directory (%v), you "+
 					"should run `rm -rf %v` to delete it manually. Error: (%v)", node.Config.DataDirectory,
 					node.Config.DataDirectory, err)))
-				return
```
lazynina, line 368: why are we removing the `return` statement here?

cmd/node_test.go
```diff
@@ -0,0 +1,259 @@
...
+	expectedRunningStatus := false
+	actualRunningStatus := testNode.IsRunning()
+	require.Equal(t, actualRunningStatus, expectedRunningStatus)
+
+	// Start the node
+	testNode.Start()
+	// expected running status should be true after the server is started
+	expectedRunningStatus = true
+	actualRunningStatus = testNode.IsRunning()
+	require.Equal(t, actualRunningStatus, expectedRunningStatus)
```
lazynina, line 30:
```suggestion
	require.True(t, testNode.IsRunning(), expectedRunningStatus)
```

cmd/node_test.go
```diff
@@ -0,0 +1,259 @@
...
+
+	// stop the node
+	testNode.Stop()
+	expectedRunningStatus = false
+	actualRunningStatus = testNode.IsRunning()
+	require.Equal(t, actualRunningStatus, expectedRunningStatus)
```
lazynina, line 36:
```suggestion
	require.False(t, testNode.IsRunning())
```

cmd/node_test.go
```diff
@@ -0,0 +1,259 @@
...
+	testConfig := GenerateTestConfig(t, 18000, testDir, 10)
+
+	testConfig.ConnectIPs = []string{"deso-seed-2.io:17000"}
+
+	testNode := NewNode(&testConfig)
+
+	// expected running status should be false before the node is started
+	expectedRunningStatus := false
+	actualRunningStatus := testNode.IsRunning()
+	require.Equal(t, actualRunningStatus, expectedRunningStatus)
```
lazynina, line 23:
```suggestion
	require.False(t, testNode.IsRunning())
```

cmd/node_test.go
```diff
@@ -0,0 +1,259 @@
...
+	actualErr = testNode.UpdateStatusRunning()
+	require.ErrorIs(t, actualErr, expErr)
+
+	// Stop the node
+	testNode.Stop()
+	// Should be able to change status to RUNNING from STOP.
+	actualErr = testNode.UpdateStatusRunning()
+	require.NoError(t, actualErr)
+	// Once the running flag is changed, the isRunning function should return true
+	require.Equal(t, true, testNode.IsRunning())
```
lazynina, line 71:
```suggestion
	require.True(t, testNode.IsRunning())
```

lazynina (changes requested):
Agreed with all feedback from @AeonSw4n . Main input from me is to move all the mutex handling into the helper functions that handle stopping and updating status so the caller doesn't have to worry about it.

cmd/node.go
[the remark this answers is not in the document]
hackintoshrao: @AeonSw4n @lazynina Makes sense. I'll abstract the lock logic from the user, and introduce it inside the helper functions.

cmd/node.go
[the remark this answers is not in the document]
hackintoshrao: I'll use a small check next time. Sorry about the typos @AeonSw4n @lazynina

cmd/node.go
```diff
@@ -273,7 +358,9 @@ func (node *Node) Start(exitChannels ...*chan struct{}) {
 		case <-syscallChannel:
 		}
 
+		node.statusMutex.Lock()
```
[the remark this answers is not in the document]
hackintoshrao: > Agreed with all feedback from @AeonSw4n . Main input from me is to move all the mutex handling into the helper functions that handle stopping and updating status so the caller doesn't have to worry about it.

Agreed.

cmd/node.go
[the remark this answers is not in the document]
hackintoshrao: Makes sense.

cmd/node.go
hackintoshrao: > Generally I prefer to have the handling of the mutexes confined to the methods that actually interact with the field we're trying to lock on. Could we move the lock/unlock on statusMutex into the `node.Stop` call?

I'll move these locks inside `node.Stop()`

cmd/node.go
hackintoshrao: @lazynina  You're right. I'll fix this.

cmd/node.go
hackintoshrao: @lazynina  Thank you for the suggestion. I guess `gofmt -s` should catch these simplifications, will run it before I push again.

cmd/node.go
hackintoshrao: @lazynina `Fatal` would any way exit right?

hackintoshrao:
@AeonSw4n @lazynina : I made the changes as per your suggestions, thank you for the great inputs and the timely reviews. Here's the summary of my changes.

> I find "get/set" more intuitive than "load/change" (though it's just my personal preference). Since this getter isn't holding a mutex, maybe we rename it to GetStatusWithoutLock 

Changed LoadStatus->GetStatus and ChangeStatus-> GetStatus

> also maybe we make this method private and wrap it with some other GetStatus() that does actually hold the status mutex? 
> would be best if all the mutex logic lives within these functions instead of handled by the callers.

created private functions which suffix `withoutLock`, and wrapper GetStatus and SetStatus which uses locks inside them. 
Moved the locks which were earlier held at multiple places to a few set location at the beginning of function definition.

> It feels like you're using statusMutex very similarly to how runningMutex is used. How about we just get rid of runningMutex and replace it with the statusMutex? In this case, we would also get rid of node.statusMutex.Lock()/Unlock() in here, since the mutex will be held inside of node.Stop()

Got rid of `runningMutex` and replaced it with `statusMutex`. Now, `statusMutex` is held inside of node.Stop() and node.Start(), this eliminates the use of locks at multiple places.

> let's move the node.statusMutex.Lock()/Unlock() and the if !node.IsRunning(){...} condition inside the node.Stop(). Since Stop() should never be called when node is already stopped it feels reasonable to isolate everything inside Stop. Same goes for the NodeErase case.

Node.Stop() automatically takes care of validating the current status of the node before stopping. Hence, removed the `!node.IsRunning` check. 

> This is the setter for the node status, similarly to my comment about the getter LoadStatus(), maybe we rename it to setNodeStatusWithoutLock()?

Renamed `changeNodeStatus` to `changeStatusWithoutLock`, and created a wrapper `ChangeStatus` which holds the lock internally.

>  big piece from my review is that we don't want to have these mutexes flying around all the calling functions, but should instead handle the lock/unlock in the helper functions as much as we can.

Mutexes don't fly anymore :) 🚀

AeonSw4n:
@hackintoshrao thanks for making the updates! Just wanted to make a quick update and let you know reviewing and testing all your code is a high priority for me, and will get to it once I finish being tunnel-visioned on access groups & group chats.

hackintoshrao:
Thanks @AeonSw4n . Just that this PR will enable me to do tons of contributions. I'm sweeping through the codebase writing tests, but I can't move forward till this is merged. 

May be I can help with access groups and chats by writing tests and manually testing them.

AeonSw4n:
Oh, that's so awesome @hackintoshrao I'm so excited to see your other contributions! Okay, I'll look into your code in the evening. Speaking of tests, I've actually been working on a new transaction unit testing framework lately. I'll PR it out of my current code so it's easier to read lol -- figured that might be interesting to you. 

So cool to see your readiness to help with access groups! If you feel like checking out the code the latest branch is in this PR #412. Let's wrap this PR, and I'll think of what could be the best use of your time. :)

hackintoshrao:
Thanks @AeonSw4n . Getting this merged will also help me fix tests for the #412 PR. Eager to check out the transaction test framework.

cmd/run.go
```diff
@@ -29,7 +29,9 @@ func Run(cmd *cobra.Command, args []string) {
 	node.Start(&shutdownListener)
 
 	defer func() {
+		node.statusMutex.Lock()
```
AeonSw4n, line 32: I think we can get rid of this guy now, right?

AeonSw4n (approved):
Just reviewed your latest code, and I've got some minor change requests, but giving you an approve because your code is really awesome! One more comment I have is, do we need these `.DS_Store` files? We didn't have them before and they feel a bit redundant.

Once again thank you for this contribution and your flexibility when faced with our feedback 😃 @lazynina @diamondhands0 can you guys also take a look so we can merge?

hackintoshrao:
> Just reviewed your latest code, and I've got some minor change requests, but giving you an approve because your code is really awesome! One more comment I have is, do we need these `.DS_Store` files? We didn't have them before and they feel a bit redundant.

Thank you @AeonSw4n :) I'll remove the .DS files, they are redundant. 

> 
> Once again thank you for this contribution and your flexibility when faced with our feedback 😃 @lazynina @diamondhands0 can you guys also take a look so we can merge?

I learned quite a bit from the reviews :)

hackintoshrao:
@AeonSw4n Found the transaction test framework inside the `block_view_test.go`. Working on some refactoring and tests using the test suite. Is there a better channel to communicate than Github comments😅

lazynina (changes requested):
@hackintoshrao - looking very good. I've opened a PR against your fork with some changes that would simplify the diff between your fork and the existing code in core. I think it'll be faster for us to get it through with a smaller diff - we can keep the existing function names and move all the testing logic into the `integration_testing` package.

AeonSw4n:
> @AeonSw4n Found the transaction test framework inside the `block_view_test.go`. Working on some refactoring on the suite to improve its usability, for instance, to try parallel transaction scenarios with multiple accounts and verify their sanity.
> 
> Is there a better channel to communicate while I work on the codebase than Github comments😅

Prallelizing tests/transactions could be fun @hackintoshrao :) Though keep in mind they will always be processed sequentially in the mempool. We are planning to build actual prallelization for transaction processing a little further on the horizon -- probably after/concurrently with PoS. I've also refactored the testing framework into another PR #422. Before it's merged I'm planning on adding a state verification for before & after connect-disconnect to Postgres (currently this only works for Badger) and just documenting everything a little better.

Re communication channel: certainly! I think you were chatting with Dylan over Discord, is that a good channel? I'll reach out to you soon.

hackintoshrao:
@AeonSw4n Discord is fine, I just need to map the Github accounts to the Discord user names! 

I'll look into #422 and look into opportunities to contribute.

hackintoshrao:
@lazynina The PR is ready as per the suggestions, do let me know if there's anything more needs to be done for this to be merged :)

lazynina:
> @lazynina The PR is ready as per the suggestions, do let me know if there's anything more needs to be done for this to be merged :)

@hackintoshrao I have been testing this out by running and stopping a node, but ran into a case where one of the status change validations returns an error and the node will not switch to running. Trying to figure out exactly what's going on and to test out more stop + restart situations to make sure we don't have nodes that get stuck. I have only reproduced it once but it's a little unclear what is happening. As I have more details, I'll share them

hackintoshrao:
Thank you @lazynina . Please do report the edge scenarios when you have them.

lazynina (approved):

lazynina:
> Thank you @lazynina . Please do report the edge scenarios when you have them.

I can't reproduce it - maybe it happened in a dream! @diamondhands0 for final review.

hackintoshrao:
Thanks @lazynina . Will for the review by @diamondhands0 .

lazynina:
@hackintoshrao - can you provide more details on how to reproduce the bug ( #413 ) and what specifically in this PR is fixing that bug? @diamondhands0

hackintoshrao:
The bug occurs when the [`shouldRestart`](https://github.com/deso-protocol/core/blob/724f6d2fb64dd0a1c3aaa138fbdf33bb619356eb/cmd/node.go#L233) flag is `true` during the first run of the server.

In this scenario, `nodeMessageChan` [receives](https://github.com/deso-protocol/core/blob/724f6d2fb64dd0a1c3aaa138fbdf33bb619356eb/cmd/node.go#L236) a `lib.NodeErase` message, but at this point, `node.IsRunning` is still set to `false`.

Hence, instead of safely recovering, it panics [here](https://github.com/deso-protocol/core/blob/724f6d2fb64dd0a1c3aaa138fbdf33bb619356eb/cmd/node.go#L359). 


Here is a quick screen recording explainer https://www.loom.com/share/410232d7a6ab4d1e8610c9259de1cbf6 .

PR #415, opened 2022-11-09
Update Minor Error Message in blockchain.go

davidjwriter:
Fixing minor copy/paste error in an error message for CreateUpdateNFTTxn. It was copied from a similar error for CreateAcceptNFTBidTxn.


AeonSw4n (approved):
Good catch, thanks!

PR #417, opened 2022-11-12
Added "IncludeComments" flag to allow to also fetch comment posts.

DeSoEngineer:
See: https://github.com/deso-protocol/dips/discussions/93#discussioncomment-4120746

⚠ Breaks backend code, unless the PR there is also accepted.


donhardman:
They do not use Postgres in production and it's still in beta and not finished state, they use BadgerDB

DeSoEngineer:
Learning new things as we go 😂

I cared for both branches, as far as I understand the code.

donhardman:
That's great; keep going! 🚀

PR #419, opened 2022-11-15
Add onlyNFTS and onlyPosts filter

erwin-willems:
This one is needed for / related to: https://github.com/deso-protocol/backend/pull/413

### DIFF
--- a/lib/block_view_post.go
+++ b/lib/block_view_post.go
@@ -450,7 +450,7 @@ func (bav *UtxoView) GetAllPosts() (_corePosts []*PostEntry, _commentsByPostHash
 	return allCorePosts, commentsByPostHash, nil
 }
 
-func (bav *UtxoView) GetPostsPaginatedForPublicKeyOrderedByTimestamp(publicKey []byte, startPostHash *BlockHash, limit uint64, mediaRequired bool, nftRequired bool) (_posts []*PostEntry, _err error) {
+func (bav *UtxoView) GetPostsPaginatedForPublicKeyOrderedByTimestamp(publicKey []byte, startPostHash *BlockHash, limit uint64, mediaRequired bool, onlyNFTs bool, onlyPosts bool) (_posts []*PostEntry, _err error) {
 	if bav.Postgres != nil {
 		var startTime uint64 = math.MaxUint64
 		if startPostHash != nil {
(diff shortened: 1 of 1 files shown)


lib/block_view_post.go
```diff
@@ -450,7 +450,7 @@ func (bav *UtxoView) GetAllPosts() (_corePosts []*PostEntry, _commentsByPostHash
 	return allCorePosts, commentsByPostHash, nil
 }
 
-func (bav *UtxoView) GetPostsPaginatedForPublicKeyOrderedByTimestamp(publicKey []byte, startPostHash *BlockHash, limit uint64, mediaRequired bool, nftRequired bool) (_posts []*PostEntry, _err error) {
+func (bav *UtxoView) GetPostsPaginatedForPublicKeyOrderedByTimestamp(publicKey []byte, startPostHash *BlockHash, limit uint64, mediaRequired bool, onlyNFTs bool, onlyPosts bool) (_posts []*PostEntry, _err error) {
```
lazynina, line 453: we should validate that `onlyNFTs` and `onlyPosts` are not both true. If they are both true, we should return an error since we're just going to spin our wheels.

lazynina (changes requested):

lib/block_view_post.go
erwin-willems: Added validation (Nice find!)

lazynina (changes requested):

lib/block_view_post.go
```diff
@@ -450,23 +450,31 @@ func (bav *UtxoView) GetAllPosts() (_corePosts []*PostEntry, _commentsByPostHash
...
```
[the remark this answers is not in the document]
lazynina: validation needs to be outside of if statement for postgres

lib/block_view_post.go
```diff
@@ -450,23 +450,34 @@ func (bav *UtxoView) GetAllPosts() (_corePosts []*PostEntry, _commentsByPostHash
...
+	}
 	if bav.Postgres != nil {
 		var startTime uint64 = math.MaxUint64
 		if startPostHash != nil {
 			startPostEntry := bav.GetPostEntryForPostHash(startPostHash)
 			startTime = startPostEntry.TimestampNanos
 		}
+		if onlyNFTs && onlyPosts {
+			return nil, fmt.Errorf("GetPostsPaginatedForPublicKeyOrderedByTimestamp: onlyNFTs and onlyPosts can not be enabled both")
+		}
```
lazynina, line 465:
```suggestion
```

PR #422, opened 2022-11-29
Transaction testing framework

AeonSw4n:


lib/block_view_test.go
```diff
@@ -61,6 +67,665 @@ var (
...
+	tes._testVectorDependency = make(map[transactionTestIdentifier][]transactionTestIdentifier)
+
+	for _, testVectorBlock := range tes.testVectorBlocks {
+		// Run all the test vectors for this block.
+		dbEntriesBefore := tes.GetAllStateDbEntries(tm)
+		tes.testConnectBlock(tm, testVectorBlock)
+		tes.testDisconnectBlock(tm, testVectorBlock)
+		dbEntriesAfter := tes.GetAllStateDbEntries(tm)
+		tes.compareBeforeAfterDbEntries(dbEntriesBefore, dbEntriesAfter)
+		glog.Infof(CLog(Yellow, "RunBadgerTest: successfully connected/disconnected block and verified db state"))
```
lazynina, line 289: Can we set a glog level such that our builds don't have all these print statements in them please? applies to all print statements.

lib/block_view_test.go
```diff
@@ -61,6 +67,665 @@ var (
...
+	for publicKey, amountNanos := range config.fundPublicKeysWithNanosMap {
+		tes.fundPublicKey(tm, publicKey, amountNanos)
+	}
+
+	return tm
+}
+
+// Fund provided public key with the desired amount of DeSo in nanos.
+func (tes *transactionTestSuite) fundPublicKey(tm *transactionTestMeta, publicKey PublicKey, amountNanos uint64) {
+	// Note we don't need to pass Postgres here, because _doBasicTransferWithVieFlush takes it from Blockchain.
```
lazynina, line 386:
```suggestion
	// Note we don't need to pass Postgres here, because _doBasicTransferWithViewFlush takes it from Blockchain.
```

lazynina (commented):
seems like this tests well, but it is a little confusing to go through this code. @diamondhands0 - feel free to look when you'd like to give some input here.

PR #423, opened 2022-12-01
Group test blocks using t.Run()

hackintoshrao:


hackintoshrao:
@AeonSw4n I just grouped the tests in #422 using t.Run().  Can you merge it into #422 if this looks good? I'll be adding more cases. Also looking into adding a state verification for before & after connect-disconnect to Postgres tests.

AeonSw4n:
Sure thing, looks great! Happy to merge :)

That's cool you've looked into the state verification. The idea is to make sure that the blockchain state (as expressed in badger/postgres) reverts to its proper previous configuration after a block is disconnected. We could turn this concept into a generic sanity-check run by the testing framework, to automatically test that the transaction _disconnect logic was correctly coded. This feature already works for the `RunBadgerTest()` function. In the code, you can see that we call the `GetAllStateDbEntries()` method twice. First, we call it to fetch the entire state before a block is connected -- which is stored under `dbEntriesBefore`. Second, we fetch the state again after connecting and disconnecting a block and store it in `dbEntriesAfter`. We then call `compareBeforeAfterDbEntries` on those two state snapshots to verify that they are identical (which is the expected behavior). 

It would be nice to have a similar functionality in `RunPostgresTest()`. I was planning on writing it at some point before merging my original PR. The tricky part is how do you get all tables in postgres to run a `SELECT * FROM` query, though we might just prefer to specify the affected tables explicitly through a field of `transactionTestVector` struct or something.

hackintoshrao:
Thanks, @AeonSw4n for the explanation. Will keep you posted.

PR #430, opened 2022-12-12
Mf/add advanced querying for associations

mattfoley8:


lib/db_utils.go
```diff
@@ -8644,55 +8692,97 @@ func DBGetUserAssociationIdsByAttributes(handle *badger.DB, snap *Snapshot, asso
...
+	}
+
+	// Map deleted UTXO AssociationIDs to keys.
+	deletedUtxoKeys := NewSet[string]([]string{})
+	for _, deletedUtxoAssociationId := range deletedUtxoAssociationIDs.ToSlice() {
+		deletedUtxoKey, err := _dbUserAssociationIdToKey(handle, snap, deletedUtxoAssociationId, prefixType)
+		if err != nil {
+			return nil, nil, errors.Wrapf(err, "DBGetUserAssociationIdsByAttributes: ")
+		}
+		if deletedUtxoKey != nil {
```
lazynina, line 8716: when would deletedUtxoKey ever be nil?

lib/db_utils.go
```diff
@@ -8573,50 +8605,66 @@ func DBGetPostAssociationByAttributesWithTxn(txn *badger.Txn, snap *Snapshot, qu
...
 		}
 	} else if associationQuery.TargetUserPKID != nil {
 		// PrefixUserAssociationByTargetUser: TargetUserPKID, AssociationType, AssociationValue, TransactorPKID
+		prefixType = Prefixes.PrefixUserAssociationByTargetUser
 		keyPrefix = append(keyPrefix, Prefixes.PrefixUserAssociationByTargetUser...)
 		keyPrefix = append(keyPrefix, associationQuery.TargetUserPKID.ToBytes()...)
 	} else {
 		// TransactorPKID == nil, TargetUserPKID == nil
-		return nil, errors.New("DBGetUserAssociationIdsByAttributes: invalid query params")
+		return nil, nil, errors.New("DBGetUserAssociationIdsByAttributes: invalid query params")
```
lazynina, line 8667: I generally like to have my error cases up front. So `if TransactorPKID == nil && TargetUserPKID == nil { throw error }`. Reducing the number of `else` and `else if` statements is encouraged

lib/db_utils.go
```diff
@@ -8785,25 +8877,98 @@ func DBGetPostAssociationIdsByAttributes(handle *badger.DB, snap *Snapshot, asso
...
+	}
+
+	// Map deleted UTXO AssociationIDs to keys.
+	deletedUtxoKeys := NewSet[string]([]string{})
+	for _, deletedUtxoAssociationId := range deletedUtxoAssociationIDs.ToSlice() {
+		deletedUtxoKey, err := _dbPostAssociationIdToKey(handle, snap, deletedUtxoAssociationId, prefixType)
+		if err != nil {
+			return nil, nil, errors.Wrapf(err, "DBGetPostAssociationIdsByAttributes: ")
+		}
+		if deletedUtxoKey != nil {
```
lazynina, line 8901: when would deletedUtxoKey be nil?

lib/db_utils.go
```diff
@@ -8785,25 +8877,98 @@ func DBGetPostAssociationIdsByAttributes(handle *badger.DB, snap *Snapshot, asso
...
+	}
+	if associationEntry == nil {
+		return nil, nil
 	}
-	return associationIDs, nil
+	key, err := DBKeyForUserAssociationByPrefix(associationEntry, prefixType)
+	if err != nil {
+		return nil, err
+	}
+	return key, nil
```
lazynina, line 8950: can we just return the response from the function?
```suggestion
	return DBKeyForUserAssociationByPrefix(associationEntry, prefixType)
```

lib/db_utils.go
```diff
@@ -8785,25 +8877,98 @@ func DBGetPostAssociationIdsByAttributes(handle *badger.DB, snap *Snapshot, asso
...
+) ([]byte, error) {
+	// Converts a PostAssociationID to a db key.
+	associationEntry, err := DBGetPostAssociationByID(handle, snap, associationID)
+	if err != nil {
+		return nil, err
+	}
+	if associationEntry == nil {
+		return nil, nil
+	}
+	key, err := DBKeyForPostAssociationByPrefix(associationEntry, prefixType)
```
lazynina, line 8967: same as above

lib/db_utils.go
```diff
@@ -8974,3 +9139,89 @@ func DBDeletePostAssociationWithTxn(txn *badger.Txn, snap *Snapshot, association
...
+
+	dbErr := db.View(func(txn *badger.Txn) error {
+		var err error
+		keysFound, valsFound, err = _enumerateKeysForPrefixWithLimitOffsetOrderWithTxn(
+			txn, prefix, limit, lastSeenKey, sortDescending, deletedUtxoKeys,
+		)
+		if err != nil {
+			return err
+		}
+		return nil
```
lazynina, line 9162: can't we just return `err`?
```suggestion
		return err
```

lib/db_utils.go
```diff
@@ -8974,3 +9139,89 @@ func DBDeletePostAssociationWithTxn(txn *badger.Txn, snap *Snapshot, association
...
+		if limit > 0 && len(keysFound) >= limit {
+			break
+		}
+		// Copy key.
+		key := nodeIterator.Item().Key()
+		keyCopy := make([]byte, len(key))
+		copy(keyCopy[:], key[:])
+		// Skip if key is the last seen key.
+		if bytes.Equal(keyCopy, lastSeenKey) {
+			continue
```
lazynina, line 9211: should we track whether or not we've seen the `lastSeenKey` so we can avoid using `bytes.Equal` on future keys?

lib/generics.go
```diff
@@ -0,0 +1,60 @@
...
+	for _, element := range elements {
+		set.Add(element)
+	}
+	return set
+}
+
+func (set *Set[T]) Add(element T) {
+	// Add element to set.
+	if !set.Includes(element) {
+		set._innerMap[element] = set.Size()
```
lazynina, line 22: hmm isn't it possible that we'll have multiple keys with the same value if we have adds and removes? Maybe a better way to do this to ensure ordering would be to have a monotonically increasing integer in the set struct and just add 1 to it with each add operation (and never decrease it with deletes).

lazynina (changes requested):

lib/db_adapter.go
```diff
@@ -62,32 +65,132 @@ func (adapter *DbAdapter) GetPostAssociationByAttributes(associationEntry *PostA
...
+	return DBGetPostAssociationIdsByAttributes(adapter.badgerDb, adapter.snapshot, associationQuery, deletedUtxoAssociationIDs)
+}
+
+func (adapter *DbAdapter) SortUserAssociationEntriesByPrefix(
+	associationEntries []*UserAssociationEntry,
+	prefixType []byte,
+	sortDescending bool,
+) ([]*UserAssociationEntry, error) {
+	// Postgres sorts results by AssociationID.
+	if adapter.postgresDb != nil {
```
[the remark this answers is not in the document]
mattfoley8: DBKeyForEntry is in-memory whereas postgres sorts by AssociationID in the database. IMO it's ok for postgres and badger to diverge in the non-consensus querying. Postgres has more lax rules around parameters (since order of badger key doesn't matter). And paginates through association entries by AssociationID instead of badger key prefix.

lib/generics.go
mattfoley8: Yeah, that's a good idea. Will add that.

lib/db_adapter.go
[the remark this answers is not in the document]
lazynina: sounds good - resolved

lib/db_adapter.go
```diff
@@ -62,32 +65,132 @@ func (adapter *DbAdapter) GetPostAssociationByAttributes(associationEntry *PostA
 	return DBGetPostAssociationByAttributes(adapter.badgerDb, adapter.snapshot, associationEntry)
 }
 
-func (adapter *DbAdapter) GetUserAssociationsByAttributes(associationQuery *UserAssociationQuery) ([]*UserAssociationEntry, error) {
+func (adapter *DbAdapter) GetUserAssociationsByAttributes(
+	associationQuery *UserAssociationQuery,
+	deletedUtxoAssociationIDs *Set[*BlockHash],
```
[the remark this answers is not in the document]
mattfoley8: That way we can return nil from any functions that want to return a set.

lib/block_view_association.go
```diff
@@ -731,32 +731,86 @@ func (bav *UtxoView) GetUserAssociationsByAttributes(associationQuery *UserAssoc
...
+		if !isNewEntry {
+			continue
+		}
+		utxoAssociationEntries = append(utxoAssociationEntries, associationEntry)
 	}
-	// Convert map to slice.
-	var associationEntries []*UserAssociationEntry
-	for _, matchingAssociationEntry := range associationEntryMap {
-		associationEntries = append(associationEntries, matchingAssociationEntry)
+	if associationQuery.SortDescending {
```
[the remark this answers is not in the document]
mattfoley8: If we sort ascending, we pull from the database first and the view second (deleting any entries that are deleted in the view). If we sort descending, we pull from the UTXO view first and the database second. So we can't do an in-memory sort.

lib/block_view_association.go
```diff
@@ -845,46 +902,100 @@ func (bav *UtxoView) _getUtxoUserAssociationEntriesByAttributes(
...
+		}
+		// Then add UTXO view association entries.
+		for _, associationEntry := range utxoAssociationEntries {
+			if associationQuery.Limit > 0 && len(associationEntries) >= associationQuery.Limit {
+				break
+			}
+			if !addedAssociationEntryIDs.Includes(associationEntry.AssociationID) {
+				associationEntries = append(associationEntries, associationEntry)
+			}
+		}
```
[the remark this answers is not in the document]
mattfoley8: Replied above.

lib/block_view_association.go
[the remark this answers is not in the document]
lazynina: well you're not guaranteed to have entries in the view be "newer" than entries in the DB, does this change how we should sort?

lib/db_utils.go
mattfoley8: If there is no AssociationEntry found for that AssociationID, then deletedUtxoKey will be nil. This can happen and a unit test caught this edge case when an association is deleted in the UTXO view and is never persisted in the database. In that case, we don't need to include the key in our set of deleted keys, since the association never made it to the database (and it won't be returned from the UTXO view either since it's deleted).

lib/db_utils.go
mattfoley8: See comment above.

lib/db_utils.go
mattfoley8: I generally agree, but appending these keys is kinda tricky since we don't break or return so have to use if/else if/elses to control logic flow. Here we have a `if A != nil else if B != nil else raise Error` which IMO is clearer than `if A == nil && B == nil raise Error; if A != nil else if B != nil` but LMK if you disagree.

lib/db_adapter.go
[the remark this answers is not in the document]
mattfoley8: But will update so don't have to do nil checks.

lib/db_utils.go
lazynina: nah it's fine, leave it.

lib/block_view_association.go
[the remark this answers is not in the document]
mattfoley8: IMO no. It feels like we just need one to return db then UTXO view and the other return UTXO view then db. Otherwise, it's weird for an end-user to query order ASC and then query order DESC and see the same entries from the UTXO view first both times for example.

Also importantly, sort ASC|DESC is not necessarily by age. It's by db prefix key for badger and by AssociationID (BlockHash) for postgres. The UTXO view follows the sorting pattern of the database.

lib/db_utils.go
```diff
@@ -8625,7 +8673,7 @@ func DBGetUserAssociationIdsByAttributes(handle *badger.DB, snap *Snapshot, asso
 		keyPrefix = append(keyPrefix, []byte{0}...) // Null terminator byte for AssociationType which can vary in length
 	} else if associationQuery.AssociationValue != "" || associationQuery.AssociationValuePrefix != "" {
 		// AssociationType == "", (AssociationValue != "" || AssociationValuePrefix != "")
-		return nil, errors.New("DBGetUserAssociationIdsByAttributes: invalid query params")
+		return nil, nil, errors.New("DBGetUserAssociationIdsByAttributes: invalid query params")
```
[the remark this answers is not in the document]
mattfoley8: I will revisit this.

lib/db_adapter.go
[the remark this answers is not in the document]
mattfoley8: Actually the Set generic only works if stored as a pointer. Otherwise there's weird intermittent behavior where the elements of the set can be re-ordered. Maybe something with copying it. Gonna just pass a pointer for now.

lib/block_view_association.go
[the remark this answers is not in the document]
lazynina: well why can't we just append all the entries from the view and the db into one slice and then apply the sort criteria and then take the slice of elements we want? I was thinking this would actually get us the most consistent set of results, adhering to the sorting order defined by the db? I think you could replace lines 827 to 873 with the snippet below.
```
	allAssociationEntries := utxoAssociationEntries
	for _, dbEntry := range dbAssociationEntries {
		if _, exists := utxoAssociationEntryMap[dbEntry.AssociationID]; !exists {
			allAssociationEntries = append(allAssociationEntries, dbEntry)
		}
	}
	allAssociationEntriesSorted, err := bav.GetDbAdapter().SortUserAssociationEntriesByPrefix(
		allAssociationEntries, prefixType, associationQuery.SortDescending)
	maxIndex := associationQuery.Limit
	if maxIndex > len(allAssociationEntriesSorted) {
		maxIndex = len(allAssociationEntriesSorted)
	}
	return allAssociationEntries[:maxIndex], nil
```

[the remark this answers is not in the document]
lazynina: you could do some sort of binary search insertion, but I think that's unnecessary rn instead of using the helper function you defined in the db adapter.

lib/block_view_association.go
```diff
@@ -877,85 +963,137 @@ func (bav *UtxoView) GetPostAssociationsByAttributes(associationQuery *PostAssoc
...
-	}
-	// Loop through UTXO view association entries. Add non-deleted association entries.
-	// Note that the map prevents duplicates. And remove deleted association entries.
-	newUtxoAssociationEntries, deletedUtxoAssociationEntryIDs := bav._getUtxoPostAssociationEntriesByAttributes(associationQuery)
+	// Loop through UTXO view association entries adding non-deleted
+	// association entries. Note that the Set prevents duplicates.
 	for _, utxoAssociationEntry := range newUtxoAssociationEntries {
+		if associationQuery.Limit > 0 && associationIDs.Size() >= associationQuery.Limit {
+			break
+		}
```
lazynina, line 1073: why are we validating the number of associationIDs we have in a count query?

lib/block_view_association.go
```diff
@@ -753,83 +787,135 @@ func (bav *UtxoView) GetUserAssociationsByAttributes(associationQuery *UserAssoc
...
-	}
-	// Loop through UTXO view association entries. Add non-deleted association entries.
-	// Note that the map prevents duplicates. And remove deleted association entries.
-	newUtxoAssociationEntries, deletedUtxoAssociationEntryIDs := bav._getUtxoUserAssociationEntriesByAttributes(associationQuery)
+	// Loop through UTXO view association entries adding non-deleted
+	// association entries. Note that the Set prevents duplicates.
 	for _, utxoAssociationEntry := range newUtxoAssociationEntries {
+		if associationQuery.Limit > 0 && associationIDs.Size() >= associationQuery.Limit {
+			break
+		}
```
lazynina, line 895: Wouldn't we want to just continue to count?

lib/db_utils.go
```diff
@@ -8644,55 +8692,97 @@ func DBGetUserAssociationIdsByAttributes(handle *badger.DB, snap *Snapshot, asso
...
+		lastSeenKey, err = _dbUserAssociationIdToKey(handle, snap, associationQuery.LastSeenAssociationID, prefixType)
+		if err != nil {
+			return nil, nil, errors.Wrapf(err, "DBGetUserAssociationIdsByAttributes: ")
+		}
+	}
+
+	// Map deleted UTXO AssociationIDs to keys.
+	deletedUtxoKeys := NewSet[string]([]string{})
+	for _, deletedUtxoAssociationId := range deletedUtxoAssociationIDs.ToSlice() {
+		deletedUtxoKey, err := _dbUserAssociationIdToKey(handle, snap, deletedUtxoAssociationId, prefixType)
```
lazynina, line 8712: I don't really like included utxo in the prefixes around here. deleted stuff is inherently only available in the view. `deleteUtxoKey` actually makes me think of a deleted `UtxoKey`, which is not what you're trying to convey here. I would go with and try to remove all usages of utxo in variable names in this PR
```
deletedUtxoKeys -> deletedAssociationKey
deletedUTxoAssociationId -> deletedAssociationId
```

lib/db_utils.go
```diff
@@ -8974,3 +9131,89 @@ func DBDeletePostAssociationWithTxn(txn *badger.Txn, snap *Snapshot, association
...
+
+	for nodeIterator.Seek(startingKey); nodeIterator.ValidForPrefix(prefix); nodeIterator.Next() {
+		// Break if at or beyond limit.
+		if limit > 0 && len(keysFound) >= limit {
+			break
+		}
+		// Copy key.
+		key := nodeIterator.Item().Key()
+		keyCopy := make([]byte, len(key))
+		copy(keyCopy[:], key[:])
```
lazynina, line 9199: we can probably push the copying of the key to right before we copy the value. Using the key directly when comparing using bytes.Equal and set.Includes should not change the key

lazynina (commented):
looking pretty clean - just a few questions

lib/db_utils.go
mattfoley8: Makes sense. I was trying to differentiate between database keys/entries and UTXO view keys/entries. Otherwise, it's kinda weird to receive a Set of deleted keys/entries to the database and consider them deleted. But maybe it's clear enough. And they don't need a `deletedUtxoViewKeys` or `deletedMempoolKey` prefix. Wdyt?

lazynina: If it's deleted, it is inherently in the view. And I think this is a bit anti-pattern for core as we don't typically send data to the DB from the view, rather grab data from DB and then merge into view, which I think is why this naming convention is bothering me. but I see that you're doing it to avoid having to go BACK to the DB again to get more results if you have a lot of collisions between the view and the db.

I think we axe the utxo from these variable names in all functions except the enumerate one you created. I still don't love the idea of going from view -> DB though, but I don't know how to get away with it without either fetching more than we need OR going back to fetch more if we don't have enough.

lib/block_view_association.go
```diff
@@ -639,11 +639,33 @@ func _isValidUserAssociationQuery(associationQuery *UserAssociationQuery) error
...
-	if (len(associationQuery.AssociationType) > 0 && len(associationQuery.AssociationTypePrefix) > 0) ||
-		(len(associationQuery.AssociationValue) > 0 && len(associationQuery.AssociationValuePrefix) > 0) {
-		return errors.New("invalid query params")
+	if len(associationQuery.AssociationType) > 0 && len(associationQuery.AssociationTypePrefix) > 0 {
+		return errors.New("invalid query params: both AssociationType and AssociationTypePrefix provided")
+	}
+	if len(associationQuery.AssociationValue) > 0 && len(associationQuery.AssociationValuePrefix) > 0 {
+		return errors.New("invalid query params: both AssociationValue and AssociationValuePrefix provided")
+	}
+	if associationQuery.Limit < 0 {
```
lazynina, line 650: do we want to enforce an upper bound on limit here?

lib/block_view_association.go
```diff
@@ -753,83 +797,132 @@ func (bav *UtxoView) GetUserAssociationsByAttributes(associationQuery *UserAssoc
...
-	for utxoAssociationEntryID := range deletedUtxoAssociationEntryIDs {
-		// Remove deleted association entries found in the UTXO view.
-		delete(associationEntryMap, utxoAssociationEntryID)
-	}
-	// Convert map to slice.
-	var associationEntries []*UserAssociationEntry
-	for _, matchingAssociationEntry := range associationEntryMap {
-		associationEntries = append(associationEntries, matchingAssociationEntry)
+	// Sort the UTXO view association entries according to the query's specified sort order.
+	sortedUtxoViewAssociationEntries, err := bav.GetDbAdapter().SortUserAssociationEntriesByPrefix(
```
lazynina, line 817: I think we want to put all the entries from both the view and the db into one slice and order it so we get the same results after we flush to the db.

lib/block_view_association.go
```diff
@@ -753,83 +797,132 @@ func (bav *UtxoView) GetUserAssociationsByAttributes(associationQuery *UserAssoc
...
 	if err != nil {
 		return 0, errors.Wrap(err, "CountUserAssociationsByAttributes: ")
 	}
+	// First pull matching association entries from the UTXO view so that
+	// we can track deleted association entries and can properly limit
+	// the number of entries retrieved from the database.
+	newAssociationEntries, deletedAssociationIdMap := bav._getUtxoViewUserAssociationEntriesByAttributes(associationQuery)
 	// Pull matching association IDs from the db.
-	associationIDs, err := bav.GetDbAdapter().GetUserAssociationIdsByAttributes(associationQuery)
+	associationIDs, _, err := bav.GetDbAdapter().GetUserAssociationIdsByAttributes(associationQuery, deletedAssociationIdMap)
```
lazynina, line 896: okay so kinda crazyyyyy idea. but really what the second param to `GetUserAssociationIdsByAttributes` is really just a set of `AssociationIds` to skip. What if we just add all the `associationIds` from `newAssociationEntries` to the `deletedAssociationIdMap` before we pass it in? then on the other side, we can really just return `len(newAssociationEntries) + len(associationIds)`. Thoughts? and we change the lingo from `deleted` to `excludeKeys` all the way down the stack to `EnumerateKeysforPrefixWithLimitOffsetOrder`.

lib/block_view_association_test.go
```diff
@@ -982,7 +1035,7 @@ func _testAssociations(t *testing.T, flushToDB bool) {
 			require.Equal(t, userAssociationEntries[0].TargetUserPKID, m3PKID)
 		} else {
 			require.Error(t, err)
-			require.Contains(t, err.Error(), "invalid query params")
+			require.Contains(t, err.Error(), "invalid query params: missing AssociationType")
```
lazynina, line 1038: Maybe it's better for us to define some string consts for these? I mean I guess you've already updated the tests, but if we were to change them again, it would be annoying. Maybe if we update the errors again, then we start defining them as consts. and these consts could live in `block_view_association.go` instead of `constants.go` since it's only really necessary to support these tests.

lazynina (changes requested):

lib/block_view_association.go
lazynina: this applies to `GetUserAssociationsByAttributes` too btw

lib/block_view_association.go
lazynina: Spoke offline - this will be addressed in a separate PR.

lib/block_view_association.go
mattfoley8: I was thinking I wouldn't in core. But would in backend. That makes it easier to change. I was thinking by default the Limit would be 100. And the max would possibly be 100, 500, or 1k depending on performance.

lazynina (approved):

PR #435, opened 2022-12-16
Mf/add associations derived key spending limits

mattfoley8:


lib/network.go
```diff
@@ -5470,6 +5525,31 @@ func (tsl *TransactionSpendingLimit) FromBytes(blockHeight uint64, rr *bytes.Rea
 		}
 	}
 
+	if MigrationTriggered(blockHeight, AssociationsMigration) {
```
lazynina, line 5528: I'd move this to the end.

lib/network.go
```diff
@@ -5488,6 +5568,7 @@ func (tsl *TransactionSpendingLimit) Copy() *TransactionSpendingLimit {
 		DAOCoinOperationLimitMap:     make(map[DAOCoinOperationLimitKey]uint64),
 		NFTOperationLimitMap:         make(map[NFTOperationLimitKey]uint64),
 		DAOCoinLimitOrderLimitMap:    make(map[DAOCoinLimitOrderLimitKey]uint64),
+		AssociationLimitMap:          make(map[AssociationLimitKey]uint64),
```
lazynina, line 5571: I think this is ok to not have it guarded by block height?

lazynina (changes requested):

lazynina (approved):

PR #439, opened 2022-12-20
Ln/add map set func

lazynina:


lib/generics.go
```diff
@@ -55,3 +55,19 @@ func (set *Set[T]) ToSlice() []T {
 	}
 	return results
 }
+
+func MapSet[T comparable, K any](set *Set[T], mapFunc func(elem T) (K, error)) ([]K, error) {
+	var results []K
+	err := set.ForEach(func(elem T) error {
+		mappedResult, innerErr := mapFunc(elem)
+		if innerErr != nil {
+			return innerErr
```
mattfoley8, line 64: I assume this works, but I'm not sure I follow how. We will end up with a mappedResult that is less than or equal to the length of the set. Less than if there are any errors. Equal if there aren't any errors. And we will end up with a map of returned errors or nils. How does the outer error know to error if e.g. `Any?([nil, nil, error, nil, nil])`.

lazynina: If any element returns an error from the map function, the whole thing errors. I can add a test for this

mattfoley8: Makes sense. +1 to a test. Would convince me definitively.

lib/db_utils.go
```diff
@@ -8772,15 +8766,9 @@ func DBGetPostAssociationsByAttributes(
...
-		associationEntry, innerErr := DBGetPostAssociationByID(handle, snap, &associationID)
-		if innerErr != nil {
-			return innerErr
-		}
-		associationEntries = append(associationEntries, associationEntry)
-		return nil
+		return DBGetPostAssociationByID(handle, snap, &associationID)
 	})
 	if err != nil {
 		return nil, nil, errors.Wrapf(err, "DBGetUserAssociationsByAttributes: problem retrieving association entry by ID: ")
```
mattfoley8, line 8774: Unrelated to your changes but this should say DBGetPostAssociationsByAttributes…

mattfoley8 (approved):

PR #451, opened 2023-01-11
Allow for freezing posts.

mattfoley8:


lib/block_view_post.go
```diff
@@ -801,9 +809,16 @@ func (bav *UtxoView) _connectSubmitPost(
 				PkToStringBoth(txn.PublicKey), spew.Sdump(GetParamUpdaterPublicKeys(blockHeight, bav.Params)))
 		}
 
-		// Modification of an NFT is not allowed (before the specified block height).
-		if existingPostEntryy.IsNFT && blockHeight < bav.Params.ForkHeights.AllowUpdatingNFTPostsBlockHeight {
-			return 0, 0, nil, errors.Wrapf(RuleErrorSubmitPostCannotUpdateNFT, "_connectSubmitPost: ")
+		if blockHeight >= bav.Params.ForkHeights.AllowUpdatingNFTPostsBlockHeight {
+			// Modification of a frozen post is not allowed after the above block height.
+			if existingPostEntryy.IsFrozen {
```
lazynina, line 814: we need to set `IsFrozen` on `newPostEntry` within the `if len(txMeta.PostHashToModify) != 0 {` statement. Currently I only see us setting `IsFrozen` on new posts

lib/block_view_post.go
```diff
@@ -755,6 +756,13 @@ func (bav *UtxoView) _connectSubmitPost(
 		copy(repostedPostHash[:], repostedPostHashBytes)
 		delete(extraData, RepostedPostHash)
 	}
+	isFrozen := false
+	if blockHeight >= bav.Params.ForkHeights.AllowUpdatingNFTPostsBlockHeight {
+		if extraDataIsFrozen, exists := extraData[IsFrozen]; exists {
+			isFrozen = bytes.Equal(extraDataIsFrozen, []byte{1})
```
lazynina, line 762: we can mimic what we did with `isQuotedRepost` above. we should define `var FrozenPostVal = []byte{1}` in constants.go so we can consistently reference it everywhere. I do prefer using bytes.Equal tho as you have it here as opposed to reflect.DeepEqual

lib/block_view_post_test.go
```diff
@@ -2072,3 +2072,100 @@ func TestDeSoDiamondErrorCases(t *testing.T) {
...
+		testMeta.txnOps = append(testMeta.txnOps, utxoOps)
+		testMeta.txns = append(testMeta.txns, txn)
+		require.NoError(t, utxoView.FlushToDb(0))
+		return txn.Hash().ToBytes(), nil
+	}
+
+	var postHash []byte
+	{
+		// Happy path: creating a frozen post should succeed.
+		postHash, err = submitPost("frozen post!", nil)
```
lazynina, line 2153: we should have a test where we do the following:
1. create a post that is not frozen
2. update that post to be frozen
3. prove that we can't update a frozen post again

mainly step 2 above is required.

lib/constants.go
```diff
@@ -334,6 +335,9 @@ type EncoderMigrationHeights struct {
 
 	// DeSoAssociations coincides with the AssociationsBlockHeight block
 	DeSoAssociations MigrationHeight
+
+	// DeSoFrozenPosts coincides with the AllowUpdatingNFTPostsBlockHeight block
+	DeSoFrozenPosts MigrationHeight
```
lazynina, line 340: ultimately we'll combine all these fork heights into one. Having separate attributes of the ForkHeight struct will make encoder migrations take longer as it appears we have MORE migrations to run. this can be handled later though in a separate PR. we'll come up with some name for the fork

lib/constants.go
```diff
@@ -353,6 +357,11 @@ func GetEncoderMigrationHeights(forkHeights *ForkHeights) *EncoderMigrationHeigh
 			Height:  uint64(forkHeights.AssociationsBlockHeight),
 			Name:    AssociationsMigration,
 		},
+		DeSoFrozenPosts: MigrationHeight{
```
lazynina, line 360: yeah we'll want to combine these migrations. more migrations = slower

lazynina (changes requested):

lib/constants.go
```diff
@@ -1169,6 +1165,8 @@ const (
 	RepostedPostHash = "RecloutedPostHash"
 	// Key in transaction's extra map -- The presence of this key indicates that this post is a repost with a quote.
 	IsQuotedRepostKey = "IsQuotedReclout"
+	// Key in transaction's extra data map that freezes a post rendering it immutable.
+	IsFrozen = "IsFrozen"
```
lazynina, line 1169: We typically add `Key` as a suffix to these constants.

lazynina (approved):
just change `const IsFrozen = "IsFrozen"` to `const IsFrozenKey = "IsFrozen"`

PR #461, opened 2023-01-19
Update access group metamask string.

mattfoley8:


lib/network.go
```diff
@@ -5300,22 +5300,17 @@ func (tsl *TransactionSpendingLimit) ToMetamaskString(params *DeSoParams) string
...
 			indentationCounter++
 			opString += _indt(indentationCounter) + "Access Group Owner Public Key: " +
 				Base58CheckEncode(accessGroupKey.AccessGroupOwnerPublicKey.ToBytes(), false, params) + "\n"
-			opString += _indt(indentationCounter) + "Access Group Scope Type: " +
-				accessGroupKey.AccessGroupScopeType.ToString() + "\n"
-			// FIXME: Verify that this string prints nicely in the event that we are asking
-			// for a derived key with the permission to create a group with {any group key name}
-			// and {any operation type}.
-			groupKeyName := string(accessGroupKey.AccessGroupKeyName.ToBytes())
+			groupKeyName := string(bytes.Trim(accessGroupKey.AccessGroupKeyName.ToBytes(), "\x00"))
```
lazynina, line 5311: we should use `AccessKeyNameDecode` instead of this other way of trimming bytes

lib/network.go
```diff
@@ -5328,25 +5323,25 @@ func (tsl *TransactionSpendingLimit) ToMetamaskString(params *DeSoParams) string
 			opString += _indt(indentationCounter) + "]\n"
 			accessGroupStr = append(accessGroupStr, opString)
 		}
+		// Ensure deterministic ordering of the transaction count limit strings by doing a lexicographical sort.
+		sortStringsAndAddToLimitStr(accessGroupStr)
```
lazynina, line 5327: good catch

lib/network.go
```diff
@@ -5328,25 +5323,25 @@ func (tsl *TransactionSpendingLimit) ToMetamaskString(params *DeSoParams) string
...
 			indentationCounter++
 			opString += _indt(indentationCounter) + "Access Group Owner Public Key: " +
 				Base58CheckEncode(accessGroupMemberKey.AccessGroupOwnerPublicKey.ToBytes(), false, params) + "\n"
-			opString += _indt(indentationCounter) + "Access Group Scope Type: " +
-				accessGroupMemberKey.AccessGroupScopeType.ToString() + "\n"
-			// FIXME: Verify that this string prints nicely in the event that we are asking
-			// for a derived key with the permission to create a group with {any group key name}
-			// and {any operation type}.
-			groupKeyName := string(accessGroupMemberKey.AccessGroupKeyName.ToBytes())
+			groupKeyName := string(bytes.Trim(accessGroupMemberKey.AccessGroupKeyName.ToBytes(), "\x00"))
```
lazynina, line 5342: use AccessKeyNameDecode

lazynina (changes requested):

PR #479, opened 2023-02-27
Use correct struct type for create post assoc

jackson-dean:


brootle:
guys, what about User Post association? please, double check and make sure all is OK there! thanks!

mattfoley8 (approved):

PR #485, opened 2023-03-17
remove explicit spend from txn fees

lazynina:


lib/block_view_dao_coin_limit_order.go
```diff
@@ -1906,3 +1941,89 @@ func ScaleFloatFormatStringToUint256(floatStr string, scaleFactor *uint256.Int)
...
+	// Validate $DESO doesn't overflow uint64.
+	if !desoNanosToFulfillOrders.IsUint64() {
+		return 0, fmt.Errorf(
+			"GetDESONanosToFillOrder: fulfilling order $DESO overflows uint64")
+	}
+
+	return desoNanosToFulfillOrders.Uint64(), nil
+}
+
+func (bav *UtxoView) ConstructTransactorOrderFromTxn(txn *MsgDeSoTxn, blockHeight uint32) *DAOCoinLimitOrderEntry {
```
mattfoley8, line 2016: I'm not exactly sure what it is, but this func name could be more clear. Right now, it's not entirely clear it pertains only to DAO coin limit orders. Like `ConvertTxnToDAOCoinLimitOrderEntry()`.

lib/block_view.go
```diff
@@ -1773,6 +1780,49 @@ func (bav *UtxoView) _connectBasicTransfer(
...
+					err,
+					"_connectBasicTransfer: Transactor PKID entry doesn't exist; this should never happen")
+			}
+			var explicitSpend uint64
+			for _, filledOrder := range bav.TxHashToFilledOrder[*txHash] {
+				if filledOrder == nil {
+					continue
+				}
+				if !filledOrder.TransactorPKID.Eq(transactorPKIDEntry.PKID) {
+					explicitSpend += filledOrder.CoinQuantityInBaseUnitsBought.Uint64()
```
mattfoley8, line 1821: missing uint246.Int --> uint64 overflow check

lib/block_view_dao_coin_limit_order.go
```diff
@@ -809,10 +826,23 @@ func (bav *UtxoView) _connectDAOCoinLimitOrder(
 						if utxoOp, err = bav._addBalance(newDESOSurplus.Uint64(), pubKey); err != nil {
 							return 0, 0, nil, err
 						}
+						// Add the DESO to the total output.
+						totalOutput += newDESOSurplus.Uint64()
```
mattfoley8, line 830: missing uint256.Int --> uint64 conversion overflow check

lib/block_view_dao_coin_limit_order.go
```diff
@@ -809,10 +826,23 @@ func (bav *UtxoView) _connectDAOCoinLimitOrder(
...
 					} else {
-						if utxoOp, err = bav._spendBalance(newDESOSurplus.Neg(newDESOSurplus).Uint64(), pubKey, blockHeight); err != nil {
+						spendAmount := newDESOSurplus.Neg(newDESOSurplus).Uint64()
```
mattfoley8, line 832: missing uint256.Int --> uint64 conversion overflow check

lib/block_view.go
```diff
@@ -109,6 +109,11 @@ type UtxoView struct {
 	// Map of PKID to the next nonce to use for a transaction.
 	PKIDToNextNonce map[PKID]uint64
 
+	// Map of TxHash to filled order. Needed to handle
+	// derived key accounting for DAOCoinLimitOrders
+	// where the transactor is selling DESO.
+	TxHashToFilledOrder map[BlockHash][]*FilledDAOCoinLimitOrder
```
mattfoley8, line 115: nit: naming TxHashToFilledOrder --> TxHashToFilledDAOCoinLimitOrder. Would it be possible that we would have other types of orders in the future? CC orders for example. More expicit ~ more future-proof. :man_shrugging:

lib/block_view.go
```diff
@@ -1773,6 +1780,54 @@ func (bav *UtxoView) _connectBasicTransfer(
 		}
 
 		utxoOpsForTxn = append(utxoOpsForTxn, newUtxoOp)
+		// Certain transaction types spend more DESO than input specified
```
mattfoley8, line 1783: Instead of computing this here, could we just pass in an additionalSpend uint64 to connectBasicTransfer? And then compute these in the corresponding connectCreatorCoin, connectNFT, connectDAOCoinLimitOrder? Downside is would be updating the connectBasicTransfer func signature which is used in a bunch of places. Upside is it better encapsulates each txn types logic.

mattfoley8: Also could avoid storing the filledOrders on the utxoView in that case.

PR #486, opened 2023-03-20
Ln/new nonce scheme

lazynina:


lib/block_view_types.go
```diff
@@ -3614,6 +3614,12 @@ type GlobalParamsEntry struct {
 
 	// The new minimum fee the network will accept
 	MinimumNetworkFeeNanosPerKB uint64
+
+	// MaxNonceExpirationBlockBuffer is maximum value nodes will
+	// tolerate for the difference between the current block height
+	// and the expiration block height specified in the nonce for a
+	// transaction.
+	MaxNonceExpirationBlockBuffer uint64
```
tholonious, line 3622: Let's make this param name a bit more self documenting. Feel free to ignore if you disagree.
```suggestion
	MaxNonceExpirationBlockHeightOffset uint64
```

lib/db_utils.go
```diff
@@ -10462,83 +10465,125 @@ func DBDeletePostAssociationWithTxn(txn *badger.Txn, snap *Snapshot, association
 // DeSo nonce mapping functions
 // -------------------------------------------------------------------------------------
 
-func _dbKeyForPKIDToNextNonce(pkid PKID) []byte {
+func _dbKeyForNonceEntry(accountNonce *DeSoNonce, pkid *PKID) []byte {
```
tholonious, line 10468:
```suggestion
func _dbKeyForTransactionNonceEntry(nonce *DeSoNonce, pkid *PKID) []byte {
```

lib/block_view_types.go
```diff
@@ -6013,3 +6026,21 @@ func (newMessageTxindexMetadata *NewMessageTxindexMetadata) GetVersionByte(block
 func (newMessageTxindexMetadata *NewMessageTxindexMetadata) GetEncoderType() EncoderType {
 	return EncoderTypeNewMessageTxindexMetadata
 }
+
+type NonceEntry struct {
+	Nonce     *DeSoNonce
+	PKID      *PKID
```
tholonious, line 6032:
```suggestion
	TransactorPKID      *PKID
```

lib/block_view_types.go
```diff
@@ -6013,3 +6026,21 @@ func (newMessageTxindexMetadata *NewMessageTxindexMetadata) GetVersionByte(block
...
+	isDeleted bool
+}
+
+func (ne *NonceEntry) ToMapKey() NonceMapKey {
```
tholonious, line 6036:
```suggestion
func (ne *TransactionNonceEntry) ToMapKey() TransactionNonceMapKey {
```

tholonious (changes requested):

lib/block_view.go
```diff
@@ -3515,40 +3540,87 @@ func (bav *UtxoView) Preload(desoBlock *MsgDeSoBlock, blockHeight uint64) error
 	return nil
 }
 
-func (bav *UtxoView) GetNextNonceForPublicKey(pkBytes []byte) (uint64, error) {
-	pkidEntry := bav.GetPKIDForPublicKey(pkBytes)
-	if pkidEntry == nil || pkidEntry.isDeleted {
-		return 0, fmt.Errorf("GetNextNonceForPublicKey: PKID entry is deleted for public key %s", PkToStringBoth(pkBytes))
+func (bav *UtxoView) GetNonceEntry(accountNonce *DeSoNonce, pkid *PKID) (*NonceEntry, error) {
```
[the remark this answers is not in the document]
lazynina: good catch - yeah I originally had it called AccountNonce as specified in original doc, so just need to purge all `accountNonce` usages

lib/block_view.go
```diff
@@ -1186,26 +1189,21 @@ func (bav *UtxoView) DisconnectTransaction(currentTxn *MsgDeSoTxn, txnHash *Bloc
...
+		nonce, err := bav.GetTransactorNonceEntry(currentTxn.TxnNonce, pkidEntry.PKID)
+		if err != nil {
+			return errors.Wrapf(err, "DisconnectTransaction: Problem getting account nonce for nonce %s and PKID %v", currentTxn.TxnNonce.String(), pkidEntry.PKID)
 		}
-		// Ensure that the currentTxn's and nextNonce line up correctly.
-		if currentTxn.TxnNonce != nextNonce-1 {
-			return fmt.Errorf(
-				"DisconnectTransaction: Txn nonce %d does not match expected nonce %d for pub key %s",
-				currentTxn.TxnNonce, nextNonce-1, PkToStringBoth(currentTxn.PublicKey))
+		// TODO: Is it possible that the nonce is already deleted from badger? I don't think so, but I'm not sure.
```
mattfoley8, line 1202: does this comment still apply since you resolved tholonious' comment below?

lib/block_view.go
```diff
@@ -3515,40 +3540,88 @@ func (bav *UtxoView) Preload(desoBlock *MsgDeSoBlock, blockHeight uint64) error
...
+			"ConstructNonceForPublicKey: No PKID entry found for public key %s",
+			PkToStringBoth(publicKey))
+	}
+	return bav.ConstructNonceForPKID(pkidEntry.PKID, blockHeight, 2)
+}
+
+func (bav *UtxoView) ConstructNonceForPKID(pkid *PKID, blockHeight uint64, depth uint8) (*DeSoNonce, error) {
+	// construct nonce
+	// TODO: what's our initial value for this in the event that the global params entry isn't set?
+	expirationBuffer := uint64(10000)
```
mattfoley8, line 3604: should move to lib/constants.go

lib/block_view_test.go
```diff
@@ -1260,6 +1316,57 @@ func TestUpdateGlobalParams(t *testing.T) {
...
+
+		require.NoError(utxoView.FlushToDb(0))
+
+		require.Equal(DbGetGlobalParamsEntry(utxoView.Handle, chain.snapshot), prevGlobalParams)
+		require.Equal(utxoView.GlobalParamsEntry, prevGlobalParams)
+
+		// Check the balance of the updater after this txn
+		require.NotEqual(0, _getBalance(t, chain, nil, moneyPkString))
+
+		// TODO: make sure mempool won't accept the transaction w/ a high expiration.
```
mattfoley8, line 1368: seems like a good test to add

lib/db_utils.go
```diff
@@ -10462,83 +10465,123 @@ func DBDeletePostAssociationWithTxn(txn *badger.Txn, snap *Snapshot, association
...
+		nonceEntries = append(nonceEntries, TransactorNonceKeyToTransactorNonceEntry(key))
+	}
+	return nonceEntries
+}
+
+func TransactorNonceKeyToTransactorNonceEntry(key []byte) *TransactorNonceEntry {
+	keyWithoutPrefix := key[1:]
+	expirationHeight := DecodeUint64(keyWithoutPrefix[:8])
+	pkid := &PKID{}
+	copy(pkid[:], keyWithoutPrefix[8:41])
```
mattfoley8, line 10574: This seems a little finicky to do the manual byte slicing we're doing here. Probably fine.

lib/block_view.go
```diff
@@ -1186,26 +1189,21 @@ func (bav *UtxoView) DisconnectTransaction(currentTxn *MsgDeSoTxn, txnHash *Bloc
...
+		if err != nil {
+			return errors.Wrapf(err, "DisconnectTransaction: Problem getting account nonce for nonce %s and PKID %v", currentTxn.TxnNonce.String(), pkidEntry.PKID)
 		}
-		// Ensure that the currentTxn's and nextNonce line up correctly.
-		if currentTxn.TxnNonce != nextNonce-1 {
-			return fmt.Errorf(
-				"DisconnectTransaction: Txn nonce %d does not match expected nonce %d for pub key %s",
-				currentTxn.TxnNonce, nextNonce-1, PkToStringBoth(currentTxn.PublicKey))
+		// TODO: Is it possible that the nonce is already deleted from badger? I don't think so, but I'm not sure.
+		if nonce == nil || nonce.isDeleted {
```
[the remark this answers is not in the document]
lazynina: Nonce clear out has moved to process block as discussed. However, I still think we MAY encounter a situation in which the nonce is deleted from the db and we're disconnecting a series of blocks.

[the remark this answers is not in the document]
lazynina: this now lives in connect / disconnect block which leads us to reverting state properly on block disconnects

lib/errors.go
```diff
@@ -485,13 +489,15 @@ const (
...
-	TxErrorUnconnectedTxnNotAllowed     RuleError = "TxErrorUnconnectedTxnNotAllowed"
+	TxErrorTooLarge                           RuleError = "TxErrorTooLarge"
+	TxErrorDuplicate                          RuleError = "TxErrorDuplicate"
+	TxErrorIndividualBlockReward              RuleError = "TxErrorIndividualBlockReward"
+	TxErrorInsufficientFeeMinFee              RuleError = "TxErrorInsufficientFeeMinFee"
+	TxErrorInsufficientFeeRateLimit           RuleError = "TxErrorInsufficientFeeRateLimit"
+	TxErrorInsufficientFeePriorityQueue       RuleError = "TxErrorInsufficientFeePriorityQueue"
+	TxErrorUnconnectedTxnNotAllowed           RuleError = "TxErrorUnconnectedTxnNotAllowed"
+	TxErrorNonceExpired                       RuleError = "TxErrorNonceExpired"
+	TxErrorNonceExpirationBlockBufferExceeded RuleError = "TxErrorNonceExpirationBlockBufferExceeded"
```
tholonious, line 500:
```suggestion
	TxErrorNonceExpirationBlockHeightOffsetExceeded RuleError = "TxErrorNonceExpirationBlockHeightOffsetExceeded"
```

lib/constants.go
```diff
@@ -1317,6 +1318,13 @@ const (
 	MinAccessGroupKeyNameCharacters = 1
 	MaxAccessGroupKeyNameCharacters = 32
 
+	// TODO: What are reasonable values for these? Or do we even need them at all?
+	// Min/Max MaxNonceExpirationBlockHeightOffset - Min/max value to which the MaxNonceExpirationBlockHeightOffset can be set.
+	MinMaxNonceExpirationBlockHeightOffset = 1
+	MaxMaxNonceExpirationBlockHeightOffset = 10000
+	// TODO: what's our initial value for this in the event that the global params entry isn't set?
```
tholonious, line 1325: Could just implicitly use MaxMaxNonceExpirationBlockHeightOffset as the default value when the entry doesn't exist. The max value is generally the safest acceptable value.

lib/block_view.go
```diff
@@ -3515,40 +3570,113 @@ func (bav *UtxoView) Preload(desoBlock *MsgDeSoBlock, blockHeight uint64) error
...
+
+	// Make sure we don't have a collision.
+	nonceEntry, err := bav.GetTransactorNonceEntry(&nonce, pkid)
+	if err != nil {
+		return nil, errors.Wrapf(err, "ConstructNonceForPKID: ")
+	}
+	if nonceEntry == nil {
+		return &nonce, nil
+	}
+	if depth == 0 {
```
tholonious, line 3676: Just for my own understanding, how do you plan on using the depth param?

tholonious (changes requested):

lib/block_view.go
lazynina: This really only serves to support transaction construction. In the event of a collision, we'll attempt to generate the nonce again.

It's always 2 on the initial call to this function. I'll add a check that depth <= 2. If we hit a collision three times when generating the nonce, we throw an error.

lib/block_view.go
```diff
@@ -1434,6 +1431,26 @@ func (bav *UtxoView) DisconnectBlock(
 				"utxoOps (%d)", numOutputs, numAddUtxoOps)
 	}
 
+	// After the balance model block height, we may have a delete expired nonces utxo operation.
+	// We need to revert this before iterating over the transactions in the block.
+	if desoBlock.Header.Height >= uint64(bav.Params.ForkHeights.BalanceModelBlockHeight) {
+		// We need to revert the delete expired nonces operation.
+		deleteExpiredNoncesUtxoOps := utxoOps[len(utxoOps)-1]
```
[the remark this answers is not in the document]
tholonious: After chatting offline, this part LGTM now. It's safe and correct to assume the OperationTypeDeleteExpiredNonces ops are at the very end of the list.

lib/block_view.go
tholonious: Ah, I missed the recursive call on line 3679. This makes sense now.

lib/block_view.go
```diff
@@ -3515,40 +3576,121 @@ func (bav *UtxoView) Preload(desoBlock *MsgDeSoBlock, blockHeight uint64) error
...
 	}
-	return bav.GetNextNonceForPKID(*pkidEntry.PKID)
-}
 
-func (bav *UtxoView) GetNextNonceForPKID(pkid PKID) (uint64, error) {
-	nextNonce, nonceFound := bav.PKIDToNextNonce[pkid]
-	if nonceFound {
-		return nextNonce, nil
+	nonceEntry, exists := bav.TransactorNonceMapKeyToTransactorNonceEntry[mapKey]
+	if exists {
```
mattfoley8, line 3592: nit: I usually do a `if exists && nonceEntry != nil` check here. I dunno how nonceEntry would ever be nil, but double-checking just in case fwiw.

mattfoley8 (approved):

lib/block_view.go
```diff
@@ -3515,40 +3576,121 @@ func (bav *UtxoView) Preload(desoBlock *MsgDeSoBlock, blockHeight uint64) error
...
+	bav.TransactorNonceMapKeyToTransactorNonceEntry[nonceEntry.ToMapKey()] = nonceEntry
+}
+
+func (bav *UtxoView) DeleteTransactorNonceEntry(nonceEntry *TransactorNonceEntry) {
+	if nonceEntry == nil {
+		glog.Errorf("DeleteTransactorNonceEntry: nil nonceEntry")
+		return
+	}
+
+	nonceEntry.isDeleted = true
```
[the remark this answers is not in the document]
lazynina: good call.

PR #488, opened 2023-03-25
WIP: State indexer

superzordon:


lib/state_change_syncer.go
```diff
@@ -0,0 +1,431 @@
...
+// StateChangeSyncer is used to keep track of the state changes that should be written to the state change file.
+type StateChangeSyncer struct {
+	// The file that the state changes are written to.
+	StateChangeFile *os.File
+	// The file that allows quick lookup of a StateChangeEntry given its index in the file.
+	// This is represented by a list of uint32s, where each uint32 is the offset of the state change entry in the state change file.
+	StateChangeIndexFile *os.File
+	StateChangeFileSize  uint32
+	DeSoParams           *DeSoParams
+	// This map is used to keep track of the bytes should be written to the state change file upon a db flush.
```
lazynina, line 161:
```suggestion
	// This map is used to keep track of the bytes that should be written to the state change file upon a db flush.
```

lib/state_change_syncer.go
```diff
@@ -0,0 +1,431 @@
...
+	// The ID of the flush is to track which entries should be written to the state change file upon flush completion.
+	// This is needed because many flushes can occur asynchronously during hypersync, and we need to make sure that
+	// we write the correct entries to the state change file.
+	// During blocksync, all flushes are synchronous, so we don't need to worry about this. As such, those flushes
+	// are given the uuid.Nil ID.
+	UnflushedBytes map[uuid.UUID]UnflushedStateSyncerBytes
```
lazynina, line 167: do we need a mutex around this map?

lib/state_change_syncer.go
```diff
@@ -0,0 +1,431 @@
...
+	ConnectedMempoolTxMap map[BlockHash]*StateChangeEntry
+}
+
+// Open a file, create if it doesn't exist.
+func openOrCreateLogFile(fileName string) (*os.File, error) {
+	file, err := os.OpenFile(fileName, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
+	if err != nil {
+		return nil, err
+	}
+	return file, nil
```
lazynina, line 182:
```suggestion
	return os.OpenFile(fileName, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
```

lib/state_change_syncer.go
```diff
@@ -0,0 +1,431 @@
...
+
+	// If the transaction is connected, add it to the connected mempool map. Upon disconnect, we'll look this up and pass
+	// it to the state change file so the consumer can use them to revert the mempool transaction.
+	if event.IsConnected {
+		// Check to see if the operation in question has a "core_state" annotation for its key.
+		if !isCoreStateKey(stateChangeEntry.KeyBytes) {
+			return
+		}
+		// Create a copy of the state change entry for the map so that the utxoOps don't get deleted when they're removed
+		//from the stateChangeEntry struct (utxoOps are only included for disconnects).
```
lazynina, line 249:
```suggestion
		// from the stateChangeEntry struct (utxoOps are only included for disconnects).
```

lib/state_change_syncer.go
```diff
@@ -0,0 +1,431 @@
...
+		mapStateChangeEntry := *stateChangeEntry
```
lazynina, line 250: i *think* this is probably a fine way to create a copy as long as we don't modify anything that references a pointer underneath.  A little concerned about UtxoOps below.

lib/state_change_syncer.go
```diff
@@ -0,0 +1,431 @@
...
+			if !isCoreStateKey(connectedMempoolTx.KeyBytes) {
+				return
+			}
+
+			stateChangeEntry.Encoder = connectedMempoolTx.Encoder
+			stateChangeEntry.KeyBytes = connectedMempoolTx.KeyBytes
+			stateChangeEntry.UtxoOps = connectedMempoolTx.UtxoOps
+		} else {
+			return
+		}
```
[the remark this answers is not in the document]
lazynina: early return and avoiding else statements where possible.

lib/state_change_syncer.go
```diff
@@ -0,0 +1,431 @@
...
+	if !isCoreStateKey(stateChangeEntry.KeyBytes) {
+		return
+	}
+
+	// Get the relevant deso encoder for this keyBytes.
+	var encoderType EncoderType
+	if isEncoder, encoder := StateKeyToDeSoEncoder(stateChangeEntry.KeyBytes); isEncoder && encoder != nil {
+		encoderType = encoder.GetEncoderType()
+	} else {
+		// If the keyBytes is not an encoder, then we decode the entry from the key value.
```
lazynina, line 316: maybe add a comment that this is expected for things such as LikeEntry where the entry is decoded from the key. the comment is a little off at the end - the phrase "key value" is a little confusing.

lib/state_change_syncer.go
```diff
@@ -0,0 +1,431 @@
...
+		}
+		encoderType = keyEncoder.GetEncoderType()
+		stateChangeEntry.Encoder = keyEncoder
+	}
+	// Set the encoder type.
+	stateChangeEntry.EncoderType = encoderType
+
+	// Encode the state change entry. We encode as a byte array, so the consumer can buffer just the bytes needed
+	// to decode this entry when reading from file.
+	entryBytes := EncodeToBytes(0, stateChangeEntry, false)
```
lazynina, line 329: hmmm we probably need a real block height in this call to `EncodeToBytes`, right? This gets passed down into `RawEncodeWithoutMetadata`

lazynina (changes requested):

lib/block_view_association.go
```diff
@@ -163,11 +160,23 @@ func (bav *UtxoView) _connectDeleteUserAssociation(
 	}
 	bav._deleteUserAssociationEntryMappings(prevAssociationEntry)
 
+	// Track state change metadata.
+	stateChangeMetadata := &DeleteUserAssociationStateChangeMetadata{
+		TargetUserPublicKeyBase58Check: PkToString(
```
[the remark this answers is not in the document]
superzordon: Correct. It also currently lives in the utxo ops for now

lib/block_view_creator_coin.go
```diff
@@ -1339,6 +1344,11 @@ func (bav *UtxoView) HelpConnectCreatorCoinSell(
 	}
 	desoLockedNanosDiff := int64(existingProfileEntry.CreatorCoinEntry.DeSoLockedNanos) - int64(prevCoinEntry.DeSoLockedNanos)
 
+	// Add a CreatorCoinStateChangeMetadata to the UtxoOperation to track the state change.
+	stateChangeMetadata := &CreatorCoinStateChangeMetadata{
+		ProfileEntry: existingProfileEntry,
```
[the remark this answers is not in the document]
superzordon: Existing profile entry is used to know the change in deso locked nanos for the CreatorCoinTxindexMetadata. The calculation that is done is `desoLockedNanosDiff = int64(stateChangeMetadata.ProfileEntry.CreatorCoinEntry.DeSoLockedNanos - prevCoinEntry.DeSoLockedNanos)`

lib/block_view_types.go
```diff
@@ -3492,6 +3535,84 @@ func (fe *FollowEntry) IsDeleted() bool {
 	return fe.isDeleted
 }
 
+func (fe *FollowEntry) RawEncodeWithoutMetadata(blockHeight uint64, skipMetadata ...bool) []byte {
+	var data []byte
+	data = append(data, EncodeToBytes(blockHeight, fe.FollowerPKID, skipMetadata...)...)
```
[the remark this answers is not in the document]
superzordon: Looking through other encoders (e.g. DeSoBalanceEntry, NFTEntry, AssociationEntries, etc), seems as though the standard way we've handled these PKID encodings is through the "EncodeToBytes" helper. We certainly could do a fixed length byte encoding here and save a couple bytes, but it may be better to stick to the helper fn for consistency's sake.

lib/block_view_types.go
```diff
@@ -1176,6 +1194,11 @@ func (op *UtxoOperation) RawEncodeWithoutMetadata(blockHeight uint64, skipMetada
 		}
 	}
 
+	// StateChangeMetadata
+	if op.StateChangeMetadata != nil {
```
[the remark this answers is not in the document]
superzordon: I've confirmed w/ @AeonSw4n that utxo operations are not checksummed

lib/state_change_syncer.go
diamondhands0: +1

lib/state_change_syncer.go
```diff
@@ -0,0 +1,726 @@
...
+		return nil, err
+	}
+	return file, nil
+}
+
+// NewStateChangeSyncer initializes necessary log files and returns a StateChangeSyncer.
+func NewStateChangeSyncer(stateChangeFilePath string, nodeSyncType NodeSyncType) *StateChangeSyncer {
+	stateChangeIndexFilePath := fmt.Sprintf("%s-index", stateChangeFilePath)
+	stateChangeMemPoolFilePath := fmt.Sprintf("%s-mempool", stateChangeFilePath)
+	stateChangeMempoolIndexFilePath := fmt.Sprintf("%s-mempool-index", stateChangeFilePath)
```
diamondhands0, line 241: So if I pass --state-change-file-path="/home/zzz/statchangedir", would it create /home/zzz/statchangedir-index as a file? Feels like maybe it should be "/home/zzz/statechangedir/index.log" instead. You could do that with filepath.Join(stateChangeFilePath, "index.log"). Join() always does the "right thing" when you're trying to make a new file with an extended name.

lib/state_change_syncer.go
```diff
@@ -0,0 +1,726 @@
...
+	if err != nil {
+		glog.Fatalf("Error opening stateChangeMempoolIndexFile: %v", err)
+	}
+
+	stateChangeMempoolFileInfo, err := stateChangeMempoolFile.Stat()
+	if err != nil {
+		glog.Fatalf("Error getting stateChangeMempoolFileInfo: %v", err)
+	}
+
+	// Check if the state change file is empty. If not, BlocksyncCompleteEntriesFlushed should be true.
```
diamondhands0, line 265: I think there is an edge case here where the sync was interrupted in the middle of flushing entries to the log file. If true, just add a comment explaining that.

diamondhands0: Also if that ever happens it would be nice if you could detect it and kill the node so that it doesn't run with a corrupted state. I'd add that as a TODO underneath.

lib/state_change_syncer.go
```diff
@@ -0,0 +1,726 @@
...
+	}
+	mempoolUtxoView.EventManager.stateSyncerFlushed(&StateSyncerFlushedEvent{
+		FlushId:          uuid.Nil,
+		Succeeded:        true,
+		IsMempoolFlush:   true,
+		BlockSyncFlushId: originalCommittedFlushId,
+	})
+
+	// Loop through all the unconnected transactions in the mempool and connect them and their utxo ops to the mempool view.
+	for txHash, unconnectedTxn := range server.mempool.unconnectedTxns {
```
diamondhands0, line 611: I'm a bit confused about what this is doing. In particular, I don't think we'll ever have txns in unconnectedTxns the way the mempool works right now. This is because, if a txn doesn't connect to the UniversalView, we just reject it. So a txn is either in the UniversalView or it isn't in the mempool at all.

Was there something you ran into that made you need unconnectedTxns?

lib/state_change_syncer.go
diamondhands0: Having read more of the code now, it seems like you're good as long as there's a StateSyncerFlushed thing somewhere.

lib/state_change_syncer.go
```diff
@@ -0,0 +1,726 @@
...
+	}
+
+	if event.IsMempoolTxn {
+		// Create a mempool flush ID if one doesn't already exist.
+		if event.FlushId == uuid.Nil && stateChangeSyncer.MempoolFlushId == uuid.Nil {
+			stateChangeSyncer.MempoolFlushId = uuid.New()
+		}
+
+		// If the event flush ID is nil, then we need to use the global mempool flush ID.
+		if event.FlushId == uuid.Nil {
```
diamondhands0, line 340: I think more bullet-proof here is `if flishId == uuid.Nil`

lib/state_change_syncer.go
```diff
@@ -0,0 +1,726 @@
...
+			flushId = stateChangeSyncer.MempoolFlushId
+		}
+
+		// The current state of the tracked mempool is stored in the MempoolKeyValueMap. If this entry is already in there
+		// then we don't need to re-write it to the state change file.
+		// Create key for op + key map
+		txKey := fmt.Sprintf("%v%v", event.StateChangeEntry.OperationType, string(event.StateChangeEntry.KeyBytes))
```
diamondhands0, line 347: I'd make this into a function and use that function everywhere rather than manually Sprintf'ing.

lib/state_change_syncer.go
```diff
@@ -0,0 +1,726 @@
...
+			delete(stateChangeSyncer.MempoolKeyValueMap, txKey)
+			delete(stateChangeSyncer.MempoolFlushKeySet, txKey)
+		} else {
+			stateChangeSyncer.MempoolKeyValueMap[txKey] = event.StateChangeEntry.EncoderBytes
+			stateChangeSyncer.MempoolFlushKeySet[txKey] = true
+		}
+	} else {
+		// If the flush ID is nil, then we need to use the global block sync flush ID.
+		if event.FlushId == uuid.Nil {
+			flushId = stateChangeSyncer.BlockSyncFlushId
```
diamondhands0, line 367: Same here. I'd check `flushId == uuid.Nil`

lib/state_change_syncer.go
```diff
@@ -0,0 +1,726 @@
...
+
+	// Get the byte index of where this transaction occurs in the unflushed bytes, and add it to the list of
+	// indexes that should be written to the index file.
+	dbOperationIndex := unflushedBytes.StateChangeBytesSize
+	unflushedBytes.StateChangeOperationIndexes = append(unflushedBytes.StateChangeOperationIndexes, dbOperationIndex)
+
+	// Update the state change file size.
+	transactionLen := uint64(len(writeBytes))
+	unflushedBytes.StateChangeBytesSize += transactionLen
+
```
diamondhands0, line 480: I don't understand why we need StateChangeBytesSize. Isn't it always just equal to len(StateChangeBytes)?

lib/state_change_syncer.go
```diff
@@ -0,0 +1,431 @@
...
+	unflushedBytes.StateChangeBytes = append(unflushedBytes.StateChangeBytes, writeBytes...)
+
+	// Get the byte index of where this transaction occurs in the unflushed bytes, and add it to the list of
+	// indexes that should be written to the index file.
+	dbOperationIndex := unflushedBytes.StateChangeBytesSize
+	unflushedBytes.StateChangeOperationIndexes = append(unflushedBytes.StateChangeOperationIndexes, dbOperationIndex)
+
+	// Update the state change file size.
+	transactionLen := uint32(len(writeBytes))
+	unflushedBytes.StateChangeBytesSize += transactionLen
```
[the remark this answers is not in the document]
diamondhands0: +1 (assuming we keep this field-- see my other comment)

lib/state_change_syncer.go
```diff
@@ -0,0 +1,726 @@
...
+
+	stateChangeSyncer.UnflushedBytes[flushId] = unflushedBytes
+}
+
+// FlushTransactionsToFile writes the bytes that have been cached on the StateChangeSyncer to the state change file.
+func (stateChangeSyncer *StateChangeSyncer) FlushTransactionsToFile(event *StateSyncerFlushedEvent) error {
+	flushId := event.FlushId
+	// Get the relevant global flush ID from the state change syncer if the flush ID is nil.
+	if event.IsMempoolFlush && event.FlushId == uuid.Nil {
```
diamondhands0, line 488: Would rewrite as follows:
```
	if event.FlushId == uuid.Nil {
		if event.IsMempoolFlush {
			flushId = stateChangeSyncer.MempoolFlushId
		} else {
			flushId = stateChangeSyncer.BlockSyncFlushId
		}
	}
```

lib/state_change_syncer.go
```diff
@@ -0,0 +1,726 @@
...
+		delete(stateChangeSyncer.UnflushedBytes, flushId)
+	}
+
+	unflushedBytes, exists := stateChangeSyncer.UnflushedBytes[flushId]
+
+	if !exists {
+		return nil
+	}
+
+	if unflushedBytes.StateChangeBytes == nil || unflushedBytes.StateChangeOperationIndexes == nil || len(unflushedBytes.StateChangeOperationIndexes) == 0 {
```
diamondhands0, line 519: With slices, prefer `len(xxx) == 0` over a nil check unless you're trying to explicitly check whether something has been initialized.

lib/state_change_syncer.go
```diff
@@ -0,0 +1,726 @@
...
+		return fmt.Errorf("Error flushing state change file: FlushId %v has nil bytes\n", flushId)
+	}
+
+	// Write the encoded StateChangeEntry bytes to the state changer file.
+	_, err := flushFile.Write(unflushedBytes.StateChangeBytes)
+
+	if err != nil {
+		return fmt.Errorf("Error writing to state change file: %v", err)
```
diamondhands0, line 527: May be useful to distinguish whether this failure happened on a mempool flush or a main db one. Latter is pretty serious so we'd want to see that in the logs.

lib/state_change_syncer.go
```diff
@@ -0,0 +1,726 @@
...
+	// The StateChangeOperationIndexes array contains the byte index of		 where each transaction occurs within the
+	// unflushed state change bytes (e.g. the first value in the slice will always be 0).
+	// We need to add the size of the state change file to each of these values to get the byte index
+	// in the state change file.
+	for _, indexBytes := range unflushedBytes.StateChangeOperationIndexes {
+		// Get the byte index of where this transaction occurs in the state change file.
+		dbOperationIndex := indexBytes + flushFileSize
+		// Convert the byte index value to a uint64 byte slice.
+		dbOperationIndexBytes := make([]byte, 8)
+		binary.LittleEndian.PutUint64(dbOperationIndexBytes, dbOperationIndex)
```
diamondhands0, line 543: Is there a reason we encode a full uint64 rather than using UintToBuf? I don't think the efficiency matters much just figured I'd mention it.

diamondhands0: For context, UintToBuf uses the minimum number of bytes to encode the value. So it would be 2-3 ish bytes for each number instead of 8.

lib/state_change_syncer.go
```diff
@@ -0,0 +1,726 @@
...
+		stateChangeIndexBuf = append(stateChangeIndexBuf, dbOperationIndexBytes...)
+	}
+
+	// Write the encoded uint32 indexes to the index file.
+	_, err = indexFile.Write(stateChangeIndexBuf)
+	if err != nil {
+		return fmt.Errorf("Error writing to state change index file: %v", err)
+	}
+	if event.IsMempoolFlush {
+		stateChangeSyncer.StateChangeMempoolFileSize += unflushedBytes.StateChangeBytesSize
```
diamondhands0, line 554: Still feels like you can get rid of this and use len(StateChangeBytes)

lib/state_change_syncer.go
superzordon: We need to be able to pinpoint where in the file the pointer lives for a particular index. If we used a dynamic-sized encoding, we'd need to iterate through every pointer until we reach the index we want. By using a standard-sized uint64, we seek straight to index * 8 to find the pointer we want.

lib/state_change_syncer.go
```diff
@@ -0,0 +1,726 @@
...
+		for server.mempool == nil || server.blockchain.chainState() != SyncStateFullyCurrent {
+			time.Sleep(1000 * time.Millisecond)
+		}
+		if !stateChangeSyncer.BlocksyncCompleteEntriesFlushed && stateChangeSyncer.SyncType == NodeSyncTypeBlockSync {
+			stateChangeSyncer.FlushAllEntriesToFile(server)
+		}
+		mempoolClosed := server.mempool.stopped
+		for !mempoolClosed {
+			// Sleep for a short while to avoid a tight loop.
+			time.Sleep(1000 * time.Millisecond)
```
diamondhands0, line 663: Wonder how low we can make this value. Would be cool if you could submit a txn and have it show up in PG in like 100ms. I'd maybe bias toward a lower value until we run into performance issues.

lib/state_change_syncer.go
```diff
@@ -0,0 +1,726 @@
...
+			mempoolClosed, err = stateChangeSyncer.SyncMempoolToStateSyncer(server)
+			if err != nil {
+				glog.Errorf("StateChangeSyncer.StartMempoolSyncRoutine: Error syncing mempool to state syncer: %v", err)
+			}
+		}
+	}()
+}
+
+func (stateChangeSyncer *StateChangeSyncer) FlushAllEntriesToFile(server *Server) error {
+	deadlock.Opts.DeadlockTimeout = 100 * time.Minute
```
diamondhands0, line 675: I think what we really want here is, rather than *extending* the deadlock timeout, we actually want to turn it off entirely. If there's a way to do that, let's do that instead. Or else set it to a very large value so that we don't hit this ceiling mysteriously in the future.

lib/db_utils.go
```diff
@@ -476,8 +477,50 @@ type DBPrefixes struct {
...
-	// NEXT_TAG: 78
+	// PrefixTxnHashToTxn is used to store transactions that have been processed by the node. This isn't actually stored
+	// in badger, but is tracked by state syncer when processing mempool transactions.
+	// The index has the following structure:
+	// 	<prefix, txnHash> -> <Transaction>
+	PrefixTxnHashToTxn []byte `prefix_id:"[78]" core_state:"true"`
+
+	// PrefixTxnHashToUtxoOps is used to store UtxoOps for transactions that have been processed by the node.
+	// This isn't actually stored in badger, but is tracked by state syncer when processing mempool transactions.
+	PrefixTxnHashToUtxoOps []byte `prefix_id:"[79]" core_state:"true"`
```
diamondhands0, line 488: It looks like this is only used in the unconnectedTransactions thing in the mempool dumping part, which I think we can delete, which would then mean that we can delete this as well.

lib/db_utils.go
```diff
@@ -6736,6 +6903,67 @@ type TransactionMetadata struct {
 	NewMessageTxindexMetadata            *NewMessageTxindexMetadata            `json:",omitempty"`
 }
 
+func (txnMeta *TransactionMetadata) GetEncoderForTxType(txnType TxnType) DeSoEncoder {
```
diamondhands0, line 6906: Looks like this is unused?

lib/state_change_syncer.go
```diff
@@ -0,0 +1,726 @@
...
+		if event.FlushId == uuid.Nil {
+			flushId = stateChangeSyncer.BlockSyncFlushId
+		}
+	}
+
+	// Get the relevant deso encoder for this keyBytes.
+	var encoderType EncoderType
+	if isEncoder, encoder := StateKeyToDeSoEncoder(stateChangeEntry.KeyBytes); isEncoder && encoder != nil {
+		// Blocks are stored in a non-DeSo-Encoder format, so we need to convert them to DeSo-Encoder format by appending metadata.
+		if encoder.GetEncoderType() == EncoderTypeBlock {
```
diamondhands0, line 375: I don't think you need to do this for blocks anymore? I noticed you added the code needed to turn MsgDeSoBlock into an actual DeSoEncoder, so it should "just work" without this right?

lib/snapshot.go
```diff
@@ -395,6 +403,10 @@ func NewSnapshot(mainDb *badger.DB, mainDbDirectory string, snapshotBlockHeightP
 	timer := &Timer{}
 	timer.Initialize()
 
+	// Limit the number of items stored in the OperationChannel. Because the snapshot chunks are 100MB each,
+	// this limits the number of operations stored at one time to 20GB.
+	const maxOperationQueueSize = 20
```
diamondhands0, line 408: Still so thrilled we figured out how to reduce memory with this.

lib/block_view.go
```diff
@@ -117,10 +117,13 @@ type UtxoView struct {
...
-	Postgres *Postgres
-	Params   *DeSoParams
-	Snapshot *Snapshot
+	Handle       *badger.DB
+	Postgres     *Postgres
+	Params       *DeSoParams
+	Snapshot     *Snapshot
+	EventManager *EventManager
+
+	IsMempoolView bool
```
diamondhands0, line 126: Let's add thorough comments to all of the above. I'm going to help where I can.

lib/state_change_syncer.go
```diff
@@ -0,0 +1,726 @@
...
+
+	stateChangeSyncer.BlockHeight = blockHeight
+
+	mempoolUtxoView, err := server.GetMempool().GetAugmentedUniversalView()
+	if err != nil {
+		return false, errors.Wrapf(err, "StateChangeSyncer.SyncMempoolToStateSyncer: ")
+	}
+	// Kill the snapshot so that it doesn't affect the original snapshot.
+	mempoolUtxoView.Snapshot = nil
+	mempoolUtxoView.IsMempoolView = true
```
diamondhands0, line 584: OK I think this is going to be a super annoying comment, but let me run it by you.

I don't love this IsMempoolView field, and I was thinking about how we could remove it. The reason I don't like it is because it's kindof confusing on the UtxoView struct, and we have to thread it through *everywhere* even though it doesn't really change over the course of a flush.

So I had the following suggestion:
1. Move this field to the EventManager object, which we're already threading through everywhere.
2. Delete it from all of the arguments we're passing it to now. This is why this comment is annoying: It means you have to go to db_utils.go and change a lot of function signatures to remove the field...
3. To make SyncMempoolToStateSyncer work, create a new EventManager struct and call stateSyncerOperation stateSyncerFlush on that in this function rather than using the existing one.
4. Then I think you can remove the field from UtxoView and db_utils.go and wherever else.

I don't want it to be something that takes a whole day, though. Can you just look at how annoying it is and make a call? If it's more than an hour ish to do it then let's not worry about it for now.

lib/block_view_types.go
```diff
@@ -110,9 +110,14 @@ const (
 	EncoderTypeDmThreadEntry                     EncoderType = 37
 	EncoderTypeDeSoNonce                         EncoderType = 38
 	EncoderTypeTransactorNonceEntry              EncoderType = 39
+	EncoderTypeStateChangeEntry                  EncoderType = 40
```
diamondhands0, line 113: Do me a favor and let's start adding comments to these. You don't need to annotate the others, but for the new ones you added just mention how they're used.

lib/state_change_syncer.go
```diff
@@ -0,0 +1,726 @@
...
+	// The type of operation that should be performed on the database.
+	OperationType StateSyncerOperationType
+	// The key that should be used for the operation.
+	KeyBytes []byte
+	// The encoder that is being captured by this state change.
+	Encoder DeSoEncoder
+	// The encoder represented in bytes. This could be created by just calling EncodeToBytes on the encoder, but
+	// during operations like hypersync, we are given the raw bytes of the encoder, which is all we need to encode the
+	// StateChangeEntry. Thus, we store the raw bytes here to avoid having to re-encode the encoder.
+	EncoderBytes []byte
```
diamondhands0, line 38: Can we name this like AncestralRecordEncoder maybe?

lib/state_change_syncer.go
```diff
@@ -0,0 +1,726 @@
...
+	MempoolFlushKeySet map[string]bool
+
+	// Tracks the flush IDs of the last block sync flush and the last mempool flush.
+	// These are not used during hypersync, as many flushes are being processed asynchronously.
+	BlockSyncFlushId uuid.UUID
+	MempoolFlushId   uuid.UUID
+
+	// Mutex to prevent concurrent writes to the state change file.
+	StateSyncerMutex *sync.Mutex
+	EntryCount       uint64
```
diamondhands0, line 217: I didn't see this used anywhere. If not, let's remove it.

lib/state_change_syncer.go
```diff
@@ -0,0 +1,726 @@
...
+			return
+		}
+
+		// Check to see if any of the keys in the mempool key set are not in the mempool key value map.
+		// If so, reset the mempool key value map, mempool key set, and mempool file, start from scratch.
+		for key, _ := range stateChangeSyncer.MempoolKeyValueMap {
+			// If any of the keys that the mempool is currently tracking weren't included in the flush, the state syncer
+			// mempool is bad and needs to be reset.
+			if _, ok := stateChangeSyncer.MempoolFlushKeySet[key]; !ok {
+				stateChangeSyncer.ResetMempool()
```
diamondhands0, line 427: So I've been trying really hard to figure out how MempoolKeyValueMap could ever be different from MempoolFlushKeySet. Right now, it seems like they'd always have the exact same stuff in them, and thus this inner thing would never get triggered. Am I missing something?

lib/state_change_syncer.go
```diff
@@ -0,0 +1,726 @@
...
+		}
+	} else {
+		// If the flush ID is nil, then we need to use the global block sync flush ID.
+		if event.FlushId == uuid.Nil {
+			flushId = stateChangeSyncer.BlockSyncFlushId
+		}
+	}
+
+	// Get the relevant deso encoder for this keyBytes.
+	var encoderType EncoderType
```
diamondhands0, line 372: More comments here. I want to better understand when something hits the first case vs the second. I think it happens whenever we have a field in the db that's a "key only" field with no Entry as a value.

lib/state_change_syncer.go
```diff
@@ -0,0 +1,726 @@
...
+	// The encoder that is being captured by this state change.
+	Encoder DeSoEncoder
+	// The encoder represented in bytes. This could be created by just calling EncodeToBytes on the encoder, but
+	// during operations like hypersync, we are given the raw bytes of the encoder, which is all we need to encode the
+	// StateChangeEntry. Thus, we store the raw bytes here to avoid having to re-encode the encoder.
+	EncoderBytes []byte
+	// The ancestral record that should be used to revert this transaction.
+	AncestralRecord DeSoEncoder
+	// The ancestral record represented in bytes.
+	AncestralRecordBytes []byte
```
diamondhands0, line 42: Add comment on when ancestral record bytes is used. I assume in hypersync it's not pupulated for example. Also I think it's only for upsert?

lib/state_change_syncer.go
```diff
@@ -0,0 +1,726 @@
...
+func (stateChangeEntry *StateChangeEntry) GetEncoderType() EncoderType {
+	return EncoderTypeStateChangeEntry
+}
+
+// UnflushedStateSyncerBytes is used to keep track of the bytes that should be written to the state change file upon a db flush.
+type UnflushedStateSyncerBytes struct {
+	StateChangeBytes     []byte
+	StateChangeBytesSize uint64
+	// This is a list of the indexes of the state change bytes that should be written to the state change index file.
+	StateChangeOperationIndexes []uint64
```
diamondhands0, line 177: Give comments on these fields that explains how they work.

lib/state_change_syncer.go
```diff
@@ -0,0 +1,726 @@
...
+	StateSyncerMutex *sync.Mutex
+	EntryCount       uint64
+
+	BlockHeight uint64
+
+	SyncType NodeSyncType
+
+	// During blocksync, we flush all entries by index to the state change file once the sync is complete.
+	// This is used to track whether this procedure has been initiated.
+	BlocksyncCompleteEntriesFlushed bool
```
diamondhands0, line 225: Explain the deeper reasons for why we do this optimization, rather than just flushing with each block. We want to give a lot of context for optimizations like these so that we can revisit them intelligently in the future. Otherwise, people are scared to change things like this for fear it will break things in unexpected ways.

cmd/node.go
```diff
@@ -141,7 +141,7 @@ func (node *Node) Start(exitChannels ...*chan struct{}) {
 
 	// Setup chain database
 	dbDir := lib.GetBadgerDbPath(node.Config.DataDirectory)
-	opts := lib.PerformanceBadgerOptions(dbDir)
+	opts := badger.DefaultOptions(dbDir)
```
diamondhands0, line 144: Revert to PerformanceOptions per our conversation

lib/snapshot.go
```diff
@@ -303,15 +309,17 @@ type Snapshot struct {
...
 
 // NewSnapshot creates a new snapshot instance.
 func NewSnapshot(mainDb *badger.DB, mainDbDirectory string, snapshotBlockHeightPeriod uint64, isTxIndex bool,
-	disableChecksum bool, params *DeSoParams, disableMigrations bool, useDefaultBadgerOptions bool) (_snap *Snapshot, _err error, _shouldRestart bool) {
+	disableChecksum bool, params *DeSoParams, disableMigrations bool, useDefaultBadgerOptions bool, eventManager *EventManager) (_snap *Snapshot, _err error, _shouldRestart bool) {
 
 	// Initialize the ancestral records database
 	snapshotDirectory := filepath.Join(GetBadgerDbPath(mainDbDirectory), "snapshot")
-	snapshotOpts := PerformanceBadgerOptions(snapshotDirectory)
+	snapshotOpts := DefaultBadgerOptions(snapshotDirectory)
```
diamondhands0, line 322: Revert to PerformanceBadgerOptions per our conversation

scripts/nodes/n0_test
```diff
@@ -25,7 +25,6 @@ rm /tmp/main.*.log
   --testnet=true  \
   --api-port=18001 \
   --protocol-port=18000 \
```
diamondhands0, line 27: Can we delete the whole nodes folder here? People should only use the ones in backend from now on. These are outdated and misleading.

lib/db_utils.go
superzordon: This is used in the consumer code (`txIndexMetadata.GetEncoderForTxType(transaction.TxnMeta.GetTxnType())`), but if you'd prefer to keep it out of core we could move it to a helper in the consumer library.

lib/state_change_syncer.go
superzordon: `PrefixBlockHashToBlock` stores each block serialized as a `DeSoMessage`, rather than a `DeSoEncoder`. The only difference between these 2 formats is the proceeding metadata, so rather than decoding the struct and re-encoding it, we just call this helper to append the `DeSoEncoder` metadata.

lib/state_change_syncer.go
superzordon: This actually isn't the AncestralRecord, it's the bytes for the encoder this state change represents. This is used during hypersync - rather than decoding the bytes we receive from the peer, passing it to state syncer, and then re-encoding the bytes, we just pass the raw bytes through this `EncoderBytes` field.

lib/state_change_syncer.go
superzordon: Yep, it's only for upsert. We only use ancestral records when syncing mempool entries. When the new block syncs, we need to revert any mempool entries we've stored in the database before applying the committed entries. We do this reversion by reverting to the ancestral record value.

scripts/nodes/n0_test
superzordon: This makes sense to me 👍

lib/state_change_syncer.go
superzordon: The intention of this code was to iterate over all transactions in the mempool and emit them to the state syncer. This has been switched from `unconnectedTxns` to `_getTransactionsOrderedByTimeAdded`.

lib/state_change_syncer.go
superzordon: `MempoolFlushKeySet` is reset every mempool flush, whereas `MempoolKeyValueMap` is reset every time a block mining occurs. The purpose of this is to detect if any entries being tracked in the mempool by state syncer become ejected. The `MempoolKeyValueMap` represents everything that has been flushed to the mempool since the last block, while the `MempoolFlushKeySet` represents only things from this latest flush. If something is in the `MempoolKeyValueMap` but is missing from `MempoolFlushKeySet`, that would indicate an entry has been ejected from the mempool since we started tracking it. The course of action here is to reset the state syncer mempool, so that the consumer can get a fresh view of the mempool.

I'll update the documentation around these fields to better represent and explain this.

lib/state_change_syncer.go
diamondhands0: Ah looks like the line numbers were messed up. I was referring to the field below this one, but it doesn't matter.

lib/state_change_syncer.go
diamondhands0: OK I get this fully now and it's a clever solution. Basically, every time we iterate over the mempool, we need to make sure we're flushing a superset of what we flushed in the past. If that's the case then it means writing only the new stuff is safe (and this is what we do in a different loop somewhere else). Otherwise, we need to reset the mempool and just write the whole thing from scratch.

diamondhands0 (approved):

lib/block_producer.go
```diff
@@ -202,7 +202,7 @@ func (desoBlockProducer *DeSoBlockProducer) _getBlockTemplate(publicKey []byte)
 
 		// Create a new view object.
 		utxoView, err := NewUtxoView(desoBlockProducer.chain.db, desoBlockProducer.params,
-			desoBlockProducer.postgres, desoBlockProducer.chain.snapshot)
+			desoBlockProducer.postgres, desoBlockProducer.chain.snapshot, nil)
```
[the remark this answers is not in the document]
diamondhands0: I think we always include it except for weird edge-cases. In this case, getBlockTemplate runs outside of the main event loop, and just generates blocks for miners. I would expect mempool may be the one other place where we don't include the EventManager.

lib/block_producer.go
```diff
@@ -224,9 +224,7 @@ func (desoBlockProducer *DeSoBlockProducer) _getBlockTemplate(publicKey []byte)
 			if err != nil {
 				return nil, nil, nil, errors.Wrapf(err, "Error copying UtxoView: ")
 			}
-			_, _, _, _, err = utxoViewCopy._connectTransaction(
-				mempoolTx.Tx, mempoolTx.Hash, int64(mempoolTx.TxSizeBytes), uint32(blockRet.Header.Height), true,
-				false /*ignoreUtxos*/)
+			_, _, _, _, err = utxoViewCopy._connectTransaction(mempoolTx.Tx, mempoolTx.Hash, int64(mempoolTx.TxSizeBytes), uint32(blockRet.Header.Height), true, false)
```
diamondhands0, line 227: It would be nice if you could do a pass to fix all the line lengths that the refactoring broke. With vim you can highlight a line and hit "gq" and it does it. But not sure what the shortcut is in other things.

lib/blockchain.go
```diff
@@ -2155,36 +2156,44 @@ func (bc *Blockchain) ProcessBlock(desoBlock *MsgDeSoBlock, verifySignatures boo
...
+				var innerErr error
+				if innerErr = PutUtxoOperationsForBlockWithTxn(txn, bc.snapshot, blockHeight, blockHash, utxoOpsForBlock, bc.eventManager); innerErr != nil {
 					return errors.Wrapf(innerErr, "ProcessBlock: Problem writing utxo operations to db on simple add to tip")
 				}
 				bc.timer.End("Blockchain.ProcessBlock: Transactions Db snapshot & operations")
-				if innerErr := bc.blockView.FlushToDbWithTxn(txn, blockHeight); innerErr != nil {
+				if innerErr = bc.blockView.FlushToDbWithTxn(txn, blockHeight); innerErr != nil {
 					return errors.Wrapf(innerErr, "ProcessBlock: Problem writing utxo view to db on simple add to tip")
 				}
+				if bc.eventManager != nil && !bc.eventManager.isMempoolManager {
```
diamondhands0, line 2190: Do me a favor and just add a comment explaining why we put this event trigger here rather than elsewhere. For example, is there a reason why we need to put it in the badger Update() vs outside of it? Just explain it, or explain that the placement doesn't matter, in case someone needs to move things around in the future.

lib/state_change_syncer.go
```diff
@@ -0,0 +1,867 @@
...
+				lastReceivedKey = dbBatchEntries[len(dbBatchEntries)-1].Key
+			}
+			for ii, dbEntry := range dbBatchEntries {
+				stateChangeEntry := &StateChangeEntry{
+					OperationType: DbOperationTypeInsert,
+					KeyBytes:      dbEntry.Key,
+					EncoderBytes:  dbEntry.Value,
+				}
+
+				fmt.Printf("Flushing entry %d to file\n", ii)
```
diamondhands0, line 825: Do you want to leave all these prints in here? Fine if they're intentional I'm just flagging it.

lazynina (approved):

PR #501, opened 2023-04-21
Mf/add block view stake

mattfoley8:


lib/block_view_types.go
```diff
@@ -1821,7 +1878,42 @@ func (op *UtxoOperation) RawDecodeWithoutMetadata(blockHeight uint64, rr *bytes.
...
+			return errors.Wrapf(err, "UtxoOperation.Decode: Problem reading PrevStakeEntries: ")
+		}
+		for ii := 0; ii < int(numPrevStakeEntries); ii++ {
+			prevStakeEntry := &StakeEntry{}
+			if exist, err := DecodeFromBytes(prevStakeEntry, rr); exist && err == nil {
+				op.PrevStakeEntries = append(op.PrevStakeEntries, prevStakeEntry)
+			} else if err != nil {
+				return errors.Wrapf(err, "UtxoOperation.Decode: Problem reading PrevStakeEntries: ")
+			}
+		}
```
lazynina, line 1903: any maybe something like this for decoding? can slot this into a separate PR so it doesn't muddy this one.
```
func DecodeSlice[T DeSoEncoder](rr *bytes.Reader) ([]T, error) {
	numItems, err := ReadUvarint(rr)
	if err != nil {
		return nil, errors.Wrapf(err, "DecodeSlice: Problem decoding numItems")
	}
	// Note: is it more efficient to do a make with specific length and set at each index?
	inputs := make([]T, numItems)
	var results []T
	for ii := uint64(0); ii < numItems; ii++ {
		prevEntry := inputs[ii].GetEncoderType().New()
		if exist, err := DecodeFromBytes(prevEntry, rr); exist && err == nil {
			results = append(results, prevEntry.(T))
		} else if err != nil {
			return nil, errors.Wrapf(err, "DecodeSlice: Problem decoding item %d of %d", ii, numItems)
		}
	}
	return results, nil
}
```

lib/block_view_validator.go
```diff
@@ -1249,11 +1249,28 @@ func (bav *UtxoView) GetValidatorByPKID(pkid *PKID) (*ValidatorEntry, error) {
 	if err != nil {
 		return nil, err
 	}
-	// Cache the ValidatorEntry from the db in the UtxoView.
-	bav._setValidatorEntryMappings(dbValidatorEntry)
+	if dbValidatorEntry != nil {
```
lazynina, line 1252: good catch

lib/network.go
```diff
@@ -5599,6 +5631,100 @@ func (tsl *TransactionSpendingLimit) ToMetamaskString(params *DeSoParams) string
...
+			opString += _indt(indentationCounter) + "Transaction Count: " + strconv.FormatUint(limit, 10) + "\n"
+
+			indentationCounter--
+			opString += _indt(indentationCounter) + "]\n"
+			unlockStakeLimitStr = append(unlockStakeLimitStr, opString)
+		}
+		// Ensure deterministic ordering of the transaction count limit strings by doing a lexicographical sort.
+		sortStringsAndAddToLimitStr(unlockStakeLimitStr)
+		indentationCounter--
+	}
```
lazynina, line 5726: I need to look through this piece again, but I think it looks good. Just a note to myself

mattfoley8: Sounds good. Also check out my test for the ToMetamaskStr() function which shows what the output would like for different stake/unstake/unlock stake derived key txn spending limit configurations.

lib/block_view_stake.go
```diff
@@ -0,0 +1,2476 @@
...
+func MakeStakeLimitKey(validatorPKID *PKID, stakerPKID *PKID) StakeLimitKey {
+	return StakeLimitKey{
+		ValidatorPKID: *validatorPKID,
+		StakerPKID:    *stakerPKID,
+	}
+}
+
+func (stakeLimitKey *StakeLimitKey) Encode() []byte {
+	var data []byte
+	data = append(data, stakeLimitKey.ValidatorPKID.ToBytes()...)
```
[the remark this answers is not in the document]
mattfoley8: I didn't use `EncodeToBytes` here only because we don't have the blockHeight + skipMetadata fields handy here to use. FWIW, this is what we also did for associations limit key [here](https://github.com/deso-protocol/core/blob/main/lib/network.go#L6655). Happy to modify though if you think this will be an issue.

lib/block_view_stake.go
```diff
@@ -0,0 +1,2496 @@
...
+	lockedStakeEntry.LockedAtEpochNumber, err = ReadUvarint(rr)
+	if err != nil {
+		return errors.Wrapf(err, "LockedStakeEntry.Decode: Problem reading LockedAtEpochNumber: ")
+	}
+
+	// ExtraData
+	lockedStakeEntry.ExtraData, err = DecodeExtraData(rr)
+	if err != nil {
+		return errors.Wrapf(err, "LockedStakeEntry.Decode: Problem reading ExtraData: ")
+	}
```
lazynina: see PR here to use DecodeDeSoEncoder: https://github.com/deso-protocol/core/pull/505

lib/block_view_stake.go
```diff
@@ -0,0 +1,2496 @@
...
+
+	// Validate the txn metadata.
+	if err := bav.IsValidUnlockStakeMetadata(txn.PublicKey, txMeta); err != nil {
+		return 0, 0, nil, errors.Wrapf(err, "_connectUnlockStake: ")
+	}
+
+	// Convert TransactorPublicKey to TransactorPKID.
+	transactorPKIDEntry := bav.GetPKIDForPublicKey(txn.PublicKey)
+	if transactorPKIDEntry == nil || transactorPKIDEntry.isDeleted {
+		return 0, 0, nil, RuleErrorInvalidStakerPKID
```
lazynina, line 1594: Let's wrap these rule errors - will help us track which type of txn we hit it in.

lib/block_view_stake.go
```diff
@@ -0,0 +1,2496 @@
...
+		return 0, 0, nil, RuleErrorInvalidUnlockStakeUnlockableStakeOverflowsUint64
+	}
+	totalUnlockedAmountNanosUint64 := totalUnlockedAmountNanos.Uint64()
+
+	// Return TotalUnlockedAmountNanos back to the transactor. We can use
+	// _addBalance here since we validate that connectUnlockStake can only
+	// occur after the BalanceModelBlockHeight.
+	utxoOp, err := bav._addBalance(totalUnlockedAmountNanosUint64, txn.PublicKey)
+	if err != nil {
+		return 0, 0, nil, errors.Wrapf(err, "_connectUnlockStake: ")
```
lazynina, line 1649: I kinda like having more in the `_connectUnlockStake` so debugging is easier.

lib/block_view_stake.go
```diff
@@ -0,0 +1,2496 @@
...
+		return RuleErrorInvalidUnlockStakeUnlockableStakeOverflowsUint64
+	}
+
+	// Restore the PrevLockedStakeEntries.
+	for _, prevLockedStakeEntry := range operationData.PrevLockedStakeEntries {
+		bav._setLockedStakeEntryMappings(prevLockedStakeEntry)
+	}
+
+	// Unadd TotalUnlockedAmountNanos from the transactor.
+	err = bav._unAddBalance(totalUnlockedAmountNanos.Uint64(), currentTxn.PublicKey)
```
lazynina, line 1729: I think we want to `_unAdd` before we revert the locked stake entries.

lib/block_view_stake.go
```diff
@@ -0,0 +1,2496 @@
...
+	if !existsLockedStakeEntries {
+		return RuleErrorInvalidUnlockStakeNoUnlockableStakeFound
+	}
+
+	return nil
+}
+
+func (bav *UtxoView) GetStakeEntry(validatorPKID *PKID, stakerPKID *PKID) (*StakeEntry, error) {
+	// First, check the UtxoView.
+	stakeMapKey := StakeMapKey{ValidatorPKID: *validatorPKID, StakerPKID: *stakerPKID}
```
lazynina, line 1854: do we want to check that validator PKID and staker PKID is not nil before we do this? just prevents a panic.

lib/block_view_stake.go
```diff
@@ -0,0 +1,2496 @@
...
+	return stakeEntry, nil
+}
+
+func (bav *UtxoView) GetLockedStakeEntry(
+	validatorPKID *PKID,
+	stakerPKID *PKID,
+	lockedAtEpochNumber uint64,
+) (*LockedStakeEntry, error) {
+	// First, check the UtxoView.
+	lockedStakeMapKey := LockedStakeMapKey{
```
lazynina, line 1880: let's check that pkids are not nil first.

lib/block_view_stake.go
```diff
@@ -0,0 +1,2496 @@
...
+
+	// First, pull matching LockedStakeEntries from the db.
+	dbLockedStakeEntries, err := DBGetLockedStakeEntriesInRange(
+		bav.Handle, bav.Snapshot, validatorPKID, stakerPKID, startEpochNumber, endEpochNumber,
+	)
+	if err != nil {
+		return nil, errors.Wrapf(err, "UtxoView.GetLockedStakeEntriesInRange: ")
+	}
+	for _, lockedStakeEntry := range dbLockedStakeEntries {
+		lockedStakeEntriesSet.Add(lockedStakeEntry)
```
lazynina, line 1922: hmm upon more thought, I'm a little concerned about using sets here. I'm worried that the comparable stuff won't work w/ pointers.

lib/block_view_validator_test.go
```diff
@@ -382,21 +381,19 @@ func _submitUnregisterAsValidatorTxn(
...
-	params.ForkHeights.DerivedKeySetSpendingLimitsBlockHeight = uint32(0)
-	params.ForkHeights.DerivedKeyTrackSpendingLimitsBlockHeight = uint32(0)
-	params.ForkHeights.DerivedKeyEthSignatureCompatibilityBlockHeight = uint32(0)
-	params.ForkHeights.ExtraDataOnEntriesBlockHeight = uint32(0)
-	params.ForkHeights.AssociationsAndAccessGroupsBlockHeight = uint32(0)
-	params.ForkHeights.BalanceModelBlockHeight = uint32(1)
 	params.ForkHeights.ProofOfStakeNewTxnTypesBlockHeight = uint32(1)
 	GlobalDeSoParams.EncoderMigrationHeights = GetEncoderMigrationHeights(&params.ForkHeights)
 	GlobalDeSoParams.EncoderMigrationHeightsList = GetEncoderMigrationHeightsList(&params.ForkHeights)
+	chain.snapshot = nil
```
lazynina, line 396: why do we need this?

lib/db_utils.go
```diff
@@ -7115,6 +7138,27 @@ func (txnMeta *TransactionMetadata) RawDecodeWithoutMetadata(blockHeight uint64,
 		} else {
 			return errors.Wrapf(err, "TransactionMetadata.Decode: Problem reading UnregisterAsValidatorTxindexMetadata")
 		}
+		// decoding StakeTxindexMetadata
+		CopyStakeTxindexMetadata := &StakeTxindexMetadata{}
+		if exist, err := DecodeFromBytes(CopyStakeTxindexMetadata, rr); exist && err == nil {
+			txnMeta.StakeTxindexMetadata = CopyStakeTxindexMetadata
+		} else {
+			return errors.Wrapf(err, "TransactionMetadata.Decode: Problem reading StakeTxindexMetadata")
+		}
```
lazynina, line 7147: I think we can use:
```suggestion
		txnMeta.StakeTxindexMetadata, err = DecodeDeSoEncoder(&StakeTxindexMetadata{}, rr)
		if err != nil {
			return errors.Wrapf(err, "TransactionMetadata.Decode: Problem reading StakeTxindexMetadata")
		}
```

lib/block_view_stake.go
lazynina: Okay so I wrote a quick test and it fails. So we will want to do something a bit different. I'd recommend creating a map of LockedStakeMapKey to LockedStakeEntry like we have in the utxo view. Here's the test for reference:

```
func TestStructSetAdd(t *testing.T) {
	lockedStakeEntry := &LockedStakeEntry{
		LockedStakeID:       &ZeroBlockHash,
		StakerPKID:          &ZeroPKID,
		ValidatorPKID:       &ZeroPKID,
		LockedAmountNanos:   uint256.NewInt(),
		LockedAtEpochNumber: 100,
	}
	set := NewSet([]*LockedStakeEntry{lockedStakeEntry})

	set.Add(lockedStakeEntry.Copy())

	require.Equal(t, set.Size(), 1)
}
```

lib/block_view_stake.go
```diff
@@ -0,0 +1,2496 @@
...
+			lockedStakeEntriesSet.Add(lockedStakeEntry)
+		}
+	}
+
+	// Convert LockedStakeEntries set to slice, sorted by LockedAtEpochNumber ASC.
+	lockedStakeEntries := lockedStakeEntriesSet.ToSlice()
+	sort.Slice(lockedStakeEntries, func(ii, jj int) bool {
+		return lockedStakeEntries[ii].LockedAtEpochNumber < lockedStakeEntries[jj].LockedAtEpochNumber
+	})
+	return lockedStakeEntries, nil
```
lazynina, line 1950:
```suggestion
	// First, pull matching LockedStakeEntries from the db.
	dbLockedStakeEntries, err := DBGetLockedStakeEntriesInRange(
		bav.Handle, bav.Snapshot, validatorPKID, stakerPKID, startEpochNumber, endEpochNumber,
	)
	if err != nil {
		return nil, errors.Wrapf(err, "UtxoView.GetLockedStakeEntriesInRange: ")
	}
	lockedStakeEntryMap := make(map[LockedStakeMapKey]*LockedStakeEntry)
	for _, lockedStakeEntry := range dbLockedStakeEntries {
		lockedStakeEntryMap[lockedStakeEntry.ToMapKey()] = lockedStakeEntry
	}

	// Then, pull matching LockedStakeEntries from the UtxoView.
	// Loop through all LockedStakeEntries in the UtxoView.
	for _, lockedStakeEntry := range bav.LockedStakeMapKeyToLockedStakeEntry {
		// Filter to matching LockedStakeEntries.
		if !lockedStakeEntry.ValidatorPKID.Eq(validatorPKID) ||
			!lockedStakeEntry.StakerPKID.Eq(stakerPKID) ||
			lockedStakeEntry.LockedAtEpochNumber < startEpochNumber ||
			lockedStakeEntry.LockedAtEpochNumber > endEpochNumber {
			continue
		}

		if lockedStakeEntry.isDeleted {
			// Remove from map if isDeleted.
			delete(lockedStakeEntryMap, lockedStakeEntry.ToMapKey())
		} else {
			// Otherwise, add to map.
			lockedStakeEntryMap[lockedStakeEntry.ToMapKey()] = lockedStakeEntry
		}
	}

	var lockedStakeEntries []*LockedStakeEntry
	// Convert LockedStakeEntries map to slice.
	for _, lockedStakeEntry := range lockedStakeEntryMap {
		lockedStakeEntries = append(lockedStakeEntries, lockedStakeEntry)
	}

	// Sort LockedStakeEntries by LockedAtEpochNumber ASC.
	sort.Slice(lockedStakeEntries, func(ii, jj int) bool {
		return lockedStakeEntries[ii].LockedAtEpochNumber < lockedStakeEntries[jj].LockedAtEpochNumber
	})
	return lockedStakeEntries, nil
```

lib/block_view_stake.go
```diff
@@ -0,0 +1,2514 @@
...
+	}
+
+	// Set StakeID only if this is a new StakeEntry.
+	stakeID := txHash
+	if prevStakeEntry != nil {
+		stakeID = prevStakeEntry.StakeID
+	}
+
+	// Calculate StakeAmountNanos.
+	stakeAmountNanos := txMeta.StakeAmountNanos
```
lazynina, line 1127: let's do a .Clone here to be safe.

lib/block_view_stake.go
```diff
@@ -0,0 +1,2514 @@
...
+	if err != nil {
+		return 0, 0, nil, errors.Wrapf(err, "_connectUnlockStake: error adding TotalUnlockedAmountNanos to TotalInput: ")
+	}
+
+	// Add TotalUnlockedAmountNanos to TotalOutput. The unlocked
+	// coins being sent to the transactor are an implicit output.
+	totalOutput, err = SafeUint64().Add(totalOutput, totalUnlockedAmountNanosUint64)
+	if err != nil {
+		return 0, 0, nil, errors.Wrapf(err, "_connectUnlockStake: error adding TotalUnlockedAmountNanos to TotalOutput: ")
+	}
```
lazynina, line 1667: can we move this just above the `_addBalance` call?

lazynina (approved):

PR #503, opened 2023-04-21
adding generic encode/decode deso encoder slices and single structs

lazynina:
If we like this, will make a separate PR to apply more broadly.

### DIFF
--- a/lib/generics.go
+++ b/lib/generics.go
@@ -111,9 +111,10 @@ func DecodeDeSoEncoderSlice[T DeSoEncoder](rr *bytes.Reader) ([]T, error) {
 		return nil, errors.Wrapf(err, "DecodeSlice: Problem decoding numItems")
 	}
 	// Note: is it more efficient to do a make with specific length and set at each index?
+	inputs := make([]T, numItems)
 	var results []T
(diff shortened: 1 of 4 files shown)


lib/generics.go
```diff
@@ -71,3 +76,47 @@ func MapSet[T comparable, K any](set *Set[T], mapFunc func(elem T) (K, error)) (
 	}
 	return results, nil
 }
+
+func DecodeDeSoEncoder[T DeSoEncoder](entry T, rr *bytes.Reader) (T, error) {
```
mattfoley8, line 80: Would be good to add some quick tests for these functions.

lazynina: Added a very basic test here: https://github.com/deso-protocol/core/pull/503/commits/713297bd3b58774802a6de3bdf9ebe506dcb9b4c

I don't think we need a much more intensive test here as encode/decode is tested on every use of `TestAPIServer` when it calls (TestDeSoEncoderSetup)[https://github.com/deso-protocol/core/blob/07a3df50a42be566bf582ed8a3b55a25e8f27cb6/lib/block_view_test.go#L306]

mattfoley8 (approved):

PR #506, opened 2023-04-24
Add tests for top validators by stake.

mattfoley8:


lib/block_view_validator_test.go
```diff
@@ -629,3 +631,260 @@ func _testValidatorRegistrationWithDerivedKey(t *testing.T) {
...
+
+	// Initialize test chain and miner.
+	chain, params, db := NewLowDifficultyBlockchain(t)
+	mempool, miner := NewTestMiner(t, chain, params, true)
+
+	// Initialize PoS fork height.
+	params.ForkHeights.ProofOfStakeNewTxnTypesBlockHeight = uint32(1)
+	GlobalDeSoParams.EncoderMigrationHeights = GetEncoderMigrationHeights(&params.ForkHeights)
+	GlobalDeSoParams.EncoderMigrationHeightsList = GetEncoderMigrationHeightsList(&params.ForkHeights)
+	chain.snapshot = nil
```
lazynina, line 651: can you remind me why we set the snapshot to nil here?

lazynina (approved):

lib/block_view_validator_test.go
mattfoley8: Can't remember where, but this was a potential lead for why some tests were intermittently failing. Piotr recommended adding this to prevent snapshotting from interfering with tests. Just an extra safety check. I removed and tests still pass. Gonna leave off and we'll see if it ever causes us issues.

PR #513, opened 2023-04-26
Mf/unstake stakers for unregistering validator

mattfoley8:


lib/block_view_stake.go
```diff
@@ -1080,7 +1106,7 @@ func (bav *UtxoView) _connectStake(
 	if err != nil {
 		return 0, 0, nil, errors.Wrapf(err, "_connectStake: ")
 	}
-	if prevValidatorEntry == nil || prevValidatorEntry.isDeleted || prevValidatorEntry.DisableDelegatedStake {
+	if prevValidatorEntry == nil || prevValidatorEntry.isDeleted {
```
mattfoley8, line 1109: This was a duplicate check. We already check for this in `IsValidStakeMetadata`.

lib/block_view_stake_test.go
```diff
@@ -49,17 +49,10 @@ func _testStaking(t *testing.T, flushToDB bool) {
 
 	_registerOrTransferWithTestMeta(testMeta, "m0", senderPkString, m0Pub, senderPrivString, 1e3)
 	_registerOrTransferWithTestMeta(testMeta, "m1", senderPkString, m1Pub, senderPrivString, 1e3)
-	_registerOrTransferWithTestMeta(testMeta, "m2", senderPkString, m2Pub, senderPrivString, 1e3)
```
mattfoley8, line 52: This is just some cleanup. No sense sending $ to test users who aren't used in the test.

lib/block_view_validator.go
```diff
@@ -292,7 +292,6 @@ type RegisterAsValidatorTxindexMetadata struct {
 	VotingPublicKey               string
 	VotingPublicKeySignature      string
 	VotingSignatureBlockHeight    uint64
-	UnstakedStakers               []*UnstakedStakerTxindexMetadata
```
mattfoley8, line 295: We no longer will have unstaked stakers when registering as a validator. Instead we validate that a user can't update DisableDelegatedStake=true if they have existing delegated stakers.

lazynina (approved):

PR #514, opened 2023-04-28
Add epoch data type.

mattfoley8:


lib/pos_epoch.go
```diff
@@ -0,0 +1,170 @@
...
+	"github.com/pkg/errors"
+)
+
+//
+// TYPE
+//
+
+type EpochEntry struct {
+	EpochNumber            uint64
+	LastBlockHeightInEpoch uint64
```
tholonious, line 15: Equally clear but a bit more concise
```suggestion
	FinalBlockHeight uint64
```

tholonious (changes requested):

lib/pos_epoch.go
```diff
@@ -0,0 +1,182 @@
...
+	if err != nil {
+		return 0, errors.Wrapf(err, "UtxoView.GetCurrentEpochNumber: ")
+	}
+	if epochEntry == nil {
+		return 0, errors.New("UtxoView.GetCurrentEpochNumber: no CurrentEpochEntry found")
+	}
+	return epochEntry.EpochNumber, nil
+}
+
+func (bav *UtxoView) SetCurrentEpochEntry(epochEntry *EpochEntry, blockHeight uint64) error {
```
lazynina, line 95: typically like to have these setters just glog the errors instead of returning them. Also these are typically private functions.

lib/pos_epoch.go
```diff
@@ -0,0 +1,182 @@
...
+	// This function should only ever be called from the OnEpochEnd hook.
+	if epochEntry == nil {
+		return errors.New("UtxoView.SetCurrentEpochEntry: called with nil EpochEntry")
+	}
+
+	// TODO: is this what we should do here?
+	if bav.Snapshot != nil {
```
lazynina, line 102: not sure if this is the correct place for this logic. I'll need to think about its proper place a little more. I think it should be handled in `FlushToDbWithTxn`

lib/pos_epoch.go
```diff
@@ -0,0 +1,182 @@
...
+	if err := DBPutCurrentEpochEntry(bav.Handle, bav.Snapshot, epochEntry, blockHeight); err != nil {
+		return errors.Wrapf(err, "UtxoView.SetCurrentEpochEntry: problem setting EpochEntry in db: ")
+	}
+
+	// Set the current EpochEntry in the UtxoView.
+	bav.CurrentEpochEntry = epochEntry.Copy()
+	return nil
+}
+
+func (bav *UtxoView) DeleteCurrentEpochEntry() {
```
lazynina, line 123: private

lazynina (changes requested):

lib/pos_epoch.go
```diff
@@ -0,0 +1,171 @@
...
+	return epochEntry, nil
+}
+
+func DBPutCurrentEpochEntryWithTxn(txn *badger.Txn, snap *Snapshot, epochEntry *EpochEntry, blockHeight uint64) error {
+	// Set EpochEntry in PrefixCurrentEpoch.
+	if epochEntry == nil {
+		// This is just a safety check that we are not accidentally overwriting an
+		// existing EpochEntry with a nil EpochEntry. This should never happen.
+		glog.Errorf("DBPutCurrentEpochEntryWithTxn: called with nil EpochEntry")
+		return nil
```
lazynina, line 162: since this function does return an error, should we just return the error instead of glogging?

lazynina (approved):

PR #517, opened 2023-05-01
Mf/validate stake lockup period

mattfoley8:


lib/block_view_stake.go
```diff
@@ -1476,8 +1481,9 @@ func (bav *UtxoView) _connectUnstake(
 		Type:                       OperationTypeUnstake,
 		PrevValidatorEntry:         prevValidatorEntry,
 		PrevGlobalStakeAmountNanos: prevGlobalStakeAmountNanos,
-		PrevStakeEntries:           []*StakeEntry{prevStakeEntry},
-		PrevLockedStakeEntries:     []*LockedStakeEntry{prevLockedStakeEntry},
+		PrevStakeEntries:           prevStakeEntries,
+		PrevLockedStakeEntries:     prevLockedStakeEntries,
+		PrevEpochNumber:            currentEpochNumber,
```
lazynina, line 1486: Why do we need `PrevEpochNumber` in the utxo op? Why can't we look it up in disconnect? jw

lib/constants.go
```diff
@@ -968,6 +972,9 @@ var DeSoMainnetParams = DeSoParams{
 	// reserve ratios.
 	CreatorCoinAutoSellThresholdNanos: uint64(10),
 
+	// Unstaked stake can be unlocked after a minimum of N elapsed epochs.
+	StakeLockupEpochDuration: uint64(3),
```
lazynina, line 976: will we want to move this to be a param updater thing eventually or nah?

lib/block_view_stake.go
mattfoley8: I think you're right. If we ever disconnect a transaction, it'll be in the same block/epoch as the connect logic was ran. I removed.

lib/constants.go
mattfoley8: Yes, we will. I added a comment. Good call-out.

lazynina (approved):

PR #518, opened 2023-05-02
Add jailed status to validator entries.

mattfoley8:


lib/db_utils.go
```diff
@@ -4216,6 +4216,25 @@ func DecodeUint64(scoreBytes []byte) uint64 {
 	return binary.BigEndian.Uint64(scoreBytes)
 }
 
+func EncodeUint16(num uint16) []byte {
```
mattfoley8, line 4219: I added EncodeUint16 instead of EncodeUint8 since binary.BigEndian only supports Uint16 not Uint8. Still saves us quite a few bytes. One wasted byte when encoding Uint8s as Uint16s. IMO fine for now since binary.BigEndian handles all the heavy lifting for us.

lib/block_view_validator.go
```diff
@@ -1412,13 +1482,13 @@ func (bav *UtxoView) GetTopValidatorsByStake(limit int) ([]*ValidatorEntry, erro
...
 	}
 	// Pull top N ValidatorEntries from the database (not present in the UtxoView).
-	validatorEntries, err := DBGetTopValidatorsByStake(bav.Handle, bav.Snapshot, limit, utxoViewValidatorEntries)
+	validatorEntries, err := DBGetTopActiveValidatorsByStake(bav.Handle, bav.Snapshot, limit, utxoViewValidatorEntries)
 	if err != nil {
-		return nil, errors.Wrapf(err, "UtxoView.GetTopValidatorsByStake: error retrieving entries from db: ")
+		return nil, errors.Wrapf(err, "UtxoView.GetTopActiveValidatorsByStake: error retrieving entries from db: ")
 	}
-	// Add !isDeleted ValidatorEntries from the UtxoView to the ValidatorEntries from the db.
+	// Add !isDeleted, active ValidatorEntries from the UtxoView to the ValidatorEntries from the db.
```
lazynina, line 1489: should we merge validatorEntries into the view?

lib/block_view_validator_test.go
```diff
@@ -924,65 +933,127 @@ func TestGetTopValidatorsByStakeMergingDbAndUtxoView(t *testing.T) {
...
+		utxoView.ValidatorMapKeyToValidatorEntry[m2ValidatorEntry.ToMapKey()].TotalStakeAmountNanos,
+		uint256.NewInt().SetUint64(300),
 	)
 
-	// Store m2's ValidatorEntry in the UtxoView.
-	m2ValidatorEntry := &ValidatorEntry{
-		ValidatorPKID:         m2PKID,
+	// Store m3's ValidatorEntry in the UtxoView with isDeleted=true.
+	m3ValidatorEntry.isDeleted = true
+	utxoView._setValidatorEntryMappings(m3ValidatorEntry)
```
lazynina, line 996: should we call `_deleteValidatorEntryMappings` instead?
```suggestion
	utxoView._deleteValidatorEntryMappings(m3ValidatorEntry)
```

lib/db_utils.go
mattfoley8: I figured out EncodeUint8. It was easy enough.

lazynina (approved):

PR #520, opened 2023-05-04
Mf/add unjail validator txn type

mattfoley8:


lib/block_view_validator.go
```diff
@@ -506,6 +528,41 @@ func (txindexMetadata *UnregisterAsValidatorTxindexMetadata) GetEncoderType() En
...
+// TYPES: UnjailValidatorTxindexMetadata
+//
+
+type UnjailValidatorTxindexMetadata struct {
+	ValidatorPublicKeyBase58Check string
+}
+
+func (txindexMetadata *UnjailValidatorTxindexMetadata) RawEncodeWithoutMetadata(blockHeight uint64, skipMetadata ...bool) []byte {
+	var data []byte
+	data = append(data, EncodeByteArray([]byte(txindexMetadata.ValidatorPublicKeyBase58Check))...)
```
lazynina, line 541: Isn't this a little unnecessary? won't it always just be the transactor? maybe we don't really need anything in here. idk

lib/block_view_validator.go
```diff
@@ -1396,13 +1682,48 @@ func (bav *UtxoView) IsValidUnregisterAsValidatorMetadata(transactorPublicKey []
...
 	}
-	if validatorEntry == nil {
+	if validatorEntry == nil || validatorEntry.isDeleted {
 		return errors.Wrapf(RuleErrorValidatorNotFound, "UtxoView.IsValidUnregisterAsValidatorMetadata: ")
 	}
 
 	return nil
 }
 
+func (bav *UtxoView) IsValidUnjailValidatorMetadata(transactorPublicKey []byte, metadata *UnjailValidatorMetadata) error {
```
lazynina, line 1692: we don't really need to pass in metadata here since it's just an empty struct.

lib/block_view_validator.go
```diff
@@ -1327,6 +1460,159 @@ func (bav *UtxoView) _disconnectUnregisterAsValidator(
...
+		return 0, 0, nil, errors.Wrapf(err, "_connectUnjailValidator: ")
+	}
+	if verifySignatures {
+		// _connectBasicTransfer has already checked that the txn is signed
+		// by the top-level public key, which we take to be the sender's
+		// public key so there is no need to verify anything further.
+	}
+
+	// Grab the txn metadata.
+	txMeta := txn.TxnMeta.(*UnjailValidatorMetadata)
```
lazynina, line 1501: we don't really need to do this since we never use the metadata as it's an empty struct.

lazynina (approved):

PR #521, opened 2023-05-05
Mf/add bls signature utils

mattfoley8:
Goes with https://github.com/deso-protocol/backend/pull/497.


bls/signature.go
```diff
@@ -0,0 +1,237 @@
...
+}
+
+func (publicKey *PublicKey) Eq(other *PublicKey) bool {
+	if publicKey.flowPublicKey == nil || other == nil {
+		return false
+	}
+	return publicKey.flowPublicKey.Equals(other.flowPublicKey)
+}
+
+func (publicKey *PublicKey) Copy() *PublicKey {
```
lazynina, line 170: not sure about this copy function, can we add a test to make sure we're not copying any pointers under the hood?

test.Dockerfile
```diff
@@ -2,7 +2,10 @@ FROM alpine:latest AS core
 
 RUN apk update
 RUN apk upgrade
-RUN apk add --update go gcc g++ vips vips-dev
+RUN apk add --update bash cmake git gcc g++ make vips vips-dev
+
+COPY --from=golang:1.20-alpine /usr/local/go/ /usr/local/go/
```
lazynina, line 7: how does this work?

.gitignore
```diff
@@ -4,6 +4,7 @@ quick_testing
 dump_txns
 compare_blocks.go
 backend
+core
```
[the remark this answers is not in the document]
mattfoley8: If you run `go build .` it creates an artifact called `core` that can get committed into git. This doesn't ignore the parent dir. But does ignore that build artifact.

test.Dockerfile
mattfoley8: It seemed like the easiest solution among these: https://stackoverflow.com/questions/52056387/how-to-install-go-in-alpine-linux. I also like that we can lock the go version using the docker tag there.

lazynina (approved):

PR #522, opened 2023-05-08
Mf/convert validator voting keys to bls keys

mattfoley8:


lib/block_view_validator.go
```diff
@@ -114,9 +114,23 @@ func (validatorEntry *ValidatorEntry) RawEncodeWithoutMetadata(blockHeight uint6
 		data = append(data, EncodeByteArray(domain)...)
 	}
 
+	// VotingPublicKey
+	var votingPublicKeyBytes []byte
+	if validatorEntry.VotingPublicKey != nil {
```
lazynina, line 119: is a nil voting public key valid?

lib/block_view_validator.go
```diff
@@ -114,9 +114,23 @@ func (validatorEntry *ValidatorEntry) RawEncodeWithoutMetadata(blockHeight uint6
...
+		votingPublicKeyBytes = validatorEntry.VotingPublicKey.ToBytes()
```
lazynina, line 120: should we implement the DeSoEncoder interface for BLS stuff? It would simplify some of these encode/decode methods

lib/constants.go
```diff
@@ -1218,6 +1229,11 @@ var DeSoTestnetParams = DeSoParams{
 	// Jailed validators can be unjailed after a minimum of N elapsed epochs.
 	ValidatorJailEpochDuration: uint64(3),
 
+	// When registering, a validator must include a VotingSignatureBlockHeight within CurrentBlockHeight
+	// to CurrentBlockHeight + ValidatorVotingSignatureBlockHeightWindow blocks. This is to prevent
+	// validator registration replay attacks.
+	ValidatorVotingSignatureBlockHeightWindow: uint64(60),
```
lazynina, line 1235: not sure if this value is exactly what we want, but I'm ok w/ it for now.

lib/block_view_validator_test.go
```diff
@@ -144,11 +151,130 @@ func _testValidatorRegistration(t *testing.T, flushToDB bool) {
...
+		}
+		_, err = _submitRegisterAsValidatorTxn(
+			testMeta, m0Pub, m0Priv, registerMetadata, nil, flushToDB,
+		)
+		require.Error(t, err)
+		require.Contains(t, err.Error(), RuleErrorValidatorInvalidVotingPublicKeySignature)
+	}
+	{
+		// RuleErrorValidatorInvalidVotingPublicKeySignature: invalid VotingSignatureBlockHeight
+		votingPublicKey, votingSignature := _generateVotingPublicKeyAndSignature(t, m0PkBytes, blockHeight)
```
lazynina, line 256: Just a question - do I need to include a voting pub key & sig when I'm updating my validator entry or ONLY when registering? if I just want to update my domains, is it okay for me to exclude this?

lib/block_view_validator.go
mattfoley8: I can't define the DeSoEncoder functions on the `bls.*` structs in lib and IMO it breaks encapsulation (and/or creates a circular dependency) to do it in the `bls/` package. For that reason, I refactored these into `Encode` and `Decode` helpers instead which covers the nil check.

lib/constants.go
mattfoley8: Yeah, I don't know exactly what value should be here, and eventually this will move to a GlobalParamsEntry, but I initially just estimated 60.

mattfoley8: I updated it to 300.

lib/block_view_validator_test.go
mattfoley8: That's correct. You need to provide Domains, Voting Public Key, Voting Signature, and Voting Block Height every time you call Register as Validator regardless if you are creating or updating a ValidatorEntry. It's like a stateless upsert.

lazynina (approved):

PR #524, opened 2023-05-10
Feature/pos txn types

mattfoley8:


lib/block_view_types_test.go
```diff
@@ -1,9 +1,12 @@
+//go:build relic
```
lazynina, line 1: do we need this tag in here?

lib/network.go
```diff
@@ -6099,11 +6400,23 @@ func (tsl *TransactionSpendingLimit) Copy() *TransactionSpendingLimit {
 		copyTSL.AccessGroupMemberMap[accessGroupMemberLimitKey] = accessGroupMemberCount
 	}
 
+	for stakeLimitKey, stakeLimitDESONanos := range tsl.StakeLimitMap {
+		copyTSL.StakeLimitMap[stakeLimitKey] = stakeLimitDESONanos
+	}
+
+	for stakeLimitKey, unstakeLimitDESONanos := range tsl.UnstakeLimitMap {
+		copyTSL.UnstakeLimitMap[stakeLimitKey] = unstakeLimitDESONanos
```
lazynina, line 6408: We should copy the uint256s - we don't need to copy the uint64s since they aren't pointers.
```suggestion
		copyTSL.StakeLimitMap[stakeLimitKey] = stakeLimitDESONanos.Clone()
	}

	for stakeLimitKey, unstakeLimitDESONanos := range tsl.UnstakeLimitMap {
		copyTSL.UnstakeLimitMap[stakeLimitKey] = unstakeLimitDESONanos.Clone()
```

lib/block_view_validator.go
```diff
@@ -0,0 +1,2137 @@
...
+		return errors.Wrapf(err, "ValidatorEntry.Decode: Problem reading ValidatorID: ")
+	}
+
+	// ValidatorPKID
+	validatorPKID := &PKID{}
+	if exist, err := DecodeFromBytes(validatorPKID, rr); exist && err == nil {
+		validatorEntry.ValidatorPKID = validatorPKID
+	} else if err != nil {
+		return errors.Wrapf(err, "ValidatorEntry.Decode: Problem reading ValidatorPKID: ")
+	}
```
lazynina, line 147: Let's use the DecodeDeSoEncoders to clean this function up a bit.
```suggestion
	// ValidatorID
	validatorEntry.ValidatorID, err = DecodeDeSoEncoder(&BlockHash{}, rr)
	if err != nil {
		return errors.Wrapf(err, "ValidatorEntry.Decode: Problem reading ValidatorID: ")
	}

	// ValidatorPKID
	validatorEntry.ValidatorPKID, err = DecodeDeSoEncoder(&PKID{}, rr)
	if err != nil {
		return errors.Wrapf(err, "ValidatorEntry.Decode: Problem reading ValidatorPKID: ")
	}
```

lib/block_view_stake.go
```diff
@@ -0,0 +1,2600 @@
...
+	data = append(data, EncodeByteArray(txnData.ValidatorPublicKey.ToBytes())...)
+	data = append(data, EncodeUint256(txnData.StakeAmountNanos)...)
+	return data, nil
+}
+
+func (txnData *StakeMetadata) FromBytes(data []byte) error {
+	rr := bytes.NewReader(data)
+
+	// ValidatorPublicKey
+	validatorPublicKeyBytes, err := DecodeByteArray(rr)
```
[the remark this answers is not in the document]
mattfoley8: Gonna punt for now. A good idea we should definitely do. But IMO enough moving parts with this PR already. Don't wanna add that sized refactor before DH's pass.

bls/signature.go
```diff
@@ -0,0 +1,243 @@
...
+	"errors"
+	flowCrypto "github.com/onflow/flow-go/crypto"
+	"strings"
+)
+
+const SigningAlgorithm = flowCrypto.BLSBLS12381
+
+// TODO: what should the domainTag param be?
+var HashingAlgorithm = flowCrypto.NewExpandMsgXOFKMAC128("deso-protocol")
+
```
diamondhands0, line 18: Leave detailed comments for all of this stuff please.

diamondhands0: In your comment, explain how large the resulting aggregated signature ends up being. For example, if I have N signatures that each signed a distinct payload, is the resulting aggregated signature larger than if the signatures all signed the same payload?

diamondhands0: Hmm... reading the other code, it seems unclear as to whether the signatures all needed to sign the same payload or whether they all can sign distinct payloads. Comment would help here.

diamondhands0: Ok, I see now that this libary is designed to support only the verification of a single payload, which makes sense. Still let's leave a lot of comments here so it's all more clear.

bls/signature.go
```diff
@@ -0,0 +1,243 @@
...
+		flowPublicKeys = append(flowPublicKeys, publicKey.flowPublicKey)
+	}
+	return flowCrypto.VerifyBLSSignatureOneMessage(flowPublicKeys, signature.flowSignature, payloadBytes, HashingAlgorithm)
+}
+
+//
+// TYPES: PrivateKey
+//
+
+type PrivateKey struct {
```
diamondhands0, line 43: Why did we make all of these structs as opposed to doing something like `type PrivateKey floCrypto.PrivateKey` or something? I think it's fine to leave it as-is, just wondering.

Also Maybe we should call these BLSPrivateKey, BLSPublicKey, BLSSignature, etc...? Up to you. Leave comments either way explaining everything.

bls/signature.go
```diff
@@ -0,0 +1,243 @@
...
+	"encoding/hex"
+	"errors"
+	flowCrypto "github.com/onflow/flow-go/crypto"
+	"strings"
+)
+
+const SigningAlgorithm = flowCrypto.BLSBLS12381
+
+// TODO: what should the domainTag param be?
+var HashingAlgorithm = flowCrypto.NewExpandMsgXOFKMAC128("deso-protocol")
```
diamondhands0, line 17: It would be nice to have a high-level explanation up here of how the hasher fits into all this. Even if it's really simple like "The hashing algorithm will be used in aggregating and verifying signatures. It needs a distinct string because XYZ" would be nice.

lib/block_view_validator.go
```diff
@@ -0,0 +1,2133 @@
...
+	ValidatorStatusJailed  ValidatorStatus = 2
+)
+
+type ValidatorMapKey struct {
+	// The MapKey has to contain all fields that are used in Badger keys.
+	// Otherwise, an update to the UtxoView will not be able to update or
+	// delete all relevant Badger rows.
+	ValidatorPKID           PKID
+	TotalStakeAmountNanos   uint256.Int
+	RegisteredAtBlockHeight uint64
```
diamondhands0, line 69: Note to self: It's a bit weird that RegisteredAt is a block height but JailedAt is an epoch. Feels like one should shift to the other.

lib/db_utils.go
```diff
@@ -476,8 +476,37 @@ type DBPrefixes struct {
...
 
+	// PrefixValidatorByStake: Retrieve the top N validators by stake.
+	// Prefix, TotalStakeAmountNanos, MaxUint64 - RegisteredAtBlockHeight, ValidatorPKID -> ValidatorPKID
+	// FIXME: @DH, should we duplicate the ValidatorPKID in the key and the value?
+	// Alternatively, we could just store and parse the ValidatorPKID from the key
+	// and store a struct{} as the value. That saves on space, but makes retrieving
+	// the ValidatorPKID from the key bytes more complex than just reading the value
+	// bytes directly since the key includes other preceding fields. Interesting
+	// trade-off. Curious your opinion.
+	PrefixValidatorByStake []byte `prefix_id:"[79]" is_state:"true"`
```
diamondhands0, line 491: In this case I would only have the ValidatorPKID in the key, and have the value be empty. The reason is that duplicating it like this can lead to bugs, and so we generally prefer to normalize wherever it's convenient to do so, as it seems to be here.

A case where I would say it's not worth normalizing is if you had an Entry as the value, since in that case having the value in the entry is useful in other places, and omitting it would be annoying. Since here you don't have that, I'd keep it only in the key.

lib/db_utils.go
```diff
@@ -476,8 +476,37 @@ type DBPrefixes struct {
...
+
+	// PrefixGlobalStakeAmountNanos: Retrieve the cumulative stake across all validators.
+	// Prefix -> *uint256.Int
+	PrefixGlobalStakeAmountNanos []byte `prefix_id:"[80]" is_state:"true"`
+
+	// PrefixStakeByValidatorByStaker: Retrieve a StakeEntry.
+	// Prefix, ValidatorPKID, StakerPKID -> StakeEntry
+	PrefixStakeByValidatorByStaker []byte `prefix_id:"[81]" is_state:"true"`
```
diamondhands0, line 499: Nitpick: We normally use "By" to denote -> and "And" to denote (A, B, C). So in this case, since you have (ValidatorPKID, StakerPKID) -> StakeEntry, I would call it `PrefixStakeByValidatorAndStaker`.

I would make the same change to the other things.

diamondhands0: I fixed it for you btw. Just look at how I did it for future reference and resolve.

lib/block_view_validator.go
```diff
@@ -0,0 +1,2133 @@
...
+const (
+	ValidatorStatusInvalid ValidatorStatus = 0
+	ValidatorStatusActive  ValidatorStatus = 1
+	ValidatorStatusJailed  ValidatorStatus = 2
+)
+
+type ValidatorMapKey struct {
+	// The MapKey has to contain all fields that are used in Badger keys.
+	// Otherwise, an update to the UtxoView will not be able to update or
+	// delete all relevant Badger rows.
```
diamondhands0, line 66: Ok so I don't think you need to change this necessarily, but I want you to noodle on sometihng.

For PKIDToProfileEntry, we map a PKID->ProfileEntry in the view, even though there are a lot of ancillary indexes.

The reason this works is we do the following trick:
* In the view, if we update a profile, then we'll have PKID->NewProfile.
* When we flush, we call this interesting function called DBDeleteProfileEntryMappingsWithTxn(pkid *PKID)
* What that function does is it first LOOKS UP the ProfileEntry *in the database* that matches the PKID, and then it *uses the DB ProfileEntry* to kill all the mappings in the DB. This works because the ProfileEntry in the DB has all the old values, which tells you how to delete it.
* Then, only after all the mappings are nuked, we call DBPutProfileEntryMappings with the new PKID.

I haven't gone deep enough to know if this would work here, but the way you would do it is you would have a DBDeleteValidatorEntryMappings(validatorPKID *PKID), which you call in a view flush before writing the new ValidatorEntry in the view.

diamondhands0: Btw the benefit of this is that the view code gets a lot simpler. You're effectively "pushing down" the complexity into the DB layer, which is really where you ideally want it (as low as possible).

diamondhands0: I fixed this for you and fixed the tests for fun. It wasn't hard, and I think it makes the view a lot cleaner.

Btw you did a great job with the tests. They caught some good stuff when I was making the change :)

lib/block_view.go
```diff
@@ -113,6 +114,21 @@ type UtxoView struct {
...
+	// Validator mappings
+	ValidatorMapKeyToValidatorEntry map[ValidatorMapKey]*ValidatorEntry
+
+	// Global stake across validators
+	GlobalStakeAmountNanos *uint256.Int
+
+	// Stake mappings
+	StakeMapKeyToStakeEntry map[StakeMapKey]*StakeEntry
+
+	// Locked stake mappings
```
diamondhands0, line 126: Add a comment here and/or in db_utils. The concept of locked stake is a bit tricky.

lib/block_view.go
```diff
@@ -3276,6 +3400,29 @@ func (bav *UtxoView) _connectTransaction(txn *MsgDeSoTxn, txHash *BlockHash,
...
+			for _, prevLockedStakeEntry := range utxoOp.PrevLockedStakeEntries {
+				totalLockedAmountNanos, err = SafeUint256().Add(
+					totalLockedAmountNanos, prevLockedStakeEntry.LockedAmountNanos,
+				)
+				if err != nil {
+					return nil, 0, 0, 0, errors.Wrapf(err, "ConnectTransaction: error computing TotalLockedAmountNanos: ")
+				}
+			}
+			desoLockedDelta = big.NewInt(0).Neg(totalLockedAmountNanos.ToBig())
+		}
```
diamondhands0, line 3425: Note to self: Does this sanity check require more than just UnlockStake? Why doesn't staking change the desoLockedDelta? How is this being accounted for.

lib/block_view_flush.go
```diff
@@ -140,6 +140,24 @@ func (bav *UtxoView) FlushToDbWithTxn(txn *badger.Txn, blockHeight uint64) error
...
+	}
+	if err := bav._flushStakeEntriesToDbWithTxn(txn, blockHeight); err != nil {
+		return err
+	}
+	if err := bav._flushLockedStakeEntriesToDbWithTxn(txn, blockHeight); err != nil {
+		return err
+	}
+	// TODO: We may want to move this into a new FlushToDb function that only flushes
+	// entries set in the OnEpochEndHook. No sense in wasting a bunch of cycles flushing
+	// all the other entries which will always be nil/empty in the OnEpochEndHook.
```
diamondhands0, line 157: Note to self: What would be the point of this? I think the flush is just a noop if we're not at the block height.

lib/block_view_validator.go
```diff
@@ -0,0 +1,2133 @@
...
+) error {
+	if validatorEntry == nil {
+		// This should never happen but is a sanity check.
+		glog.Errorf("DBPutValidatorWithTxn: called with nil ValidatorEntry")
+		return nil
+	}
+
+	// Set ValidatorEntry in PrefixValidatorByPKID.
+	key := DBKeyForValidatorByPKID(validatorEntry)
+	if err := DBSetWithTxn(txn, snap, key, EncodeToBytes(blockHeight, validatorEntry)); err != nil {
```
diamondhands0, line 696: FIXME: I don't think we should be using this block height for anything related to consensus. This is just the blockheight at which the values were flushed, which can vary depending on *when* the flush was called. Right now, if we connect 2 blocks and flush, we'll have a different value than if we connected one at a time, which isn't the case for anything else in our code as of today.

lib/block_view_validator.go
```diff
@@ -0,0 +1,2133 @@
...
+		return errors.Wrapf(
+			err, "DBPutValidatorWithTxn: problem storing ValidatorEntry in index PrefixValidatorByPKID",
+		)
+	}
+
+	// Set ValidatorEntry.PKID in PrefixValidatorByStake.
+	key = DBKeyForValidatorByStake(validatorEntry)
+	if err := DBSetWithTxn(txn, snap, key, EncodeToBytes(blockHeight, validatorEntry.ValidatorPKID)); err != nil {
```
diamondhands0, line 704: Same comment as previous.

lib/block_view_validator.go
diamondhands0: Nevermind. I see it's OK because it's just an encoder.

lib/block_view_stake.go
```diff
@@ -0,0 +1,2600 @@
...
+	"github.com/pkg/errors"
+	"sort"
+)
+
+//
+// TYPES: StakeEntry
+//
+
+type StakeEntry struct {
+	StakeID          *BlockHash
```
diamondhands0, line 18: I don't think we need StakeID for anything. A StakeEntry is uniquely identified by a (ValidatorPKID, StakerPKID) pair. Feels like including this StakeID just increases risk for bugs. Going to try and delete it and see what happens.

diamondhands0: I deleted it, and also deleted LockedStakeID, and it seems like a very positive change. In particular, in the case of LockedStakeID, there was a place where we were confusingly setting LockedStakeID to prevStakeEntry.StakeID, which was a bit weird. Overall, the fact that we can delete this without any negative repercussions, and a slight benefit in that one place, seems to indicate that the information is unnecessarily redundant, since a (ValidatorPKID, StakerPKID) pair uniquely identifies a StakeEntry, and since a (ValidatorPKID, StakerPKID, EpochNumber) uniquely identifies a LockedStakeEntry.

lib/block_view_stake.go
```diff
@@ -0,0 +1,2600 @@
...
+
+//
+// TYPES: LockedStakeEntry
+//
+
+type LockedStakeEntry struct {
+	LockedStakeID       *BlockHash
+	StakerPKID          *PKID
+	ValidatorPKID       *PKID
+	LockedAmountNanos   *uint256.Int
```
diamondhands0, line 115: I like that we're starting to use uint256'es everywhere. Will be good to eventually convert everything to this...

lib/block_view_stake.go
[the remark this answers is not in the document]
diamondhands0: +1 mattfoley8

lib/block_view_stake.go
```diff
@@ -0,0 +1,2600 @@
...
+
+	// Convert ValidatorPublicKey to ValidatorPKID.
+	validatorPKIDEntry := bav.GetPKIDForPublicKey(txMeta.ValidatorPublicKey.ToBytes())
+	if validatorPKIDEntry == nil || validatorPKIDEntry.isDeleted {
+		return 0, 0, nil, errors.Wrapf(RuleErrorInvalidValidatorPKID, "_connectUnlockStake: ")
+	}
+
+	// Retrieve the PrevLockedStakeEntries. These will be restored if we disconnect this txn.
+	prevLockedStakeEntries, err := bav.GetLockedStakeEntriesInRange(
+		validatorPKIDEntry.PKID, transactorPKIDEntry.PKID, txMeta.StartEpochNumber, txMeta.EndEpochNumber,
```
diamondhands0, line 1646: So... I might be missing something. But don't you need to verify that the prevLockedStakeEntries are actually allowed to be UNlocked? Right now, it looks like I can unstake and then immmediately unlock that stake right after, without a cooldown.

In addition, as I was reading this, it felt like UnlockStake should just unlock all the stake that is allowed to be unlocked. So you can get rid of StartEpochNumber and EndEpochNumber. Instead, GetLockedStakeENtriesInRange can just be GetUnlockableLockedStakeEntries, and it can search for LockedStakeEntries that have (LockedAtEpochNumber < CurrentEpochNumber+CooldownEpochs).

lib/block_view_stake.go
```diff
@@ -0,0 +1,2600 @@
...
+}
+
+//
+// TYPES: UnlockStakeMetadata
+//
+
+type UnlockStakeMetadata struct {
+	ValidatorPublicKey *PublicKey
+	StartEpochNumber   uint64
+	EndEpochNumber     uint64
```
diamondhands0, line 305: See my comment in the _connect. I think we can simplify and just have this function unlock ALL unlockable stake when you call it. Hard to imagine someone wanting anything other than this.

lib/block_view_stake.go
```diff
@@ -0,0 +1,2600 @@
...
+	lockedStakeEntry, err := DecodeDeSoEncoder(&LockedStakeEntry{}, rr)
+	if err != nil {
+		return nil, errors.Wrapf(
+			err, "DBGetLockedStakeEntry: problem decoding LockedStakeEntry: ",
+		)
+	}
+	return lockedStakeEntry, nil
+}
+
+func DBGetLockedStakeEntriesInRange(
```
diamondhands0, line 667: Would delete these if you decide to make _connectUnlock just unlock whatever the user is capable of unlocking.

lib/block_view_stake.go
```diff
@@ -0,0 +1,2600 @@
...
+	validatorEntry, err := bav.GetValidatorByPublicKey(metadata.ValidatorPublicKey)
+	if err != nil {
+		return errors.Wrapf(err, "UtxoView.IsValidUnlockStakeMetadata: ")
+	}
+	if validatorEntry == nil || validatorEntry.isDeleted {
+		return errors.Wrapf(RuleErrorInvalidValidatorPKID, "UtxoView.IsValidUnlockStakeMetadata: ")
+	}
+
+	// Validate StartEpochNumber and EndEpochNumber.
+	if metadata.StartEpochNumber > metadata.EndEpochNumber {
```
diamondhands0, line 1875: See other comments. I'd get rid of StartEpochNumber and EndEpochNumber and just automatically unlock any LockedStakeEntry that has `lockedStakeEntry.LockEpochNumber+bav.Params.StakeLockupEpochDuration > currentEpochNumber`

lib/block_view_stake.go
diamondhands0: Found it in the IsValidUnlockStake thing.

diamondhands0, line 1646: I think you could acutally just call this function with `(validatorPKID, transactorPKID, 0, currentEpochNumber-StakeLockupEpochDuration)`. Notice that when you set the `start = 0` and the `end = currentEpochNumner - StakeLockupEpochDuration`, I think you organically get all the LockupEntries that you can unlock right now.

lib/block_view_stake.go
```diff
@@ -0,0 +1,2600 @@
...
+		return 0, 0, nil, errors.Wrapf(RuleErrorInvalidUnlockStakeUnlockableStakeOverflowsUint64, "_connectUnlockStake: ")
+	}
+	totalUnlockedAmountNanosUint64 := totalUnlockedAmountNanos.Uint64()
+
+	// Add TotalUnlockedAmountNanos to TotalInput. The unlocked coins are an
+	// implicit input even though they do not come from a specific public key.
+	totalInput, err = SafeUint64().Add(totalInput, totalUnlockedAmountNanosUint64)
+	if err != nil {
+		return 0, 0, nil, errors.Wrapf(err, "_connectUnlockStake: error adding TotalUnlockedAmountNanos to TotalInput: ")
+	}
```
diamondhands0, line 1690: As a side-note, in Rosetta we're going to have to ues the "sub-account" thing like we did with Creator Coins to handle this.

lib/block_view_types.go
```diff
@@ -4804,6 +4912,11 @@ func DecodeMapStringUint64(rr *bytes.Reader) (map[string]uint64, error) {
 	return nil, nil
 }
 
+// EncodeUint256 is useful for space-efficient encoding of uint256s.
+// It does not guarantee fixed-width encoding, so should not be used
+// in BadgerDB keys. Use EncodeOptionalUint256 instead, which does
+// guarantee fixed-width encoding. Both EncodeUint256 and
+// EncodeOptionalUint256 can handle nil inputs.
 func EncodeUint256(number *uint256.Int) []byte {
```
diamondhands0, line 4920: This is confusing because every other EncodeUintX function is fixed-width, which means we're guaranteed someone is going to produce a bug where they assumed it's fixed-width. I changed the name to VariableEncodeUint256 (and decode).

lib/constants.go
```diff
@@ -581,6 +596,22 @@ type DeSoParams struct {
...
+	// ValidatorJailEpochDuration is the number of epochs that a validator must
+	// wait after being jailed before submitting an UnjailValidator txn.
+	// TODO: Move this to GlobalParamsEntry.
+	ValidatorJailEpochDuration uint64
+
+	// When registering, a validator must include a VotingSignatureBlockHeight within CurrentBlockHeight
+	// to CurrentBlockHeight + ValidatorVotingSignatureBlockHeightWindow blocks. This is to prevent
+	// validator registration replay attacks.
+	// TODO: Move this to GlobalParamsEntry.
+	ValidatorVotingSignatureBlockHeightWindow uint64
```
diamondhands0, line 613: I don't think this is needed since the transaction will natively include a nonce with an expiration block height right? I think you can delete this. If they want to prevent replay, they can just set a smaller ExpirationBlockHeight in their nonce.

lib/constants.go
```diff
@@ -949,6 +984,17 @@ var DeSoMainnetParams = DeSoParams{
 	// reserve ratios.
 	CreatorCoinAutoSellThresholdNanos: uint64(10),
 
+	// Unstaked stake can be unlocked after a minimum of N elapsed epochs.
+	StakeLockupEpochDuration: uint64(3),
+
+	// Jailed validators can be unjailed after a minimum of N elapsed epochs.
+	ValidatorJailEpochDuration: uint64(3),
```
diamondhands0, line 991: Do we want 3 or 2? Feels like 2 could work?

lib/pos_epoch.go
```diff
@@ -0,0 +1,170 @@
...
+//
+
+func (bav *UtxoView) GetCurrentEpochEntry() (*EpochEntry, error) {
+	var epochEntry *EpochEntry
+	var err error
+
+	// First, check the UtxoView.
+	epochEntry = bav.CurrentEpochEntry
+	if epochEntry != nil {
+		return epochEntry.Copy(), nil
```
diamondhands0, line 70: I think normally we don't Copy() before returning from the UtxoView Getters. I don't think it matters, but probably good to keep the same convention as everywhere else.

lib/pos_epoch.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	// If not found, check the database.
+	epochEntry, err = DBGetCurrentEpochEntry(bav.Handle, bav.Snapshot)
+	if err != nil {
+		return nil, errors.Wrapf(err, "UtxoView.GetCurrentEpoch: problem retrieving EpochEntry from db: ")
+	}
+	if epochEntry != nil {
+		// Cache in the UtxoView.
+		bav.CurrentEpochEntry = epochEntry.Copy()
```
diamondhands0, line 80: Same here. I think we'd normally return the actual value that's set in the view.

lib/block_view_validator.go
```diff
@@ -0,0 +1,2133 @@
...
+	"net/url"
+	"sort"
+)
+
+//
+// TYPES: ValidatorEntry
+//
+
+type ValidatorEntry struct {
+	ValidatorID   *BlockHash
```
diamondhands0, line 23: Don't think we need this. Going to delete.

lib/block_view_validator.go
```diff
@@ -0,0 +1,2133 @@
...
+	ValidatorPKID *PKID
+	// Note: if someone is updating their ValidatorEntry, they need to include
```
diamondhands0, line 25: All of these fields need good comments.

lib/block_view_validator.go
```diff
@@ -0,0 +1,2133 @@
...
+	// all domains. The Domains field is not appended to. It is overwritten.
+	Domains                    [][]byte
+	DisableDelegatedStake      bool
+	VotingPublicKey            *bls.PublicKey
+	VotingPublicKeySignature   *bls.Signature
+	VotingSignatureBlockHeight uint64
```
diamondhands0, line 31: I think the nonce obviates the need for this field. Going to delete.

lib/block_view_validator.go
```diff
@@ -0,0 +1,2133 @@
...
+//
+// TYPES: RegisterAsValidatorMetadata
+//
+
+type RegisterAsValidatorMetadata struct {
+	Domains                    [][]byte
+	DisableDelegatedStake      bool
+	VotingPublicKey            *bls.PublicKey
+	VotingPublicKeySignature   *bls.Signature
+	VotingSignatureBlockHeight uint64
```
diamondhands0, line 232: Don't think this field is needed because of the nonce. Going to delete.

lib/block_view_validator.go
```diff
@@ -0,0 +1,2133 @@
...
+	return key
+}
+
+func DBKeyForValidatorByStake(validatorEntry *ValidatorEntry) []byte {
+	key := append([]byte{}, Prefixes.PrefixValidatorByStake...)
+	key = append(key, EncodeUint8(uint8(validatorEntry.Status()))...)
+	// TotalStakeAmountNanos will never be nil here, but EncodeOptionalUint256
+	// is used because it provides a fixed-width encoding of uint256.Ints.
+	key = append(key, EncodeOptionalUint256(validatorEntry.TotalStakeAmountNanos)...)         // Highest stake first
+	key = append(key, EncodeUint64(math.MaxUint64-validatorEntry.RegisteredAtBlockHeight)...) // Oldest first
```
diamondhands0, line 567: Do we need this? I still haven't seen a good reason to include the RegisteredAtBlockHeight in the index key.

lib/block_view_validator.go
```diff
@@ -0,0 +1,2133 @@
...
+const RuleErrorValidatorNotFound RuleError = "RuleErrorValidatorNotFound"
+const RuleErrorValidatorMissingVotingPublicKey RuleError = "RuleErrorValidatorMissingVotingPublicKey"
+const RuleErrorValidatorMissingVotingPublicKeySignature RuleError = "RuleErrorValidatorMissingVotingPublicKeySignature"
+const RuleErrorValidatorInvalidVotingPublicKeySignature RuleError = "RuleErrorValidatorInvalidVotingPublicKeySignature"
+const RuleErrorValidatorInvalidVotingSignatureBlockHeight RuleError = "RuleErrorValidatorInvalidVotingSignatureBlockHeight"
+const RuleErrorValidatorDisablingExistingDelegatedStakers RuleError = "RuleErrorValidatorDisablingExistingDelegatedStakers"
+const RuleErrorUnjailingNonjailedValidator RuleError = "RuleErrorUnjailingNonjailedValidator"
+const RuleErrorUnjailingValidatorTooEarly RuleError = "RuleErrorUnjailingValidatorTooEarly"
+
+const MaxValidatorNumDomains int = 12
```
diamondhands0, line 2133: Hmm... I'm not sure we need a protocol-level limit on number of domains. Is there a downside to just making thi sunlimited?

lib/constants.go
```diff
@@ -949,6 +984,17 @@ var DeSoMainnetParams = DeSoParams{
...
+
+	// When registering, a validator must include a VotingSignatureBlockHeight within CurrentBlockHeight
+	// to CurrentBlockHeight + ValidatorVotingSignatureBlockHeightWindow blocks. This is to prevent
+	// validator registration replay attacks.
+	ValidatorVotingSignatureBlockHeightWindow: uint64(300),
```
diamondhands0, line 996: Wonder if we should make this value larger. When we shift to PoS we'll be doing 1 block per second so feels like 3600 might be better. On the fence, though, because 300s is also not bad.

diamondhands0: I think leave it as-is for now.

lib/block_view_validator.go
```diff
@@ -0,0 +1,2133 @@
...
+	if err != nil {
+		return 0, 0, nil, errors.Wrapf(err, "_connectUnregisterAsValidator: ")
+	}
+	if prevValidatorEntry == nil || prevValidatorEntry.isDeleted {
+		return 0, 0, nil, errors.Wrapf(RuleErrorValidatorNotFound, "_connectUnregisterAsValidator: ")
+	}
+	bav._deleteValidatorEntryMappings(prevValidatorEntry)
+
+	// Sanity check that TotalUnstakedAmountNanos == PrevValidatorEntry.TotalStakedAmountNanos.
+	if !totalUnstakedAmountNanos.Eq(prevValidatorEntry.TotalStakeAmountNanos) {
```
diamondhands0, line 1331: Note to self: Consider running a sanity-check like this more regularly somehow.

lib/block_view_validator.go
```diff
@@ -0,0 +1,2133 @@
...
+	}
+
+	// Retrieve CurrentEpochNumber.
+	currentEpochNumber, err := bav.GetCurrentEpochNumber()
+	if err != nil {
+		return errors.Wrapf(err, "UtxoView.IsValidUnjailValidatorMetadata: error retrieving CurrentEpochNumber: ")
+	}
+
+	// Validate sufficient epochs have elapsed for validator to be unjailed.
+	if validatorEntry.JailedAtEpochNumber+bav.Params.ValidatorJailEpochDuration > currentEpochNumber {
```
diamondhands0, line 1728: I wonder if we want to make this UnjailEpochNumber rather than JailedAt. This would allow us the flexibility to jail validators for different amounts of time depending on what they did. Not sure if we want to make this change now or later. Maybe fine to leave as-is and revisit if we have scenarios where we want to jail differently.

lib/block_view_validator.go
```diff
@@ -0,0 +1,2133 @@
...
+		return 0, 0, nil, errors.Wrapf(err, "_connectUnregisterAsValidator: ")
+	}
+	if verifySignatures {
+		// _connectBasicTransfer has already checked that the txn is signed
+		// by the top-level public key, which we take to be the sender's
+		// public key so there is no need to verify anything further.
+	}
+
+	// Validate the transactor.
+	if err = bav.IsValidUnregisterAsValidatorMetadata(txn.PublicKey); err != nil {
```
diamondhands0, line 1230: I notice that we allow a validator to Unregister even if they're currently jailed. This seems to create a loophole where I can Unregister in order to escape jail sooner than I otherwise could.

Imagine the following scenario:
- I'm jailed for 48 epochs, but unregistering can be done in 3 epochs.
- I staked to myself, which means that it's no trouble for me to Unregister (versus if I had a lot of other people staked to me)
- If the above is true, then I could Unregister and Re-register to escape jail.

I think a reasonable thing would be to disallow Unregister if the validator is jailed by adding the check to IsValidUnregisterAsvalidatorMetadata()

lib/block_view_validator.go
```diff
@@ -0,0 +1,2133 @@
...
+}
+
+func (bav *UtxoView) _deleteValidatorEntryMappings(validatorEntry *ValidatorEntry) {
+	// This function shouldn't be called with nil.
+	if validatorEntry == nil {
+		glog.Errorf("_deleteValidatorEntryMappings: called with nil entry, this should never happen")
+		return
+	}
+	// Create a tombstone entry.
+	tombstoneEntry := *validatorEntry
```
diamondhands0, line 1882: Should we use .Copy() for all of these to be safe?

lib/block_view_stake.go
diamondhands0: No action on this yet.

lib/block_view_validator.go
diamondhands0: Edit: I put a FIXME at the top for one big mega-comment. Same for block_view_stake.go

lib/block_view_validator.go
diamondhands0: Left it there. I think it's fine.

lib/block_view_validator.go
diamondhands0: I ended up bumping it to 100 as a compromise. Can revisit later.

lib/block_view_validator.go
mattfoley8: It looks like in other _delete funcs, we don't copy when creating a tombstone entry. Leaving as is for now. We do de-reference with the `*`.

PR #530, opened 2023-05-16
Diamondhands/pos txn types review

diamondhands0:


lib/block_view_stake.go
```diff
@@ -1469,6 +1457,9 @@ func (bav *UtxoView) _connectUnstake(
 		}
 	}
 	// 3. Delete the PrevLockedStakeEntry, if exists.
+	//
+	// Note that we don't technically need to do this since the flush will naturall delete
```
lazynina, line 1461:
```suggestion
	// Note that we don't technically need to do this since the flush will naturally delete
```

lib/block_view_validator_test.go
```diff
@@ -163,7 +163,9 @@ func _testValidatorRegistration(t *testing.T, flushToDB bool) {
 		require.Contains(t, err.Error(), RuleErrorValidatorMissingVotingPublicKey)
 	}
 	{
-		// RuleErrorValidatorInvalidVotingSignatureBlockHeight: missing
+		// Got rid of VotingSignatureBlockHeight but going to leave this test here.
```
mattfoley8, line 166: nit: IMO we should also delete this test.

lib/block_view_validator.go
```diff
@@ -561,9 +512,9 @@ func DBKeyForValidatorByPKID(validatorEntry *ValidatorEntry) []byte {
 func DBKeyForValidatorByStake(validatorEntry *ValidatorEntry) []byte {
 	key := append([]byte{}, Prefixes.PrefixValidatorByStake...)
 	key = append(key, EncodeUint8(uint8(validatorEntry.Status()))...)
-	// TotalStakeAmountNanos will never be nil here, but EncodeOptionalUint256
+	// TotalStakeAmountNanos will never be nil here, but FixedWidthEncodeUint256
```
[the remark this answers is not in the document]
diamondhands0: +1

lib/block_view_validator_test.go
diamondhands0: +1

lib/block_view_stake.go
```diff
@@ -3,19 +3,22 @@ package lib
...
 
+// FIXME: Add a big comment here explaining how Stake/Unstake/Unlock
+// all fit together at a high level.
+
 //
 // TYPES: StakeEntry
 //
 
 type StakeEntry struct {
-	StakeID          *BlockHash
```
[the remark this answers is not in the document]
diamondhands0: Copying reply from Slack:

I would say that if we want that then we should make it HASH(actualUniqueIdentifier) or something. The problem with an extra ID is now there's more than one way to query for something, and it's more likely to cause bugs. If we want UUIDs for graphql, we could just add a field like HASH(ValidatorPKID, StakerPKID) or something as a "StakeID" in Postgres, but I'm a bit averse to putting that extraneous data into the core when it's not really needed.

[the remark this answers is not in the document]
lazynina: for completeness, it's fine to compute the ID however you'd like - a hash is fine by me. The problem is creating associations. If we don't have the entry indexed by it's ID (whether that's a hash or some other random identifier we create), we can't reasonably creating an association. I think we decided we WON'T be adding associations for this EVER. If we do, we will certainly need a reindex which is becoming increasing unfeasible to perform. We would have to blocksync to properly reindex these items.

lib/block_view_validator.go
```diff
@@ -1660,6 +1603,11 @@ func (bav *UtxoView) IsValidRegisterAsValidatorMetadata(
 		metadata.DisableDelegatedStake { // Updating DisableDelegatedStake = true
 
 		// Retrieve existing StakeEntries.
+		//
+		// TODO: This is currently a very inefficient way to check to see if a validator has any
```
lazynina, line 1607: we probably should address this now

lib/block_view_stake.go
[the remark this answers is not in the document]
mattfoley8: We could do associations with ValidatorEntries and just use PKIDs as the foreign key. We don't have a standardized UUID column, but we do have entity-specific identifiers we can use. Trickier when the foreign key is multiple columns like for StakeEntries, but IMO use-cases for associations on StakeEntries are a little more obscure, while associations on ValidatorEntries could make a lot of sense to flag, endorse, tag, etc.

lib/block_view_validator.go
mattfoley8: Addressed in https://github.com/deso-protocol/core/pull/531.

lib/block_view_validator_test.go
mattfoley8: Addressed in https://github.com/deso-protocol/core/pull/531.

lib/block_view_validator.go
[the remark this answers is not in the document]
mattfoley8: Addressed in https://github.com/deso-protocol/core/pull/531.

lib/block_view_stake.go
mattfoley8: Addressed in https://github.com/deso-protocol/core/pull/531.

PR #531, opened 2023-05-17
Mf/pos txn types review

mattfoley8:


lib/block_view_validator.go
```diff
@@ -1680,6 +1704,72 @@ func (bav *UtxoView) IsValidUnjailValidatorMetadata(transactorPublicKey []byte)
...
+	// Sanity check that there are no existing StakeEntries for the validator.
+	stakeEntries, err := bav.GetStakeEntriesForValidatorPKID(utxoOp.PrevValidatorEntry.ValidatorPKID)
+	if err != nil {
+		return errors.Wrapf(err, "SanityCheckUnregisterAsValidatorTxn: error retrieving StakeEntries: ")
+	}
+	if len(stakeEntries) != 0 {
+		return errors.New("SanityCheckUnregisterAsValidatorTxn: StakeEntries for ValidatorEntry still exist")
+	}
+
+	// Sanity check the deleted StakeEntries.
```
mattfoley8, line 1743: There's one more validation I could do here which is to ensure that the created/updated LockedStakeEntries have the correct LockedAmountNanos. I punted on that since we already do a check in `_connectUnregisterAsValidator`, but mostly because we don't have a `GetLockedStakeEntriesForValidatorAndEpoch` getter function. Our LockedStakeEntries getter requires a ValidatorPKID + StakerPKID + EpochNumber, but we would want to retrieve them across StakerPKIDs in this sanity check. I could add such a getter. But it's also non-trivial logic to keep track of the PrevLockedStakeEntry existing LockedAmountNanos versus what was added by this unregistering txn in this sanity check. All do-able but felt like overkill for this sanity check. LMK.

lib/block_view_validator.go
```diff
@@ -1772,25 +1862,20 @@ func (bav *UtxoView) GetTopActiveValidatorsByStake(limit int) ([]*ValidatorEntry
...
 func (bav *UtxoView) GetGlobalStakeAmountNanos() (*uint256.Int, error) {
-	var globalStakeAmountNanos *uint256.Int
-	var err error
 	// Read the GlobalStakeAmountNanos from the UtxoView.
 	if bav.GlobalStakeAmountNanos != nil {
-		globalStakeAmountNanos = bav.GlobalStakeAmountNanos.Clone()
+		return bav.GlobalStakeAmountNanos, nil
 	}
 	// If not set, read the GlobalStakeAmountNanos from the db.
-	// TODO: Confirm if the GlobalStakeAmountNanos.IsZero() that we should look in the db.
```
mattfoley8, line 1782: Sanity check caught that we *don't* want to go to the db if the `UtxoView.GlobalStakeAmountNanos.IsZero()`. To illustrate, imagine a case where the existing global stake is 1 $DESO, but the one and only validator unstakes resulting in a global stake in the UtxoView of 0 $DESO. If we fetch the current global stake, we will see it's zero in the UtxoView so go to the database where it's still 1 $DESO, because we haven't flushed the UtxoView to the db yet. For that reason, we only want to go to the db if the global stake is `nil` in the UtxoView.

lib/block_view_validator.go
```diff
@@ -16,27 +16,63 @@ import (
...
+	// a validator sets DisableDelegatedStake to true, then they can still
+	// stake with themselves, but all other users will receive an error if they
+	// try to stake with this validator.
+	DisableDelegatedStake bool
+	// The VotingPublicKey is a BLS PublicKey that is used in consensus messages.
+	// A validator signs consensus messages with their VotingPrivateKey and then
+	// other validators can reliably prove the message came from this validator
+	// by verifying against their VotingPublicKey.
+	VotingPublicKey *bls.PublicKey
+	// The VotingPublicKeySiganture is the signature of the SHA256(VotingPublicKey).
```
lazynina, line 57:
```suggestion
	// The VotingPublicKeySignature is the signature of the SHA256(VotingPublicKey).
```
Also, couldn't some other validator just use the same voting public key and signature since we're only hashing the voting public key? Are we sure we don't want to include the validator's deso public key in the hash?

lib/db_utils.go
```diff
@@ -743,7 +732,7 @@ func StatePrefixToDeSoEncoder(prefix []byte) (_isEncoder bool, _encoder DeSoEnco
 		return true, &ValidatorEntry{}
 	} else if bytes.Equal(prefix, Prefixes.PrefixValidatorByStake) {
 		// prefix_id:"[79]"
-		return true, &PKID{}
+		return false, nil
```
lazynina, line 735: this is an important fix, good catch.

lib/block_view_stake.go
```diff
@@ -1896,6 +1938,244 @@ func (bav *UtxoView) IsValidUnlockStakeMetadata(transactorPkBytes []byte, metada
...
+	if currentStakeEntry == nil {
+		currentStakeEntry = &StakeEntry{StakeAmountNanos: uint256.NewInt()}
+	}
+	stakeEntryStakeAmounsNanosDecrease, err := SafeUint256().Sub(
+		prevStakeEntry.StakeAmountNanos, currentStakeEntry.StakeAmountNanos,
+	)
+	if err != nil {
+		return errors.Wrapf(err, "SanityCheckUnstakeTxn: error calculating StakeAmountNanos decrease: ")
+	}
+	if !stakeEntryStakeAmounsNanosDecrease.Eq(amountNanos) {
```
lazynina, line 2079: Amouns -> Amount

lazynina (commented):
just a few comments - mainly typos

lib/db_utils.go
mattfoley8: Yeah. The value was a PKID until we now only store the ValidatorPKID in the key for this index.

lib/block_view_validator.go
```diff
@@ -1772,25 +1862,20 @@ func (bav *UtxoView) GetTopActiveValidatorsByStake(limit int) ([]*ValidatorEntry
 }
 
 func (bav *UtxoView) GetGlobalStakeAmountNanos() (*uint256.Int, error) {
-	var globalStakeAmountNanos *uint256.Int
-	var err error
 	// Read the GlobalStakeAmountNanos from the UtxoView.
 	if bav.GlobalStakeAmountNanos != nil {
-		globalStakeAmountNanos = bav.GlobalStakeAmountNanos.Clone()
+		return bav.GlobalStakeAmountNanos, nil
```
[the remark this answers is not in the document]
mattfoley8: No, I'm not. I reverted. In the case of an entry, we normally don't `.Copy()` or `.Clone()` in a UtxoView getter. But in the case of a `*uint256.Int`, I can't see any downside in playing it safe and doing a `.Clone()` here.

lib/block_view_validator.go
mattfoley8: If we just hashed and signed the VotingPublicKey, you would be right. But we hash and sign the TransactorPublicKey not the VotingPublicKey. The VotingPublicKey is used to confirm the signature corresponds to the VotingPrivateKey which signed the SHA256(TransactorPublicKey).

mattfoley8: So the signature proves both that 1) you own the private key associated with the VotingPublicKey and 2) that this VotingPublicKey is associated with the current TransactorPublicKey.

mattfoley8: Oh, I see now. I wrote it wrong in this comment. Updating...

lib/block_view_validator.go
```diff
@@ -16,27 +16,63 @@ import (
...
-	ExtraData                map[string][]byte
-	isDeleted                bool
+	// TotalStakeAmountNanos is a cached value of this validator's total stake, calculated
+	// by summing all the corresponding StakeEntries assigned to this validator. We cache
+	// the value here to avoid the O(N) operation of recomputing when determining a
+	// validator's total stake. This way it is an O(1) operation instead.
+	TotalStakeAmountNanos *uint256.Int
+	// LastActiveAtEpochNumber is the last epoch in which this validator either 1) participated in
+	// consensus by voting or proposing blocks, or 2) unjailed themselves. If a validator is
+	// inactive for too long, then they are jailed.
```
diamondhands0, line 68: Interesting. So is the plan to update this value for all validators who vote in every block every time they vote? Not crazy, though it is O(NumValidators) I guess.

lib/block_view_validator.go
diamondhands0: Wow did I miss this? Good catch guys.

PR #533, opened 2023-05-19
Mf/remove jailed stake from global active stake

mattfoley8:
This PR removes any jailed stake from our GlobalActiveStakeAmountNanos value. If a validator is jailed, we do not want to consider their stake in the total active stake across validators used in determining if we have a 2/3rd consensus on votes. This PR makes the following changes:

1. Renames `GlobalStakeAmountNanos` to `GlobalActiveStakeAmountNanos`. Note the addition of the keyword `Active`. This value describes the total active stake across validators, excluding any jailed validators' stake.
2. When a user stakes with a validator, we only add the additional stake to the `GlobalActiveStakeAmountNanos` if the validator isn't jailed.
3. When a user unstakes from a validator, we only subtract the decreased stake from the `GlobalActiveStakeAmountNanos` if the validator isn't jailed.
4. When a validator is jailed, we subtract their `TotalStakeAmountNanos` from the `GlobalActiveStakeAmountNanos`.
5. When a validator unjails himself, we add back their `TotalStakeAmountNanos` to the `GlobalActiveStakeAmountNanos`.
6. When a validator unregisters, we only subtract their now unstaked `TotalStakeAmountNanos` from the `GlobalActiveStakeAmountNanos` it they weren't jailed.

### DIFF
--- a/lib/block_view.go
+++ b/lib/block_view.go
@@ -118,8 +118,8 @@ type UtxoView struct {
 	// Validator mappings
 	ValidatorPKIDToValidatorEntry map[PKID]*ValidatorEntry
 
-	// Global stake across all validators
-	GlobalStakeAmountNanos *uint256.Int
+	// Global active stake across all validators
+	GlobalActiveStakeAmountNanos *uint256.Int
 
 	// Stake mappings
 	StakeMapKeyToStakeEntry map[StakeMapKey]*StakeEntry
@@ -222,10 +222,10 @@ func (bav *UtxoView) _ResetViewMappingsAfterFlush() {
 	// ValidatorEntries
 	bav.ValidatorPKIDToValidatorEntry = make(map[PKID]*ValidatorEntry)
 
-	// Global stake across validators. We deliberately want this to initialize to nil and not zero
-	// since a zero value will overwrite an existing GlobalStakeAmountNanos value in the db, whereas
-	// a nil GlobalStakeAmountNanos value signifies that this value was never set.
-	bav.GlobalStakeAmountNanos = nil
+	// Global active stake across validators. We deliberately want this to initialize to nil and not zero
+	// since a zero value will overwrite an existing GlobalActiveStakeAmountNanos value in the db, whereas
+	// a nil GlobalActiveStakeAmountNanos value signifies that this value was never set.
+	bav.GlobalActiveStakeAmountNanos = nil
 
 	// StakeEntries
 	bav.StakeMapKeyToStakeEntry = make(map[StakeMapKey]*StakeEntry)
@@ -493,9 +493,9 @@ func (bav *UtxoView) CopyUtxoView() (*UtxoView, error) {
 		newView.ValidatorPKIDToValidatorEntry[entryKey] = entry.Copy()
 	}
 
-	// Copy the GlobalStakeAmountNanos.
-	if bav.GlobalStakeAmountNanos != nil {
-		newView.GlobalStakeAmountNanos = bav.GlobalStakeAmountNanos.Clone()
+	// Copy the GlobalActiveStakeAmountNanos.
+	if bav.GlobalActiveStakeAmountNanos != nil {
+		newView.GlobalActiveStakeAmountNanos = bav.GlobalActiveStakeAmountNanos.Clone()
 	}
 
 	// Copy the StakeEntries
--- a/lib/block_view.go
+++ b/lib/block_view.go
@@ -118,7 +118,7 @@ type UtxoView struct {
 	// Validator mappings
 	ValidatorPKIDToValidatorEntry map[PKID]*ValidatorEntry
 
-	// Global active stake across all validators
+	// The global active stake is the sum of all stake across validators who have Status = Active.
 	GlobalActiveStakeAmountNanos *uint256.Int
 
 	// Stake mappings
--- a/lib/block_view_stake.go
+++ b/lib/block_view_stake.go
@@ -1341,8 +1341,8 @@ func (bav *UtxoView) _disconnectStake(
 		bav._setStakeEntryMappings(operationData.PrevStakeEntries[0])
 	}
 
-	// Restore the PrevGlobalActiveStakeAmountNanos if the PrevValidatorEntry was active.
-	if prevValidatorEntry.Status() == ValidatorStatusActive {
+	// Restore the PrevGlobalActiveStakeAmountNanos, if exists.
+	if operationData.PrevGlobalActiveStakeAmountNanos != nil {
 		bav._setGlobalActiveStakeAmountNanos(operationData.PrevGlobalActiveStakeAmountNanos)
 	}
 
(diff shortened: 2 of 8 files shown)


lib/block_view_stake.go
```diff
@@ -1338,8 +1341,10 @@ func (bav *UtxoView) _disconnectStake(
 		bav._setStakeEntryMappings(operationData.PrevStakeEntries[0])
 	}
 
-	// Restore the PrevGlobalStakeAmountNanos.
-	bav._setGlobalStakeAmountNanos(operationData.PrevGlobalStakeAmountNanos)
+	// Restore the PrevGlobalActiveStakeAmountNanos if the PrevValidatorEntry was active.
+	if prevValidatorEntry.Status() == ValidatorStatusActive {
```
lazynina, line 1345: we could just check if `PrevGlobalActiveStakeAmountNanos` is non-nil. I generally prefer that as opposed to relying on a different field. The prevs generally means "revert to this state when disconnecting".

lib/block_view_stake.go
```diff
@@ -2021,22 +2031,24 @@ func (bav *UtxoView) SanityCheckStakeTxn(
...
-	}
-	globalStakeAmountNanosIncrease, err := SafeUint256().Sub(
-		currentGlobalStakeAmountNanos, utxoOp.PrevGlobalStakeAmountNanos,
-	)
-	if err != nil {
-		return errors.Wrapf(err, "SanityCheckStakeTxn: error calculating GlobalStakeAmountNanos increase: ")
-	}
-	if !globalStakeAmountNanosIncrease.Eq(amountNanos) {
-		return errors.New("SanityCheckStakeTxn: GlobalStakeAmountNanos increase does not match")
+	// Validate GlobalActiveStakeAmountNanos increase if validator is active.
```
lazynina, line 2034: we could also add a check that `PrevGlobalActiveStakeAmountNanos` is nil if the validator is jailed - this is a good check that ensures we don't modify the value unintentionally.

lib/block_view_validator.go
```diff
@@ -1514,10 +1512,24 @@ func (bav *UtxoView) _connectUnjailValidator(
 	// Set the CurrentValidatorEntry.
 	bav._setValidatorEntryMappings(currentValidatorEntry)
 
+	// Increase the GlobalActiveStakeAmountNanos.
```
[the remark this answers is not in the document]
mattfoley8: That's correct. See link [here](https://github.com/deso-protocol/core/blob/mf/remove-jailed-stake-from-global-active-stake/lib/block_view_validator.go#L1706-L1709). It also validates that sufficient epochs have elapsed for the validator to be unjailed.

lib/block_view_stake.go
```diff
@@ -1226,19 +1226,22 @@ func (bav *UtxoView) _connectStake(
...
-	prevGlobalStakeAmountNanos, err := bav.GetGlobalStakeAmountNanos()
-	if err != nil {
-		return 0, 0, nil, errors.Wrapf(err, "_connectStake: error retrieving GlobalStakeAmountNanos: ")
-	}
-	globalStakeAmountNanos, err := SafeUint256().Add(prevGlobalStakeAmountNanos, txMeta.StakeAmountNanos)
-	if err != nil {
-		return 0, 0, nil, errors.Wrapf(err, "_connectStake: error adding StakeAmountNanos to GlobalStakeAmountNanos: ")
+	// Increase the GlobalActiveStakeAmountNanos if the validator is active.
+	var prevGlobalActiveStakeAmountNanos *uint256.Int
+	if currentValidatorEntry.Status() == ValidatorStatusActive {
```
diamondhands0, line 1231: I'm sure you've thougth about this, but do we want to make it so that you can't stake to a jailed validator? You would just add an `else if` block here to error if the status is jailed.

If you did this, I'd also add an `else` for "Unknown validator status" error.

diamondhands0: I thought about it more, and I think what you're doing here is more elegant. If we were to disallow staking to jailed validators, we'd break symmetry with Unstake, which I think adds complexity.

lib/block_view_types.go
```diff
@@ -1876,11 +1876,9 @@ func (op *UtxoOperation) RawDecodeWithoutMetadata(blockHeight uint64, rr *bytes.
 			return errors.Wrapf(err, "UtxoOperation.Decode: Problem reading PrevValidatorEntry: ")
 		}
 
-		// PrevGlobalStakeAmountNanos
-		if prevGlobalStakeAmountNanos, err := VariableDecodeUint256(rr); err == nil {
-			op.PrevGlobalStakeAmountNanos = prevGlobalStakeAmountNanos
-		} else {
-			return errors.Wrapf(err, "UtxoOperation.Decode: Problem reading PrevGlobalStakeAmountNanos: ")
+		// PrevGlobalActiveStakeAmountNanos
+		if op.PrevGlobalActiveStakeAmountNanos, err = VariableDecodeUint256(rr); err != nil {
```
diamondhands0, line 1880: Nit: No need to fix this here, but I tend to prefer breaking assignments out of if statements like the following. I know it's not "Go-like" but I find it easier to read and less error-prone:

```
		// PrevGlobalActiveStakeAmountNanos
        op.PrevGlobalActiveStakeAmountNanos, err = VariableDecodeUint256(rr)
		if err != nil {
			return errors.Wrapf(err, "UtxoOperation.Decode: Problem reading PrevGlobalActiveStakeAmountNanos: ")
		}
```

lib/block_view_validator.go
```diff
@@ -1571,6 +1583,13 @@ func (bav *UtxoView) _disconnectUnjailValidator(
 	}
 	bav._setValidatorEntryMappings(prevValidatorEntry)
 
+	// Restore the PrevGlobalActiveStakeAmountNanos.
+	prevGlobalActiveStakeAmountNanos := operationData.PrevGlobalActiveStakeAmountNanos
+	if prevGlobalActiveStakeAmountNanos == nil {
+		return errors.New("_disconnectUnjailValidator: PrevGlobalActiveStakeAmountNanos is nil")
```
diamondhands0, line 1589: Nit: Whenever something should never happen, like I think is the case here, I like to add "This should never happen" or "sanity-check failed" to indicate that in the logs for whoever ends up having to debug it.

lib/block_view.go
```diff
@@ -118,8 +118,8 @@ type UtxoView struct {
 	// Validator mappings
 	ValidatorPKIDToValidatorEntry map[PKID]*ValidatorEntry
 
-	// Global stake across all validators
-	GlobalStakeAmountNanos *uint256.Int
+	// Global active stake across all validators
```
lazynina, line 121: can we expound upon this comment? Something like `The global active stake is the sum of all DESO staked to validators who are not jailed.`

lib/block_view_validator.go
lazynina: Let's add this^

lazynina (approved):

PR #537, opened 2023-05-22
Block Reward Patch - prevent free tx fees for recipient of block reward output

lazynina:


lib/block_view.go
```diff
@@ -3356,6 +3356,30 @@ func (bav *UtxoView) ConnectBlock(
...
+		// Make sure the first transaction is a block reward.
+		if desoBlock.Txns[0].TxnMeta.GetTxnType() != TxnTypeBlockReward {
+			return nil, errors.Wrap(RuleErrorFirstTxnMustBeBlockReward, "ConnectBlock: First transaction in block is not a block reward")
+		}
+		// Ensure that the block reward transaction has exactly one output.
+		if len(desoBlock.Txns[0].TxOutputs) != 1 {
+			return nil, errors.Wrap(RuleErrorBlockRewardTxnMustHaveOneOutput, "ConnectBlock: Block reward transaction must have exactly one output")
+		}
+		var err error
+		if blockRewardOutputPublicKey, err =
```
diamondhands0, line 3377: In cases like this, I'd break the `if err != nil` into its own line.

lib/block_view.go
```diff
@@ -3938,6 +3984,11 @@ func (bav *UtxoView) GetSpendableDeSoBalanceNanosForPublicKey(pkBytes []byte,
 				return 0, errors.Wrapf(err, "GetSpendableDeSoBalanceNanosForPublicKey: Problem adding "+
 					"block reward (%d) to immature block rewards (%d)", blockRewardForPK, immatureBlockRewards)
 			}
+			// This is the specific line that causes the bug. We should be using blockNode.Header.PrevBlockHash
```
diamondhands0, line 3987: Add a TODO

PR #538, opened 2023-05-22
Mf/generate pos leader schedules

mattfoley8:


lib/pos_leader_schedule.go
```diff
@@ -0,0 +1,106 @@
...
+			// If the sum of the stake seen so far is less than the RandomUint256, skip this validator.
+			if sumStakeAmountNanos.Lt(randomUint256) {
+				continue
+			}
+
+			// If we get to this point, the current validator is the
+			// one we should add to the leader schedule next.
+
+			// Add the current ValidatorEntry to the leaderSchedule.
+			leaderSchedule = append(leaderSchedule, validatorEntry)
```
lazynina, line 88: could we just remove the validatorEntry from the slice of validatorEntries once its been added to the leader schedule? we could get rid of the set with this change.

mattfoley8: That's the way I initially had it like `validatorEntries = append(validatorEntries[:index], validatorEntries[index+1:])`, but after reading into it, IMO the edge case handling around slice indexing is tricky and the space/time complexity of doing that slice concatenation is no better than using the set IMO. LMK what you think.

mattfoley8: Put another way, we have a trade-off here between space vs. time. We can triplicate the list of ValidatorEntries: 1) the input validator entries, 2) the leader schedule slice, and 3) the set of validator entries already added to the leader schedule. But look-ups into the set of already added validators is O(1). Alternatively, we can save ourselves from also keeping that set in memory and instead remove the validator from the input slice of validator entries. This saves space but costs us some time/runtime since the algorithm to remove an element from a slice is not O(1). IMO since the number of validators we consider in our leader schedule generation algorithm is capped (currently to 100), it's worth taking up more space to save on some time. Note that the algorithm is already O(n^2).

lazynina (approved):

lib/pos_leader_schedule.go
```diff
@@ -0,0 +1,106 @@
...
+
+func (bav *UtxoView) GenerateLeaderSchedule() ([]*PKID, error) {
+	// Retrieve CurrentRandomSeedHash.
+	currentRandomSeedHash, err := bav.GetCurrentRandomSeedHash()
+	if err != nil {
+		return nil, errors.Wrapf(err, "UtxoView.GenerateLeaderSchedule: error retrieving CurrentRandomSeedHash: ")
+	}
+
+	// Retrieve top, active validators ordered by stake.
+	validatorEntries, err := bav.GetTopActiveValidatorsByStake(int(bav.Params.LeaderScheduleMaxNumValidators))
```
diamondhands0, line 17: Never use int. Always be clear about the size of the int by using `int32` or `int64`

diamondhands0, line 17: TODO: Need to make sure that this function returns the top validators in a deterministic order. If it has any non-determinism in it, then nodes will have different leader schedules.

lib/pos_leader_schedule.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	// Pseudocode for leader-selection algorithm:
+	// Note this is an O(N^2) algorithm where N is the number of validators we include.
+	// While len(LeaderSchedule) < len(ValidatorEntries):
+	//   Hash the CurrentRandomSeedHash and generate a new RandomUint256.
+	//   Take RandomUint256 modulo TotalStakeAmountNanos.
+	//   For each ValidatorEntry:
+	//     Skip if ValidatorPKID has already been added to the leader schedule.
+	//     If ValidatorEntry.TotalStakeAmountNanos >= RandomUint256:
```
diamondhands0, line 41: I think this comment is incorrect. You're tracking the SUM of the TotalStakeAmountNanos.

lib/pos_leader_schedule.go
```diff
@@ -0,0 +1,106 @@
...
+		// Hash the CurrentRandomSeedHash each iteration. This generates
+		// multiple predictable pseudorandom values from the same seed.
+		currentRandomSHA256 := sha256.Sum256(currentRandomSeedHash.ToBytes())
+		currentRandomSeedHash, err = (&RandomSeedHash{}).FromBytes(currentRandomSHA256[:])
+		if err != nil {
+			return nil, errors.Wrapf(err, "UtxoView.GenerateLeaderSchedule: error hashing CurrentRandomSeedHash: ")
+		}
+
+		// Take RandomUint256 % TotalStakeAmountNanos.
+		randomUint256 := uint256.NewInt().Mod(currentRandomSeedHash.ToUint256(), totalStakeAmountNanos)
```
diamondhands0, line 62: TODO: Make sure Mod is efficient. I'm almos certain it is.

lib/pos_leader_schedule.go
```diff
@@ -0,0 +1,106 @@
...
+
+			// The current validator has been added to the leader schedule.
+			// Break out of this inner loop, generate a new RandomUint256,
+			// and find the next stake-weighted validator to add to the
+			// leader schedule.
+			break
+		}
+	}
+
+	return leaderSchedule, nil
```
diamondhands0, line 105: Seems pretty clean. Good work!

lib/pos_leader_schedule.go
mattfoley8: `uint256.Mod()` uses `uint256.QuoRem()` under the hood which looks to me like it's O(1). Link to Mod [code](https://github.com/Pilatuz/bigx/blob/v2.0.1/uint256/uint256.go#L324). Link to QuoRem [code](https://github.com/Pilatuz/bigx/blob/v2.0.1/uint256/uint256.go#L342).

lib/pos_leader_schedule.go
mattfoley8: Addressed in: https://github.com/deso-protocol/core/pull/540.

mattfoley8: I have a test where I regenerate a leader schedule using the same CurrentRandomSeedHash 10x times. It generates the same leader schedule each time. With that test, I'm fairly confident in this algorithm's determinism.

PR #542, opened 2023-05-31
Mf/stub out epoch complete hook

mattfoley8:


lib/pos_snapshot_entries.go
```diff
@@ -0,0 +1,643 @@
...
+//
+// SnapshotValidatorEntry
+//
+
+type SnapshotValidatorMapKey struct {
+	SnapshotAtEpochNumber uint64
+	ValidatorPKID         PKID
+}
+
+func (bav *UtxoView) SnapshotCurrentValidators(snapshotAtEpochNumber uint64) error {
```
mattfoley8, line 128: I don't love this but I thought it may be good enough for now. I'm not sure how many validators we are expecting. I load all validators into memory to merge between the database and the UtxoView before snapshotting them. Alternatively, I could copy over all validators in the database one-by-one (without even byte-decoding them), and then just insert or delete any validators in the UtxoView. This would increase the number of db inserts/deletes, but reduce the memory footprint. Happy to make that change if others think it's way better. Also happy to punt on that change until our validator set gets unwieldy to load into memory.

lib/pos_snapshot_entries.go
```diff
@@ -0,0 +1,643 @@
...
+		)
+	}
+	if validatorEntry != nil {
+		// Cache the result in the UtxoView.
+		bav._setSnapshotValidatorEntry(validatorEntry, snapshotAtEpochNumber)
+	}
+	return validatorEntry, nil
+}
+
+func (bav *UtxoView) GetSnapshotTopActiveValidatorsByStake(limit uint64, snapshotAtEpochNumber uint64) ([]*ValidatorEntry, error) {
```
mattfoley8, line 171: There's some code duplication between e.g. GetTopActiveValidatorsByStake and GetSnapshotTopActiveValidatorsByStake. Not enough to IMO refactor into one function, but just calling it out. We could have the active (non-snapshot set) default to `SnapshotAtEpochNumber = 0` and then reuse the indices + functions. So basically, if we are updating the current active validator set in consensus, we assume the `SnapshotAtEpochNumber = 0`, that's actually where we store the current validator set. And any rows in that index where `SnapshotAtEpochNumber > 1` are actually snapshot values. I think there's something nice to having them split apart (more future-proof in case their logic diverges), but I could do a pretty big refactor to combine the current + snapshot ValidatorEntries into a single index, one for by PKID and one for by stake.

mattfoley8: The same could be done for GlobalActiveStakeAmountNanos.

lib/pos_epoch_complete_hook.go
```diff
@@ -0,0 +1,84 @@
...
+		return errors.Wrapf(err, "RunEpochCompleteHook: problem generating leader schedule: ")
+	}
+	for index, validatorPKID := range leaderSchedule {
+		if index > math.MaxUint8 {
+			return errors.Errorf("RunEpochCompleteHook: LeaderIndex %d overflows uint8", index)
+		}
+		bav._setSnapshotLeaderScheduleValidator(validatorPKID, uint8(index), currentEpochEntry.EpochNumber)
+	}
+
+	// TODO: Jail inactive validators.
```
mattfoley8, line 73: I'll address these in a future PR.

lib/pos_snapshot_entries.go
```diff
@@ -0,0 +1,643 @@
...
+	for mapKey, validatorEntry := range bav.SnapshotValidatorEntries {
+		if mapKey.SnapshotAtEpochNumber == snapshotAtEpochNumber &&
+			!validatorEntry.isDeleted &&
+			validatorEntry.Status() == ValidatorStatusActive &&
+			!validatorEntry.TotalStakeAmountNanos.IsZero() {
+			validatorEntries = append(validatorEntries, validatorEntry)
+		}
+	}
+	// Sort the ValidatorEntries DESC by TotalStakeAmountNanos.
+	sort.Slice(validatorEntries, func(ii, jj int) bool {
```
lazynina, line 210: we should probably be using sort.SliceStable - we also probably need a tie breaker if two validator have the same TotalStakeAmountNanos so we don't have non-determinism issues in the future.

lib/pos_snapshot_entries.go
```diff
@@ -0,0 +1,708 @@
...
+	"sort"
+)
+
+//
+// SnapshotGlobalParamsEntry
+//
+
+const SnapshotLookbackNumEpochs uint64 = 2
+
+func (bav *UtxoView) GetSnapshotGlobalParam(field string) (uint64, error) {
```
lazynina, line 20: should we make an enum for field instead of using a string?

lib/pos_snapshot_entries.go
```diff
@@ -0,0 +1,708 @@
...
+	if globalActiveStakeAmountNanos != nil {
+		// Cache the result in the UtxoView.
+		bav._setSnapshotGlobalActiveStakeAmountNanos(globalActiveStakeAmountNanos, snapshotAtEpochNumber)
+	}
+	return globalActiveStakeAmountNanos, nil
+}
+
+func (bav *UtxoView) _setSnapshotGlobalActiveStakeAmountNanos(globalActiveStakeAmountNanos *uint256.Int, snapshotAtEpochNumber uint64) {
+	if globalActiveStakeAmountNanos == nil {
+		glog.Errorf("_setSnapshotGlobalActiveStakeAmountNanos: called with nil entry, this should never happen")
```
lazynina, line 500:
```suggestion
		glog.Errorf("_setSnapshotGlobalActiveStakeAmountNanos: called with nil entry, this should never happen")
		return
```
return after this error?

lib/pos_snapshot_entries.go
```diff
@@ -0,0 +1,708 @@
...
+	if globalParamsEntry != nil {
+		// Cache the result in the UtxoView.
+		bav._setSnapshotGlobalParamsEntry(globalParamsEntry, snapshotAtEpochNumber)
+	}
+	return globalParamsEntry, nil
+}
+
+func (bav *UtxoView) _setSnapshotGlobalParamsEntry(globalParamsEntry *GlobalParamsEntry, snapshotAtEpochNumber uint64) {
+	if globalParamsEntry == nil {
+		glog.Errorf("_setSnapshotGlobalParamsEntry: called with nil entry, this should never happen")
```
lazynina, line 97:
```suggestion
		glog.Errorf("_setSnapshotGlobalParamsEntry: called with nil entry, this should never happen")
		return
```

lazynina (changes requested):

lib/db_utils.go
```diff
@@ -529,7 +529,28 @@ type DBPrefixes struct {
...
+	// Note: we parse the ValidatorPKID from the key and the value is nil to save space.
+	PrefixSnapshotValidatorByStatusAndStake []byte `prefix_id:"[87]" is_state:"true"`
+
+	// PrefixSnapshotGlobalActiveStakeAmountNanos: Retrieve a snapshot GlobalActiveStakeAmountNanos by SnapshotAtEpochNumber.
+	// Prefix, SnapshotAtEpochNumber -> *uint256.Int
+	PrefixSnapshotGlobalActiveStakeAmountNanos []byte `prefix_id:"[88]" is_state:"true"`
+
+	// PrefixSnapshotLeaderSchedule: Retrieve a ValidatorPKID by <SnapshotAtEpochNumber, LeaderIndex>.
+	// Prefix, SnapshotAtEpochNumber, LeaderIndex -> ValidatorPKID
+	PrefixSnapshotLeaderSchedule []byte `prefix_id:"[89]" is_state:"true"`
```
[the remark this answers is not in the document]
mattfoley8: The ValidatorPKID is not included in the key. This is deliberate. We want to be able to say, who is the first validator in the leader schedule that was snapshot in epoch = N. `SnapshotAtEpochNumber = N, LeaderIndex = 0`. This way we are guaranteed there will every be one such row, with the ValidatorPKID being the value. If we included the ValidatorPKID in the key, we would either have to query with the ValidatorPKID (already knowing which validator is there which is kinda a catch-22) or there could be a weird case where there are multiple validators assigned to the same index within the same epoch since multiple keys can exist for the same SnapshotAtEpochNumber + LeaderIndex.

lazynina (approved):

PR #543, opened 2023-05-31
Mf/param updater update pos global params

mattfoley8:
Goes with https://github.com/deso-protocol/backend/pull/496.

### DIFF
--- a/lib/block_view_stake.go
+++ b/lib/block_view_stake.go
@@ -1955,7 +1955,7 @@ func (bav *UtxoView) IsValidUnlockStakeMetadata(transactorPkBytes []byte, metada
 	}
 
 	// Validate EndEpochNumber + StakeLockupEpochDuration <= CurrentEpochNumber.
-	if metadata.EndEpochNumber+bav.Params.StakeLockupEpochDuration > currentEpochNumber {
+	if metadata.EndEpochNumber+bav.GetStakeLockupEpochDuration() > currentEpochNumber {
 		return errors.Wrapf(RuleErrorInvalidUnlockStakeMustWaitLockupDuration, "UtxoView.IsValidUnlockStakeMetadata: ")
 	}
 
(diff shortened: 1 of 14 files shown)


lib/block_view_stake.go
```diff
@@ -1955,7 +1955,7 @@ func (bav *UtxoView) IsValidUnlockStakeMetadata(transactorPkBytes []byte, metada
 	}
 
 	// Validate EndEpochNumber + StakeLockupEpochDuration <= CurrentEpochNumber.
-	if metadata.EndEpochNumber+bav.Params.StakeLockupEpochDuration > currentEpochNumber {
+	if metadata.EndEpochNumber+bav.GetStakeLockupEpochDuration(0) > currentEpochNumber {
```
lazynina, line 1958: Should we be fetching the StakeLockupEpochDuration based on a snapshot?

lib/block_view_validator.go
```diff
@@ -1715,7 +1715,7 @@ func (bav *UtxoView) IsValidUnjailValidatorMetadata(transactorPublicKey []byte)
 	}
 
 	// Validate sufficient epochs have elapsed for validator to be unjailed.
-	if validatorEntry.JailedAtEpochNumber+bav.Params.ValidatorJailEpochDuration > currentEpochNumber {
+	if validatorEntry.JailedAtEpochNumber+bav.GetValidatorJailEpochDuration(0) > currentEpochNumber {
```
lazynina, line 1718: same thing here about using a snapshot

lib/block_view_stake.go
mattfoley8: Yes. We should be fetching those based on a snapshot. For now, I'm just keeping the existing behavior. I'll update exactly which snapshot we should fetch once I know it. That'll just be swapping out the `0` for the correct EpochNumber relative to the CurrentEpochNumber.

lib/block_view_validator.go
mattfoley8: Yup, you are correct. Same response as above.

lib/block_view_stake.go
mattfoley8: Yes. We should be fetching those based on a snapshot. For now, I'm just keeping the existing behavior. I'll update exactly which snapshot we should fetch once I know it. That'll just be swapping out the `0` for the correct EpochNumber relative to the CurrentEpochNumber.

lazynina (approved):
Just check that extra data isn't nil, but otherwise this looks good.

PR #546, opened 2023-06-05
Consolidate PoS block heights.

mattfoley8:


lib/constants.go
```diff
@@ -335,11 +345,12 @@ type MigrationHeight struct {
...
-	UnlimitedDerivedKeysMigration        MigrationName = "UnlimitedDerivedKeysMigration"
-	AssociationsAndAccessGroupsMigration MigrationName = "AssociationsAndAccessGroupsMigration"
-	BalanceModelMigration                MigrationName = "BalanceModelMigration"
-	ProofOfStakeNewTxnTypesMigration     MigrationName = "ProofOfStakeNewTxnTypesMigration"
+	DefaultMigration                       MigrationName = "DefaultMigration"
+	UnlimitedDerivedKeysMigration          MigrationName = "UnlimitedDerivedKeysMigration"
+	AssociationsAndAccessGroupsMigration   MigrationName = "AssociationsAndAccessGroupsMigration"
+	BalanceModelMigration                  MigrationName = "BalanceModelMigration"
+	ProofOfStake1StateSetupMigration       MigrationName = "ProofOfStake1StateSetupMigration"
+	ProofOfStake2ConsensusCutoverMigration MigrationName = "ProofOfStake2ConsensusCutoverMigration"
```
lazynina, line 353: Are any encoders actually changing  with this block height? If not, we don't need an encoder migration. Adding an encoder migration will cause the node to iterate over all the index to compute the post-migration state. It's a bit of an expensive operation so if we can only have one migration, that's preferable.

lib/constants.go
```diff
@@ -645,7 +665,10 @@ var RegtestForkHeights = ForkHeights{
 	// For convenience, we set the block height to 1 since the
 	// genesis block was created using the utxo model.
 	BalanceModelBlockHeight:            uint32(1),
-	ProofOfStakeNewTxnTypesBlockHeight: uint32(1),
+	ProofOfStake1StateSetupBlockHeight: uint32(1),
+
+	// FIXME: set to real block height when ready
+	ProofOfStake2ConsensusCutoverBlockHeight: uint32(math.MaxUint32),
```
lazynina, line 671: This is just a placeholder for when we actually write logic for PoS consensus mechanisms, correct? I don't see this fork height used in this PR

lib/constants.go
mattfoley8: I'll delete the ConsensusCutoverMigration for now. It's not used yet, but I think we will need it. But trivial to add back when we do. We use the StateSetupMigration for e.g. gating the new fields on a UtxoOperation.

lazynina: I guess my thinking is that if we KNOW we are going to be performing a migration later, we can perform it now without consequence. For example, if we want to start including a new field, but it isn't relevant until the 2nd block height, we can just encode its default value.

lazynina (approved):

PR #550, opened 2023-06-08
Mf/jail inactive validators

mattfoley8:
This PR jails inactive validators. A validator stays active by either voting on blocks or proposing new blocks. When they participate in consensus by performing either of those actions, we update their LastActiveAtEpochNumber. If they fall inactive beyond our grace period (a param stored on the GlobalParamsEntry) we jail them. We run this logic in the OnEpochCompleteHook.

### DIFF
--- a/lib/block_view_validator.go
+++ b/lib/block_view_validator.go
@@ -1929,6 +1929,12 @@ func (bav *UtxoView) ShouldJailValidator(validatorEntry *ValidatorEntry, blockHe
 	// there would be an edge case where all validators will get jailed
 	// after we deploy the StateSetup block height, but before we deploy
 	// the ConsensusCutover block height.
+	//
+	// We do another check below to make sure enough blocks have passed even
+	// after we cut-over to PoS, but since this check is so quick to perform,
+	// we keep this one here as well, since this will catch all OnEpochCompleteHooks
+	// after the StateSetup block height and before the CutoverConsensus block height
+	// and saves us a few look-ups and computations.
 	if blockHeight < uint64(bav.Params.ForkHeights.ProofOfStake2ConsensusCutoverBlockHeight) {
 		return false, nil
 	}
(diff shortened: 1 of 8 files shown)


lib/block_view_validator.go
```diff
@@ -1924,6 +1924,36 @@ func (bav *UtxoView) GetGlobalActiveStakeAmountNanos() (*uint256.Int, error) {
...
+func (bav *UtxoView) ShouldJailValidator(validatorEntry *ValidatorEntry) (bool, error) {
+	// Return false if the validator is already jailed. We do not want to jail
+	// them again as we want to retain their original JailedAtEpochNumber so
+	// that they can eventually unjail themselves.
+	if validatorEntry.Status() == ValidatorStatusJailed {
+		return false, nil
+	}
+
+	// Retrieve the CurrentEpochNumber.
+	currentEpochNumber, err := bav.GetCurrentEpochNumber()
```
lazynina, line 1936: how do we know if they satisfied some criteria such that they shouldn't be jailed? Seems like we'd end up with every validator being jailed as it stands?

lib/pos_snapshot_entries.go
```diff
@@ -207,6 +214,24 @@ func (bav *UtxoView) SnapshotCurrentValidators(snapshotAtEpochNumber uint64) err
 		if !validatorEntry.isDeleted {
 			// We only want to snapshot !isDeleted ValidatorEntries.
 			bav._setSnapshotValidatorEntry(validatorEntry, snapshotAtEpochNumber)
+
```
lazynina, line 217: should we flip this around such that we pull in everything from the db and merge it into the view and then do one loop over the view?

lazynina (commented):
Overall looks good - just two questions for you

lib/block_view_validator.go
mattfoley8: That is correct as it stands now. I just updated the description above with this line:
> A validator stays active by either voting on blocks or proposing new blocks. When they participate in consensus by performing either of those actions, we update their LastActiveAtEpochNumber.
We haven't implemented the consensus logic to handle VOTES or new BLOCK_PROPOSALS. But in that logic, we will be updating the `ValidatorEntry.LastActiveAtEpochNumber = CurrentEpochNumber`.

lib/pos_snapshot_entries.go
mattfoley8: IMO no since this is actually the more efficient operation. If we pull everything from the db, we are loading the ValidatorEntries already in the UtxoView into memory a second time and have a small overhead to loop through those ValidatorEntries and just skip them since they're already in the UtxoView (but that's more negligible than just the double-loading them into memory issue IMO). But LMK if you think the simplicity of the other method outweighs the efficiency gain.

lazynina: nah keep it.

lazynina (approved):

PR #552, opened 2023-06-09
Fix bugs in updating global params.

mattfoley8:


lib/pos_snapshot_entries.go
```diff
@@ -43,6 +43,10 @@ func (snapshotGlobalParam SnapshotGlobalParam) ToString() string {
 	return string(snapshotGlobalParam)
 }
 
+func (bav *UtxoView) GetCurrentGlobalParam(field SnapshotGlobalParam) uint64 {
```
mattfoley8, line 46: This is used by backend when we update the GlobalParamsEntry. We only want to update if the passed in value differs from the current value. So we need a getter for the current value (and not just the snapshot value like we had before).

PR #555, opened 2023-06-14
PoS Add P2P vote and timeout messages

tholonious:
Reference schema: https://github.com/deso-protocol/hotstuff_pseudocode/blob/afd4337e8aaad1be56b12c6173e4f82d79b362e0/fast_hotstuff_bls.go

PR #558, opened 2023-06-15
PoS Add QC fields to block header schema

tholonious:


tholonious:
* feature/pos-messaging-schema
  * **PR #555** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/555" target="_blank">Graphite</a> 
    * **PR #558** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/558" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/558?utm_source=stack-comment).

lib/constants.go
```diff
@@ -101,7 +101,24 @@ const (
 	//
 	// At the time of this writing, the intent is to deploy it in a backwards-compatible
 	// fashion, with the eventual goal of phasing out blocks with the previous version.
-	HeaderVersion1       = uint32(1)
+	HeaderVersion1 = uint32(1)
```
lazynina, line 104: should we convert these HeaderVersions to some type now?

lib/network_test.go
```diff
@@ -104,14 +105,76 @@ var expectedBlockHeader = &MsgDeSoHeader{
...
+		0x32, 0x33,
+	}
+	testMerkleRoot := BlockHash{
+		0x00, 0x35, 0x36, 0x37, 0x38, 0x39, 0x40, 0x41, 0x42, 0x43,
+		0x44, 0x45, 0x46, 0x47, 0x48, 0x49, 0x50, 0x51, 0x52, 0x53,
+		0x54, 0x55, 0x56, 0x57, 0x58, 0x59, 0x60, 0x61, 0x62, 0x63,
+		0x64, 0x65,
+	}
+
+	_, testBLSSignature := _generateValidatorVotingPublicKeyAndSignature(t)
```
lazynina, line 124: can we define this somewhere such that postgres tests compile? we don't run this test in postgres CI so the function can simply panic.

lib/pos_network.go
```diff
@@ -404,3 +404,119 @@ func DecodeAggregatedBLSSignature(rr io.Reader) (*AggregatedBLSSignature, error)
 
 	return &sig, nil
 }
+
+// TimeoutAggregateQuorumCertificate is an aggregation of timeout messages from 2/3rds
+// of all validators, weighted by stake, which indicates that these validators want to
+// time out a particular view.
+//
+// When validators want to time a view, they send their high QCs to the block proposer
```
lazynina, line 412:
```suggestion
// When validators want to time out a view, they send their high QCs to the block proposer
```

lib/pos_network.go
```diff
@@ -404,3 +404,119 @@ func DecodeAggregatedBLSSignature(rr io.Reader) (*AggregatedBLSSignature, error)
...
+
+	for i := 0; i < len(aggQC.ValidatorsTimeoutHighQCViews); i++ {
+		if aggQC.ValidatorsTimeoutHighQCViews[i] != other.ValidatorsTimeoutHighQCViews[i] {
+			return false
+		}
+	}
+
+	return aggQC.TimedOutView == other.TimedOutView &&
+		aggQC.ValidatorsHighQC.Eq(other.ValidatorsHighQC) &&
+		aggQC.ValidatorsTimeoutAggregatedSignature.Eq(other.ValidatorsTimeoutAggregatedSignature)
```
lazynina, line 462: Is it more efficient to perform this check before looping over the aggQC.ValidatorsTimeoutHighQCViews?

lib/block_view_types.go
```diff
@@ -4647,6 +4647,43 @@ func DecodeByteArray(reader io.Reader) ([]byte, error) {
...
+	return data
+}
+
+func DecodeUint64Array(reader io.Reader) ([]uint64, error) {
+	pkLen, err := ReadUvarint(reader)
+	if err != nil {
+		return nil, errors.Wrapf(err, "DecodeUint64Array: Problem reading array length")
+	}
+
+	if pkLen <= 0 {
```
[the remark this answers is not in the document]
tholonious: Returning nil matches the pattern in all other decodes in this file, so returning nil is more for consistency. An empty golang slice has the same observable behavior as an empty slice, so it shouldn't make much difference: https://stackoverflow.com/questions/44305170/nil-slices-vs-non-nil-slices-vs-empty-slices-in-go-language

lib/network.go
```diff
@@ -2151,6 +2247,63 @@ func DecodeHeaderVersion1(rr io.Reader) (*MsgDeSoHeader, error) {
...
+	// during both encoding and decoding.
+	retHeader.Nonce = 0
+	retHeader.ExtraNonce = 0
+
+	// ValidatorsVoteQC
+	validatorsVoteQC, err := DecodeQuorumCertificate(rr)
+	if err != nil {
+		return nil, errors.Wrapf(err, "MsgDeSoHeader.FromBytes: Problem decoding ValidatorsVoteQC")
+	}
+	retHeader.ValidatorsVoteQC = validatorsVoteQC
```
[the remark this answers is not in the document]
tholonious: Ah, good catch. This works

lib/constants.go
tholonious: Sure, it's a simple enough change to do update the type here. I created a new type `MsgDeSoHeaderVersion` and set it as an alias to uint32 so everything compiles.

lib/pos_network.go
tholonious: Hmm, the code reads a bit more cleanly with your suggestion. Not much difference in terms of efficiency though since `aggQC.ValidatorsTimeoutAggregatedSignature.Eq` is an O(n) check. I've updated the code to reorder the checks

lib/network_test.go
tholonious: Sounds good. I had to create a new file for it with a `!relic` tag. LMK if you want me to place the file anywhere specific

lib/network.go
```diff
@@ -2151,6 +2247,61 @@ func DecodeHeaderVersion1(rr io.Reader) (*MsgDeSoHeader, error) {
...
+		retHeader.Height = binary.BigEndian.Uint64(scratchBytes[:])
+	}
+
+	// The Nonce and ExtraNonce fields are unused in version 2. We skip them
+	// during both encoding and decoding.
+	retHeader.Nonce = 0
+	retHeader.ExtraNonce = 0
+
+	// ValidatorsVoteQC
+	retHeader.ValidatorsVoteQC, err = DecodeQuorumCertificate(rr)
```
AeonSw4n, line 2291: I've noticed you're using methods for encoders and static functions for decoders. E.g. here `ToBytes` is a method and `Decode...` is a static function. I think we tend to make both of them methods in the codebase. I don't have a preference if we're consistent or not but just want to make sure we're aware of changing the pattern here.

lib/pos_network.go
```diff
@@ -400,3 +400,129 @@ func DecodeAggregatedBLSSignature(rr io.Reader) (*AggregatedBLSSignature, error)
...
+	// Note: The aggregated signature is made up of partial signatures for all
+	// present validators, each of whom signed a payload with the pair
+	// (current view, the validator's local HighQC.View).
+	ValidatorsTimeoutHighQCViews         []uint64
+	ValidatorsTimeoutAggregatedSignature *AggregatedBLSSignature
+}
+
+// Performs a deep equality check between two TimeoutAggregateQuorumCertificate, and
+// returns true if the values of the two are identical.
+func (aggQC *TimeoutAggregateQuorumCertificate) Eq(
```
AeonSw4n, line 435: This is a nice in-depth deep comparison. But there is also a dumber method: encode both objects to bytes and see if they are equal. Assuming encoding is bijective that should be chill. The current code is fine to keep, but maybe in the future, the dumb comparison will save u time.

lib/constants.go
```diff
@@ -101,7 +103,24 @@ const (
...
+	// This version introduces the transition from Proof of Work to Proof of Stake blocks.
+	// It includes several changes to the header format:
+	// - Nonce field is deprecated
+	// - ExtraNonce field is deprecated
+	// - ValidatorsVoteQC field is added
+	// - ValidatorsTimeoutAggregateQC field is added
+	//
+	// This format change is a breaking change that is not backwards-compatible with
+	// versions 0 and 1.
+	HeaderVersion2 = MsgDeSoHeaderVersion(2)
```
AeonSw4n, line 116: I was also adding a new header version in one of my PRs. Will merge with your branch to make it consistent. One thing I considered is naming it something PoS-labeled, e.g. `HeaderPoSVersion1` or `PoSHeaderVersion1`. You know, adds respect :D

AeonSw4n (changes requested):
Yeah looks good! Only minor stuff. I'm setting "Request changes" because my local tests weren't passing for:
- TestHeaderConversionAndReadWriteMessage
- TestHeaderBundleSerialization

Can you double-check if these are passing for you, and I'll approve?

tholonious:
> Yeah looks good! Only minor stuff. I'm setting "Request changes" because my local tests weren't passing for:
> 
> * TestHeaderConversionAndReadWriteMessage
> * TestHeaderBundleSerialization
> 
> Can you double-check if these are passing for you, and I'll approve?

@AeonSw4n, you need to add `-tags relic` when compiling core to run all unit tests locally. The two tests above both run MsgDeSoHeader serialization/deserialization with the PoS QC fields. Both tests pass for me locally, and pass in CI.

lib/network.go
tholonious: Makes sense. Great catch!

lib/pos_network.go
tholonious: Ah this is clever. It shouldn't be possible for the same byte encoded object to decode into two different struct values, so this will work. Thanks for the suggestion.

lib/constants.go
tholonious: Open to doing either. Slight preference to having the const name purely handle version incrementing and contextualize how to compare header versions. Ex: it's clear from the names that HeaderVersion2 comes after HeaderVersion0, whereas we lose that if we start from PoSHeaderVersion1 again.

If we need to identify PoS header versions explicitly, we can define a `MsgDeSoHeaderVersion.IsPoSHeader()` that returns true/false.

lib/pos_network.go
tholonious: Ah, this would be nice! The one downside is that `aggQC .ToBytes()` may return an error if the aggQC has nil fields that cannot serialize to bytes. In that case, two otherwise equivalent `TimeoutAggregateQuorumCertificate` structs values may end up with an undefined equality result. We need to keep the existing field by field deep comparison.

lib/constants.go
lazynina: Agree w/ @tholonious on this. header versioning shouldn't take into account our consensus mechanism, even in naming form.

lib/block_view_types.go
```diff
@@ -4647,6 +4647,43 @@ func DecodeByteArray(reader io.Reader) ([]byte, error) {
...
+
+func DecodeUint64Array(reader io.Reader) ([]uint64, error) {
+	arrLen, err := ReadUvarint(reader)
+	if err != nil {
+		return nil, errors.Wrapf(err, "DecodeUint64Array: Problem reading array length")
+	}
+
+	if arrLen <= 0 {
+		return nil, nil
+	}
```
lazynina, line 4669: if arrLen == 0, should we return an empty slice instead of a nil value? Also, a uint can't be less than 0.
```suggestion
	if arrLen == 0 {
		return []uint64{}, nil
	}
```

lib/network.go
```diff
@@ -2151,6 +2247,61 @@ func DecodeHeaderVersion1(rr io.Reader) (*MsgDeSoHeader, error) {
...
+		if err != nil {
+			return nil, errors.Wrapf(err, "MsgDeSoHeader.FromBytes: Problem decoding TstampSecs")
+		}
+		retHeader.TstampSecs = binary.BigEndian.Uint64(scratchBytes[:])
+	}
+
+	// Height
+	{
+		scratchBytes := [8]byte{}
+		_, err := io.ReadFull(rr, scratchBytes[:])
```
lazynina, line 2278: maybe we should write a big endian encode/decode?

lib/pos_network.go
```diff
@@ -384,19 +384,143 @@ func (sig *AggregatedBLSSignature) ToBytes() ([]byte, error) {
...
+
+	// Here we include a list of the HighQC.View values we got from each of the
+	// validators in the ValidatorsTimeoutHighQCViews field. In addition, for each
+	// unique HighQC.View value we received, we combine all the partial signatures
+	// for that HighQC.View into a single BLSMultiSignature.
+	//
+	// Note: The aggregated signature is made up of partial signatures for all
+	// present validators, each of whom signed a payload with the pair
+	// (current view, the validator's local HighQC.View).
+	ValidatorsTimeoutHighQCViews         []uint64
```
lazynina, line 428: How is this slice ordered? do we need to be concerned about non-determinism in how these uint64s are ordered? or does each index map to a validator somehow? If so, can we add a little bit of context about the ordering please?

lib/network_test.go
lazynina: I like how you handled it - good stuff.

lazynina (changes requested):

lib/pos_network.go
tholonious: The ordering is decoupled from the schema here. I can add a comment here for context though.

The ordering of view numbers here will correspond to the ordering of the active validators in the `SnapshotValidatorByStatusAndStake` badger index. The index orders validators by `EpochNumber > Status > TotalStakeAmountNanos > ValidatorPKID`.

This array will have the highest stake active validator index 0, and list the remaining active validators in decreasing order of stake.

lib/block_view_types.go
tholonious: Returning nil matches the decoder pattern elsewhere. Nil slice and an initialized empty slice have the same observable behavior in golang. No practical difference between the two. See https://stackoverflow.com/questions/44305170/nil-slices-vs-non-nil-slices-vs-empty-slices-in-go-language.

AeonSw4n (approved):
All passed this time!

lib/network.go
tholonious: The code uses `binary.BigEndian.PutUint64(...)` to do the encoding and `binary.BigEndian.Uint64(...)` for decoding.

lazynina (approved):

PR #561, opened 2023-06-21
PoS Mempool Data Structures

AeonSw4n:

PR #562, opened 2023-06-21
PoS Mempool Transaction Register

AeonSw4n:


AeonSw4n:
* feature/pos-mempool
  * **PR #561** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/561" target="_blank">Graphite</a> 
    * **PR #562** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/562" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/562?utm_source=stack-comment).

lib/blockchain.go
```diff
@@ -1465,7 +1465,7 @@ func _FindCommonAncestor(node1 *BlockNode, node2 *BlockNode) *BlockNode {
 	return node1
 }
 
-func CheckTransactionSanity(txn *MsgDeSoTxn, blockHeight uint32, params *DeSoParams) error {
+func CheckTransactionSanity(txn *MsgDeSoTxn, blockHeight uint64, params *DeSoParams) error {
```
tholonious, line 1468: Rather than maving the uint32 to uint64 switch here, I'd suggestion adding a TODO a making the switch in a separate PR. BlockHeight is passed around as a uint32 in many places, so it will be less error prone if we do the switch in one pass holistically across the codebase where needed.

lib/pos_mempool.go
```diff
@@ -0,0 +1,446 @@
...
+}
+
+// ========================
+//	FeeBucket
+// ========================
+
+// FeeBucket is a data structure storing time buckets. Time buckets, in turn, store timestamp-ordered mempool transactions
+// that have neighbouring fees. When a transaction is added to the FeeBucket, an appropriate TimeBucket is found or created,
+// depending on the transaction's fee, and the transactions is inserted into it.
+type FeeBucket struct {
```
tholonious, line 108: Do we really need a separate FeeBucket data structure? It seems like you can simplify things by merging the FeeBucket data structure with the TransactionRegister data structure. Namely, my suggestion would be that what we actually want is two data structures only:
- A FeeTimeBucket data structure that's identical to your existing TimeBucket struct. This construct  knows its min/max fee rates, and maintains a time ordered treeset of txns for a single fee time ordered bucket
- A TransactionRegister data structure, which maintains a treeset of FeeTimeBuckets ordered by fee bucket

It makes sense that why you would wanted to break things up into three separate structs. But while reading the code, the middle FeeBucket struct ended up actually obfuscating what was happening. Collapsing things into 2 data structures, and updating TimeBucket name to FeeTimeBucket would make it more intuitive to the reader how txn storage is organized.

lib/pos_mempool.go
```diff
@@ -0,0 +1,446 @@
...
+	return &FeeBucket{
+		timeBucketSet:   timeBucketSet,
+		timeBucketIndex: make(map[int]*TimeBucket),
+		globalParams:    globalParams,
+	}
+}
+
+// FeeBucketComparator is a comparator function for TimeBuckets. It is used to order TimeBuckets in
+// the FeeBucket's timeBucketSet.
+func FeeBucketComparator(a, b interface{}) int {
```
tholonious, line 129: This is really a fee comparator for TimeBuckets. The name "FeeBucketComparator" is a bit misleading. See my suggestion above on collapsing the three data structures into two. If you do that, then you can rename this function to FeeTimeBucketComparator, which is more in line with its intended use.

lib/pos_mempool.go
```diff
@@ -0,0 +1,446 @@
...
+		index:        index,
+		txnsSet:      txnsSet,
+		globalParams: globalParams,
+	}
+}
+
+// TimeBucketComparator is a comparator function for MempoolTx transactions stored inside a TimeBucket. The comparator
+// orders the transactions by smallest timestamp. In case of a tie, transactions are ordered by greatest fee rate. Finally,
+// in case of another tie, transactions are ordered by their hash.
+func TimeBucketComparator(a, b interface{}) int {
```
tholonious, line 312: This function is a time order comparator for txns. It would be clearer to rename it to something like `MempoolTxTimeOrderComparator`

lib/pos_mempool.go
```diff
@@ -0,0 +1,446 @@
...
+}
+
+// FeeBucketComparator is a comparator function for TimeBuckets. It is used to order TimeBuckets in
+// the FeeBucket's timeBucketSet.
+func FeeBucketComparator(a, b interface{}) int {
+	aVal := a.(*TimeBucket)
+	bVal := b.(*TimeBucket)
+
+	// Determine the TimeBucket with a higher fee range. We can easily find out by comparing TimeBucket indices.
+	if aVal.index < bVal.index {
```
tholonious, line 134: Comparing the indices assumes that a and b are part of the same parent FeeBucket instance. Should we compare their fee ranges instead so that this function always works regardless of where a and b were initialized?

lib/pos_mempool.go
```diff
@@ -0,0 +1,446 @@
...
+//	TimeBucket
+// ========================
+
+// TimeBucket is a data structure storing MempoolTx with similar fee rates. It has an index, which determines the
+// range of fee rates that are accepted by the TimeBucket.
+type TimeBucket struct {
+	// feeMin is the minimum fee rate (inclusive) accepted by the TimeBucket, in nanos per KB.
+	feeMin uint64
+	// feeMax is the maximum fee rate (inclusive) accepted by the TimeBucket, in nanos per KB.
+	feeMax uint64
```
tholonious, line 287: Nit, update these fields so their units are contextualized in the name
```suggestion
	minFeeDESONanosPerKB uint64
	// feeMax is the maximum fee rate (inclusive) accepted by the TimeBucket, in nanos per KB.
	maxFeeDESONanosPerKB uint64
```

lib/pos_mempool.go
```diff
@@ -0,0 +1,446 @@
...
+func NewTimeBucket(index int, globalParams *GlobalParamsEntry) *TimeBucket {
+	txnsSet := treeset.NewWith(TimeBucketComparator)
+
+	feeMin, feeMax := ComputeTimeBucketRangeFromIndex(index, globalParams)
+	return &TimeBucket{
+		feeMin:       feeMin,
+		feeMax:       feeMax,
+		index:        index,
+		txnsSet:      txnsSet,
+		globalParams: globalParams,
```
tholonious, line 305: Any reason to store globalParams as a field on TimeBucket? Doesn't seem like the field is used anywhere else.

lib/pos_mempool.go
```diff
@@ -0,0 +1,446 @@
...
+		return -1
+	} else {
+		if aVal.FeePerKB < bVal.FeePerKB {
+			return 1
+		} else if aVal.FeePerKB > bVal.FeePerKB {
+			return -1
+		} else {
+			return HashToBigint(aVal.Hash).Cmp(HashToBigint(bVal.Hash))
+		}
+	}
```
tholonious, line 328: No need to nest the if blocks. if the first 2 conditions fails, then it's guaranteed for all remaining cases that the two txns have the same timestamp.
```suggestion
	} else if aVal.FeePerKB < bVal.FeePerKB {
		return 1
	} else if aVal.FeePerKB > bVal.FeePerKB {
		return -1
	} else {
		return HashToBigint(aVal.Hash).Cmp(HashToBigint(bVal.Hash))
	}
```

tholonious (changes requested):

lib/pos_mempool.go
```diff
@@ -0,0 +1,446 @@
...
+type TransactionRegister struct {
+	// buckets is the data structure storing the transactions in fee-time order.
+	buckets *FeeBucket
+	// txnMembership is a map of transaction hashes. It is used to determine existence of a transaction in the register.
+	txnMembership map[BlockHash]struct{}
+	// totalTxnSize is the total size of all transactions in the register.
+	totalTxnSize uint64
+
+	params       *DeSoParams
+	globalParams *GlobalParamsEntry
```
tholonious, line 28: We store the full globalParams here, but we really only need to store the the base fee rate and multiplier. IMO, you can extract those two fields, compute their big.Float values, and store them as separate fields here instead.

That way, if the GlobalParamsEntry is mutated externally, it won't break the fee bucket computation.

lib/pos_mempool.go
```diff
@@ -0,0 +1,446 @@
...
+// ComputeTimeBucketRangeFromIndex takes a fee range index and the GlobalParamsEntry, and returns the
+// [feeMin, feeMax] of the fee range.
+func ComputeTimeBucketRangeFromIndex(index int, globalParams *GlobalParamsEntry) (_min uint64, _max uint64) {
+	feeMin := computeTimeBucketMinFromIndex(index, globalParams)
+	feeMax := computeTimeBucketMinFromIndex(index+1, globalParams) - 1
+	return feeMin, feeMax
+}
+
+// ComputeTimeBucketIndexFromFeePerKBNanos takes a fee rate and the GlobalParamsEntry, and returns the fee range index.
+func ComputeTimeBucketIndexFromFeePerKBNanos(feePerKBNanos uint64, globalParams *GlobalParamsEntry) int {
```
tholonious, line 417:
```suggestion
func ComputeTimeBucketIndexFromFeePerKBNanos(feePerKBNanos uint64, feeBucketBaseRateNanos *big.Float, feeBucketMultiplier *big.Float) uint64 {
```

lib/pos_mempool.go
```diff
@@ -0,0 +1,446 @@
...
+	multiplier := BigFloatPow(feeBucketMultiplier, pow)
+	fee := NewFloat().Mul(feeBucketBaseRate, multiplier)
+
+	feeUint64, _ := fee.Uint64()
+	return feeUint64
+}
+
+// ComputeTimeBucketRangeFromIndex takes a fee range index and the GlobalParamsEntry, and returns the
+// [feeMin, feeMax] of the fee range.
+func ComputeTimeBucketRangeFromIndex(index int, globalParams *GlobalParamsEntry) (_min uint64, _max uint64) {
```
tholonious, line 410:
```suggestion
func ComputeTimeBucketRangeFromIndex(index uint64, feeBucketBaseRateNanos *big.Float, feeBucketMultiplier *big.Float) (_min uint64, _max uint64) {
```

lib/pos_mempool.go
```diff
@@ -0,0 +1,446 @@
...
+
+	subFee := Sub(logFeeFloat, logBaseRate)
+	// If the fee is less than the base rate, return 0.
+	if subFee.Cmp(NewFloat().SetFloat64(0)) < 0 {
+		return 0
+	}
+
+	divFee := Div(subFee, logMultiplier)
+	timeBucketIndexUint64, _ := divFee.Uint64()
+	timeBucketIndex := int(timeBucketIndexUint64)
```
tholonious, line 434: Why cast to int? It shouldn't be possible for the index to be < 0.

lib/pos_mempool.go
```diff
@@ -0,0 +1,446 @@
...
+
+	feeBucketBaseRate := NewFloat().SetUint64(globalParams.FeeBucketBaseRate)
+	feeBucketMultiplier := NewFloat().SetUint64(10000 + globalParams.FeeBucketMultiplierBasisPoints)
+	feeBucketMultiplier.Quo(feeBucketMultiplier, NewFloat().SetUint64(10000))
+	return feeBucketBaseRate, feeBucketMultiplier
+}
+
+// computeTimeBucketMinFromIndex takes a fee range index and the GlobalParamsEntry, and returns the minimum fee rate
+// of this fee range.
+func computeTimeBucketMinFromIndex(index int, globalParams *GlobalParamsEntry) uint64 {
```
tholonious, line 390: No need to pass in the entire global params here. The caller can extract just the required params and pass them in directly. This way, the function signature becomes self documenting re: what data it actually needs. 
```suggestion
func computeTimeBucketMinFromIndex(index uint64, feeBucketBaseRateNanos *big.Float, feeBucketMultiplier *big.Float) uint64 {
```

lib/pos_mempool.go
```diff
@@ -0,0 +1,446 @@
...
+// ========================
+//	TimeBucket
+// ========================
+
+// TimeBucket is a data structure storing MempoolTx with similar fee rates. It has an index, which determines the
+// range of fee rates that are accepted by the TimeBucket.
+type TimeBucket struct {
+	// feeMin is the minimum fee rate (inclusive) accepted by the TimeBucket, in nanos per KB.
+	feeMin uint64
+	// feeMax is the maximum fee rate (inclusive) accepted by the TimeBucket, in nanos per KB.
```
tholonious, line 286: It shouldn't be possible for both the feeMin and feeMax to be inclusive, right?. Ex, say we have two buckets:
- bucket one has the inclusive fee range [0, 2]
- bucket two has the inclusive fee range [2, 4]
If we receive a transaction with fee rate 2, then we don't want it to be ambiguous which bucket it falls into.

lib/pos_mempool.go
```diff
@@ -0,0 +1,446 @@
...
+	feeMin := computeTimeBucketMinFromIndex(index, globalParams)
+	feeMax := computeTimeBucketMinFromIndex(index+1, globalParams) - 1
+	return feeMin, feeMax
+}
+
+// ComputeTimeBucketIndexFromFeePerKBNanos takes a fee rate and the GlobalParamsEntry, and returns the fee range index.
+func ComputeTimeBucketIndexFromFeePerKBNanos(feePerKBNanos uint64, globalParams *GlobalParamsEntry) int {
+	feeBucketBaseRate, feeBucketMultiplier := computeFeeBucketBaseAndMultiplierFromGlobalParams(globalParams)
+
+	// Compute index = (log(feePerKBNanos) - log(feeBucketBaseRate)) / log(feeBucketMultiplier).
```
tholonious, line 420: Neat! 💯

lib/pos_mempool.go
```diff
@@ -0,0 +1,446 @@
...
+	feeFloat := NewFloat().SetUint64(feePerKBNanos)
+	logFeeFloat := BigFloatLog(feeFloat)
```
tholonious, line 422: You can perform this check earlier before taking the logs of the base rate and multiplier
```suggestion
	// If the fee is less than the base rate, return 0.
	if feeFloat.Cmp(feeBucketBaseRate) < 0 {
		return 0
	}	
	logFeeFloat := BigFloatLog(feeFloat)
```

lib/pos_mempool.go
```diff
@@ -0,0 +1,446 @@
...
+	logBaseRate := BigFloatLog(feeBucketBaseRate)
+	logMultiplier := BigFloatLog(feeBucketMultiplier)
+
+	subFee := Sub(logFeeFloat, logBaseRate)
+	// If the fee is less than the base rate, return 0.
+	if subFee.Cmp(NewFloat().SetFloat64(0)) < 0 {
+		return 0
+	}
```
tholonious, line 430:
```suggestion
```

lib/pos_mempool.go
AeonSw4n: Good point. By "inclusive" here, I meant that any comparison uses sharp inequality (>= or <=). So in this case, we must ensure bucket ranges are mutually exclusive. Fortunately, we do it. To find the range of a bucket with index n, we use computeTimeBucketMinFromIndex, which gives you feeMin(n). We then set feeMax(n) = feeMin(n+1) - 1. This prevents any overlap between ranges.

lib/block_view_types.go
```diff
@@ -3772,6 +3772,16 @@ type GlobalParamsEntry struct {
...
+
+	// TODO: Properly handle the newly added GlobalParamsEntry fields
+	// FeeBucketBaseRate is the minimal FeeRatePerKB a transaction can have.
+	// If a transaction has a lower fee than FeeBucketBaseRate, it will be
+	// rejected by the node's mempool.
+	FeeBucketBaseRate uint64
+
+	// FeeBucketMultiplierBasisPoints is the rate of growth of the fee bucket ranges.
+	// The multiplier is given as basis points.
+	FeeBucketMultiplierBasisPoints uint64
```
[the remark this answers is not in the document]
AeonSw4n: This sounds reasonable. I used basis points to parallel the creator coin convention. But hundredths seem sufficient enough. The other thing is, I think diamondhands was a proponent of using the basis point level of granularity. To make sure we don't switch back and forth, I added this item to our PoS discussion docs.

lib/blockchain.go
AeonSw4n: Good point. This change is a leftover from splitting my mempool code lol. I will remove the change from this PR.

lib/pos_mempool.go
AeonSw4n: I like!

lib/pos_mempool.go
AeonSw4n: Yup! I think I replied to this off-gh, so yeah, I totally agree.

lib/pos_mempool.go
```diff
@@ -0,0 +1,446 @@
...
+	return txns
+}
+
+//============================================
+//	Fee-Time Bucket Math
+//============================================
+
+// computeFeeBucketBaseAndMultiplierFromGlobalParams takes the fee base rate and multiplier for the GlobalParamsEntry,
+// and returns them as big.Floats.
+func computeFeeBucketBaseAndMultiplierFromGlobalParams(globalParams *GlobalParamsEntry) (
```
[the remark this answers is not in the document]
AeonSw4n: Reasonable.

lib/pos_mempool.go
AeonSw4n: Thanks for the more detailed comment :D totally using it

lib/pos_mempool.go
AeonSw4n: I've added this check to make sure this never happens:

```
	if timeBucketIndex < 0 {
		return 0
	}
```

To be eeextra careful, we could change timeBucketIndex to uint64. Though I doubt it will ever overflow int to the point that it's negative. Wyt, should we change it?

lib/pos_mempool.go
```diff
@@ -0,0 +1,397 @@
...
+		// If this condition is true, then the computed index is overestimated by 1.
+		return feeTimeBucketIndex - 1
+	} else if computeTimeBucketMinFromIndex(feeTimeBucketIndex+1,
+		feeBucketBaseRateNanos, feeBucketMultiplier) <= feePerKBNanos {
+		// This condition gets triggered exactly on fee bucket boundaries, i.e. when feePerKBNanos = FeeMax for some bucket.
+		// The float rounding makes the number slightly smaller like 5.9999999991 instead of 6.0.
+		return feeTimeBucketIndex + 1
+	} else {
+		return feeTimeBucketIndex
+	}
```
lazynina, line 396: avoid else when possible 😄 maebeam beat this into my head.
```suggestion
	}
	if computeTimeBucketMinFromIndex(feeTimeBucketIndex+1,
		feeBucketBaseRateNanos, feeBucketMultiplier) <= feePerKBNanos {
		// This condition gets triggered exactly on fee bucket boundaries, i.e. when feePerKBNanos = FeeMax for some bucket.
		// The float rounding makes the number slightly smaller like 5.9999999991 instead of 6.0.
		return feeTimeBucketIndex + 1
	}
	return feeTimeBucketIndex
```

lib/pos_mempool_test.go
```diff
@@ -0,0 +1,323 @@
...
+	}
+	fmt.Println()
+
+	set.Remove(22)
+
+	it = set.Iterator()
+	for it.Next() {
+		fmt.Printf("%v ", it.Value())
+	}
+	fmt.Println()
```
lazynina, line 27: let's remove print statements? or use some asserts to make this test more useful.

lib/pos_mempool_test.go
```diff
@@ -0,0 +1,323 @@
...
+			} else {
+				timeComparison := MempoolTxTimeOrderComparator(txnPool[i], txnPool[j])
+				if timeComparison == 1 {
+					return false
+				} else if timeComparison == -1 {
+					return true
+				} else {
+					return true
+				}
+			}
```
lazynina, line 184: reduce complexity of conditionals. we really only want to check if timeComparison != 1
```suggestion
			}
			if feeComparison == -1 {
				return true
			}
			timeComparison := MempoolTxTimeOrderComparator(txnPool[i], txnPool[j])
			return timeComparison != 1
```

lib/pos_mempool_test.go
```diff
@@ -0,0 +1,323 @@
...
+			if timeComparison == 1 {
+				return false
+			} else if timeComparison == -1 {
+				return true
+			} else {
+				return true
+			}
+		})
+	}
+	return txnPool
```
lazynina, line 199: we can simplify this one a lot.
```suggestion
	return txnPool
	}
	sort.Slice(txnPool, func(i, j int) bool {
		timeComparison := MempoolTxTimeOrderComparator(txnPool[i], txnPool[j])
		return timeComparison != 1
	})
	return txnPool
```

lazynina, line 199: ok this is probably the best way to simplify the whole function
```suggestion
	sort.Slice(txnPool, func(i, j int) bool {
		if !timeOnly {
			timeBucketI := _testMapMempoolTxnToTimeBucket(txnPool[i], globalParams)
			timeBucketJ := _testMapMempoolTxnToTimeBucket(txnPool[j], globalParams)

			feeComparison := FeeTimeBucketComparator(timeBucketI, timeBucketJ)
			if feeComparison == 1 {
				return false
			}
			if feeComparison == -1 {
				return true
			}
		}
		timeComparison := MempoolTxTimeOrderComparator(txnPool[i], txnPool[j])
		return timeComparison != 1
	})
	return txnPool
```

lazynina (changes requested):

lib/pos_mempool_test.go
```diff
@@ -0,0 +1,323 @@
...
+
+	verifyFeeBucket := func(n int, fee uint64) bool {
+		feeMin, feeMax := ComputeTimeBucketRangeFromIndex(n, baseRate, bucketMultiplier)
+		if fee < feeMin {
+			return false
+		}
+		if fee > feeMax {
+			return false
+		}
+		return true
```
lazynina, line 314:
```suggestion
		return fee > feeMin && fee < feeMax
```

tholonious (changes requested):

lib/pos_mempool.go
```diff
@@ -0,0 +1,418 @@
...
+// TransactionRegister is the primary transaction store of the PoS Mempool. The register is responsible for determining
+// the Fee-Time ordering of transactions. The operations supported by the register are: adding a transaction, removing
+// a transaction, iterating through all transactions in fee-time order, and retrieving all transactions ordered in
+// fee-time. The TransactionRegister doesn't perform any validation on the transactions, it just accepts the provided
+// MempoolTx and adds it to the appropriate FeeTimeBucket.
+type TransactionRegister struct {
+	// feeTimeBucketSet is a set of FeeTimeBucket objects. The set is ordered by the FeeTimeBucket's ranges, based on FeeTimeBucketComparator.
+	feeTimeBucketSet *treeset.Set
+	// feeTimeBucketsByMinFeeMap is a map of FeeTimeBucket minimum fees to FeeTimeBucket objects. It is used to quickly find
+	// a FeeTimeBucket given its fee. Note that using a map here is preferable over e.g. a slice, since the index of a
```
tholonious, line 25:
```suggestion
	// a FeeTimeBucket given its min fee. Note that using a map here is preferable over e.g. a slice, since the index of a
```

lib/pos_mempool.go
```diff
@@ -0,0 +1,418 @@
...
+		totalTxnSize:                       0,
+		params:                             params,
+		minimumNetworkFeeNanosPerKB:        minNetworkFee,
+		feeBucketRateMultiplierBasisPoints: bucketMultiplier,
+	}
+}
+
+// FeeTimeBucketComparator is a comparator function for FeeTimeBucket objects. It is used to order FeeTimeBucket objects
+// in the TransactionRegister's feeTimeBucketSet based on fee ranges (higher fee ranges are ordered first).
+func FeeTimeBucketComparator(a, b interface{}) int {
```
tholonious, line 60: Nit, scoping. This function is an implementation detail of the TransactionRegister, so we don't want to be expored outside of the current package.
```suggestion
func feeTimeBucketComparator(a, b interface{}) int {
```

lib/pos_mempool.go
```diff
@@ -0,0 +1,418 @@
...
+	if tr.txnMembership.Includes(*txn.Hash) {
+		return fmt.Errorf("TransactionRegister.AddTransaction: Transaction already exists in register")
+	}
+
+	// If the transaction overflows the maximum mempool size, reject it.
+	if tr.totalTxnSize+txn.TxSizeBytes > tr.params.MaxMempoolPosSizeBytes || tr.totalTxnSize > math.MaxUint64-txn.TxSizeBytes {
+		return fmt.Errorf("TransactionRegister.AddTransaction: Transaction size exceeds maximum mempool size")
+	}
+
+	// Determine the index of the bucket based on the transaction's fee rate.
```
tholonious, line 95:
```suggestion
	// Determine the min fee of the bucket based on the transaction's fee rate.
```

lib/pos_mempool.go
```diff
@@ -0,0 +1,418 @@
...
+		}
+	} else {
+		// If the bucket doesn't exist, create it and add the transaction to it.
+		newBucket := NewFeeTimeBucket(bucketMinFeeNanosPerKb, bucketMaxFeeNanosPerKB)
+		if err := newBucket.AddTransaction(txn); err != nil {
+			return errors.Wrapf(err, "TransactionRegister.AddTransaction: Error adding transaction to bucket: %v", err)
+		}
+		tr.feeTimeBucketSet.Add(newBucket)
+		tr.feeTimeBucketsByMinFeeMap[bucketMinFeeNanosPerKb] = newBucket
+	}
```
tholonious, line 111: Similar line of thought as as Nina's comments. Always good to avoid else cases, especially when they result in some duplicated logic
```suggestion
	bucket, bucketExists := tr.feeTimeBucketsByMinFeeMap[bucketMinFeeNanosPerKb]

	// If the bucket doesn't exist, create an empty one and store it.	
	if !bucketExists {
		bucket = NewFeeTimeBucket(bucketMinFeeNanosPerKb, bucketMaxFeeNanosPerKB)
		tr.feeTimeBucketSet.Add(bucket)
		tr.feeTimeBucketsByMinFeeMap[bucketMinFeeNanosPerKb] = bucket
	}

	// Add the transaction to the to the bucket
	if err := bucket.AddTransaction(txn); err != nil {
		return errors.Wrapf(err, "TransactionRegister.AddTransaction: Error adding transaction to bucket: ")
	}
```

lib/pos_mempool.go
```diff
@@ -0,0 +1,397 @@
...
+// TransactionRegister is the primary transaction store of the PoS Mempool. The register is responsible for determining
+// the Fee-Time ordering of transactions. The operations supported by the register are: adding a transaction, removing
+// a transaction, iterating through all transactions in fee-time order, and retrieving all transactions ordered in
+// fee-time. The TransactionRegister doesn't perform any validation on the transactions, it just accepts the provided
+// MempoolTx and adds it to the appropriate FeeTimeBucket.
+type TransactionRegister struct {
+	// timeBucketSet is a set of FeeTimeBucket objects. The set is ordered by the FeeTimeBucket's ranges, based on FeeTimeBucketComparator.
+	timeBucketSet *treeset.Set
+	// timeBucketIndex is a map of FeeTimeBucket indices to FeeTimeBucket objects. It is used to quickly find a FeeTimeBucket given its index.
+	timeBucketIndex map[int]*FeeTimeBucket
```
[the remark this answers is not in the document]
tholonious: Marking as resolved. Using a map is the right solution given the gaps in indices (now updated to min fees).

lib/pos_mempool.go
```diff
@@ -0,0 +1,418 @@
...
+
+	// We will set the mempoolTxIterator to point at the first transaction in the new FeeTimeBucket.
+	fti.mempoolTxIterator = nextFeeTimeBucket.txnsSet.Iterator()
+	if !fti.mempoolTxIterator.Next() {
+		// This shouldn't happen, but if it does, we've reached the end of the TransactionRegister.
+		return false
+	}
+
+	return true
+}
```
tholonious, line 229: I realize it's not possible to have empty buckets based on the rules in the TransactionRegister.AddTransaction and TransactionRegister.RemoveTransaction functions. That said, you can make the iterator more bullet proof by having it assume as little as possible about the implementations in those functions. This surprisingly makes the code safer, and IMO the logic much more concise
```suggestion
func (fti *FeeTimeIterator) Next() bool {
	// If we have remaining txns in the current FeeTimeBucket, then we're done. We can advance
	// to the next txn within the current FeeTimeBucket and exit early here.
	if  fti.mempoolTxIterator != nil && fti.mempoolTxIterator.Next() {
		return true
	}

	// If we have no more txns in the current FeeTimeBucket then we need to advance to the
	// next non-empty FeeTimeBucket and advance to the first transaction within that FeeTimeBucket.
	
	// Keep advancing to the next FeeTimeBucket until we find a non-empty one. Once we find a
	// non-empty FeeTimeBucket, then store a newly created iterator for that FeeTimeBucket
	for !fti.bucketIterator.Next() {		
		// Get the current FeeTimeBucket
		nextFeeTimeBucket := fti.bucketIterator.Value().(*FeeTimeBucket)
		// Initialize an iterator for the current FeeTimeBucket
		fti.mempoolTxIterator = nextFeeTimeBucket.txnsSet.Iterator()
		
		// Check if the newly found FeeTimeBucket is empty. If it's not empty, then we're done.
		if fti.mempoolTxIterator.Next() {
			return true
		}
	} 

	// We have no remaining txns in the current FeeTimeBucket and no remaining FeeTimeBuckets.
	return false
}
```

lib/pos_mempool.go
```diff
@@ -0,0 +1,418 @@
...
+
+// Value returns the transaction that the iterator is pointing at.
+func (fti *FeeTimeIterator) Value() (*MempoolTx, bool) {
+	if !fti.initialized {
+		return nil, false
+	}
```
tholonious, line 235:
```suggestion
	if fti. mempoolTxIterator == nil {
		return nil, false
	}
```

tholonious (changes requested):

tholonious:
Alright @AeonSw4n, one last round. Everything LGTM at a high level. Main comments are for the FeeTimeIterator. If you can bullet-proof it a bit and add more unit tests around its edge cases, it'll make everything quite more resilient.

lib/pos_mempool.go
tholonious, line 1: Consider renaming this file to `pos_transaction_register.go` so we have a dedicated file that houses all the fee time bucketing math.

lib/pos_mempool_test.go
```diff
@@ -0,0 +1,298 @@
...
+	_testBucketStandardAddTest(t, txnPool, globalParams, true,
+		func(tx *MempoolTx) {
+			require.Equal(nil, timeBucket.AddTransaction(tx))
+		},
+		func() []*MempoolTx {
+			return timeBucket.GetTransactions()
+		},
+	)
+}
+
```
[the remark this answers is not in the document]
AeonSw4n: I've added a test called `TestSanityCheckTransactionRegister()` that checks for these edge-cases. The last case:

> TransactionRegister with multiple FeeTimeBuckets with multiple txns

Is checked in the other tests pretty thoroughly.

lib/pos_mempool.go
```diff
@@ -0,0 +1,418 @@
...
+
+	// This condition gets triggered exactly on fee bucket boundaries, i.e. when feeNanosPerKB = FeeMax for some bucket.
+	// The float rounding makes the number slightly smaller like 5.9999999991 instead of 6.0.
+	if ComputeFeeTimeBucketMinFromIndex(feeTimeBucketIndex+1, minimumNetworkFeeNanosPerKB, feeBucketMultiplier) <= feeNanosPerKB {
+		return feeTimeBucketIndex + 1
+	}
+
+	// If we get here, then the computed index is correct.
+	return feeTimeBucketIndex
```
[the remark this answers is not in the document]
AeonSw4n: I replaced all the occurrences of "index" with "exponent." I like this terminology better.

lib/pos_mempool_test.go
```diff
@@ -0,0 +1,298 @@
...
+			timeComparison := mempoolTxTimeOrderComparator(txnPool[i], txnPool[j])
+			if timeComparison == 1 {
+				return false
+			} else if timeComparison == -1 {
+				return true
+			}
+			return true
+		})
+	}
+	return txnPool
```
lazynina, line 180: I think we can simplify the sorting logic here.
```suggestion
	sort.Slice(txnPool, func(i, j int) bool {
		if !timeOnly {
			timeBucketI := _testMapMempoolTxnToTimeBucket(txnPool[i], globalParams)
			timeBucketJ := _testMapMempoolTxnToTimeBucket(txnPool[j], globalParams)

			feeComparison := feeTimeBucketComparator(timeBucketI, timeBucketJ)
			if feeComparison == 1 {
				return false
			} else if feeComparison == -1 {
				return true
			}
		}
		
		timeComparison := mempoolTxTimeOrderComparator(txnPool[i], txnPool[j])
		return timeComparison != 1
	})
	return txnPool
```

lib/pos_mempool.go
```diff
@@ -0,0 +1,418 @@
...
+		}
+	} else {
+		// If the bucket doesn't exist, create it and add the transaction to it.
+		newBucket := NewFeeTimeBucket(bucketMinFeeNanosPerKb, bucketMaxFeeNanosPerKB)
+		if err := newBucket.AddTransaction(txn); err != nil {
+			return errors.Wrapf(err, "TransactionRegister.AddTransaction: Error adding transaction to bucket: %v", err)
+		}
+		tr.feeTimeBucketSet.Add(newBucket)
+		tr.feeTimeBucketsByMinFeeMap[bucketMinFeeNanosPerKb] = newBucket
+	}
```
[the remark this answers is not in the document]
tholonious: @lazynina I like it. It's a little easier to read

lib/pos_transaction_register.go
```diff
@@ -0,0 +1,426 @@
...
+	feeTimeBucketSet *treeset.Set
+	// feeTimeBucketsByMinFeeMap is a map of FeeTimeBucket minimum fees to FeeTimeBucket objects. It is used to quickly find
+	// a FeeTimeBucket given its min fee.
+	feeTimeBucketsByMinFeeMap map[uint64]*FeeTimeBucket
+	// txnMembership is a set of transaction hashes. It is used to determine existence of a transaction in the register.
+	txnMembership *Set[BlockHash]
+	// totalTxnSize is the total size of all transactions in the register.
+	totalTxnSize uint64
+
+	params *DeSoParams
```
tholonious, line 32: Similar comment as with the GlobalParams. It's cleaner to pull the max mempool size params in the `NewTransactionRegister` constructor and store it locally.
```suggestion
	maxTxnSizeBytes uint64
```

lib/pos_transaction_register.go
```diff
@@ -0,0 +1,426 @@
...
+type TransactionRegister struct {
+	// feeTimeBucketSet is a set of FeeTimeBucket objects. The set is ordered by the FeeTimeBucket's ranges, based on feeTimeBucketComparator.
+	feeTimeBucketSet *treeset.Set
+	// feeTimeBucketsByMinFeeMap is a map of FeeTimeBucket minimum fees to FeeTimeBucket objects. It is used to quickly find
+	// a FeeTimeBucket given its min fee.
+	feeTimeBucketsByMinFeeMap map[uint64]*FeeTimeBucket
+	// txnMembership is a set of transaction hashes. It is used to determine existence of a transaction in the register.
+	txnMembership *Set[BlockHash]
+	// totalTxnSize is the total size of all transactions in the register.
+	totalTxnSize uint64
```
tholonious, line 30: Always good to contextualize the units in the field name
```suggestion
	totalTxnSizeBytes uint64
```

lib/pos_transaction_register.go
```diff
@@ -0,0 +1,426 @@
...
+		return fmt.Errorf("TransactionRegister.RemoveTransaction: Transaction with transaction hash %v does not "+
+			"exist in the register", txn.Hash.String())
+	}
+
+	if tr.totalTxnSize < txn.TxSizeBytes {
+		return fmt.Errorf("TransactionRegister.RemoveTransaction: Transaction with transaction hash %v size %v "+
+			"exceeds total mempool size %v", txn.Hash.String(), txn.TxSizeBytes, tr.totalTxnSize)
+	}
+
+	// Determine the exponent of the bucket based on the transaction's fee rate.
```
tholonious, line 147:
```suggestion
	// Determine the min fee of the bucket based on the transaction's fee rate.
```

lib/pos_transaction_register.go
```diff
@@ -0,0 +1,426 @@
...
+	}
+
+	if txn, ok := fti.mempoolTxIterator.Value().(*MempoolTx); ok {
+		return txn, true
+	}
+	return nil, false
+}
+
+// Initialized returns true if the iterator is initialized, and false otherwise.
+func (fti *FeeTimeIterator) Initialized() bool {
```
tholonious, line 248: Nice 💯

tholonious (commented):
Looks awesome! Just one last tiny set of comments. We're good to approve and merge after these

lib/constants.go
```diff
@@ -1356,8 +1363,10 @@ var (
 		// We initialize the CreateProfileFeeNanos to 0 so we do not assess a fee to create a profile until specified by ParamUpdater.
 		CreateProfileFeeNanos: 0,
 		// We initialize the CreateNFTFeeNanos to 0 so we do not assess a fee to create an NFT until specified by ParamUpdater.
-		CreateNFTFeeNanos: 0,
-		MaxCopiesPerNFT:   0,
+		CreateNFTFeeNanos:                  0,
+		MaxCopiesPerNFT:                    0,
+		FeeBucketBaseRateNanosPerKB:        0,
+		FeeBucketRateMultiplierBasisPoints: 0,
```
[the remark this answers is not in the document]
tholonious: Agreed with @AeonSw4n. The prod and testnet configuration we use have non-zero  `MinimumNetworkFeeNanosPerKB ` initialized, so this is generally safe.

tholonious (approved):
💯

PR #563, opened 2023-06-21
PoS Transaction Register Prune

AeonSw4n:


AeonSw4n:
* feature/pos-mempool
  * **PR #561** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/561" target="_blank">Graphite</a> 
    * **PR #562** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/562" target="_blank">Graphite</a> 
      * **PR #563** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/563" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/563?utm_source=stack-comment).

lib/pos_transaction_register.go
```diff
@@ -186,6 +223,74 @@ func (tr *TransactionRegister) GetFeeTimeTransactions() []*MempoolTx {
...
+
+	// Remove the transactions from prunedTxns.
+	for _, txn := range prunedTxns {
+		if err := tr.removeTransactionNoLock(txn); err != nil {
+			return nil, errors.Wrapf(err, "TransactionRegister.Prune: Error removing transaction %v", txn.Hash.String())
+		}
+	}
+	return prunedTxns, nil
+}
+
```
tholonious, line 293: Rather than having a nested for loop with two separate `break` statements, you can split the logic in this function into a helper function called `getTransactionsToPrune` that handles the nested iteration and returns as soon as it discovers the first n transactions to prune. This will read a bit more cleanly.

```suggestion
	prunedTxns, prunedBytes, err := tr.getTransactionsToPrune(minBytesToPrune)
	if err != nil {
		return nil, errors.Wrapf(err, "TransactionRegister.Prune: Error computings transactions to prune: ")
	}

	// Remove the transactions from prunedTxns.
	for _, txn := range prunedTxns {
		if err := tr.removeTransactionNoLock(txn); err != nil {
			return nil, errors.Wrapf(err, "TransactionRegister.Prune: Error removing transaction %v", txn.Hash.String())
		}
	}
	return prunedTxns, nil
}

func (tr *TransactionRegister) getTransactionsToPrune(minBytesPruned uint64) (_prunedTxns []*MempoolTx, _prunedBytes uint64, _err error) {

	// If the minimum number of bytes to prune is 0, return.
	if minBytesPruned == 0 {
		return nil, nil
	}

	// If the register is empty, return.
	if tr.Empty() {
		return nil, nil
	}

	prunedBytes := uint64(0)
	prunedTxns := []*MempoolTx{}

	// Find the FeeTime bucket at the end of the Set. It'll have the smallest fee among the buckets in the register.
	it := tr.feeTimeBucketSet.Iterator()
	it.End()
	// Iterate through the buckets in reverse order so that we drop transactions ordered by least-to-highest priority.
	for it.Prev() {
		bucket, ok := it.Value().(*FeeTimeBucket)
		if !ok {
			return nil, fmt.Errorf("TransactionRegister.getTransactionsToPrune: Error casting value of FeeTimeBucket")
		}

		// Iterate through the transactions in the current FeeTime bucket. We iterate in reverse order, starting from the
		// end, so that we drop transactions ordered by least-to-highest priority.
		bucketIt := bucket.GetIterator()
		bucketIt.End()
		for bucketIt.Prev() {
			txn, ok := bucketIt.Value().(*MempoolTx)
			if !ok {
				return nil, fmt.Errorf("TransactionRegister.getTransactionsToPrune: Error casting value of MempoolTx")
			}
			// Make sure that, somehow, the number of pruned bytes doesn't overflow uint64.
			if prunedBytes > math.MaxUint64-txn.TxSizeBytes {
				return nil, fmt.Errorf("TransactionRegister.getTransactionsToPrune: Error pruning %v bytes from register: "+
					"prunedBytes %v exceeds max uint64", minBytesPruned, prunedBytes)
			}
			// Add the transaction to the prunedTxns list.
			prunedTxns = append(prunedTxns, txn)
			prunedBytes += txn.TxSizeBytes
			// If we've pruned sufficiently many bytes, we can break.
			if prunedBytes >= minBytesPruned {
				return prunedTxns, prunedBytes, nil
			}
		}
	}
	
	return nil, 0, nil
}
```

tholonious (changes requested):

lib/pos_transaction_register.go
```diff
@@ -19,6 +20,7 @@ import (
 // fee-time. The TransactionRegister doesn't perform any validation on the transactions, it just accepts the provided
 // MempoolTx and adds it to the appropriate FeeTimeBucket.
 type TransactionRegister struct {
+	sync.RWMutex
```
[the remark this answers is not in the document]
tholonious: Makes sense. And agreed that the current mutex usage is safe (no risk of deadlock). My only comment is that the mutex within this struct doesn't guarantee thread safety for usage of FeeTimeIterator as-is since the RWMutex is released after the FeeTimeIterator is constructed in GetFeetimeIterator().

To guarantee thread safety for the FeeTimeIterator's usage, the higher level Mempool struct will need to hold the TransactionRegister's read lock for the entire duration of time it's using the iterator. As long as that's the case within the Mempool struct, then defining the mutex here will be bulletproof.

lib/pos_transaction_register.go
```diff
@@ -186,6 +249,91 @@ func (tr *TransactionRegister) GetFeeTimeTransactions() []*MempoolTx {
...
+			txn, ok := bucketIt.Value().(*MempoolTx)
+			if !ok {
+				return nil, fmt.Errorf("TransactionRegister.getTransactionsToPrune: " +
+					"Error casting value of MempoolTx")
+			}
+			// Make sure that, somehow, the number of pruned bytes doesn't overflow uint64.
+			if prunedBytes > math.MaxUint64-txn.TxSizeBytes {
+				return nil, fmt.Errorf("TransactionRegister.getTransactionsToPrune: "+
+					"Error pruning %v bytes from register: prunedBytes %v exceeds max uint64", minPrunedBytes, prunedBytes)
+			}
```
tholonious, line 318: IMO you can remove this consistency check. Preventing size uint64 overflow is one of guarantees of the TransactionRegister struct. And you validate that each txn doesn't cause a uint64 overflow of the TransactionRegister's size before adding it to the register. Since all those checks are encapsulated in this struct, it's safe to assume here that this condition can never be hit.
```suggestion
```

lib/pos_transaction_register.go
```diff
@@ -186,6 +249,91 @@ func (tr *TransactionRegister) GetFeeTimeTransactions() []*MempoolTx {
...
+			// Add the transaction to the prunedTxns list.
+			prunedTxns = append(prunedTxns, txn)
+			prunedBytes += txn.TxSizeBytes
+			// If we've pruned sufficiently many bytes, we can break.
+			if prunedBytes >= minPrunedBytes {
+				break
```
tholonious, line 324:
```suggestion
			// If we've pruned sufficiently many bytes, then we're done. We can return early.
			if prunedBytes >= minPrunedBytes {
				return prunedTxns, nil
```

lib/pos_transaction_register.go
```diff
@@ -186,6 +249,91 @@ func (tr *TransactionRegister) GetFeeTimeTransactions() []*MempoolTx {
...
+			}
+		}
+
+		// If we've pruned sufficiently many bytes, we can break.
+		if prunedBytes >= minPrunedBytes {
+			break
+		}
```
tholonious, line 331: If you return early in line 324, then this condition isn't necessary. You can simplify the loop by removing this.
```suggestion
```

lib/pos_transaction_register.go
```diff
@@ -186,6 +249,91 @@ func (tr *TransactionRegister) GetFeeTimeTransactions() []*MempoolTx {
...
+	}
+
+	return prunedTxns, nil
```
tholonious, line 334:
```suggestion
	// If we reach this point, it means that we've iterated through the entire TransactionRegister 
	// and have no remaining transactions. We can return the txns we've found so far.
	return prunedTxns, nil
```

tholonious (changes requested):
Looks great! Just some minor comments to simplify the logic for the new reverse iteration nested loop. We're good to merge after that.

lib/pos_transaction_register.go
```diff
@@ -161,13 +168,66 @@ func (tr *TransactionRegister) RemoveTransaction(txn *MempoolTx) error {
 	return nil
 }
 
+func (tr *TransactionRegister) addBucket(bucket *FeeTimeBucket) {
+	if bucket == nil {
```
lazynina, line 172: do we need to hold the lock for add/remove bucket? I think we only use it in contexts where we already hold a write lock, but we should name it such that it's clear that it does not hold a lock

lib/pos_transaction_register_test.go
```diff
@@ -26,21 +27,225 @@ func TestSanityCheckTransactionRegister(t *testing.T) {
...
+
+	// Try pruning 0 bytes
+	txns, err := txnRegister.PruneToSize(txnRegister.totalTxnsSizeBytes)
+	require.Nil(err)
+	require.Equal(0, len(txns))
+
+	// Remove a single transaction
+	txns, err = txnRegister.PruneToSize(txnRegister.totalTxnsSizeBytes - 1)
+	require.Nil(err)
+	require.Equal(1, len(txns))
```
lazynina, line 155: You could use `require.Len` instead here if you prefer that syntax - I'm okay w/ either, just make it a little clearer that this is a length check

lib/pos_transaction_register.go
```diff
@@ -186,6 +249,91 @@ func (tr *TransactionRegister) GetFeeTimeTransactions() []*MempoolTx {
...
+
+	// If the maximum number of bytes is greater or equal to the current size of the register, return.
+	if maxSizeBytes >= tr.totalTxnsSizeBytes {
+		return nil, nil
+	}
+
+	// If the register is empty, return.
+	if tr.feeTimeBucketSet.Empty() {
+		return nil, nil
+	}
```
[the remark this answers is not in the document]
AeonSw4n: Things can break when we transition between the two locks. Though, it depends where exactly we change the locks. Imo holding an overused single writelock makes things easier.

lib/pos_transaction_register.go
```diff
@@ -294,42 +455,78 @@ func mempoolTxTimeOrderComparator(a, b interface{}) int {
...
 // FeeTimeBucket's fee range, or if the transaction hash is nil.
 func (tb *FeeTimeBucket) AddTransaction(txn *MempoolTx) error {
+	tb.Lock()
+	defer tb.Unlock()
+
 	if txn == nil || txn.Hash == nil {
 		return fmt.Errorf("FeeTimeBucket.AddTransaction: Transaction or transaction hash is nil")
 	}
 
+	if tb.txnMembership.Includes(*txn.Hash) {
```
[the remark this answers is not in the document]
AeonSw4n: I would say this is similar to the PruneToSize case. Switching locks in a concurrent environment runs a risk that some instruction will sneak its way in between.

E.g. let's say we call `AddTransaction(txn)` a bunch of times concurrently. If multiple contexts acquire the readlock at the same time, they will then sneak to the writelock and add the same transaction multiple times.

lib/pos_transaction_register_test.go
AeonSw4n: Cool! Didn't know about this one. I'm gonna start using it.

lib/pos_transaction_register.go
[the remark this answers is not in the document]
lazynina: yep - please disregard the comments about separately acquiring read and write locks.

lib/pos_transaction_register.go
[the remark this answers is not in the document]
lazynina: yeah just keep the write locks

tholonious (changes requested):

lazynina (approved):

PR #566, opened 2023-06-27
PoS Simplify PublicKey field names in P2P messages

tholonious:
Simple PR that renames fields in the MsgValidatorVote and MsgValidatorTimeout message types. It renames:
- ValidatorPublicKey -> PublicKey
- ValidatorVotingPublicKey -> VotingPublicKey

The "Validator" prefix in the field names isn't needed in the field names since it's already contextualized in the message type.

PR #565, opened 2023-06-21
PoS Add Bitset to messaging schema

tholonious:

PR #564, opened 2023-06-21
PoS Implement Bitset data structure

tholonious:

PR #567, opened 2023-06-27
PoS Add BLS public key and signature to BlockProducerInfo

tholonious:
This PR makes three changes:
1. It migrates the existing ECDSA PublicKey fields in the BlockProducerInfo from a byte slice to the strongly typed `PublicKey` type
2. It introduces versioning to BlockProducerInfo, with a byte encoder and decoder that support the new versioned schema. The existing un-versioned schema is still supported by the legacy encoder and decoder
3. It adds BLS VotingPUblicKey and VotePartialSignature fields to BlockProducerInfo that are populated in newly created Version=1 format

### DIFF
--- a/lib/network.go
+++ b/lib/network.go
@@ -2355,7 +2355,7 @@ type BlockProducerInfo struct {
 	Version BlockProducerInfoVersion
 
 	// ECDSA public key for the block producer.
-	PublicKey []byte
+	PublicKey *PublicKey
 	// The block producer's ECDSA signature for the block. This field is used in
 	// BlockProducerInfo version 0, and is deprecated from version 1 onwards.
 	Signature *btcec.Signature
@@ -2385,10 +2385,10 @@ func (bpi *BlockProducerInfo) ToBytes() ([]byte, error) {
 	encodedBytes = append(encodedBytes, bpi.Version)
 
 	// Required ECDSA PublicKey
-	if len(bpi.PublicKey) == 0 {
+	if bpi.PublicKey == nil {
 		return nil, fmt.Errorf("BlockProducerInfo.ToBytes: PublicKey is required")
 	}
-	encodedBytes = append(encodedBytes, EncodeByteArray(bpi.PublicKey)...)
+	encodedBytes = append(encodedBytes, bpi.PublicKey.ToBytes()...)
 
 	// The ECDSA Signature is redundant, and is removed in BlockProducerInfo version 1 and above
 
@@ -2426,7 +2426,7 @@ func (bpi *BlockProducerInfo) FromBytes(rr *bytes.Reader) error {
 	}
 
 	// Required ECDSA PublicKey
-	bpi.PublicKey, err = DecodeByteArray(rr)
+	bpi.PublicKey, err = ReadPublicKey(rr)
 	if err != nil {
 		return errors.Wrapf(err, "BlockProducerInfo.FromBytes: Problem reading PublicKey")
 	}
@@ -2455,7 +2455,7 @@ func (bpi *BlockProducerInfo) FromBytes(rr *bytes.Reader) error {
 func (bpi *BlockProducerInfo) Serialize_Legacy() []byte {
 	data := []byte{}
 	data = append(data, UintToBuf(uint64(len(bpi.PublicKey)))...)
-	data = append(data, bpi.PublicKey...)
+	data = append(data, bpi.PublicKey.ToBytes()...)
 
 	sigBytes := []byte{}
 	if bpi.Signature != nil {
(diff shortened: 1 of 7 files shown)


tholonious:
* feature/pos-messaging-schema
  * **PR #555** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/555" target="_blank">Graphite</a> 
    * **PR #558** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/558" target="_blank">Graphite</a> 
      * **PR #564** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/564" target="_blank">Graphite</a> 
        * **PR #565** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/565" target="_blank">Graphite</a> 
          * **PR #566** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/566" target="_blank">Graphite</a> 
            * **PR #567** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/567" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/567?utm_source=stack-comment).

lib/network.go
```diff
@@ -2351,11 +2352,101 @@ func (msg *MsgDeSoHeader) String() string {
...
+		return nil, fmt.Errorf("BlockProducerInfo.ToBytes: VotePartialSignature is required")
+	}
+	encodedBytes = append(encodedBytes, EncodeByteArray(bpi.VotePartialSignature.ToBytes())...)
+
+	return encodedBytes, nil
+}
+
+// Byte decoder for the BlockProducerInfo, with support for versioning. The encoder only
+// supports BlockProducerInfo version 1 and above. For the legacy version 0, use the
+// BlockProducerInfo.Serialize_Legacy() method instead.
```
lazynina, line 2412:
```suggestion
// BlockProducerInfo.Deserialize_Legacy() method instead.
```

lib/network.go
```diff
@@ -2351,11 +2352,101 @@ func (msg *MsgDeSoHeader) String() string {
...
+func (bpi *BlockProducerInfo) FromBytes(rr *bytes.Reader) error {
+	var err error
+
+	// Required Version field
+	bpi.Version, err = rr.ReadByte()
+	if err != nil {
+		return errors.Wrapf(err, "BlockProducerInfo.FromBytes: Problem reading Version")
+	}
+
```
lazynina, line 2421: should we throw some error is Version is 0?

lib/network.go
```diff
@@ -2351,11 +2352,101 @@ func (msg *MsgDeSoHeader) String() string {
...
+	encodedBytes := []byte{}
+
+	// Required Version field
+	encodedBytes = append(encodedBytes, bpi.Version)
+
+	// Required ECDSA PublicKey
+	if len(bpi.PublicKey) == 0 {
+		return nil, fmt.Errorf("BlockProducerInfo.ToBytes: PublicKey is required")
+	}
+	encodedBytes = append(encodedBytes, EncodeByteArray(bpi.PublicKey)...)
```
lazynina, line 2391: I think all public keys should have the same length. If so, we can just concat the bytes instead of prepending w/ the length of the byte array.

lib/network.go
```diff
@@ -2436,10 +2533,21 @@ type MsgDeSoBlock struct {
...
+	// This BlockProducerInfo field describes the producer of the block and their signature
+	// for the block.
+	//
+	// In Proof of Work blocks, this field is optional and provides the producer of the block
+	// the ability to sign the block with their private key. Doing this proves that this block
+	// was produced by a particular entity, which can be useful for nodes that want to restrict
+	// who they accept blocks from.
+	//
+	// In Proof of Stake blocks, this field is required and serves two purposes:
+	// 1. It's allows the block producer to sign their block with their ECDSA or BLS private key.
```
lazynina, line 2545:
```suggestion
	// 1. It allows the block producer to sign its block with its ECDSA or BLS private key.
```

lib/network.go
```diff
@@ -2436,10 +2533,21 @@ type MsgDeSoBlock struct {
...
+	// This allows validators to verify that the block was produced by the expected leader for the
+	// current block height and view.
+	// 2. It contains the block producer's BLS partial signature, which acts as their vote on the
+	// block. This way, their vote can be aggregated into a QC by the next block proposer in the leader
```
lazynina, line 2549:
```suggestion
	// block. This way, its vote can be aggregated into a QC by the next block proposer in the leader
```

lazynina (changes requested):

lib/network.go
tholonious: Hmm, yeah it's fair to return an error in that case. We know the BlockProducerInfo is in an unexpected format

lib/network.go
tholonious: Sounds good. We'll update the fields type the the strongly typed `PublicKey` so it can reuse the type byte encoder.

lazynina (approved):

lib/network.go
AeonSw4n: I wonder, maybe we should throw an error if `bpi.Version != BlockProducerInfoVersion1`? Potentially, if someone sets the `BlockProducerInfoVersion` to something like 2, it can conflict with future updates to `BlockProducerInfo`.

AeonSw4n (changes requested):
Nice! I just have one comment regarding the forward-compatibility of BlockProducerInfo.Version.

lib/network.go
lazynina: yeah we should validate that the version "is valid"

tholonious: Good catch. These ToBytes/FromBytes functions should verify that `bpi.Version == BlockProducerInfoVersion1` every time and error otherwise.

lazynina (approved):

PR #570, opened 2023-06-28
feature/pos-messaging-schema

tholonious:
Summary of changes in this PR:
- New `MsgDeSoValidatorVote` message schema
  - VOTE message between validators
- New `MsgDeSoValidatorTimeout` message schema
  - TIMEOUT message between validators
- New `Bitset` Data Structure Implementation
  - Simple implementation of a bitmask that we can use to encode the existence of each validator's signature in a timeout aggregate QC
- Updated `MsgDeSoHeader` message schema
  - New QC, TimeoutQC, block proposer public keys, signature, and view fields
  - Updated hash function for PoS block headers to use `Sha256DoubleHash`
- New multi-payload BLS signature verification in the `bls` package


tholonious:
* feature/proof-of-stake
  * **PR #570** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/570" target="_blank">Graphite</a>  👈
    * **PR #571** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/571" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/570?utm_source=stack-comment).

lib/network.go
```diff
@@ -1880,16 +1893,48 @@ type MsgDeSoHeader struct {
 	// The height of the block this header corresponds to.
 	Height uint64
 
+	// Nonce is only used for Proof of Work blocks, and will only be populated
```
diamondhands0, line 1896: Do you mean PRE proof of work blocks?

diamondhands0: Whoops sorry nevermind; I understand.

lib/pos_network.go
```diff
@@ -0,0 +1,557 @@
...
+	// validators, each of whom signed a payload with the pair
+	// (current view, the validator's local HighQC.View).
+	//
+	// The ordering of high QC views and validators in the aggregate signature will
+	// match the ordering of active validators in descending order of stake for the
+	// timed out view's epoch. I.e. index 0 will correspond to the highest-staked active
+	// validator in the epoch, index 1 will correspond to the second-highest-staked active
+	// validator, ...
+	ValidatorsTimeoutHighQCViews         []uint64
+	ValidatorsTimeoutAggregatedSignature *AggregatedBLSSignature
```
diamondhands0, line 455: I think there may have been a misunderstanding of this field. Let's suppose that we have the following setup:
- Validator_01: (TimedOutView = 11, HighQCView = 10)
- Validator_02: (TimedOutView = 11, HighQCView = 10)
- Validator_03: (TimedOutView = 11, HighQCView = 9)
- Validator_04: (TimedOutView = 11, HighQCView = 9)
- Validator_05: (TimedOutView = 11, HighQCView = 8)

If this is the case, then I think we need three aggregated signatures, one for each (TimedOutView, HighQCView) pair. And in the TimeoutAggregateQC, it would look more like:
- ValidatorsUniqueTimeoutHighQCViews []uint64 // <-- Notice Unique added to the name. This would have [8, 9, 10]
- ValidatorsTimeoutAggregatedSignatures []*AggregatedBLSSignature // <-- Notice it's a list of AggregatedBLSSignature. It would look as follows: [Signature((11, 8)), Signature([(11, 9), (11, 9)]), Signature([(11, 10), (11, 10)])]

Let me know if I'm missing something with regard to this. The key is that AggregatedBLSSignature can only sign a single payload, which means that we need a different aggregate signature for each HighQCView that validators have.

diamondhands0: Link looks broken. Just go to fast_hotstuff_bls.go#L195 in the research code.

lib/network.go
```diff
@@ -2018,13 +2063,75 @@ func (msg *MsgDeSoHeader) EncodeHeaderVersion1(preSignature bool) ([]byte, error
...
+	// TransactionMerkleRoot
+	transactionMerkleRoot := msg.TransactionMerkleRoot
+	if transactionMerkleRoot == nil {
+		transactionMerkleRoot = &BlockHash{}
+	}
+	retBytes = append(retBytes, transactionMerkleRoot[:]...)
+
+	// TstampSecs: this field can be encoded to take up the full 64 bits now
+	// that MsgDeSoHeader version 2 does not need to be backwards compatible.
+	retBytes = append(retBytes, UintToBuf(msg.TstampSecs)...)
```
diamondhands0, line 2092: This is good. We want uint64 for this field and for the height. There are a few places where the height is casted to uint32 and it would be good to convert those to uint64 as we upgrade.

tholonious:
* feature/proof-of-stake
  * **PR #570** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/570" target="_blank">Graphite</a>  👈
    * **PR #575** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/575" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/570?utm_source=stack-comment).

lib/pos_network.go
tholonious: BLS natively supports signature aggregation for multiple public keys, each with a unique payload. So the schema as defined in the code will actually work. 

See PR https://github.com/deso-protocol/core/pull/575 for concrete examples with unit tests that do multi-payload signature aggregation & verification.

bls/signature.go
```diff
@@ -8,23 +8,25 @@ import (
...
 	flowCrypto "github.com/onflow/flow-go/crypto"
+	"github.com/onflow/flow-go/crypto/hash"
 )
 
-// The SigningAlgorithm for BLS keys is BLSBLS12381 which is BLS on the BLS 12-381 curve.
+// The signingAlgorithm for BLS keys is BLSBLS12381 which is BLS on the BLS 12-381 curve.
 // This is the only supported BLS signing algorithm in the flowCrypto package.
 // BLS is used such that we can aggregate signatures into one signature.
-const SigningAlgorithm = flowCrypto.BLSBLS12381
+const signingAlgorithm = flowCrypto.BLSBLS12381
```
tholonious, line 21: This `signingAlgorithm` const and `hashingAlgorithm` variable are implementation details of how this package wraps the flow BLS scheme implementation underneath.

We don't want this package to expose them externally, so it's best to make them private.

lib/pos_network.go
tholonious: Making this as resolved. I've merged PR https://github.com/deso-protocol/core/pull/575 which supports BLS signature aggregation and verification on multiple payloads.

bls/signature.go
```diff
@@ -41,15 +43,30 @@ func AggregateSignatures(signatures []*Signature) (*Signature, error) {
...
-		flowPublicKeys = append(flowPublicKeys, publicKey.flowPublicKey)
+func VerifyAggregateSignatureSinglePayload(publicKeys []*PublicKey, signature *Signature, payloadBytes []byte) (bool, error) {
+	flowPublicKeys := extractFlowPublicKeys(publicKeys)
+	return flowCrypto.VerifyBLSSignatureOneMessage(flowPublicKeys, signature.flowSignature, payloadBytes, hashingAlgorithm)
+}
+
+// VerifyAggregateSignatureMultiplePayloads takes in a slice of bls.PublicKeys, a bls.Signature, and a slice of payloads.
+// It returns true if each bls.PublicKey at index i has signed its respective payload at index i in the payloads slice.
+// The input bls.Signature is the aggregate signature of each public key's partial bls.Signatures for its respective payload.
+func VerifyAggregateSignatureMultiplePayloads(publicKeys []*PublicKey, signature *Signature, payloadsBytes [][]byte) (bool, error) {
```
diamondhands0, line 57: Hmm... I had a question about this. When I imagined it, it seemed like the signature would take a [][]*PublicKey to parallel the payloadBytes. But here I see it's only a single list of public keys.

If this is the case, then how would verification of the QCs work? My understanding was we'd have a situation like the following:
* HighQC1 is signed by {pk1, pk2, pk3}
* HighQC2 is signed by {pk4, pk5}
* HighQC3 is signed by {pk6, pk7, pk8}

In which case I would expect something more like the following:
publicKeys := [][]*PublicKey{
    {pk1, pk2, pk3},
    {pk4, pk5}
    {pk6, pk7, pk8}
}
payloads := [][]byte{
    HighQC1,
    HighQC2,
    HighQC3
}
verified, err := VerifyAggregateSignatureMultiplePayloads(publicKeys, signature, payloads)

I think I'm missing something though. Lmk what you think.

lib/constants.go
```diff
@@ -85,9 +85,11 @@ const (
 	NetworkType_TESTNET NetworkType = 2
 )
 
+type MsgDeSoHeaderVersion = uint32
```
diamondhands0, line 88: Nit: I saw this the first time and thought it's probably fine, but isn't the more normal syntax `type MsgDeSoHeaderVersion uint32` with no `=`?

lib/network.go
```diff
@@ -1880,16 +1893,63 @@ type MsgDeSoHeader struct {
...
+	// ValidatorsTimeoutAggregateQC is only used for Proof of Stake blocks, starting with
+	// MsgDeSoHeader version 2. For all earlier versions, this field will default to nil.
+	//
+	// In the event of a timeout, this field will contain the aggregate QC constructed from
+	// timeout messages from 2/3 of validators weighted by stake, and proves that they have
+	// timed out. This value is set to null in normal cases where a regular block vote has
+	// taken place.
+	ValidatorsTimeoutAggregateQC *TimeoutAggregateQuorumCertificate
+
+	// ProposerVotePartialSignature is only used for Proof of Stake blocks, starting with
```
diamondhands0, line 1946: Nit: I think this comment is stale.

lib/network.go
```diff
@@ -1880,16 +1893,63 @@ type MsgDeSoHeader struct {
...
+	// MsgDeSoHeader version 2. For all earlier versions, this field will default to nil.
+	//
+	// The block proposer's partial BLS signature of the (ProposedInView, BlockHash) pair
+	// for the block. This signature proves that a particular validator proposed the block,
+	// and also acts as the proposer's vote for the block.
+	ProposerVotePartialSignature *bls.Signature
```
diamondhands0, line 1952: Side-note: Just wanted to say your comments overall are awesome, man. They're really going to help us and future engineers figure out what we were thinking. Great stuff.

lib/network.go
```diff
@@ -2018,13 +2078,104 @@ func (msg *MsgDeSoHeader) EncodeHeaderVersion1(preSignature bool) ([]byte, error
...
+	// DecodeHeaderVersion2.
+	proposerSignatureBytes := []byte{}
+	if !preSignature {
+		// If the the preSignature flag is set to false, then the caller intends to encode the signature.
+		// The signature must be non-nil.
+		if msg.ProposerVotePartialSignature == nil {
+			return nil, fmt.Errorf("EncodeHeaderVersion2: ProposerVotePartialSignature must be non-nil")
+		}
+		proposerSignatureBytes = msg.ProposerVotePartialSignature.ToBytes()
+	}
```
diamondhands0, line 2165: Noting that we need to encode/decode the new fields here.

lib/network.go
```diff
@@ -2139,6 +2290,84 @@ func DecodeHeaderVersion1(rr io.Reader) (*MsgDeSoHeader, error) {
...
+	proposerSignatureBytes, err := DecodeByteArray(rr)
+	if err != nil {
+		return nil, errors.Wrapf(err, "MsgDeSoHeader.FromBytes: Problem decoding ProposerVotePartialSignature")
+	}
+	if len(proposerSignatureBytes) != 0 {
+		retHeader.ProposerVotePartialSignature, err = (&bls.Signature{}).FromBytes(proposerSignatureBytes)
+		if err != nil {
+			return nil, errors.Wrapf(err, "MsgDeSoHeader.FromBytes: Problem decoding ProposerVotePartialSignature")
+		}
+	}
```
diamondhands0, line 2366: Need to encode/decode the new fields here.

lib/pos_network.go
```diff
@@ -0,0 +1,569 @@
...
```
diamondhands0, line 467: This goes with my other comment on aggregating signatures. I think I might be missing something, but if my previous comment's assumptions aren't wrong, then the fields hree seem like they'd be more like the following:
```
ValidatorsTimeoutHighQCViews         []uint64

ValidatorsTimeoutHighQCPublicKeys    []BitSet // Same list size as the previous field. Each bitset in this field represents the pubkeys corresponding to a particular HighQCView.

ValidatorsTimeoutAggregatedSignature *Signature // I think this would be a vanilla "Signature" from signature.go that you would pass to VerifyAggregateSignatureMultiplePayloads
```

So then you would do something like this to verify a timeout qc:
```
// Assume tqc is a TimeoutAggregateQuorumCertificate
publicKeys = [][]*PublicKey{}
utxoView := NewUtxoView() // Have to get a view to convert the bitsets properly
for bs := range tqc.TimeoutAggregateQuorumCertificate {
    publicKeys = append(publicKeys, utxoView.BitSetToPubKeys(bs))
}

payloads = tqc.ValidatorsTimeoutHighQCViews // Maybe these need to be hashed or whatever but you get the idea

verified, err := VerifyAggregateSignatureMultiplePayloads(publicKeys, tqc.ValidatorsTimeoutAggregatedSignature, payloads)
```

Again I feel like I'm missing something. Just typing out what I was thinking so I can get up to speed.

bls/signature.go
tholonious: I see. Is the intention here compression of the payloads? Or is there a limitation in BLS signature verification that requires payload deduplication?

My understanding of all of this is purely based off the UBC researchers' reference implementation here: https://github.com/deso-protocol/hotstuff_pseudocode/blob/diamondhands/REV-02-Jun-06-2023/fast_hotstuff_bls.go#L186. 

So I may be missing something too. 

My understanding is that, whenever those public keys timeout, it doesn't sign its high QC, but rather its high QC's view. 

Using your example with pk1, pk2, pk3... pk8, when each public key times out for view N, it signs the following:
- pk1 signs the pair <N, HighQC1.View>
- pk2 signs the pair <N, HighQC1.View>
- pk3 signs the pair <N, HighQC2.View>
- ...
- pk8 signs the pair <N, HighQC3.View>

When the leader aggregates a timeout QC for view N, it constructs the timeout QC as follows:
- publicKeys = [pk1, pk2, pk3, pk4, pk5, pk6, pk7, pk8]
  - all of those public keys have timed out
- payloads = [HighQC1.View, HighQC1.View, HighQC1.View, HighQC2.View, HighQC2.View, HighQC3.View, HighQC3.View, HighQC3.View]
  - the index of each element in this slice corresponds to the public key that sent it
- signature = aggregation of all of the partial signatures from pk1, pk2, pk3,...

diamondhands0: OK I get what you're doing now. Your understanding of what we need to do is correct, I was just imagining a different schema that "deduplicates" the payload, if that makes sense.

Basically, what you're saying is you aggregate a bunch of (PublicKey, payload) pairs, which I think is the most elegant thing to do. The less elegant thing that I was suggesting is we aggregate ([]PublicKey, payload) pairs, where each tuple will have a *unique* payload plus all the public keys that signed that particular unique payload. This amounts to the same thing, just grouped differently. Lmk if this doesn't make sense.

I think the way you have it is how we should keep it. It's technically more redundant because we repeat the same view number a lot in ValidatorsTimeoutHighQCViews but we can compress it easily later if it ever becomes an issue (e.g. when we have 10k+ validators or something...).

lib/pos_network.go
diamondhands0: Ok so adding to my previous comment, let me confirm my understanding here.

Suppose we have the following situation (notice I reordered some pks to make things interesting):
```
HighQC1 is signed by {pk4, pk5, pk3}
HighQC2 is signed by {pk1, pk2}
HighQC3 is signed by {pk6, pk7, pk8}
```

I think then the fields would be as follows:
```
ValidatorsTimeoutHighQCViews = [2, 2, 1, 1, 1, 3, 3, 3]
ValidatorsTimeoutAggregatedSignature = AggregatedBLSSignature{
    // The bitmap would be set so that it's converted as follows...
    bitmap = [pk1, pk2, pk3, pk4, pk5, pk6, pk7, pk8]
    // Then the signature aggregates (pk1, pk1HighQCView), (pk2, pk2HighQCView), ...
}
```

With the above, the verification would look as follows:
```
VerifyAggregateSignatureSinglePayload([pk1, ..., pk8], signature *Signature, [2, 2, 1, 1, 1, 3, 3, 3])
```

Is that roughly correct?

bls/signature.go
tholonious: Correct. And you're right that the payloads are duplicated in my current scheme. I'm actually aligned with you that deduplicating should be the end-game of what we want to implement.

What's interesting is that, even if we do the deduplication as you suggest, we'll still need to unravel the public keys and payloads into individual (publickey, payload) pairs to do the BLS signature verification. This means that the in-memory `TimeoutAggregateQuorumCertificate` struct's schema is best left as-is.

This also means that the biggest value of deduplicating the payloads is the compression it provides when sending timeout QCs as network messages. 

What nice about the above is that, it means we can implement the compression in the ToBytes/FromBytes functions for the timeout message without changing the struct fields at all. When we get to the point where we want to compress the views, we can increment the `MsgDeSoHeader` version and implement new byte encoding that hs the compression.

lib/pos_network.go
tholonious: Yup, that's correct!

lib/constants.go
tholonious: You're right. Ideally we define the type as `type MsgDeSoHeaderVersion uint32` without the `=`.

There's actually a fundamental difference between the two that makes the current definition with the `=` easier to work with for us:
- Defining the type with the `=` defines it as a type alias for `uint32`. The two are interchangeable, and values of either type can be assigned to the other. The two types are equivalent which means that all of our strongly typed uint32 byte encoding/decoding functions work as-is
- Defining the type without the `=` defines it as a brand new 4 byte unsigned integer type. Go's type system sees it as a distinct type from `uint32` type. 
  - The advantage of defining the new type is that the type system enforces usage of just the valid values of the new type. You can't assign any uint32 to the version field
  - The downside is that we would have to implement new byte encoders and decoders for every newly defined type. Or we would have to unsafe casting when encoding and decoding it

So, it seems like defining it as a type alias is the less bad approach. LMK though

lib/network.go
tholonious: Comment's up to date. Definitely open to changing the `ProposerVotePartialSignature` if you have another preference.

tholonious:
* feature/proof-of-stake
  * **PR #570** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/570" target="_blank">Graphite</a>  👈
    * **PR #580** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/580" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/570?utm_source=stack-comment).

lib/network.go
tholonious: I've added a new `EncodeOptionalBLSSignature(sig)` function for this

lib/network.go
tholonious: Added a new `DecodeOptionalBLSSignature(rr)` function for this too

lazynina (approved):

lib/constants.go
iamnotnader: Ah wow I just read through this. Didn't know the type keyword worked like that, but makes sense. Great that we didn't have to rewrite all the code.

bls/signature.go
iamnotnader: Just read through this and I love the idea of pushing the compression into the ToBytes/FromBytes. That actually seems like the perfect place to put it because its benefit is to the network layer, no to the actual logic like you mention.

PR #574, opened 2023-06-29
Fix copy paste mistake in function comment

pelmers:
This commit writes a comment for `disconnectAtBlockHeight` that more closely matches what it does.

Background: I'm doing a research study on function comments that may be incorrect. You can find more info and a short and optional anonymous survey [here](https://forms.office.com/e/yHv4PRXM2i).
 
Thanks for reviewing!


AeonSw4n (approved):

AeonSw4n:
Thank you for fixing the comment!

PR #577, opened 2023-07-05
Add block proposer public keys and signature to block header

tholonious:


tholonious:
* feature/proof-of-stake
  * **PR #570** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/570" target="_blank">Graphite</a> 
    * **PR #577** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/577" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/577?utm_source=stack-comment).

lib/network.go
```diff
@@ -2090,12 +2106,26 @@ func (msg *MsgDeSoHeader) EncodeHeaderVersion2(preSignature bool) ([]byte, error
 	// that MsgDeSoHeader version 2 does not need to be backwards compatible.
 	retBytes = append(retBytes, UintToBuf(msg.TstampSecs)...)
 
-	// Height
+	// Height: similar to the field above, this field can be encoded to take
+	// up the full 64 bits now that MsgDeSoHeader version 2 does not need to
```
lazynina, line 2110: It doesn't always take up all 64 bits. UintToBuf is a variable length encoding.

lib/network.go
```diff
@@ -2344,17 +2413,31 @@ func (msg *MsgDeSoHeader) GetMsgType() MsgType {
...
+	// Compute the specialized PoW hash for header versions 0 and 1.
+	if msg.Version == HeaderVersion0 || msg.Version == HeaderVersion1 {
+		return ProofOfWorkHash(headerBytes, msg.Version), nil
+	}
+
+	// TODO: Do we need a new specialized hash function for Proof of Stake
+	// block headers? A simple SHA256 hash seems like it would be sufficient.
+	// The use of ASICS is no longer a consideration, so we should be able to
+	// simplify the hash function used.
+	return Sha256DoubleHash(headerBytes), nil
```
tholonious, line 2440: @AeonSw4n any suggestions here? Seems like the hash function used to produce block hashes for PoS blocks can be simplified. 

Alternative we can just revert this change to use the existing ProofOfWorkHash function.

lib/network.go
```diff
@@ -2344,17 +2413,31 @@ func (msg *MsgDeSoHeader) GetMsgType() MsgType {
...
 }
 
-// Hash is a helper function to compute a hash of the header. Note that the header
-// hash is special in that we always hash it using the ProofOfWorkHash rather than
-// Sha256DoubleHash.
+// Hash is a helper function to compute a hash of the header. For Proof of Work
+// blocks headers, which have header version 0 or 1, this uses the specialized
+// ProofOfWorkHash, which mining difficulty and hardware into consideration.
+//
+// For For Proof of Stake block headers, which start header versions 2, the
```
lazynina, line 2420:
```suggestion
// For Proof of Stake block headers, which starts at header versions 2,
```

lib/network.go
```diff
@@ -2344,17 +2413,31 @@ func (msg *MsgDeSoHeader) GetMsgType() MsgType {
...
 		return nil, errors.Wrap(err, "MsgDeSoHeader.Hash: ")
 	}
 
-	return ProofOfWorkHash(headerBytes, msg.Version), nil
+	// Compute the specialized PoW hash for header versions 0 and 1.
+	if msg.Version == HeaderVersion0 || msg.Version == HeaderVersion1 {
+		return ProofOfWorkHash(headerBytes, msg.Version), nil
+	}
+
+	// TODO: Do we need a new specialized hash function for Proof of Stake
```
lazynina, line 2436: I think a sha 256 double hash should be fine here! DH can confirm this tho, so I would leave this comment in there so he thinks on it.

lib/network.go
```diff
@@ -2509,11 +2592,17 @@ func (msg *MsgDeSoBlock) EncodeBlockVersion1(preSignature bool) ([]byte, error)
...
+func (msg *MsgDeSoBlock) EncodeBlockVersion2(preSignature bool) ([]byte, error) {
+	return msg.EncodeBlockCommmon(preSignature)
+}
+
 func (msg *MsgDeSoBlock) ToBytes(preSignature bool) ([]byte, error) {
 	if msg.Header.Version == HeaderVersion0 {
 		return msg.EncodeBlockVersion0(preSignature)
 	} else if msg.Header.Version == HeaderVersion1 {
 		return msg.EncodeBlockVersion1(preSignature)
+	} else if msg.Header.Version == HeaderVersion2 {
```
lazynina, line 2604: maybe we switch this to a switch statement? whenever I see if-else-if-else-if all checking the same fields, I generally opt for the switch statement.

lazynina (changes requested):

lib/network.go
tholonious: It can take up to 64 bits. The field is encoded as variable length, but it has 64 bits of resolution.

lib/network.go
```diff
@@ -2509,12 +2592,19 @@ func (msg *MsgDeSoBlock) EncodeBlockVersion1(preSignature bool) ([]byte, error)
...
 }
 
+func (msg *MsgDeSoBlock) EncodeBlockVersion2(preSignature bool) ([]byte, error) {
+	return msg.EncodeBlockCommmon(preSignature)
+}
+
 func (msg *MsgDeSoBlock) ToBytes(preSignature bool) ([]byte, error) {
-	if msg.Header.Version == HeaderVersion0 {
+	switch msg.Header.Version {
+	case HeaderVersion0:
```
lazynina, line 2601:
```suggestion
```

lib/network.go
```diff
@@ -2509,12 +2592,19 @@ func (msg *MsgDeSoBlock) EncodeBlockVersion1(preSignature bool) ([]byte, error)
...
 		return msg.EncodeBlockVersion0(preSignature)
-	} else if msg.Header.Version == HeaderVersion1 {
+	case HeaderVersion1:
```
lazynina, line 2603:
```suggestion
	case HeaderVersion0, HeaderVersion1:
```

lazynina: case statements don't "fallthrough" like they do in js.

lazynina (changes requested):

lazynina (approved):

lib/network.go
```diff
@@ -2119,11 +2149,25 @@ func (msg *MsgDeSoHeader) EncodeHeaderVersion2(preSignature bool) ([]byte, error
...
+	proposerSignatureBytes := []byte{}
+	if !preSignature {
+		// If the the preSignature flag is set to false, then the caller intends to encode the signature.
+		// The signature must be non-nil.
+		if msg.ProposerVotePartialSignature == nil {
+			return nil, fmt.Errorf("EncodeHeaderVersion2: ProposerVotePartialSignature must be non-nil")
+		}
+		proposerSignatureBytes = msg.ProposerVotePartialSignature.ToBytes()
+	}
+	retBytes = append(retBytes, EncodeByteArray(proposerSignatureBytes)...)
```
AeonSw4n, line 2165: I think this append should only happen inside the `if !preSignature {}` condition. Otherwise, it will append `[]byte{0}` which seems problematic. The `[]byte{0}` is the result of encoding array lenght in `EncodeToBytes`.

To see why this could be problematic, let's say that `X := EncodeHeaderVersion2(true)`. Now say `y := EncodeByteArray(proposerSignatureBytes)`. If you look at `X[: len(X) - len(y)]` I think it's not equal to `EncodeHeaderVersion2(false)`, which is what we supposedly want, because that's what we will be signing. Unless I'm misunderstanding how the preSignature flag is meant to work.

lib/network.go
AeonSw4n: `Sha256DoubleHash` works well here!

lib/network.go
```diff
@@ -2344,17 +2413,31 @@ func (msg *MsgDeSoHeader) GetMsgType() MsgType {
...
+//
+// For Proof of Stake block headers, which start header versions 2, it uses the
+// simpler Sha256DoubleHash function.
 func (msg *MsgDeSoHeader) Hash() (*BlockHash, error) {
-	preSignature := false
-	headerBytes, err := msg.ToBytes(preSignature)
+	// The preSignature flag is unused during byte encoding in
+	// in header versions 0 and 1. We set it to true to ensure that
+	// it's forward compatible for versions 2 and beyond.
+	headerBytes, err := msg.ToBytes(true)
```
AeonSw4n, line 2426: Ah yes so my initial comment re preSignature flag is applicable. We should add tests for signing the Hash to make sure everything is bulletproof.

AeonSw4n (changes requested):
Looks good! Got one comment regarding the handling of `preSignature` flag. In addition, I think we should increase our test coverage around `MsgDeSoHeader`, in particular testing the signing of the header hash.

lib/network.go
tholonious: We need the append outside of the `if !preSignature {}` so that the DecodeHeaderVersion2 function can decode all possible outputs produced by the EncodeHeaderVersion2 function. Without it, the byte encoding/decoding won't be bijective.

You are correct that it's always the case that `len(EncodeHeaderVersion2(false)) != len(EncodeHeaderVersion2(true)) + len(proposerSignatureBytes)`. This is a result of using variable length ints to store the signature length. IMO, the above isn't problematic though.

The `preSignature` flag usage gives us the critical properties we want:
1. It gives us a deterministic method to encode the header to bytes excluding the signature. The placeholder `[]byte{0}` will always be encoded when `preSignature=true`. This is straightforward to decode if needed.
2. It give us a deterministic method to encode the header including the signature. In this case the end of the encoded bytes hare `[]byte{len(signature)} ... proposerSignatureBytes`. This is also straightforward to decode.
3. Given a header with the signature field populated, it gives us a way to verify the signature field as follows:
  i. Compute `preSignatureBytes = header.ToBytes(preSignature=true)`
  ii. Compute `headerHash = Hash(preSignatureBytes)`
  iii. Check `VerifyBLSSignature(header.ProposerVotingPublicKey, header.ProposerVotePartialSignature, headerHash)`


Can you elaborate on the risks you see with the above?

AeonSw4n: Looks like you got this covered. As long as we're consistent that header byte encoding for hashes always has the extra fictitious []byte{0} at the end, then we're good. I guess "preSignature" is slightly misleading. What it really means is "nill signature", which means an encoding that will not be a slice of the encoding "with signature" included (they differ on that last []byte{0}). Since this is a convention we've used in other place, let's go forward with this.

AeonSw4n (approved):

PR #578, opened 2023-07-05
PoS Mempool

AeonSw4n:


AeonSw4n:
* feature/proof-of-stake
  * **PR #578** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/578" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/578?utm_source=stack-comment).

lib/pos_transaction_register.go
```diff
@@ -0,0 +1,425 @@
...
+	// If this condition is true, then the computed exponent is overestimated by 1.
+	if computeFeeTimeBucketMinFromExponent(feeTimeBucketExponent, minimumNetworkFeeNanosPerKB, feeBucketMultiplier) > feeNanosPerKB {
+		return feeTimeBucketExponent - 1
+	}
+
+	// This condition gets triggered exactly on fee bucket boundaries, i.e. when feeNanosPerKB = FeeMax for some bucket.
+	// The float rounding makes the number slightly smaller like 5.9999999991 instead of 6.0.
+	if computeFeeTimeBucketMinFromExponent(feeTimeBucketExponent+1, minimumNetworkFeeNanosPerKB, feeBucketMultiplier) <= feeNanosPerKB {
+		return feeTimeBucketExponent + 1
+	}
```
diamondhands0, line 421: Such a tricky tricky set of conditions to check for this last part. I thought about it for a bit and I think there may be *one case* where this function gets it wrong, with *I think* an easy solution...

We have three possible edge cases for feeNanosPerKB:
* n-epsilon:
  - feeNanosPerKB = base*multiplier^(n-epsilon)
  - In this case, we want to return bucket n
* n
  - feeNanosPerKB = base*multiplier^(n)
  - In this case, we want to return bucket n
* n+epsilon:
  - feeNanosPerKB = base*multiplier^(n+epsilon)
  - In this case, we want to return bucket n+1

Now, the problem is that we *also* have three possibilities for what feeTimeBucketExponent could end up being because of rounding error. These cases are: feeTimeBucketExponent = {n-1, n, n+1}. The reason for this is simply that the error in BigFloatLog could wind up returning n+1 even when we have (n-epsilon), and vice versa returning n-1 even when we have n+epsilon.

Below I work out the logic of the last two checks in all cases. Notice it breaks when you have feeTimeBucketExponent = n+1 with n-epsilon as the input:

n-epsilon
* Assume:
  - fee = base*multiplier^(n-epsilon) (want bucket n-1)
* compute n-1
  - base*multiplier^(n-1) > base*multiplier^(n-epsilon) = FALSE
  - base*multiplier^((n-1)+1=n) <= base*multiplier^(n-epsilon) = FALSE
  - return n-1 = CORRECT
* compute n
  - base*multiplier^(n) > base*multiplier^(n-epsilon) = TRUE
    * n -> n-1
  - base*multiplier^((n-1)+1=n) <= base*multiplier^(n-epsilon) = FALSE
  - return n-1 CORRECT
* compute n+1
  - base*multiplier^(n+1) > base*multiplier^(n-epsilon) = TRUE
    * n+1 -> n
  - base*multiplier^((n)+1=n+1) <= base*multiplier^(n-epsilon) = FALSE
  - **return n = WRONG**

n
* Assume:
  - fee = base*multiplier^n (want bucket n)
* compute n-1
  - base*multiplier^(n-1) > base*multiplier^(n) = FALSE
  - base*multiplier^((n-1)+1) <= base*multiplier^(n) = TRUE
    * n-1 -> n
  - return n
* compute n
  - base*multiplier^(n) > base*multiplier^(n) = FALSE
  - base*multiplier^((n)+1) <= base*multiplier^(n) = FALSE
  - return n
* compute n+1
  - base*multiplier^(n+1) > base*multiplier^(n) = TRUE
    * n+1 -> n
  - base*multiplier^((n)+1) <= base*multiplier^(n) = FALSE
  - return n

n+epsilon
* Assume:
  - fee = base*multiplier^(n+epsilon) (want bucket n)
* compute n-1
  - base*multiplier^(n-1) > base*multiplier^(n+epsilon) = FALSE
  - base*multiplier^((n-1)+1) <= base*multiplier^(n+epsilon) = TRUE
    * n-1 -> n
  - return n CORRECT
* compute n
  - base*multiplier^(n) > base*multiplier^(n+epsilon) = FALSE
  - base*multiplier^((n)+1) <= base*multiplier^(n+epsilon) = FALSE
  - return n CORRECT
* compute n+1
  - base*multiplier^(n+1) > base*multiplier^(n+epsilon) = TRUE
    * n+1 -> n
  - base*multiplier^((n)+1) <= base*multiplier^(n+epsilon) = FALSE
  - return n CORRECT

If the above is correct, I think the solution is actually pretty simple. In particular, I think if you just do the first check *twice* you will cover that one annoying case. The issue is basically that, in the situation where you have n-epsilon as the exponent, and ONLY in that case, you could be off by *two*. And so you need to check the >feeNanosPerKB thing twice...

AeonSw4n: If I understand your argument correctly, then I think we should actually be chill. I don't think we can overestimate `n` by a whole 1, but rather by a tiny fraction like 0.00001. In this case, the `n-epsilon` path will not cause us to arrive at `n+1`, though `n` might be possible. If we've somehow got over `n`, like `n + 0.0001`, then the uint cast should round it down to `n`.

I was also worried about these edge cases. I made a test called `TestComputeFeeBucketWithFee` in `pos_transaction_register_test.go`, where I test for the correct computation of `n` for every `feeNanosPerKB` in the range from 1000 to 100,000. It felt like a sufficiently large range to test for all sorts of float-rounding shenanigans.

Do you still have concerns about this exponent computation? We can try to bulletproof it more if you'd like.

AeonSw4n:
Add mutex for status checks.

AeonSw4n:
* feature/proof-of-stake
  * **PR #578** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/578" target="_blank">Graphite</a>  👈
    * **PR #647** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/647" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/578?utm_source=stack-comment).

tholonious (approved):

PR #579, opened 2023-07-07
PoS Staker Snapshotting and Rewards

tholonious:


tholonious:
* feature/proof-of-stake
  * **PR #579** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/579" target="_blank">Graphite</a>  👈
    * **PR #581** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/581" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/579?utm_source=stack-comment).

lib/block_view_validator.go
```diff
@@ -1919,39 +1854,61 @@ func (bav *UtxoView) GetTopActiveValidatorsByStake(limit uint64) ([]*ValidatorEn
...
+	if blockHeight < uint64(bav.Params.ForkHeights.ProofOfStake2ConsensusCutoverBlockHeight) {
+		return nil
 	}
-	// If not set, read the GlobalActiveStakeAmountNanos from the db.
-	globalActiveStakeAmountNanos, err := DBGetGlobalActiveStakeAmountNanos(bav.Handle, bav.Snapshot)
+
+	snapshotGlobalParams, err := bav.GetSnapshotGlobalParamsEntry()
 	if err != nil {
-		return nil, errors.Wrapf(err, "UtxoView.GetGlobalActiveStakeAmountNanos: ")
+		return errors.Wrapf(err, "UtxoView.JailInactiveSnapshotValidators: ")
```
lazynina, line 1867: Let's add some more details to the errors in this function so it's easier to pinpoint where they came from if we hit them.

tholonious:
* feature/proof-of-stake
  * **PR #579** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/579" target="_blank">Graphite</a>  👈
    * **PR #625** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/625" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/579?utm_source=stack-comment).

tholonious:
* feature/proof-of-stake
  * **PR #579** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/579" target="_blank">Graphite</a>  👈
    * **PR #626** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/626" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/579?utm_source=stack-comment).

lib/pos_epoch.go
```diff
@@ -84,7 +98,16 @@ func (bav *UtxoView) GetCurrentEpochEntry() (*EpochEntry, error) {
 
 	// If still not found, return the GenesisEpochEntry. This will be the
 	// case prior to the first execution of the OnEpochCompleteHook.
-	return &EpochEntry{EpochNumber: 0, FinalBlockHeight: math.MaxUint64}, nil
+	//
+	// TODO: Should FinalBlockHeight be ProofOfStake1StateSetupBlockHeight for epoch 0?
+	// The ProofOfStake1StateSetupBlockHeight fork height is exactly when epoch 0 ends.
+	// Epoch 1 begins at the next height.
```
tholonious, line 104: We currently don't read the FinalBlockHeight for epoch 0 so this shouldn't matter. That said, `ProofOfStake1StateSetupBlockHeight` is technically the correct value to return and IMO will avoid confusion for the long run. I'm leaving this as a TODO here since it's not clear to me what the motivation was to return `math.MaxUint64` and if it has some special purpose

lib/block_view_validator.go
```diff
@@ -592,83 +629,63 @@ func DBGetValidatorByPKIDWithTxn(txn *badger.Txn, snap *Snapshot, pkid *PKID) (*
...
 	validatorEntriesToSkip []*ValidatorEntry,
 ) ([]*ValidatorEntry, error) {
 	var validatorEntries []*ValidatorEntry
 
-	// Convert ValidatorEntriesToSkip to ValidatorEntryKeysToSkip.
-	validatorKeysToSkip := NewSet([]string{})
+	// Convert validatorEntriesToSkip to the ValidatorPKIDs we need to skip.
+	validatorPKIDsToSkip := NewSet([]PKID{})
 	for _, validatorEntryToSkip := range validatorEntriesToSkip {
-		validatorKeysToSkip.Add(string(DBKeyForValidatorByStake(validatorEntryToSkip)))
```
tholonious, line 606: There was a subtle issue here in cases where a validator had its TotalStakeAmountNanos field updated in the view. Since the TotalStakeAmountNanos field is used in the DB key and is a mutable value, it is unreliable to use for filtering during DB seeks. 

This seek would have returned a validators that were intended to be filtered out.

lib/pos_snapshot_entries.go
```diff
@@ -51,6 +52,21 @@ func (bav *UtxoView) GetCurrentGlobalParamsEntry() *GlobalParamsEntry {
...
+//     be in agreement on the size of the validator set, leader schedule, stakes to reward, and epoch duration.
+//   - When transitioning to a new epoch, we use the snapshot GlobalParams to determine the length of the next
+//     epoch. All validators need to be in agreement ahead of time on what length of the next epoch will be before
+//     the epoch begins.
+//
+// For all other uses, only the CurrentGlobalParamsEntry is appropriate to use. This includes all transaction connect
+// logic and end of epoch operations that mutate the validator entries and stake entries BEFORE they are
+// snapshotted. This approach ensures that whenever we create a snapshot of the validator set, leader schedule,
+// and stakes to reward... the GlobalParams used to create the snapshots are snapshotted along with that data, and
+// live alongside them.
```
diamondhands0, line 69: Love seeing long thoughtful comments like this.

lib/pos_staking_rewards.go
```diff
@@ -0,0 +1,282 @@
...
+}
+
+func convertAPYBasisPointsToFloat(apyBasisPoints uint64) *big.Float {
+	apyBasisPointsAsFloat := NewFloat().SetUint64(apyBasisPoints)
+	return apyBasisPointsAsFloat.Quo(apyBasisPointsAsFloat, _basisPointsAsFloat)
+}
+
+func computeFractionOfYearAsFloat(nanoSecs uint64) *big.Float {
+	nanoSecsAsFloat := NewFloat().SetUint64(nanoSecs)
+	return nanoSecsAsFloat.Quo(nanoSecsAsFloat, _nanoSecsPerYearAsFloat)
```
diamondhands0, line 259: Always prefer NewFloat().Quo(nanoSecsAsFloat, _nanoSecsPerYearAsFloat)

lib/pos_staking_rewards.go
```diff
@@ -0,0 +1,282 @@
...
+// computeStakingReward uses float math to compute the compound interest on the stake amounts based on the
+// elapsed time since the last staking reward distribution and the APY.
+//
+// It produces the result for: stakeAmount * [e ^ (apy * elapsedTime / 1 year) - 1]
+func computeStakingReward(stakeAmountNanos *uint256.Int, elapsedFractionOfYear *big.Float, apy *big.Float) *big.Float {
+	stakeAmountFloat := NewFloat().SetInt(stakeAmountNanos.ToBig())
+	growthExponent := NewFloat().Mul(elapsedFractionOfYear, apy)                // apy * elapsedTime / 1 year
+	growthMultiplier := BigFloatPow(bigE, growthExponent)                       // e ^ (apy * elapsedTime / 1 year)
+	finalStakeAmountNanos := NewFloat().Mul(stakeAmountFloat, growthMultiplier) // stakeAmount * [e ^ (apy * elapsedTime / 1 year)]
+	return finalStakeAmountNanos.Sub(finalStakeAmountNanos, stakeAmountFloat)   // stakeAmount * [e ^ (apy * elapsedTime / 1 year) - 1]
```
diamondhands0, line 271: I don't think this is possible... but because floats are imprecise, is it ever possible that finalStakeAmountNanos could be LESS than stakeAmountFloat? Might be worth clamping this value to zero just in case.

lib/pos_staking_rewards.go
```diff
@@ -0,0 +1,282 @@
...
+	_stakerRewardNanos uint64,
+	_validatorCommissionNanos uint64,
+	_err error,
+) {
+	// Compute the staker's reward amount using a big float math, and immediately convert it to big int
+	// so we can do the remainder of the math using integer operations. This is the only operation where
+	// we need float math.
+	stakerRewardNanos := convertBigFloatToBigInt(
+		computeStakingReward(snapshotStakeEntry.StakeAmountNanos, elapsedFractionOfYear, apy),
+	)
```
diamondhands0, line 118: No action on this, but I am a bit sad that DESO is still in nanos rather than a 1e18 uint256 thing. It means that sufficiently small stakes with a sufficiently small epoch window will round down to zero. This being said, I checked, and with a 1bp APY, which is the lowest we can do above zero right now, and with our epoch time being an hour, you'd still get 11 nanos for a single DESO per epoch. So hopefully we can continue limping along with this 1e9 situation.

lib/pos_staking_rewards.go
diamondhands0: I see you check it after calling this function. I'd just add the clamp here as well.

lib/pos_staking_rewards.go
```diff
@@ -0,0 +1,282 @@
...
+	// staked to themselves, then there's no reason to compute the validator commission.
+	if !snapshotStakeEntry.StakerPKID.Eq(snapshotStakeEntry.ValidatorPKID) {
+		// Fetch the ValidatorEntry that the stake is delegated to. The validator is guaranteed to be in the
+		// snapshot validator set, because only stakes from the snapshot validator set are eligible to receive
+		// rewards.
+		validatorEntry, err := bav.GetSnapshotValidatorSetEntryByPKID(snapshotStakeEntry.ValidatorPKID)
+		if err != nil {
+			return 0, 0, errors.Wrapf(err, "computeStakerRewardAndValidatorCommission: problem fetching validator entry: ")
+		}
+		if validatorEntry == nil {
```
diamondhands0, line 143: nil or isDeleted right?

lib/pos_staking_rewards.go
```diff
@@ -0,0 +1,282 @@
...
+}
+
+// computeValidatorCommission uses integer math to compute the validator's commission amount based on the staker's
+// reward amount and the validator's commission rate. Wherever possible, we rely on integer math so that rounding
+// errors are simpler to reason through.
+//
+// It produces the integer result for: floor[(stakerReward * validatorCommissionBasisPoints) / 1e4]
+func computeValidatorCommission(stakerRewardNanos *big.Int, validatorCommissionBasisPoints uint64) *big.Int {
+	scaledStakerReward := big.NewInt(0).Mul(stakerRewardNanos, big.NewInt(int64(validatorCommissionBasisPoints)))
+	return scaledStakerReward.Div(scaledStakerReward, _basisPointsAsInt)
```
diamondhands0, line 281: We usually prefer `big.NewInt(0).Div...`

lib/pos_staking_rewards.go
```diff
@@ -0,0 +1,282 @@
...
+		}
+
+		if validatorEntry.DelegatedStakeCommissionBasisPoints > 0 {
+			// We use integer math to compute the validator's commission. The commission is computed as:
+			// floor(stakerReward * validatorCommissionBasisPoints / 10000)
+			validatorCommissionNanos = computeValidatorCommission(
+				stakerRewardNanos, validatorEntry.DelegatedStakeCommissionBasisPoints,
+			)
+
+			if validatorCommissionNanos.Cmp(stakerRewardNanos) > 0 {
```
diamondhands0, line 156: Good sanity check

lib/pos_staking_rewards.go
```diff
@@ -0,0 +1,282 @@
...
+				// This should never happen. If the validator's commission is greater than the total staker reward amount,
+				// then something has gone wrong.
+				return 0, 0, errors.Errorf(
+					"computeStakerRewardAndValidatorCommission: validator commission is greater than staker reward amount",
+				)
+			}
+
+			// Deduct the validator commission from the staker's reward.
+			stakerRewardNanos.Sub(stakerRewardNanos, validatorCommissionNanos)
```
diamondhands0, line 165: Prefer `stakerRewardNanos = big.NewInt(0).Sub(stakerRewardNanos, validatorCommissionNanos)`

The reason we prefer this format is it's less error-prone. Technically, it's more efficient to modify a pointer in-place like you have it here, but it's harder to read and runs the risk of causing bugs.

lib/pos_staking_rewards.go
```diff
@@ -0,0 +1,282 @@
...
+	}
+
+	// At this point, there are three possible cases:
+	// 1. The stake entry still exists and wants to restake their rewards.
+	// 2. The stake entry still exists and does not want to restake their rewards.
+	// 3. The stake entry has unstaked since the snapshot was taken.
+
+	// For case 1, we distribute the rewards by adding them to the staker's staked amount.
+	if stakeEntry != nil && stakeEntry.RewardMethod == StakingRewardMethodRestake {
+		stakeEntry.StakeAmountNanos.Add(stakeEntry.StakeAmountNanos, uint256.NewInt().SetUint64(rewardNanos))
```
diamondhands0, line 194: Prefer `stakeEntry = big.NewInt(0).Add(stakeEntry.StakeAmountNanos, uint256.NewInt().SetUint64(rewardNanos))`

lib/pos_staking_rewards.go
```diff
@@ -0,0 +1,282 @@
...
+	}
+
+	// For cases 2 and 3, the staker does not want their rewards restaked. The staker is still
+	// eligible to receive rewards because their stake was used to secure the network. So we pay out
+	// the rewards directly to the staker's wallet.
+
+	stakerPublicKey := bav.GetPublicKeyForPKID(stakerPKID)
+	if _, err = bav._addBalance(rewardNanos, stakerPublicKey); err != nil {
+		return errors.Wrapf(err, "distributeStakingReward: problem adding rewards to staker's DESO balance: ")
+	}
```
diamondhands0, line 207: Just a note that this is going to need some extra work in Rosetta to make sure its accounting tracks this. I'm not sure the right solution for this yet, and we don't need to resolve it here. Just something to keep in mind.

Likely we will need to output something to Rosetta that allows it to figure out what's going on. Or, possibly even crazier, we could rewrite Rosetta to use state syncer, which may potentially be easier and eliminate edge cases... But probably not. We'll probably just want to hack this together.

lib/pos_staking_rewards.go
```diff
@@ -0,0 +1,282 @@
...
+	// This approach has a few advantages:
+	// 1. It gives validators an easy opt-in feature to restake their commissions. This is useful for validators that want to maximize
+	// their staking rewards over the long run. Validators can opt out of it by disabling reward restaking on their own StakeEntry.
+	// 2. It simplifies the validator commission distribution code by re-using the same code path for distributing staking
+	// rewards. By requiring the validator to already have a StakeEntry for themselves if they want to restake their commissions,
+	// this approach allows us to avoid manually creating new StakeEntries for the validator specifically for restaking commissions.
+	//
+	// TODO: The downside of the above is that it couples the restaking behavior for validator commissions and the validator's own
+	// staking reward. This is fine though, because if the validator wants to restake their own rewards but not their commissions, then
+	// they can stake to themselves using a separate wallet and only enable reward restaking for that StakeEntry.
```
diamondhands0, line 230: Clever approach that simplifies the code a lot. Good work.

lib/pos_leader_schedule.go
```diff
@@ -13,14 +14,13 @@ func (bav *UtxoView) GenerateLeaderSchedule() ([]*PKID, error) {
...
 
-	// Retrieve the SnapshotGlobalParamsEntry.LeaderScheduleMaxNumValidators.
-	snapshotGlobalParamsEntry, err := bav.GetSnapshotGlobalParamsEntry()
-	if err != nil {
-		return nil, errors.Wrapf(err, "UtxoView.GenerateLeaderSchedule: error retrieving SnapshotGlobalParamsEntry: ")
-	}
+	// Retrieve the LeaderScheduleMaxNumValidators from the current GlobalParams. We are safe to use the current
+	// global params because this generates a new leader schedule from the current validator entries, in preparation
+	// to snapshot the leader schedule.
+	currentGlobalParamsEntry := bav.GetCurrentGlobalParamsEntry()
```
diamondhands0, line 20: So I guess this was a bug before?

lib/pos_epoch_complete_hook.go
```diff
@@ -45,62 +65,168 @@ func (bav *UtxoView) RunEpochCompleteHook(blockHeight uint64) error {
...
-	globalActiveStakeAmountNanos, err := bav.GetGlobalActiveStakeAmountNanos()
-	if err != nil {
-		return errors.Wrapf(err, "RunEpochCompleteHook: problem retrieving GlobalActiveStakeAmountNanos: ")
+	// TODO: Evict old snapshots when safe to do so.
+
+	// Step 3: Roll Over to The Next Epoch
+	if err := bav.runEpochCompleteEpochRollover(currentEpochEntry.EpochNumber, blockHeight, blockTimestampNanoSecs); err != nil {
+		return errors.Wrapf(err, "RunEpochCompleteHook: ")
 	}
-	bav._setSnapshotGlobalActiveStakeAmountNanos(globalActiveStakeAmountNanos, currentEpochEntry.EpochNumber)
```
diamondhands0, line 90: I love this. Extremely clean. Great work.

lib/block_view_validator.go
diamondhands0: Oh wow good catch.

lib/block_view_validator.go
```diff
@@ -1724,15 +1683,14 @@ func (bav *UtxoView) IsValidUnjailValidatorMetadata(transactorPublicKey []byte)
...
 
-	// Retrieve the SnapshotGlobalParamsEntry.ValidatorJailEpochDuration.
-	snapshotGlobalParamsEntry, err := bav.GetSnapshotGlobalParamsEntry()
-	if err != nil {
-		return errors.Wrapf(err, "UtxoView.IsValidUnjailValidatorMetadata: error retrieving SnapshotGlobalParamsEntry: ")
-	}
+	// Retrieve the ValidatorJailEpochDuration from the current global params. It's safe to use the current global
+	// params here because the changes made to locked stake do not affect the PoS consensus until they are
+	// snapshotted.
+	currentGlobalParamsEntry := bav.GetCurrentGlobalParamsEntry()
```
diamondhands0, line 1689: Interesting. Makes sense.

lib/block_view_validator.go
```diff
@@ -1794,27 +1752,6 @@ func (bav *UtxoView) SanityCheckUnregisterAsValidatorTxn(
 		return errors.New("SanityCheckUnregisterAsValidatorTxn: TotalUnstakedAmountNanos doesn't match")
 	}
```
diamondhands0, line 1753: Feels good we don't need this sanity check anymore :)

lib/pos_leader_schedule.go
tholonious: Correct. This was a subtle bug. It would have resulted in an extra 2 epoch delay for leader schedule size changes to take effect. In that 2 epoch window, we would have had a mismatch between the leader schedule's actual size, and what the snapshotted global params say leader schedule size should be.

We would have hit this eventually when implementing the consensus.

lib/block_view_stake.go
```diff
@@ -32,9 +32,17 @@ import (
 // TYPES: StakeEntry
 //
 
+type StakingRewardMethod = uint8
+
+const (
+	StakingRewardMethodPayToBalance StakingRewardMethod = iota
```
diamondhands0, line 38: I'd avoid using iota. If you're going to use it I think the proper way is to then not define the other numbers. So like this:
```
StakingRewardMethodPayToBalance StakingRewardMethod = iota
StakingRewardMethodRestake      StakingRewardMethod
StakingRewardMethodWhatever1      StakingRewardMethod
StakingRewardMethodWhatever2      StakingRewardMethod
```

Obviously I hate this because you can't immediately look at the code and know what the number is without, at minimum, help from an IDE.

So we instead just go 0, 1, 2, ... and avoid using iota entirely.

lib/pos_staking_rewards.go
```diff
@@ -0,0 +1,282 @@
...
+}
+
+// computeStakingReward uses float math to compute the compound interest on the stake amounts based on the
+// elapsed time since the last staking reward distribution and the APY.
+//
+// It produces the result for: stakeAmount * [e ^ (apy * elapsedTime / 1 year) - 1]
+func computeStakingReward(stakeAmountNanos *uint256.Int, elapsedFractionOfYear *big.Float, apy *big.Float) *big.Float {
+	stakeAmountFloat := NewFloat().SetInt(stakeAmountNanos.ToBig())
+	growthExponent := NewFloat().Mul(elapsedFractionOfYear, apy)                // apy * elapsedTime / 1 year
+	growthMultiplier := BigFloatPow(bigE, growthExponent)                       // e ^ (apy * elapsedTime / 1 year)
```
diamondhands0, line 269: Let's get rid of `bigE` and use `BigFloatExp` here, which should already exist.

lib/deso_math.go
```diff
@@ -99,6 +100,7 @@ var (
 	bigOne            = NewFloat().SetUint64(1)
 	bigTwo            = NewFloat().SetUint64(2)
 	bigSqrt2Over2     = NewFloat().Quo(bigSqrt2, bigTwo)
+	bigE              = NewFloat().SetFloat64(2.71828182845904523536028747135266249775724709369995957496696763)
```
diamondhands0, line 103: Let's use BigFloatExp instead and delete bigE. I don't remember why but I think it's much more efficient than BigFloatPow for some reason. Not that it matters, as I doubt this will be a bottleneck.

lib/network.go
```diff
@@ -1890,6 +1890,11 @@ type MsgDeSoHeader struct {
 	// mined.
 	TstampSecs uint64
 
+	// TODO: Add a new TstampNanoSecs field that will have nanosecond resolution.
+	// For backwards compatibility with existing backends and frontends, we will keep
+	// the existing TstampSecs field and populate it as TstampNanoSecs / 1e9 when decoding
+	// block headers from bytes.
```
diamondhands0, line 1896: I was thinking about this before reading this comment. I think this approach makes sense. Maybe we rename TstampSecs to LegacyTstampSecs.

lib/pos_epoch.go
diamondhands0: I think we just wanted a way to detect that we were at the genesis. So it's kindof like a "marker" value. That said, I agree the ProofOfStake1StateSetupBlockHeight is a better thing to return. Feel free to change it. Just make sure you also change what calls it so that they're properly picking up the market value.

lib/pos_epoch_complete_hook.go
```diff
@@ -45,62 +65,168 @@ func (bav *UtxoView) RunEpochCompleteHook(blockHeight uint64) error {
...
+
+	// Filter the top n stake entries by the current validator set. We do not want to reward stakes that are
+	// not in the current validator set, so we pre-filter them here before snapshotting them.
+	validatorSetPKIDs := NewSet([]PKID{})
+	for _, validatorEntry := range validatorSet {
+		validatorSetPKIDs.Add(*validatorEntry.ValidatorPKID)
+	}
+	topStakesInValidatorSet := collections.SliceFilter(topStakeEntries, func(s *StakeEntry) bool {
+		return validatorSetPKIDs.Includes(*s.ValidatorPKID)
+	})
```
diamondhands0, line 224: So just wanted to clarify: I think what this will do is we will get back the max stake entries, say we get back 10k, and then cull out stake entries that are attached to either jailed validators or validators who don't fit in the top validator set.

I think this makes sense. We should definitely cut out the jailed validators. I'm just kinda wondering if we should still reward people who stake to validators who don't participate in consensus (i.e. who are beyond the max validators). The only reason I can see for incentivizing that is to prevent a situation where people only want to stake to the top ten validators or something, for fear of not being rewarded. For example, this rule would certainly hurt the Nth validator, and probably up to the N-100th or so.

Anyway I think it's fine. Let's leave it as is. Just lmk if I misunderstood anything above.

tholonious: @diamondhands0 you've understood correctly. The fact that we additionally filter stakers after fetching the top 10k has the downside you mentioned.

An alternative approach I was considering was to make the DB seek over the top stake entries dynamic. I.e. while doing the seek, we can filter out stake entries that are not part of the top n validators as well. This way, we're guaranteed to get 10k stake entries and they are ALL part of the active validator set. 

The downside of the alternative approach is that, it results in an uncapped DB seek. Meaning, if we have 10 million stake entries, but we want to fetch the top 10k that are part of the active validator set, we may end up seeking very deep into the list of 10 mil stake entries that satisfy the active validator set filter. This could be costly.

diamondhands0: I was messaging in Slack but going to consolidate my thoughts here.

I think that leaving it as-is is fine for now. Once we hit the 10k stake threshold, this is easy enough to fix with a hard fork.

The above being said, if we wanted to do this the "right" way, then I think it would actually involve the following:
1. Include the validator status in `PrefixStakeByStakeAmount` at the beginning like we do for validators themselves. This will make it so that the 10k stakers we fetch are attached to unjailed validators.
2. Then, if you want, you can eliminate the check on validators being in the top N. This would resolve the perverse incentive to avoid smaller validators, since you get paid as long as your boy isn't jailed.

diamondhands0: Nevermind. I realized you can't limit to the validators in the top N because then you wouldn't be able to figure out their commissions. I think every chain has this issue though so it's nbd.

I'd just leave this as is. Once we hit the 10k threshold we can apply whatever the proper fix is.

tholonious: Updated up here to consolidate our chat on Slack too. The solution I proposed above expresses quite simply in the code actually. You can check it out here: https://github.com/deso-protocol/core/pull/632

After seeing this solution in action, I'd be pretty happy to just ship it now and have one less nuance to think about in the top 10k filtering for staker snapshots.

lib/network.go
tholonious: This code comment is resolved here: https://github.com/deso-protocol/core/pull/631
Marking this as resolved since you've reviewed PR 631 already and like the solution there.

lib/pos_staking_rewards.go
tholonious: Nah, definitely possible. The imperfect precision may lead to the quotient becoming a minuscule positive value, but it will never underflow zero. I'll add the check here anyway

lib/pos_staking_rewards.go
tholonious: Sounds good. I've made a not of this as a requirement for Rosetta when we eventually get to it.

lib/block_view_stake.go
tholonious: Ah gotchu. I'll update to use explicit the values 0 and 1 then

lib/pos_staking_rewards.go
tholonious: Not the end of the world here. In a practical case where the APY is ex: 5%, the hourly reward on 1 DESO would be 5707.78 nanos which rounds down to 5707. And 5707/5707.78 = 1bp. So the 1e9 precision doesn't result in a meaningful loss for the small staker.

That said, I totally agree that we're OK to limp along for now, but we are starting to push the boundary on 1e9 scaling becoming less practical.

lazynina (approved):

PR #581, opened 2023-07-07
PoS Decouple Validator Jailing from Validator Snapshotting

tholonious:


tholonious:
* feature/proof-of-stake
    * **PR #581** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/581" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/581?utm_source=stack-comment).

lib/block_view_validator.go
```diff
@@ -1937,6 +1937,65 @@ func (bav *UtxoView) GetGlobalActiveStakeAmountNanos() (*uint256.Int, error) {
...
+		if shouldJailValidator {
+			if err = bav.JailValidator(validatorEntry); err != nil {
+				return errors.Wrapf(
+					err, "JailAllInactiveValidators: problem jailing validator %v: ", validatorEntry.ValidatorPKID,
+				)
+			}
+		}
+
+		// We don't want to retrieve any ValidatorEntries from the db that are present in the UtxoView.
+		utxoViewValidatorPKIDs = append(utxoViewValidatorPKIDs, validatorEntry.ValidatorPKID)
```
lazynina, line 1969: I think we want to add all validator PKIDs to the utxo validator PKIDs so we don't bother fetching them from the DB even if the validator entry is deleted? also we can reduce the nested if.
```suggestion
		// We don't want to retrieve any ValidatorEntries from the db that are present in the UtxoView.
		utxoViewValidatorPKIDs = append(utxoViewValidatorPKIDs, validatorEntry.ValidatorPKID)
		
		if validatorEntry.isDeleted {
			continue
		}

		// Check if we should jail the validator.
		shouldJailValidator, err := bav.ShouldJailValidator(validatorEntry, blockHeight)
		if err != nil {
			return errors.Wrapf(
				err,
				"JailAllInactiveValidators: problem determining if should jail validator %v: ",
				validatorEntry.ValidatorPKID,
			)
		}

		// Jail them if so.
		if !shouldJailValidator {
			continue
		}
		
		if err = bav.JailValidator(validatorEntry); err != nil {
			return errors.Wrapf(
				err, "JailAllInactiveValidators: problem jailing validator %v: ", validatorEntry.ValidatorPKID,
			)
		}
```

lib/block_view_validator.go
```diff
@@ -1937,6 +1937,65 @@ func (bav *UtxoView) GetGlobalActiveStakeAmountNanos() (*uint256.Int, error) {
...
+		// Check if we should jail the validator.
+		shouldJailValidator, err := bav.ShouldJailValidator(validatorEntry, blockHeight)
+		if err != nil {
+			return errors.Wrapf(
+				err, "JailAllInactiveValidators: problem determining if should jail validator %v: ", validatorEntry.ValidatorPKID,
+			)
+		}
+		// Jail them if so.
+		if shouldJailValidator {
+			if err = bav.JailValidator(validatorEntry); err != nil {
```
lazynina, line 1988: reduce nested if? up to you really, but I prefer to `continue` instead of the nested if err pattern

lazynina (changes requested):

lib/block_view_validator.go
tholonious: Great catch. Thanks!

lib/block_view_validator.go
tholonious: Agreed. Good suggestion

lib/block_view_validator.go
```diff
@@ -1937,6 +1937,69 @@ func (bav *UtxoView) GetGlobalActiveStakeAmountNanos() (*uint256.Int, error) {
...
+		shouldJailValidator, err := bav.ShouldJailValidator(validatorEntry, blockHeight)
+		if err != nil {
+			return errors.Wrapf(
+				err,
+				"JailAllInactiveValidators: problem determining if should jail validator %v: ",
+				validatorEntry.ValidatorPKID,
+			)
+		}
+
+		// Jail them if so.
```
lazynina, line 1962:
```suggestion
		// If this validator should not be jailed, continue to the next validator.
```

lib/block_view_validator.go
```diff
@@ -1937,6 +1937,69 @@ func (bav *UtxoView) GetGlobalActiveStakeAmountNanos() (*uint256.Int, error) {
...
+		if !shouldJailValidator {
+			continue
+		}
+
```
lazynina, line 1966:
```suggestion
// If we're here, this validator should be jailed.
```

or something like that.

lib/block_view_validator.go
```diff
@@ -1937,6 +1937,69 @@ func (bav *UtxoView) GetGlobalActiveStakeAmountNanos() (*uint256.Int, error) {
...
+	for _, validatorEntry := range dbValidatorEntries {
+		// Check if we should jail the validator.
+		shouldJailValidator, err := bav.ShouldJailValidator(validatorEntry, blockHeight)
+		if err != nil {
+			return errors.Wrapf(
+				err, "JailAllInactiveValidators: problem determining if should jail validator %v: ", validatorEntry.ValidatorPKID,
+			)
+		}
+		// Jail them if so.
```
lazynina, line 1988: comment clean up

lazynina (changes requested):

lazynina (approved):

PR #599, opened 2023-07-14
PoS Mempool Balance Ledger

AeonSw4n:


AeonSw4n:
* feature/proof-of-stake
      * **PR #599** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/599" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/599?utm_source=stack-comment).

lib/pos_mempool_ledger.go
```diff
@@ -0,0 +1,101 @@
...
+func (pml *BalanceLedger) DecreaseBalance(publicKey PublicKey, amount uint64) {
+	pml.Lock()
+	defer pml.Unlock()
+
+	reservedBalance, exists := pml.reservedBalancesNanos[publicKey]
+	if !exists {
+		return
+	}
+	if amount > reservedBalance {
+		pml.reservedBalancesNanos[publicKey] = 0
```
tholonious, line 80:
```suggestion
		delete(bl.reservedBalancesNanos, publicKey)
```

lib/pos_mempool_ledger.go
```diff
@@ -0,0 +1,101 @@
...
+// GetReservedBalanceNanos returns the user's reserved balance in nanos.
+func (pml *BalanceLedger) GetReservedBalanceNanos(publicKey PublicKey) uint64 {
+	pml.RLock()
+	defer pml.RUnlock()
+
+	reservedBalance, _ := pml.reservedBalancesNanos[publicKey]
+	return reservedBalance
+}
+
+func (pml *BalanceLedger) Reset() {
```
tholonious, line 96:
```suggestion
func (bl *BalanceLedger) Reset() {
```

tholonious (changes requested):
The logic LGTM overall. Just left some naming suggestions in-line. When reading through the code, I found the terms "Increase" and "Decrease" in the public interface to be ambiguous: i.e. does "Increase" reserve more coins, or does it free coins to the pk's spendable balance so they can be spent?

Since the role of the ledge is to track "reserved" balance, I suggest using more explicit verbs like "Reserve" and "Free" so it's dead-simple for readers to understand what the public functions are doing.

lib/pos_mempool_ledger.go
```diff
@@ -0,0 +1,101 @@
...
+	return nil
+}
+
+// IncreaseBalance increases the user's reserved balance by the given amount.
+func (pml *BalanceLedger) IncreaseBalance(publicKey PublicKey, amount uint64) {
+	pml.Lock()
+	defer pml.Unlock()
+
+	reservedBalance, _ := pml.reservedBalancesNanos[publicKey]
+	pml.reservedBalancesNanos[publicKey] = reservedBalance + amount
```
lazynina, line 67: Overflow check?

lib/pos_mempool_ledger.go
```diff
@@ -0,0 +1,101 @@
...
+	reservedBalance, exists := pml.reservedBalancesNanos[publicKey]
+	if !exists {
+		return
+	}
+	if amount > reservedBalance {
+		pml.reservedBalancesNanos[publicKey] = 0
+		return
+	}
+
+	pml.reservedBalancesNanos[publicKey] = reservedBalance - amount
```
lazynina, line 84: Underflow check

lazynina (changes requested):

lib/pos_mempool_ledger.go
```diff
@@ -0,0 +1,101 @@
...
+	"sync"
+)
+
+// BalanceLedger is a simple in-memory ledger of reserved balances for user public keys. The values in the ledger can be
+// increased or decreased, as long as user's reserved balance doesn't exceed the user's spendable balance.
+type BalanceLedger struct {
+	sync.RWMutex
+
+	// Map of public keys to reserved balances in nanos.
+	reservedBalancesNanos map[PublicKey]uint64
```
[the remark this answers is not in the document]
AeonSw4n: Cool idea! Abstracting away a thread-safe map into a reusable struct is a clean solution. However, as I've started implementing this generic and using it in BalanceLedger, I've encountered some issues. The main issue occurred when I tried rewriting the `IncreaseBalance` function. In this case, we need to perform a read and then a write. We first read the user's balance and then write the balance + amount. The problem is that the `RwLockMap` functions `Get` and `Set` release locks. So, theoretically, some undesirable instruction could sneak its way in between and 180 the balance. We would need something that does a Get and a Set under a single lock.

To solve this, I came up with an `Update` method:
```
func (m *RWLockMap[K, V]) Update(key K, expression func(V, bool) V) {
	m.Lock()
	defer m.Unlock()

	value, exists := m._innerMap[key]
	newValue := expression(value, exists)

	m._innerMap[key] = newValue
}
```
This method works nicely because we can encapsulate the "+ amount" logic inside the `expression` argument. But even with this function, some things become problematic. Let's say we want to check for the overflow after a read but before the write. In such a case, we want to return an error. However, this means we would need to do this inside the `expression`, which should, in turn, actually return `(V, error)`. Similarly, `Update` should now return an `error`. More complexity. 

Though I'm chill with complexity. What do you think? If you're fine with having the `Update` method, I'm happy to add this generic.

lib/pos_mempool_ledger.go
```diff
@@ -0,0 +1,101 @@
...
+	reservedBalance, exists := pml.reservedBalancesNanos[publicKey]
+
+	// Check for reserved balance overflow.
+	if exists && amountNanos > math.MaxUint64-reservedBalance {
+		return errors.Errorf("CheckBalanceIncrease: Reserved balance overflow")
+	}
+
+	newReservedBalance := reservedBalance + amountNanos
+	if newReservedBalance > spendableBalanceNanos {
+		return errors.Errorf("PosMempool.AddTransaction: Not enough balance to cover txn fees "+
```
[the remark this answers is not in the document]
AeonSw4n: Oops

lib/pos_mempool_ledger.go
AeonSw4n: Oh, I think we got one just above :)

lib/pos_mempool_ledger.go
AeonSw4n: Good call, added!

lib/pos_mempool_ledger.go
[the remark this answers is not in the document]
tholonious: In this case, migrating to a generic RWLockMap data structure may be overkill. You make the call on whether to implement it or not @AeonSw4n. IMO, the code reads clearly as-is and the amount of code de-duplication we gain by implementing the RWLockMap is minimal.

lib/pos_mempool_ledger.go
```diff
@@ -0,0 +1,110 @@
...
+	balance, exists := bl.balances[publicKey]
+
+	// Check for balance overflow.
+	if exists && amount > math.MaxUint64-balance {
+		return errors.Errorf("CheckBalanceIncrease: balance overflow")
+	}
+
+	newBalance := balance + amount
+	if newBalance > maxBalance {
+		return errors.Errorf("PosMempool.AddTransaction: Not enough balance to cover txn fees "+
```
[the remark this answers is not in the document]
AeonSw4n: My bad, I thought I've updated this. No clue why it was still here lol

lib/pos_mempool_ledger.go
```diff
@@ -0,0 +1,110 @@
...
+	// Check for balance underflow.
+	if amount > balance {
+		delete(bl.balances, publicKey)
+		return
+	}
+
+	bl.balances[publicKey] = balance - amount
+}
+
+// GetBalance returns the user's balance in nanos.
```
tholonious, line 96:
```suggestion
// GetBalance returns the user's balance.
```

tholonious (approved):

PR #600, opened 2023-07-14
PoS Mempool Transaction Validation

AeonSw4n:

PR #601, opened 2023-07-14
PoS Database Interface

AeonSw4n:


AeonSw4n:
* feature/proof-of-stake
          * **PR #601** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/601" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/601?utm_source=stack-comment).

lib/db_bolt.go
```diff
@@ -0,0 +1,237 @@
+package lib
```
tholonious, line 1: @AeonSw4n  You can deprioritize the BoltDB exploration. We don't intend to merge this code for now anyway. You can move the code to a separate experimental branch.

For context, Nina & Zordon have been separately looking at reducing Badger memory usage by tuning the badger config options. Let's see where that ends up, and what the holistic memory usage of the final PoS node is before introducing a new KV data store.

lib/db_badger.go
```diff
@@ -0,0 +1,213 @@
+package lib
```
tholonious, line 1: This file has no dependencies on the lib directory! You can move the files in this PR to their own packages under the root directory of the project. I suggest a directory structure like this: `core/storage/badger`.

This way, the code in this PR can decoupled from the lib package, and imported appropriately where needed.

tholonious (changes requested):

lib/db_bolt.go
AeonSw4n: Moved to #617

go.mod
```diff
@@ -48,6 +48,7 @@ require (
 
 require (
 	github.com/Microsoft/go-winio v0.4.16 // indirect
+	github.com/boltdb/bolt v1.3.1 // indirect
```
lazynina, line 51: Let's delete boltdb deps from this PR
```suggestion
```

storage/badger.go
```diff
@@ -0,0 +1,230 @@
...
+// ==========================
+// BadgerIterator
+// ==========================
+
+type BadgerIterator struct {
+	it  *badger.Iterator
+	ctx *BadgerContext
+}
+
+func NewBadgerIterator(it *badger.Iterator, ctx *BadgerContext) *BadgerIterator {
```
lazynina, line 153: Does this iterator need to support reverse iteration?

storage/database_test.go
```diff
@@ -0,0 +1,471 @@
...
+	t.Logf("BadgerDB directory: %s\nIt should be automatically removed at the end of the test", dir)
+	require.NoError(err)
+
+	opts := DefaultBadgerOptions(dir)
+	db := NewBadgerDatabase(opts, true)
+	require.NoError(db.Setup())
+	defer db.Erase()
+	defer db.Close()
+
+	badgerCtx := db.GetContext([]byte{})
```
lazynina, line 250: It seems like a lot of these tests do exactly the same thing. can we refactor a bit so it's easier to read? something like:
```
func _doBadgerTest(testConfig *TestConfig, dir string, useDefaultOptions bool)
```

or just pass badger options in. This way we don't have so much clutter. 

Also, do we really need this many tests for opening performance badger DBs and writing to disk? Are we going to run these tests in CI? If we're running them in CI, I would recommend we reduce the number of tests to what is absolutely necessary so we're not clogging up CI runs. Additionally, limiting the number of usages of performance options (when running in CI) would be great.

lazynina (changes requested):

storage/badger.go
```diff
@@ -0,0 +1,230 @@
...
+}
+
+func NewBadgerTransaction(txn *badger.Txn, wb *badger.WriteBatch) *BadgerTransaction {
+	return &BadgerTransaction{
+		txn: txn,
+		wb:  wb,
+	}
+}
+
+func (btx *BadgerTransaction) Set(key []byte, value []byte, ctx Context) error {
```
tholonious, line 95: Why pass in a `Context` here? The caller should know the exact key they want to store the value in, so having an additional `Context` will further obfuscate the operation to the caller (and future devs who use this struct).

I understand that you have a more important role for the `Context` struct in mind for BoltDB. We do not currently plan to merge any of the BoltDB code, and haven't made plans to do so. So, it's worth keeping this interface as simple as possible until we cross that bridge.

My suggestion would be to remove the Context parameter here.

storage/badger.go
```diff
@@ -0,0 +1,230 @@
...
+		return errors.Wrapf(err, "Set:")
+	}
+
+	if btx.wb != nil {
+		return btx.wb.Set(prefixedKey, value)
+	}
+	return btx.txn.Set(prefixedKey, value)
+}
+
+func (btx *BadgerTransaction) Delete(key []byte, ctx Context) error {
```
tholonious, line 107: Same suggestion as above to remove the Context parameter. The caller should know the prefix they want to write to.
```suggestion
func (btx *BadgerTransaction) Delete(key []byte) error {
```

storage/badger.go
```diff
@@ -0,0 +1,230 @@
...
+		return errors.Wrapf(err, "Delete:")
+	}
+
+	if btx.wb != nil {
+		return btx.wb.Delete(prefixedKey)
+	}
+	return btx.txn.Delete(prefixedKey)
+}
+
+func (btx *BadgerTransaction) Get(key []byte, ctx Context) ([]byte, error) {
```
tholonious, line 119: Ditto on removing the Context param here.
```suggestion
func (btx *BadgerTransaction) Get(key []byte) ([]byte, error) {
```

storage/badger.go
```diff
@@ -0,0 +1,230 @@
...
+func (btx *BadgerTransaction) GetIterator(ctx Context) (Iterator, error) {
+	badgerCtx, err := AssertContext[*BadgerContext](ctx, BADGERDB)
+	if err != nil {
+		return nil, err
+	}
+	opts := badger.DefaultIteratorOptions
+	it := btx.txn.NewIterator(opts)
+	it.Seek(badgerCtx.prefix)
+	return NewBadgerIterator(it, badgerCtx), nil
+}
```
tholonious, line 142: What's interesting about the `GetIterator(ctx)` interface function is that we generally always want to create a new iterator given a prefix and transaction. Golang's type system is also quite limited (no support for variance) which is partly why it forces you to implement the `AssertContext` function. 

My general suggestion for golang dev is to avoid having functions that require you to pass in interface types that you internally need to validate and cast.

As an alternative here, what you can do is update the constructor for NewBadgerIterator and have it incapsulate the logic for creating the iterator:
```suggestion
func NewBadgerIterator(txn *BadgerTransaction, prefix []byte) *BadgerIterator {
	opts := badger.DefaultIteratorOptions
	it := txn.NewIterator(opts)
	it.Seek(prefix)
	return &BadgerIterator{
		it:  it,
	}
}
```

storage/badger.go
```diff
@@ -0,0 +1,230 @@
...
+
+// ==========================
+// BadgerIterator
+// ==========================
+
+type BadgerIterator struct {
+	it  *badger.Iterator
+	ctx *BadgerContext
```
tholonious, line 150: Similar to the my comments above, there's no need to store a separate context object here. The iterator can encapsulate all the implementation details of what it needs to iterate over (all keys with the given prefix).

This way, you can eliminate one more layer and moving part in the implementation of the iterator.
```suggestion
	prefix []byte
```

storage/database_test.go
```diff
@@ -0,0 +1,471 @@
...
+}
+
+// TestBadger_Default_Experiment_10MB_Batch is a BadgerDB test in which we write 10MB of data to the database.
+// In the test, we:
+//  1. Write 100 equal size KV items to the database.
+//  2. Remove 20 items from the database.
+//  3. Retrieve 20 items from the database.
+//  4. Iterate over 20 items in the database.
+//  5. Iterate over all items in the database.
+func TestBadger_Default_Experiment_10MB_Batch(t *testing.T) {
```
tholonious, line 69: This is totally on your radar, but Im dropping the comment here so it's contextualized in the PR too. We only need one test using Default options and one test using the Performance options. And in general, both tests need to write the minimum amount of data to disk to test the logic you've built in the wrapper structs above. 

These test will execute in every CI run, so we want to avoid writing large amounts of data to disk and to instead use the minimum overhead needed to test the new logic.

tholonious (changes requested):

storage/database_test.go
lazynina: also note that we should update `test.Dockerfile` to run tests in this directory by adding `"github.com/deso-protocol/core/storage"` on line 32

storage/badger.go
AeonSw4n: The mempool use-case doesn't really need reverse iteration so I think, for now, we can stick with the simple iterator. We will definitely need it in the future.

storage/database_test.go
AeonSw4n: Agreed, I think just one generic test is sufficient to stress test the badger db implementation. So I got rid of all of them but one.

storage/database.go
```diff
@@ -0,0 +1,69 @@
+package storage
+
+type DatabaseId byte
+
+const (
+	BADGERDB DatabaseId = 0
+)
```
tholonious, line 7: This is unused. Should be fine to delete
```suggestion
```

storage/badger.go
```diff
@@ -0,0 +1,186 @@
...
+		return fn(T)
+	})
+}
+
+func (bdb *BadgerDatabase) Close() error {
+	return bdb.db.Close()
+}
+
+func (bdb *BadgerDatabase) Erase() error {
+	return os.RemoveAll(bdb.opts.Dir)
```
tholonious, line 77: Would it be safer to use `bdb.db. DropAll()` instead? https://pkg.go.dev/github.com/dgraph-io/badger#DB.DropAll

Not sure what the intricacies are of how badger caches data in memory, and if manually deleting the data directory is the move. If `os.RemoveAll` is safer, it's worth adding a comment here on why we made the decision to use it here

storage/badger.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+	if bdb.useWriteBatch {
+		err = wb.Flush()
+		if err != nil {
+			return errors.Wrapf(err, "Update: Problem flushing write batch")
+		}
+		wb.Cancel()
+	}
+	return nil
```
tholonious, line 62: Just a nit. It's worth getting into the habit of un-nesting if blocks where possible.
```suggestion
	if !bdb.useWriteBatch {
		return nil
	}
	
	if err = wb.Flush(); err != nil {
		return errors.Wrapf(err, "Update: Problem flushing write batch")
	}
	
	wb.Cancel()
```

lazynina: agreed^

storage/badger.go
```diff
@@ -0,0 +1,186 @@
...
+}
+
+func (btx *BadgerTransaction) Get(key []byte) ([]byte, error) {
+	var value []byte
+
+	item, err := btx.txn.Get(key)
+	if err != nil {
+		return value, errors.Wrapf(err, "Get:")
+	}
+	return item.ValueCopy(nil)
```
lazynina, line 117:
```suggestion
	item, err := btx.txn.Get(key)
	if err != nil {
		return nil, errors.Wrapf(err, "Get:")
	}
	return item.ValueCopy(nil)
```
I don't think we need to declare this `value` variable

lazynina:
after looking through the other PRs in this stack, I'm not sure I see why we need this DB interface necessarily since it's merely wrapping existing badger functions and moves a lot of code around. I'm not sure the additional complexity by adding this layer in is offset by the value of the abstraction at this time and we should consider removing this piece from the stack if possible.

PR #602, opened 2023-07-14
PoS Mempool Persister

AeonSw4n:


AeonSw4n:
* feature/proof-of-stake
            * **PR #602** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/602" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/602?utm_source=stack-comment).

lib/pos_mempool_persister.go
```diff
@@ -0,0 +1,205 @@
...
+	mp.Lock()
+	defer mp.Unlock()
+
+	if len(mp.updateBatch) == 0 {
+		return nil
+	}
+
+	localContext := mp.dbCtx.GetContext([]byte(DbMempoolContextId))
+	err := mp.dbCtx.Update(localContext, func(txn storage.Transaction, ctx storage.Context) error {
+		for _, event := range mp.updateBatch {
```
tholonious, line 124: This function and the next one are the only ones that read/write to the DB. And the DB operations they make are quite simple. Set, Delete, and Iteration+Read. Performing the same operations using Badger's API would be just as straightforward. 

If that's the case, then the DB interface in the previous PR doesn't add much (any?) value is making it easier to swap in a new DB. But the interface is a quite large middle layer we need to maintain.

There's a case to be made for deprecating the DB interface, and moving it to your experimental branch. And to instead perform the DB operations here using Badger's API directly.

lib/event_manager.go
```diff
@@ -82,3 +89,13 @@ func (em *EventManager) blockAccepted(event *BlockEvent) {
 		handler(event)
 	}
 }
+
+func (em *EventManager) OnMempoolEvent(handler MempoolEventFunc) {
```
tholonious, line 93: The use of the event manager to abstract away the persister logic is 🔥

lib/pos_mempool_persister.go
```diff
@@ -0,0 +1,205 @@
...
+	mp.reset()
+	mp.stopGroup.Add(1)
+	mp.startGroup.Add(1)
+	go mp.run()
+	mp.startGroup.Wait()
+	mp.status = MempoolPersisterStatusRunning
+}
+
+func (mp *MempoolPersister) run() {
+	mp.startGroup.Done()
```
tholonious, line 71: Seems like you can just set the status to running here. If you do that, then the `startGroup` wait group becomes redundant and you can remove it. And all status changes happen within this function so they're guaranteed to be on the same thread
```suggestion
	mp.status = MempoolPersisterStatusRunning
```

lib/pos_mempool_persister.go
```diff
@@ -0,0 +1,205 @@
...
+	})
+	mp.updateBatch = nil
+
+	if err != nil {
+		return errors.Wrapf(err, "MempoolPersister: Error persisting batch")
+	}
+	return nil
+}
+
+func (mp *MempoolPersister) RetrieveTransactions() ([]*MempoolTx, error) {
```
tholonious, line 155: This only returns the persisted transactions (not the ones in the updateBatch). So maybe rename to that to make the usage clearer.

Also, is there ever a scenario where you would want this to return all txns (both from the DB, and from the current un-persisted batch)? Doesn't seem like it since we would really only run this function on node boot-up
```suggestion
func (mp *MempoolPersister) GetPersistedTransactions() ([]*MempoolTx, error) {
```

lib/pos_mempool_persister.go
```diff
@@ -0,0 +1,205 @@
...
+		case event := <-mp.eventQueue:
+			switch event.Type {
+			case MempoolEventAdd, MempoolEventRemove:
+				mp.Lock()
+				mp.updateBatch = append(mp.updateBatch, event)
+				mp.Unlock()
+
+			case MempoolEventExit:
+				mp.status = MempoolPersisterStatusNotRunning
+				mp.stopGroup.Done()
```
tholonious, line 83: I think you also want to close the event queue channel here, no? It signals to the caller that the channel won't accept any more events. I know you `reset()` everything later on, but the existing instance of the channel is still open until it's garbage collected
```suggestion
				close(mp.eventQueue)
				mp.stopGroup.Done()
```

lib/pos_mempool_persister.go
```diff
@@ -0,0 +1,205 @@
...
+// MempoolPersister is responsible for persisting transactions in the mempool to the database. Whenever a transaction is
+// added or removed from the mempool, the MempoolPersister is notified via the OnMempoolEvent callback. The Persister
+// will then add the event to a queue. Periodically, the transaction queue is flushed to the database and all the cached
+// transactions are persisted.
+type MempoolPersister struct {
+	sync.Mutex
+	stopGroup  sync.WaitGroup
+	startGroup sync.WaitGroup
+
+	frequencyMilliseconds int
```
tholonious, line 41: Always nice to clarify naming
```suggestion
	batchPersistFrequencyMilliseconds int
```

tholonious (changes requested):

lib/pos_mempool_persister.go
AeonSw4n: Ah this is an interesting little synchronization problem. Since `run()` is executed in a goroutine, it's possible that it'll not execute right away. If somehow we publish events to the persister before the `run()` goroutine gets scheduled, then these events will be lost. Having the startGroup ensures that the run loop is up and running.

lib/pos_mempool_persister.go
AeonSw4n: Feels like in such a case mempool will be available and we could use it to retrieve the full view of transactions.

lib/pos_mempool_transaction_test.go
```diff
@@ -0,0 +1 @@
+package lib
```
lazynina, line 1: delete this file or add a relevant test please

lib/pos_mempool_persister.go
tholonious: Ah, makes sense. It's worth leaving a comment here to explain that edge for future readers. Tricky stuff.

lib/pos_mempool_persister.go
```diff
@@ -0,0 +1,239 @@
...
+		return nil
+	})
+	if err != nil {
+		return nil, errors.Wrapf(err, "MempoolPersister: Error retrieving transactions")
+	}
+	return mempoolTxns, nil
+}
+
+// OnMempoolEvent is used to add a transaction event to the eventQueue.
+func (mp *MempoolPersister) OnMempoolEvent(event *MempoolEvent) {
```
tholonious, line 225: Nit: clearer naming that decouples the persister's naming from the event based flow you have above.
```suggestion
// EnqueueEvent is used to add a transaction event to the eventQueue.
func (mp *MempoolPersister) EnqueueEvent(event *MempoolEvent) {
```

tholonious (approved):

PR #603, opened 2023-07-14
PoS Mempool Struct

AeonSw4n:


AeonSw4n:
* feature/proof-of-stake
            * **PR #602** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/602" target="_blank">Graphite</a> 
              * **PR #603** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/603" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/603?utm_source=stack-comment).

AeonSw4n:
* feature/proof-of-stake
            * **PR #602** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/602" target="_blank">Graphite</a> 
              * **PR #603** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/603" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/603?utm_source=stack-comment).

lib/pos_mempool.go
```diff
@@ -0,0 +1,394 @@
...
+
+func (dmp *PosMempool) addTransactionNoLock(txn *MempoolTx) error {
+	userPk := NewPublicKey(txn.Tx.PublicKey)
+	txnFee := txn.Tx.TxnFeeNanos
+	latestBlockHeight := dmp.latestBlockNode.Height
+
+	// Validate that the user has enough balance to cover the transaction fees.
+	spendableBalanceNanos, err := dmp.latestBlockView.GetSpendableDeSoBalanceNanosForPublicKey(userPk.ToBytes(), latestBlockHeight)
+	if err != nil {
+		return errors.Wrapf(err, "CheckBalanceIncrease: Problem getting spendable balance")
```
lazynina, line 218: I think `CheckBalanceIncrease` is confusing here. Is there a better string for this error message?

lib/pos_mempool.go
```diff
@@ -0,0 +1,394 @@
...
+	}
+	if err := dmp.ledger.CanIncreaseBalance(*userPk, txnFee, spendableBalanceNanos); err != nil {
```
lazynina, line 220: Isn't this actually a balance DECREASE? Okay so I see what the ledger is doing now - can we rename these functions so it's more clear what they do. It hink something like `CanIncreaseTxnFeeBalance` or something of that nature.

lib/pos_mempool.go
```diff
@@ -0,0 +1,394 @@
...
+		}
+		dmp.eventManager.mempoolEvent(event)
+	}
+
+	return nil
+}
+
+// RemoveTransaction is the main function for removing a transaction from the mempool.
+func (dmp *PosMempool) RemoveTransaction(txn *MempoolTx) error {
+	if dmp.status == PosMempoolStatusNotRunning {
```
lazynina, line 248: Should we use the IsRunning function here instead?

lib/pos_mempool.go
```diff
@@ -0,0 +1,394 @@
...
+		}
+		dmp.eventManager.mempoolEvent(event)
+	}
+
+	return nil
+}
+
+// GetTransaction returns the transaction with the given hash if it exists in the mempool. This function is thread-safe.
+func (dmp *PosMempool) GetTransaction(txHash *BlockHash) *MempoolTx {
+	if dmp.status == PosMempoolStatusNotRunning {
```
lazynina, line 283: Should we use the IsRunning function here instead?

lib/pos_mempool.go
```diff
@@ -0,0 +1,394 @@
...
+
+	dmp.RLock()
+	defer dmp.RUnlock()
+
+	return dmp.txnRegister.GetTransaction(txHash)
+}
+
+// GetTransactions returns all transactions in the mempool ordered by the Fee-Time algorithm. This function is thread-safe.
+func (dmp *PosMempool) GetTransactions() []*MempoolTx {
+	if dmp.status == PosMempoolStatusNotRunning {
```
lazynina, line 295: Should we use the IsRunning function here instead?

lib/pos_mempool.go
```diff
@@ -0,0 +1,394 @@
...
+//
+//	for it.Next() {
+//		if txn, ok := it.Value(); ok {
+//			// Do something with txn.
+//		}
+//	}
+//
+// Note that the iteration pattern is not thread-safe. Another lock should be used to ensure thread-safety.
+func (dmp *PosMempool) GetIterator() MempoolIterator {
+	if dmp.status == PosMempoolStatusNotRunning {
```
lazynina, line 316: Should we use the IsRunning function here instead?

lib/pos_mempool.go
```diff
@@ -0,0 +1,394 @@
...
+		if err := dmp.removeTransactionNoLock(prunedTxn); err != nil {
+			glog.Errorf("PosMempool.prune: Problem removing transaction from mempool: %v", err)
+		}
+	}
+	return nil
+}
+
+// UpdateLatestBlock updates the latest block view and latest block node in the mempool.
+func (dmp *PosMempool) UpdateLatestBlock(blockView *UtxoView, blockNode *BlockNode) {
+	if dmp.status == PosMempoolStatusNotRunning {
```
lazynina, line 348: Should we use the IsRunning function here instead?

lib/pos_mempool.go
```diff
@@ -0,0 +1,394 @@
...
+	dmp.latestBlockView = blockView
+	dmp.latestBlockNode = blockNode
+}
+
+// UpdateGlobalParams updates the global params in the mempool. Changing GlobalParamsEntry can impact the validity of
+// transactions in the mempool. For example, if the minimum network fee is increased, transactions with a fee below the
+// new minimum will be removed from the mempool. To safely handle this, this method re-creates the TransactionRegister
+// with the new global params and re-adds all transactions in the mempool to the new register.
+func (dmp *PosMempool) UpdateGlobalParams(globalParams *GlobalParamsEntry) {
+	if dmp.status == PosMempoolStatusNotRunning {
```
lazynina, line 364: Should we use the IsRunning function here instead?

lib/pos_mempool.go
```diff
@@ -0,0 +1,394 @@
...
+		// If we get here, it means that the transaction is no longer valid. We remove it from the mempool.
+		removedTxnHashes = append(removedTxnHashes, mempoolTx.Hash.String())
+		if err := dmp.removeTransactionNoLock(mempoolTx); err != nil {
+			glog.Errorf("PosMempool.UpdateGlobalParams: Problem removing txn with hash %v from register: %v",
+				mempoolTx.Hash.String(), err)
+		}
+	}
+
+	if len(removedTxnHashes) > 0 {
+		glog.Errorf("PosMempool.UpdateGlobalParams: Transactions with the following hashes were removed: %v",
```
lazynina, line 389: Should this be glog.Infof instead? I think this is not an error case.

lib/pos_mempool_test.go
```diff
@@ -0,0 +1,109 @@
...
+	newPool := NewPosMempool(params, globalParams, latestBlockView, latestBlockNode, dir, eventManager)
+	require.NoError(newPool.Start())
+	require.Equal(PosMempoolStatusRunning, newPool.status)
+	newPoolTxns := newPool.GetTransactions()
+	require.Equal(2, len(newPoolTxns))
+	newPool.Stop()
+}
+
+// TODO: Add a test for pruning and a lot of transactions
+// TODO: Add a test for removing transactions
```
lazynina, line 109: Do you want to add these tests in a separate PR? I want to see pruning txns tested as well as updating global params + pruning.

lib/pos_transaction_register.go
```diff
@@ -249,6 +250,18 @@ func (tr *TransactionRegister) GetFeeTimeTransactions() []*MempoolTx {
...
 }
 
+// GetTransaction returns the transaction with the given hash if it exists in the register, or nil otherwise.
+func (tr *TransactionRegister) GetTransaction(hash *BlockHash) *MempoolTx {
+	tr.RLock()
+	defer tr.RUnlock()
+
+	if hash == nil {
+		return nil
+	}
```
lazynina, line 260:
```suggestion
	if hash == nil {
		return nil
	}
	tr.RLock()
	defer tr.RUnlock()
```
We can acquire the read lock after handling the nil case. This way we won't have to wait for writes in the nil block hash case.

lib/block_view.go
```diff
@@ -1607,6 +1607,10 @@ func _isEntryImmatureBlockReward(utxoEntry *UtxoEntry, blockHeight uint32, param
 	return false
 }
 
+func (bav *UtxoView) VerifySignature(txn *MsgDeSoTxn, blockHeight uint32) (_derivedPkBytes []byte, _err error) {
```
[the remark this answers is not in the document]
AeonSw4n: We need it to be public if we intend to separate UtxoView and PosMempool into their own modules -- which I think we should do eventually. I chose to make the wrapper because renaming `_verifySignature` to `VerifySignature` will result in a lot of changes to existing code, unnecessarily.

lib/pos_mempool.go
```diff
@@ -0,0 +1,391 @@
...
+
+	// Add the transaction to the mempool.
+	if err := dmp.AddTransaction(mempoolTx); err != nil {
+		return errors.Wrapf(err, "PosMempool.AddTransaction: Problem adding transaction to mempool")
+	}
+	return nil
+}
+
+// AddTransaction is the main function for adding a new transaction to the mempool.
+func (dmp *PosMempool) AddTransaction(txn *MempoolTx) error {
```
tholonious, line 187: Is there ever a scenario where an external caller will want to use AddTransaction and not ProcessMsgDeSoTxn? IMO, it shouldn't be possible for callers to add txns to the mempool, bypassing internal validations. If you agree, then we may as well rename ProcessMsgDeSoTxn to `AddTransaction(txn *MsgDeSoTxn)`, and collapse the two functions into one to simplify the interface

lib/pos_mempool.go
```diff
@@ -0,0 +1,391 @@
...
+			Type: MempoolEventAdd,
+		}
+		dmp.persister.EnqueueEvent(event)
+	}
+
+	return nil
+}
+
+// RemoveTransaction is the main function for removing a transaction from the mempool.
+func (dmp *PosMempool) RemoveTransaction(txn *MempoolTx) error {
```
tholonious, line 244: Would it make sense to update the input parameter here to be the txn hash? In practice, an external caller likely won't have the entire MempoolTx available on-hand. In order to remove a txn, they would first have to fetch up the txn from the mempool by its hash, and pass the txn here. We may as well simplify it for them and have them pass in the txn hash here anyway
```suggestion
func (dmp *PosMempool) RemoveTransaction(txnHash *BlockHash) error {
```

lib/pos_mempool.go
```diff
@@ -0,0 +1,391 @@
...
+
+func (dmp *PosMempool) addTransactionNoLock(txn *MempoolTx) error {
+	userPk := NewPublicKey(txn.Tx.PublicKey)
+	txnFee := txn.Tx.TxnFeeNanos
+	latestBlockHeight := dmp.latestBlockNode.Height
+
+	// Validate that the user has enough balance to cover the transaction fees.
+	spendableBalanceNanos, err := dmp.latestBlockView.GetSpendableDeSoBalanceNanosForPublicKey(userPk.ToBytes(), latestBlockHeight)
+	if err != nil {
+		return errors.Wrapf(err, "CheckBalanceIncrease: Problem getting spendable balance")
```
tholonious, line 215:
```suggestion
		return errors.Wrapf(err, "PosMempool.AddTransaction: Problem getting spendable balance")
```

lib/pos_mempool.go
```diff
@@ -0,0 +1,391 @@
...
+	for _, prunedTxn := range prunedTxns {
+		if err := dmp.removeTransactionNoLock(prunedTxn); err != nil {
+			glog.Errorf("PosMempool.prune: Problem removing transaction from mempool: %v", err)
+		}
+	}
+	return nil
+}
+
+// UpdateLatestBlock updates the latest block view and latest block node in the mempool.
+func (dmp *PosMempool) UpdateLatestBlock(blockView *UtxoView, blockNode *BlockNode) {
```
tholonious, line 344: Same comment as above. May as well just pass in the block height since that's the piece of info we care about
```suggestion
func (dmp *PosMempool) UpdateLatestBlock(blockView *UtxoView, blockHeight uint64) {
```

lib/pos_mempool.go
```diff
@@ -0,0 +1,391 @@
...
+	}
+
+	dmp.RLock()
+	defer dmp.RUnlock()
+
+	return dmp.txnRegister.GetTransaction(txHash)
+}
+
+// GetTransactions returns all transactions in the mempool ordered by the Fee-Time algorithm. This function is thread-safe.
+func (dmp *PosMempool) GetTransactions() []*MempoolTx {
```
tholonious, line 291: Same comment as above. Would it make sense to return MsgDesoTxns here for separation of concerns?
```suggestion
func (dmp *PosMempool) GetTransactions() []*MsgDeSoTxn {
```

lib/pos_mempool.go
```diff
@@ -0,0 +1,391 @@
...
+			Type: MempoolEventRemove,
+		}
+		dmp.persister.EnqueueEvent(event)
+	}
+
+	return nil
+}
+
+// GetTransaction returns the transaction with the given hash if it exists in the mempool. This function is thread-safe.
+func (dmp *PosMempool) GetTransaction(txHash *BlockHash) *MempoolTx {
```
tholonious, line 279: Just curious. Would it make sense to return *MsgDeSoTxn here instead? In a way, the MempoolTx construct is an implementation detail of the mempool. If we can get it to a point where consensus doesn't know about the struct, then we get better separation of concerns, which will make it easier for us decouple/move the mempool code into its own package later on.
```suggestion
func (dmp *PosMempool) GetTransaction(txHash *BlockHash) *MsgDeSoTxn {
```

lib/pos_mempool.go
```diff
@@ -0,0 +1,391 @@
...
+	GetTransaction(txHash *BlockHash) *MempoolTx
+	GetTransactions() []*MempoolTx
+	GetIterator() MempoolIterator
+	UpdateLatestBlock(blockView *UtxoView, blockNode *BlockNode)
+	UpdateGlobalParams(globalParams *GlobalParamsEntry)
+}
+
+type MempoolIterator interface {
+	Next() bool
+	Value() (*MempoolTx, bool)
```
tholonious, line 35: Same comment as below. Should this return *MsgDeSoTxn to hide the MempoolTx struct for external callers? It should make it easier for us to refactor the mempool implementation into its own package later on.
```suggestion
	Value() (*MsgDeSoTxn, bool)
```

lib/pos_mempool.go
```diff
@@ -0,0 +1,391 @@
...
+	// checks if the transaction has a valid signature and if the transaction's sender has enough funds to cover the fee.
+	// The latestBlockView should be updated whenever a new block is added to the blockchain via UpdateLatestBlock.
+	latestBlockView *UtxoView
+	// latestBlockNode is used to infer the latest block height. The latestBlockNode should be updated whenever a new
+	// block is added to the blockchain via UpdateLatestBlock.
+	latestBlockNode *BlockNode
+}
+
+func NewPosMempool(params *DeSoParams, globalParams *GlobalParamsEntry, latestBlockView *UtxoView,
+	latestBlockNode *BlockNode, dir string) *PosMempool {
```
tholonious, line 78: May as well just pass in the block height for simplicity
```suggestion
	latestBlockHeight uint64, dir string) *PosMempool {
```

lib/pos_mempool.go
tholonious, line 291: Wdyt about this for naming clarity?
```suggestion
func (dmp *PosMempool) GetFeeTimeTransactions() []*MempoolTx {
```

tholonious (changes requested):

lib/pos_mempool_test.go
AeonSw4n: No worries, tests are here now. I just didn't want to block code review on me adding a couple of small tests.

lib/pos_mempool.go
```diff
@@ -0,0 +1,391 @@
...
+// mempool ordered by the Fee-Time algorithm. Transactions can be fetched with the following pattern:
+//
+//	for it.Next() {
+//		if txn, ok := it.Value(); ok {
+//			// Do something with txn.
+//		}
+//	}
+//
+// Note that the iteration pattern is not thread-safe. Another lock should be used to ensure thread-safety.
+func (dmp *PosMempool) GetIterator() MempoolIterator {
```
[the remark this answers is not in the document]
AeonSw4n: Ah yes I was thinking that too. I kinda prefer the simpler `GetIterator()`. The Fee-Time ordering is compartmentalized in TransactionRegister and PosMempool doesn't have to necesserily be aware of it. Also I feel like `GetIterator()` is more aligned with the generic `Mempool` interface.

lib/pos_mempool.go
```diff
@@ -0,0 +1,391 @@
...
+		latestBlockView: latestBlockView,
+		latestBlockNode: latestBlockNode,
+	}
+}
+
+func (dmp *PosMempool) Start() error {
+	if dmp.IsRunning() {
+		return nil
+	}
+
```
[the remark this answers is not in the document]
AeonSw4n: This is a fun one. I think a mutex here will cause a deadlock. For instance, `GetPersistedTransactions` will deadlock. Rather than hacking these methods to support a no-lock variant, I think we can actually get away without locks in `Start` and `Stop`. The reason is that `status` is only changed at the very end of `Start` and `Stop`. This means that `Start` will only ever execute if node actually stoped or it was never run in the first place. Similarly, `Stop` will never execute, unless `Start` was run and finished executing. Because of this, I think `Start` and `Stop` are actually thread-safe and don't need locks. But see if you can think of any edge-case.

lib/pos_mempool.go
```diff
@@ -0,0 +1,391 @@
...
+	dmp.status = PosMempoolStatusRunning
+
+	return nil
+}
+
+func (dmp *PosMempool) Stop() {
+	if !dmp.IsRunning() {
+		return
+	}
+
```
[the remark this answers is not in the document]
AeonSw4n: Check my comment regarding holding locks in `Start`

lib/pos_mempool.go
AeonSw4n: I think I like it more as `GetTransactions`. This way the Mempool interface is more generic.

lib/pos_mempool.go
AeonSw4n: Hmm I think it should be an increase. We're trying to add a new transaction to the mempool, so we're checking if the user's reserved balance (in fees) + the new txn fee is within user's spendable balance.

lazynina: I think the name is a little confusing is all. @tholonious - how do you feel about the naming convention here? If it's clear to you, then I'm happy to keep it as is so we don't have a long and unwieldy name. I know what it's doing now, but the name didn't make it clear to me at first glance.

lib/pos_mempool.go
```diff
@@ -0,0 +1,417 @@
...
+		return errors.Wrapf(MempoolErrorNotRunning, "PosMempool.RemoveTransaction: ")
+	}
+
+	dmp.Lock()
+	defer dmp.Unlock()
+
+	// Get the transaction from the register.
+	txn := dmp.txnRegister.GetTransaction(txnHash)
+	if txn == nil {
+		return nil
```
lazynina, line 262: Should this be an error if we try to remove a transaction that isn't found in the register? jw

lib/pos_mempool.go
```diff
@@ -0,0 +1,417 @@
...
+	}
+
+	prunedTxns, err := dmp.txnRegister.PruneToSize(dmp.params.MaxMempoolPosSizeBytes)
+	if err != nil {
+		return errors.Wrapf(err, "PosMempool.pruneNoLock: Problem pruning mempool")
+	}
+	for _, prunedTxn := range prunedTxns {
+		if err := dmp.removeTransactionNoLock(prunedTxn, true); err != nil {
+			glog.Errorf("PosMempool.pruneNoLock: Problem removing transaction from mempool: %v", err)
```
lazynina, line 363: Should we return the error here instead of glogging it?

lib/pos_mempool_test.go
```diff
@@ -0,0 +1,249 @@
...
+	m0PubBytes, _, _ := Base58CheckDecode(m0Pub)
+	m1PubBytes, _, _ := Base58CheckDecode(m1Pub)
+
+	latestBlockView, err := NewUtxoView(db, params, nil, nil)
+	require.NoError(err)
+
+	dir, err := os.MkdirTemp("", "badgerdb-mempool")
+	require.NoError(err)
+	t.Logf("BadgerDB directory: %s\nIt should be automatically removed at the end of the test", dir)
+	defer os.RemoveAll(dir)
```
lazynina, line 90: Should we move this to a helper function and use t.Cleanup? This way we can reduce code duplication across tests. Maybe even a `NewTestPoSMempool` function that takes in the appropriate attributes and then cleans itself up properly upon test completion.

cmd/config.go
```diff
@@ -41,6 +41,9 @@ type Config struct {
 	SnapshotBlockHeightPeriod uint64
 	DisableEncoderMigrations  bool
 
+	// Mempool
+	BatchPersistFrequencyMilliseconds uint64
```
tholonious, line 45: This is a mempool param that's defined outside of the mempool. It's good to contextualize by prefixing the name
```suggestion
	MempoolBatchPersistFrequencyMilliseconds uint64
```

cmd/config.go
```diff
@@ -102,6 +105,9 @@ func LoadConfig() *Config {
 	config.SnapshotBlockHeightPeriod = viper.GetUint64("snapshot-block-height-period")
 	config.DisableEncoderMigrations = viper.GetBool("disable-encoder-migrations")
 
+	// Mempool
+	config.BatchPersistFrequencyMilliseconds = viper.GetUint64("batch-persist-frequency-milliseconds")
```
tholonious, line 109:
```suggestion
	config.MempoolBatchPersistFrequencyMilliseconds = viper.GetUint64("mempool-batch-persist-frequency-milliseconds")
```

cmd/node.go
```diff
@@ -85,6 +85,9 @@ func (node *Node) Start(exitChannels ...*chan struct{}) {
 		node.Params.EnableRegtest()
 	}
 
+	// Set the mempool's batch persist frequency.
+	node.Params.BatchPersistFrequencyMilliseconds = node.Config.BatchPersistFrequencyMilliseconds
```
tholonious, line 89:
```suggestion
	node.Params.MempoolBatchPersistFrequencyMilliseconds = node.Config.MempoolBatchPersistFrequencyMilliseconds
```

cmd/run.go
```diff
@@ -79,6 +79,11 @@ func SetupRunFlags(cmd *cobra.Command) {
 		- hypersync: Will sync by downloading historical state, and will NOT
 		  download historical blocks. Can only be set if HyperSync is true.`)
 
+	// Mempool
+	cmd.PersistentFlags().Uint64("batch-persist-frequency-milliseconds", 30000,
```
tholonious, line 83:
```suggestion
	cmd.PersistentFlags().Uint64("mempool-batch-persist-frequency-milliseconds", 30000,
```

lib/constants.go
```diff
@@ -1110,7 +1113,8 @@ var DeSoMainnetParams = DeSoParams{
 	EncoderMigrationHeights:     GetEncoderMigrationHeights(&MainnetForkHeights),
 	EncoderMigrationHeightsList: GetEncoderMigrationHeightsList(&MainnetForkHeights),
 
-	MaxMempoolPosSizeBytes: 3 << 30, // 3GB
+	MaxMempoolPosSizeBytes:            3 << 30, // 3Gb
+	BatchPersistFrequencyMilliseconds: 30000,
```
tholonious, line 1117:
```suggestion
	MempoolBatchPersistFrequencyMilliseconds: 30000,
```

lib/constants.go
```diff
@@ -1370,7 +1374,8 @@ var DeSoTestnetParams = DeSoParams{
 	EncoderMigrationHeights:     GetEncoderMigrationHeights(&TestnetForkHeights),
 	EncoderMigrationHeightsList: GetEncoderMigrationHeightsList(&TestnetForkHeights),
 
-	MaxMempoolPosSizeBytes: 3 << 30, // 3GB
+	MaxMempoolPosSizeBytes:            3 << 30, // 3Gb
+	BatchPersistFrequencyMilliseconds: 30000,
```
tholonious, line 1378:
```suggestion
	MempoolBatchPersistFrequencyMilliseconds: 30000,
```

lib/constants.go
```diff
@@ -706,6 +706,9 @@ type DeSoParams struct {
 
 	// The maximum aggregate number of bytes of transactions included in the PoS mempool.
 	MaxMempoolPosSizeBytes uint64
+
+	// BatchPersistFrequencyMilliseconds is the frequency with which pos mempool persists transactions to disk.
+	BatchPersistFrequencyMilliseconds uint64
```
tholonious, line 711:
```suggestion
	MempoolBatchPersistFrequencyMilliseconds uint64
```

lib/pos_mempool.go
tholonious: Returning nil here is the desired behavior. When successfully processing a block, the server will try to naively remove the block's txns from the mempool to ensure that a connected txn is no longer in the mempool. So as long as a txn isn't in the mempool at the end of this operation, we consider the remove a success.

lib/pos_mempool.go
tholonious: I guess the assumption here is that this loop will never error because the txn no longer exists in the transaction register, in which case the `dmp.removeTransactionNoLock()` call will never error.

@AeonSw4n, IMO it's fine to replace the `dmp.removeTransactionNoLock()` call with a helper function that just does the following:
```
	// Decrease the appropriate ledger's balance by the transaction fee.
	dmp.ledger.DecreaseBalance(*userPk, txn.Fee)

	// Emit an event for the removed transaction.
	if persistToDb {
		event := &MempoolEvent{
			Txn:  txn,
			Type: MempoolEventRemove,
		}
		dmp.persister.EnqueueEvent(event)
	}
```

This would guarantee there's no error returned, and clarify for us in the future how pruneNoLock() differs from removeTransactionNoLock().

AeonSw4n: Yes, this glog should never happen, except maybe if `prunedTxn` or `prunedTxn.Hash` were nil. Either way, it's mostly a dummy error handler.

@tholonious re the code split, I prefer making the call to `removeTransactionNoLock`. It's nice that both "removeTransaction" and "removeTransactioNoLock" erase transactions. This property will not hold for the new "sub-remove" function. 

Imo it's also intuitive to call an internal remove function for each pruned txn. Perhaps making assumptions that `dmp.removeTransactionNoLock()` doesn't panic on non-existing transactions isn't that bad?

tholonious: @AeonSw4n, that seems fine.  The implementation of `dmp.removeTransactionNoLock()` is encapsulated within the same struct anyway so that's a reasonable assumption. 

Definitely add a comment in the code here that this function call will never error so it's clearer to readers

lib/pos_mempool_test.go
AeonSw4n: Nice! It's cool how `t.Cleanup` is like a better `defer`.

PR #604, opened 2023-07-14
Limit number of chunks in operation channel to reduce hypersync memory

superzordon:
…y consumption

### DIFF
--- a/cmd/node.go
+++ b/cmd/node.go
@@ -141,7 +141,7 @@ func (node *Node) Start(exitChannels ...*chan struct{}) {
 
 	// Setup chain database
 	dbDir := lib.GetBadgerDbPath(node.Config.DataDirectory)
-	opts := lib.DefaultBadgerOptions(dbDir)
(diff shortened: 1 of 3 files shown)


cmd/node.go
```diff
@@ -141,7 +141,7 @@ func (node *Node) Start(exitChannels ...*chan struct{}) {
 
 	// Setup chain database
 	dbDir := lib.GetBadgerDbPath(node.Config.DataDirectory)
-	opts := lib.PerformanceBadgerOptions(dbDir)
+	opts := badger.DefaultOptions(dbDir)
```
AeonSw4n, line 144: This is a significant change that reduces memory by 30-40 GB. But also comes at tradeoffs of smaller Badger transactions. Could you blocksync with this change and see if everything works?

AeonSw4n (changes requested):

diamondhands0 (approved):

lazynina (approved):

PR #611, opened 2023-07-18
Run State Mutating Operations Before Snapshotting in End of Epoch Hook

tholonious:


tholonious:
* feature/proof-of-stake
    * **PR #611** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/611" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/611?utm_source=stack-comment).

lib/pos_epoch_complete_hook.go
```diff
@@ -72,45 +92,95 @@ func (bav *UtxoView) RunEpochCompleteHook(blockHeight uint64) error {
 		return errors.New("RunEpochCompleteHook: CurrentEpochEntry is nil, this should never happen")
 	}
 
-	currentGlobalParamsEntry := bav.GetCurrentGlobalParamsEntry()
-
 	// Snapshot the current GlobalParamsEntry.
-	bav._setSnapshotGlobalParamsEntry(bav.GlobalParamsEntry, currentEpochEntry.EpochNumber)
+	bav._setSnapshotGlobalParamsEntry(bav.GetCurrentGlobalParamsEntry(), currentEpochEntry.EpochNumber)
```
lazynina, line 96: I might move all this snapshotting logic to a separate function. This way we can unit test the snapshotting logic if we want and will help keep this function nice and short. I can see us adding more snapshotting at some point and would be good to confine that to its own function.

lib/pos_epoch_complete_hook.go
```diff
@@ -119,50 +189,16 @@ func (bav *UtxoView) RunEpochCompleteHook(blockHeight uint64) error {
...
 	for _, stakeEntry := range topStakesInValidatorSet {
 		snapshotStakeEntry := SnapshotStakeEntry{
-			SnapshotAtEpochNumber: currentEpochEntry.EpochNumber,
+			SnapshotAtEpochNumber: epochNumber,
 			ValidatorPKID:         stakeEntry.ValidatorPKID,
 			StakerPKID:            stakeEntry.StakerPKID,
 			StakeAmountNanos:      stakeEntry.StakeAmountNanos,
 		}
 		bav._setSnapshotStakeToReward(&snapshotStakeEntry)
 	}
```
lazynina, line 202: so we are no longer pruning old snapshots. This is fine for now I think.

lazynina (changes requested):

lib/pos_epoch_complete_hook.go
tholonious: We never pruned old snapshots. It's still TBD whether/when it's safe to prune them, so we're keeping them indefinitely for now

lib/pos_epoch_complete_hook.go
tholonious: Sure. I'll move each of the three categories of operations into their own top-level functions

lazynina (approved):

PR #615, opened 2023-07-20
Implement Staking Rewards APY Computations

tholonious:


tholonious:
* feature/proof-of-stake
    * **PR #614** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/614" target="_blank">Graphite</a> 
      * **PR #615** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/615" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/615?utm_source=stack-comment).

lib/deso_math.go
```diff
@@ -99,6 +100,7 @@ var (
 	bigOne            = NewFloat().SetUint64(1)
 	bigTwo            = NewFloat().SetUint64(2)
 	bigSqrt2Over2     = NewFloat().Quo(bigSqrt2, bigTwo)
+	bigE              = NewFloat().SetFloat64(2.71828182845904523536028747135266249775724709369995957496696763)
```
lazynina, line 103: should we use `math.E` instead? or do we like how we have precision controlled w/ the constant here?

lib/pos_staking_rewards.go
```diff
@@ -1,5 +1,271 @@
...
+		return nil
+	}
+
+	// Retrieve the current EpochEntry.
+	currentEpochEntry, err := bav.GetCurrentEpochEntry()
+	if err != nil {
+		return errors.Wrapf(err, "DistributeStakingRewardsToSnapshotStakes: problem retrieving current EpochEntry: ")
+	}
+
+	// Check if the current epoch's timestamp is somehow greater than the block timestamp. This should never as long
```
lazynina, line 23:
```suggestion
	// Check if the current epoch's timestamp is somehow greater than the block timestamp. This should never happen as long
```

lib/pos_staking_rewards.go
```diff
@@ -1,5 +1,271 @@
...
+	// as timestamps are moving forward when connecting each block.
+	if currentEpochEntry.CreatedAtBlockTimestampNanoSecs >= blockTimestampNanoSecs {
+		return errors.Errorf("DistributeStakingRewardsToSnapshotStakes: current EpochEntry's CreatedAtBlockTimestampNanoSecs "+
```
lazynina, line 26: Should this wrap a `RuleError` instead? Here's the comment about the purpose of RuleError from the core codebase:
```
// RuleError is an error type that specifies an error occurred during
// block processing that is related to a consensus rule. By checking the
// type of the error the caller can determine that the error was due to
// a consensus rule and determine which consensus rule caused the issue.
```

lib/pos_staking_rewards.go
```diff
@@ -1,5 +1,271 @@
...
+	// their validator.
+	for _, snapshotStakeEntry := range snapshotStakesToReward {
+
+		// Compute the staker's portion of the staking reward, and the validator's commission.
+		stakerReward, validatorCommission, err := bav.computeStakerRewardAndValidatorCommission(snapshotStakeEntry, elapsedFractionOfYear, apy)
+		if err != nil {
+			return errors.Wrapf(err, "DistributeStakingRewardsToSnapshotStakes: problem computing staker reward and validator commission: ")
+		}
+
+		// If either the staker reward or the validator commission is zero, then there's nothing to be done. Move on to the next staker.
```
lazynina, line 76:
```suggestion
		// If both the staker reward or the validator commission is zero, then there's nothing to be done. Move on to the next staker.
```

lib/pos_staking_rewards.go
```diff
@@ -1,5 +1,271 @@
...
+	// their staking rewards over the long run. Validators can opt out of it by disabling reward restaking on their own StakeEntry.
+	// 2. It simplifies the validator commission distribution code by re-using the same code path for distributing staking
+	// rewards. By requiring the validator to already have a StakeEntry for themselves if they want to restake their commissions,
+	// this approach allows us to avoid manually creating new StakeEntries for the validator specifically for restaking commissions.
+	//
+	// TODO: The downside of the above is that it couples the restaking behavior for validator commissions and the validator's own
+	// staking reward. This is fine though, because if the validator wants to restake their own rewards but not their commissions, then
+	// they can stake to themselves using a separate wallet and only enable reward restaking for that StakeEntry.
+	//
+	// If the above isn't desired the behavior, then we can alternatively always pay out validator's commission directly to their wallet.
```
lazynina, line 221: one of the other long term benefits of how this code is set up is that we can track how much a validator got from commissions from a specific staker (if we simply pass in the staker's PKID when we call this function) when we do some utxo operation refactoring work.

lib/pos_staking_rewards.go
```diff
@@ -1,5 +1,271 @@
...
+// computeStakingReward uses float math to compute the compound interest on the stake amounts based on the
+// elapsed time since the last staking reward distribution and the APY.
+//
+// It produces the result for: stakeAmount * [e ^ (apy * elapsedTime / 1 year) - 1]
+func computeStakingReward(stakeAmount *uint256.Int, elapsedFractionOfYear *big.Float, apy *big.Float) *big.Float {
+	stakeAmountFloat := NewFloat().SetInt(stakeAmount.ToBig())
+	growthExponent := NewFloat().Mul(elapsedFractionOfYear, apy)           // apy * elapsedTime / 1 year
+	growthMultiplier := BigFloatPow(bigE, growthExponent)                  // e ^ (apy * elapsedTime / 1 year)
+	finalStakeAmount := NewFloat().Mul(stakeAmountFloat, growthMultiplier) // stakeAmount * [e ^ (apy * elapsedTime / 1 year) - 1]
+	return finalStakeAmount.Sub(finalStakeAmount, stakeAmountFloat)        // stakeAmount * [e ^ (apy * elapsedTime / 1 year) - 1]
```
lazynina, line 260: Technically should the comment on line 259 be `stakeAmount * [e ^ (apy *elapsedTime / 1 year)` since we don't sub the original amount until line 260?

lib/pos_staking_rewards.go
tholonious: Good idea. This can be a RuleError since the blockTimestamp is passed from the block header that's currently being connected

lib/pos_staking_rewards.go
tholonious: You're right. Good catch.

lazynina (approved):

AeonSw4n (approved):

lib/pos_staking_rewards.go
```diff
@@ -1,5 +1,276 @@
...
+	_stakerRewardNanos uint64,
+	_validatorCommissionNanos uint64,
+	_err error,
+) {
+	// Compute the staker's reward amount using a big float math, and immediately convert it to big int
+	// so we can do the remainder of the math using integer operations. This is the only operation where
+	// we need float math.
+	stakerRewardNanos := convertBigFloatToBigInt(
+		computeStakingReward(snapshotStakeEntry.StakeAmountNanos, elapsedFractionOfYear, apy),
+	)
```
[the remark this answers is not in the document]
tholonious: The implementation for Float.Int says the result is nil if the input float is +/- infinity, which is impossible based on how we construct the float. Either way, I've added the nil check

PR #631, opened 2023-08-08
PoS Block Header Followup changes

tholonious:


lib/network.go
```diff
@@ -1886,9 +1886,13 @@ type MsgDeSoHeader struct {
...
-	// The unix timestamp (in seconds) specifying when this block was
-	// mined.
-	TstampSecs uint64
+	// The original TstampSecs struct field is deprecated and replaced by the higher resolution
+	// TstampNanoSecs field. The deprecation is backwards compatible for all existing header
+	// versions and byte encodings. To read or write timestamps with the old 1-second resolution,
+	// use the SetTstampSecs() and GetTstampSecs() public methods.
+
+	// The unix timestamp (in nanoseconds) specifying when this block was produced.
+	TstampNanoSecs uint64
```
diamondhands0, line 1895: Super smart. Love it.

lib/constants.go
```diff
@@ -50,6 +50,19 @@ const (
...
+const (
+	NanoSecondsPerSecond = uint64(1000000000)
+)
+
+func SecondsToNanoSeconds(secs uint64) uint64 {
+	return secs * NanoSecondsPerSecond
+}
+
+func NanoSecondsToSeconds(nanos uint64) uint64 {
+	return nanos / NanoSecondsPerSecond
```
diamondhands0, line 63: Careful about overflow. Just triple-check that overflow is impossible in all places where this is used.

lib/postgres.go
```diff
@@ -1288,7 +1289,7 @@ func (postgres *Postgres) UpsertBlockTx(tx *pg.Tx, blockNode *BlockNode) error {
 
 		TxMerkleRoot: blockNode.Header.TransactionMerkleRoot,
 		Version:      blockNode.Header.Version,
-		Timestamp:    blockNode.Header.TstampSecs,
+		Timestamp:    blockNode.Header.TstampNanoSecs,
```
diamondhands0, line 1292: No need for action here, just note that all this code is deprecated in favor of state syncer, which has a similar logic in it that we can update anytime we want.

migrate/20230803154200_scale_timestamp_to_nanos.go
```diff
@@ -0,0 +1,25 @@
+package migrate
```
diamondhands0, line 1: There's no real need to worry about this because all this Postgres code is going to get deleted once state syncer is complete, but I don't think it matters much either way.

tholonious: I added this code here for completeness, so it's one less breaking change in the interim. Nina's and my expectation was that Postgres in its current form will be deprecated long before this new code ships to prod.

lib/postgres.go
tholonious: Fully expecting to deprecate this once the state syncer work is merged. This change is here for completeness so the code still behaves as expected

lib/constants.go
tholonious: Copying response from slack to consolidate discussion here:

I checked all call-sites for this function and verified that every legacy callsite sets the previous TstampSecs to "seconds since Jan 1 1970". What's nice about that value is that it will always cleanly scale by 1e9 without overflowing. The only scenario where we get an overflow is if a caller passes in a value that's orders of magnitude higher than "seconds since Jan 1 1970", in which case the timestamp is wrong by 300+ years and is invalid anyway. It would fail block header validation. So, you're correct that we'll never hit that issue in practice. If we're REALLY concerned about it, what we can do is use a uint256 for the new TstampNanosSecs field so that it doesn't overflows when scaled by 1e9 in-memory (this has its own trade-offs btw). Viable solution though. I'm personally fine punting on adding this workaround until we see a block in the wild with such an invalid timestamp that fails to decode and scale properly. If one exists, we're guaranteed to see it when syncing a node with the latest code long before we ship PoS.

lazynina (approved):

PR #636, opened 2023-08-11
Update badger sync settings to optimize memory usage during hypersync

superzordon:
Update hypersync to use default badger settings, switch to default settings once hypersync completes

### DIFF
--- a/lib/server.go
+++ b/lib/server.go
@@ -148,6 +148,11 @@ type Server struct {
 	// timer is a helper variable that allows timing events for development purposes.
 	// It can be used to find computational bottlenecks.
 	timer *Timer
+
+	// DbMutex protects the badger database from concurrent access when it's being closed & re-opened.
+	// This is necessary because the database is closed & re-opened when the node finishes hypersyncing in order
+	// to change the database options from Default options to Performance options.
+	DbMutex deadlock.Mutex
 }
 
 func (srv *Server) HasProcessedFirstTransactionBundle() bool {
@@ -365,7 +370,8 @@ func NewServer(
 	_trustedBlockProducerStartHeight uint64,
 	eventManager *EventManager,
 	_nodeMessageChan chan NodeMessage,
-	_forceChecksum bool) (
+	_forceChecksum bool,
+	_hypersyncMaxQueueSize uint32) (
 	_srv *Server, _err error, _shouldRestart bool) {
 
 	var err error
@@ -376,7 +382,7 @@ func NewServer(
 	archivalMode := false
 	if _hyperSync {
 		_snapshot, err, shouldRestart = NewSnapshot(_db, _dataDir, _snapshotBlockHeightPeriod,
-			false, false, _params, _disableEncoderMigrations, false)
+			false, false, _params, _disableEncoderMigrations, _hypersyncMaxQueueSize)
 		if err != nil {
 			panic(err)
 		}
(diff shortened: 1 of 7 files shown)


lib/server.go
```diff
@@ -1400,6 +1415,27 @@ func (srv *Server) _handleSnapshot(pp *Peer, msg *MsgDeSoSnapshotData) {
 	srv.GetBlocks(pp, int(headerTip.Height))
 }
 
+// updateDbOpts closes the current badger DB instance and re-opens it with the provided new options.
+func (srv *Server) updateDbOpts(opts badger.Options) {
+	// Make sure that a mempool process doesn't try to access the DB while we're closing and re-opening it.
+	srv.mempool.mtx.RLock()
+	defer srv.mempool.mtx.RUnlock()
```
diamondhands0, line 1422: You may actually want Lock() here not RLock(). RLock() is non-exclusive meaning that other threads could be reading the mempool db (but not writing to it). If you do Lock() it kicks out readers in addition to writers.

lib/server.go
```diff
@@ -1400,6 +1415,27 @@ func (srv *Server) _handleSnapshot(pp *Peer, msg *MsgDeSoSnapshotData) {
...
+	// Make sure that a server process doesn't try to access the DB while we're closing and re-opening it.
+	srv.DbMutex.Lock()
+	defer srv.DbMutex.Unlock()
```
diamondhands0, line 1425: Hmm.. I'm a bit confused. I think DbMutex is only used here, which means that if anything else has a handle on the db then it won't exclude it from reading from that handle. So does this really do anything? It seems like the only thing it prevents is calling updateDbOpts in multiple threads, but nothing else.

lib/server.go
```diff
@@ -1400,6 +1415,27 @@ func (srv *Server) _handleSnapshot(pp *Peer, msg *MsgDeSoSnapshotData) {
...
+	db, err := badger.Open(opts)
+	if err != nil {
+		// If we can't open the DB with the new options, we need to exit the process.
+		glog.Fatalf("Server._handleSnapshot: Problem switching badger db to performance opts, error: (%v)", err)
+	}
+	srv.blockchain.db = db
+	srv.snapshot.mainDb = srv.blockchain.db
+	srv.mempool.bc.db = srv.blockchain.db
+	srv.mempool.backupUniversalUtxoView.Handle = srv.blockchain.db
+	srv.mempool.universalUtxoView.Handle = srv.blockchain.db
```
diamondhands0, line 1436: This is dirty but I'm trying to think about how we would clean it up. We could have a wrapper on the DB that this thing can update... but that's dirty too. Ideally you could change the db without having to update all of these handles manually... I'm noodling.

lib/server.go
superzordon: This is utilized in backend in a few recurring jobs, such as the hot feed routine.

lib/server.go
superzordon: Agreed, not exactly a pretty solution, but the only other option I could think of was the wrapper struct, which you've mentioned.

lib/server.go
diamondhands0: Hmm I see it's used in backend.

lib/server.go
```diff
@@ -148,6 +148,11 @@ type Server struct {
 	// timer is a helper variable that allows timing events for development purposes.
 	// It can be used to find computational bottlenecks.
 	timer *Timer
+
+	// DbMutex protects the badger database from concurrent access when it's being closed & re-opened.
+	// This is necessary because the database is closed & re-opened when the node finishes hypersyncing in order
+	// to change the database options from Default options to Performance options.
+	DbMutex deadlock.Mutex
```
diamondhands0, line 155: So we're doing a couple things here:
1. We reduce the hypersync queue size. Seems like a strict win.
2. We swap the db options on badger after we finish syncing. This one creates some messiness.

Is (2) a significant improvement to memory usage? Like if we merge (1) but not (2) does it go above 32gb?

superzordon: 2 does win us a substantial amount of memory usage. The intense I/O requirements of hypersync married with our very high performance settings results in a massive amount of memory usage. 

An alternate approach here would be to fix the way we store utxo ops in the backend db. My hunch is that if we were to fix that index to only store a single utxo op per record, we would be able to run default settings in block sync. This would involve moving the block + transaction to the key of that index, rather than having a single record that contains a bundle with every op for every transaction in a block.

diamondhands0 (approved):
I didn't run it locally, but the overall change looks good with the commentary that I added.

PR #640, opened 2023-08-23
Implement PoS Consensus Interfaces and Initialization

tholonious:


tholonious:
  * **PR #640** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/640" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/640?utm_source=stack-comment).

consensus/types.go
```diff
@@ -1,53 +1,158 @@
...
+//   - Signaling its caller when it has timed out the current view
+//   - Signaling its caller when it has a QC for the current block
+//   - Signaling its caller when it has a timeout QC for the current view
+//
+// When a new block is connected to the chain, the caller is expected to update the chain tip. The module
+// resets all internal data structures and timers to handle all of the above based on the new chain tip.
+//
+// This module is very simply and only houses the logic that decide what action to perform next given the
+// current chain tip. The module does not track the history of blocks, and instead needs its caller to
+// provide update the block at the current chain tip. It expects its caller to maintain the block chain,
```
lazynina, line 109:
```suggestion
// update the block at the current chain tip. It expects its caller to maintain the block chain,
```

consensus/types.go
```diff
@@ -1,53 +1,158 @@
...
+// the index of all past blocks, to perform QC validations for incoming blocks, to handle the commit rule,
+// and only then and to pass the validated chain tip. Note: this module takes the provided chain tip as a
+// trusted input and does NOT validate any incoming blocks. This also mean the module expect its caller to
```
lazynina, line 112:
```suggestion
// trusted input and does NOT validate any incoming blocks. This also mean the module expects its caller to
```

consensus/types.go
```diff
@@ -1,53 +1,158 @@
...
 
-	internalTimersUpdated chan interface{}
-	quit                  chan interface{}
+	blockConstructionCadence time.Duration
+	timeoutBaseDuration      time.Duration
+
+	nextBlockConstructionTime time.Time
+	nextTimeoutTime           time.Time
+
+	// The latest block accepted by the caller. We only keep track of and build on top of the the chain
```
lazynina, line 124:
```suggestion
	// The latest block accepted by the caller. We only keep track of and build on top of the chain
```

consensus/types.go
```diff
@@ -1,53 +1,158 @@
...
+	// next block height.
+	validators []Validator
 
-	// votesSeen and timeoutsSeen are in-memory maps of all the votes and timeouts we've
-	// seen so far, organized by their block hash and then by the public key of the voter.
-	votesSeen    map[BlockHash]map[bls.PublicKey]*bls.Signature // TODO: this should be a vote message
-	timeoutsSeen map[BlockHash]map[bls.PublicKey]*bls.Signature // TODO: this should be a timeout message
+	// votesSeen is an in-memory map of all the votes we've seen so far, organized by the public key
+	// of the sender and the block hash of the block that was voted on.
+	votesSeen map[BlockHash]map[bls.PublicKey]VoteMessage
```
lazynina, line 137: We could make a struct to represent this key if we feel that's cleaner.
```
type VotesSeenMapKey struct {
    BlockHash
    bls.PublicKey
}
```

consensus/types.go
```diff
@@ -1,53 +1,158 @@
...
 
+	// timeoutsSeen is an in-memory map of all the timeout messages we've seen so far, organized by
+	// the public key of the sender and the view that was timed out.
+	timeoutsSeen map[uint64]map[bls.PublicKey]TimeoutMessage
```
lazynina, line 141: Same as above. It may be simpler to check the maps if we don't have it nested.

lazynina (changes requested):

consensus/types.go
```diff
@@ -1,53 +1,158 @@
...
+//   - ConstructTimeoutQC: The consensus has a timeout QC for a view and is ready to construct
+//     an empty block with the timeout QC at the next block height and the current view
 
 type ConsensusEventType byte
 
 const (
-	ConsensusEventTypeTimeout ConsensusEventType = iota
-	ConsensusEventTypeBlockProposal
-	ConsensusEventTypeVote
+	ConsensusEventTypeVote ConsensusEventType = iota
```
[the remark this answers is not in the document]
tholonious: Sure. It's clearer to have explicit values for these, although not too cirtical since they're part of a new type definition.

consensus/types.go
tholonious: Great ideal. The nest map thing was definitely gross

consensus/consensus.go
```diff
@@ -1,25 +1,97 @@
...
+		return errors.New("Consensus instance is already running")
+	}
+
+	// Validate the inputs
+	if blockConstructionCadence <= 0 {
+		return errors.New("Block construction duration must be > 0")
+	}
+	if timeoutBaseDuration <= 0 {
+		return errors.New("Timeout base duration must be > 0")
+	}
```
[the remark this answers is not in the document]
tholonious: Not going to make a material difference in performance. I slightly prefer grabbing the lock at the top of the function so it's clearer to the reader up-front that that this function needs the write lock and will release it at the end

consensus/consensus.go
```diff
@@ -1,25 +1,97 @@
...
+// the validator set. The functions expects the following for the input params:
+//   - Block construction duration must be > 0
+//   - Timeout base duration must be > 0
+//   - The input block must have a valid block hash, block height, view, and QC
+//   - The validators param must be sorted in decreasing order of stake, with a
+//     consistent tie breaking scheme. The validator set is expected to be valid for
+//     validating votes and timeouts for the next block height.
+//
+// Given the above, This function does two things:
+//   - Initializes all internal data structures so used to track incoming votes and timeout messages
```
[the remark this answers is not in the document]
lazynina: This comment still read wrong to me.

consensus/consensus.go
```diff
@@ -31,35 +80,81 @@ func (pc *FastHotStuffConsensus) HandleTimeoutMessage( /* TODO */ ) {
...
+	fc.lock.Lock()
+	if fc.status == consensusStatusRunning {
+		// Nothing to do here. The consensus instance is already running.
+		fc.lock.Unlock()
+		return
+	}
+
+	// Update the consensus status to mark it as running.
+	fc.status = consensusStatusRunning
+	fc.lock.Unlock()
```
lazynina, line 101: we don't need to hold the lock while we update the next block construction time or the next timeout time?

consensus/types.go
```diff
@@ -1,53 +1,168 @@
...
+//   - Aggregation of timeout messages for the current view
+//   - Signaling its caller when it can vote on the current chain tip
+//   - Signaling its caller when it has timed out the current view
+//   - Signaling its caller when it has a QC for the current block
+//   - Signaling its caller when it has a timeout QC for the current view
+//
+// When a new block is connected to the chain, the caller is expected to update the chain tip. The module
+// resets all internal data structures and timers to handle all of the above based on the new chain tip.
+//
+// This module is very simply and only houses the logic that decide what action to perform next given the
```
lazynina, line 107:
```suggestion
// This module is very simple and only houses the logic that decides what action to perform next given the
```

consensus/types.go
```diff
@@ -1,53 +1,168 @@
...
-	nextTimeoutTime       time.Time
+	lock sync.RWMutex
+
+	blockConstructionCadence time.Duration
+	timeoutBaseDuration      time.Duration
 
-	internalTimersUpdated chan interface{}
-	quit                  chan interface{}
+	nextBlockConstructionTime time.Time
+	nextTimeoutTime           time.Time
```
lazynina, line 122: should we name these to be `nextBlockConstructionTimestamp` and `nextTimeoutTimestamp`? `Timestamp` seems a little less ambiguous than `Time`

consensus/types.go
```diff
@@ -1,53 +1,168 @@
...
+	// the timeout signal will be triggered for this view.
+	currentView uint64
+	// The validator set sorted in decreasing order of stake amount, with a consistent tie-breaking
+	// scheme. This validator set is expected to be valid for validating votes and timeouts for the
+	// next block height.
+	validators []Validator
 
+	// votesSeen is an in-memory map of all the votes we've seen so far, organized by the public key
+	// of the sender and the block hash of the block that was voted on.
+	votesSeen map[votesSeenMapKey]VoteMessage
```
lazynina, line 137: I was wrong about these - we want the nested maps. sorry

lazynina (changes requested):

consensus/types.go
tholonious: Reverted.

consensus/types.go
tholonious: Sure

consensus/consensus.go
tholonious: Ah good catch. We still need the lock

consensus/consensus.go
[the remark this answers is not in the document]
tholonious: Fixed now.

consensus/consensus.go
```diff
@@ -1,65 +1,180 @@
...
+// Initializes the consensus instance with the latest known valid block in the blockchain, and
+// the validator set for the next block height. The functions expects the following for the input
+// params:
+//   - Block construction duration must be > 0
+//   - Timeout base duration must be > 0
+//   - The input block must have a valid block hash, block height, view, and QC
+//   - The validators param must be sorted in decreasing order of stake, with a
+//     consistent tie breaking scheme. The validator set is expected to be valid for
+//     validating votes and timeouts for the next block height.
+//
```
AeonSw4n, line 25: > //   - Block construction duration must be > 0
 //   - Timeout base duration must be > 0
 //   - The input block must have a valid block hash, block height, view, and QC
 //   - The validators param must be sorted in decreasing order of stake, with a
 //     consistent tie breaking scheme. The validator set is expected to be valid for
 //     validating votes and timeouts for the next block height.
 //


Maybe we instead make it:


```
 //   - blockConstructionCadence: block construction duration must be > 0
 //   - timeoutBaseDuration: base duration must be > 0
 //   - chainTip: The input block must have a valid block hash, block height, view, and QC
 //   - validators: The validators param must be sorted in decreasing order of stake, with a
 //     consistent tie breaking scheme. The validator set is expected to be valid for
 //     validating votes and timeouts for the next block height.
 //

```

Or something like this, so it's more clear.

consensus/consensus.go
```diff
@@ -1,65 +1,180 @@
...
-func (fc *FastHotStuffConsensus) HandleBlockProposal( /* TODO */ ) {
+func (fc *FastHotStuffConsensus) ConstructVoteQC( /* TODO */ ) {
 	// TODO
 }
 
+func (fc *FastHotStuffConsensus) ConstructTimeoutQC( /* TODO */ ) {
+	// TODO
+}
+
+// Sets the initial times for the the block construction and timeout timers and starts
```
AeonSw4n, line 99: remove one "the"

consensus/consensus.go
```diff
@@ -1,65 +1,180 @@
...
+// the event loop building off of the current chain tip.
 func (fc *FastHotStuffConsensus) Start() {
```
AeonSw4n, line 101: I think `Start` should kick off the for loop in a go routine. Maybe we should use the same pattern with `startGroup` as in `Mempool`'s Start. This way the caller doesn't have to worry about calling `Start` as `go Start()`. This complexity is pushed inside the `FastHotStuffConsensus` struct. In addition, the `startGroup` will ensure that the goroutine is spawned properly and begun running before we leave the context of `Start`. Lmk if you'd like me to elaborate on how the `startGroup` works.

Moreover, we should update the `fc.status` to `consensusStatusRunning` at the end of `Start`, once `startGroup` for the goroutined loop returns. When we do all of this, we also get a nice simplification of the locks, since we will be able to do a simple lock, defer-unlock.

consensus/consensus.go
```diff
@@ -1,65 +1,180 @@
...
 func (fc *FastHotStuffConsensus) Stop() {
-	fc.quit <- struct{}{}
+	fc.lock.Lock()
+	defer fc.lock.Unlock()
+
+	// Grabbing the lock first and checking the status ensures that we can never push a stop
+	// signal once the channel has been closed. It's OK if we push multiple stop signals while
+	// the channel is still open.
+	if fc.status == consensusStatusRunning {
+		fc.stopSignal <- struct{}{}
```
AeonSw4n, line 166: Could you add an exitGroup here? This will ensure that the Start's loop really finishes before we return from this function.

consensus/types.go
```diff
@@ -1,53 +1,161 @@
...
+// ConsensusEvent is a way for FastHotStuffConsensus to send messages back to the Server.
+// There are four types of events that can be sent:
+//   - Vote: The consensus is ready to vote on a block at a given block height and view
 //   - Timeout: The consensus has timed out on a view
-//   - BlockProposal: The consensus has a QC for a block and is ready to propose the next
-//     block
+//   - ConstructVoteQC: The consensus has a QC for a block and is ready to construct the
+//     next block at the next block height and the current view
+//   - ConstructTimeoutQC: The consensus has a timeout QC for a view and is ready to construct
+//     an empty block with the timeout QC at the next block height and the current view
```
AeonSw4n, line 19: I had a similar comment earlier, maybe we name the events by name? I.e. instead of `Vote` we say `ConsensusEventTypeVote`. This will make the comment more clear to the reader.

consensus/types.go
```diff
@@ -1,53 +1,161 @@
...
+	// This is a hybrid function that returns the QC from the block.
+	// - If the block is a normal block, this returns the validators' votes for the previous block
+	// - If the block contains a timeout QC, this returns the validators' high QC aggregated from
+	//   validators' timeout messages
+	// We are able to simplify the GetQC() to this behavior because this QC is only needed to construct
+	// a timeout QC for the next block in the event of a timeout. So, this QC will always be the latest QC
+	// at the current chain's tip that subsequent blocks will build on top of.
+	GetQC() QuorumCertificate
+}
+
```
AeonSw4n, line 87: Very cool!

AeonSw4n (changes requested):
So fun! Now I want to complete block production ASAP and hop on the consensus train lol. Left some comments, mainly about concurrency control. The overall structure looks great 😄

consensus/types.go
tholonious: Ah good catch. I forgot to update these names. Resolved now

AeonSw4n (approved):
❤️

PR #641, opened 2023-08-25
Start Unit Tests for Fast HotStuff Consensus

tholonious:

PR #645, opened 2023-08-28
Implement FastHotStuffEventLoop.AdvanceView(...)

tholonious:


tholonious:
  * **PR #640** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/640" target="_blank">Graphite</a> 
    * **PR #641** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/641" target="_blank">Graphite</a> 
      * **PR #642** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/642" target="_blank">Graphite</a> 
        * **PR #643** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/643" target="_blank">Graphite</a> 
          * **PR #645** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/645" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/645?utm_source=stack-comment).

lazynina (approved):

consensus/consensus.go
```diff
@@ -137,8 +138,50 @@ func (fc *FastHotStuffConsensus) UpdateChainTip(chainTip Block, validators []Val
...
+// to the next view number. This can only happen in the event of a local timeout. In that case, the block at
+// the chain tip is unchanged, the validator set for the next block height is unchanged, and it is only the
+// next view number that must be changed.
+//
+// Given an unchanged chain tip and validator set, this function updates the view and resets the internal
+// timers and state of the Fast HotStuff consensus that determine the next action.
+//
+// The input view number must be greater than the current view number. If the input view number is less than
+// or equal to the current chain tip's view number, then this function returns an error.
+func (fc *FastHotStuffConsensus) UpdateCurrentView(view uint64) error {
```
AeonSw4n, line 151: Hmm, is there a case where we would increase the view other than seeing a higher block with a QC (forward-skip), or by advancing a view by +1 in case of a block proposal or timeout? If these are the only cases, then I'm not sure we need this public-facing function and we can instead handle all view advances privately. In particular, any non-forward-skip view advance should only increment by +1. Maybe you want to make a function that parallels `AdvanceView` or the `Timer.onTimeout` in the [Hotstuff reference](https://github.com/deso-protocol/hotstuff_pseudocode/pull/7)? In general, I think we should make `FastHotStuffConsensus` as close to the reference as possible. Consensus code is very tricky in that even seemingly benign deviations from the reference could lead to disastrous consequences. You can trust my experience of ratholling over 1-liners for hours with others.

I also might be missing the bigger picture of how you plan to use this function. If it's only called during the timeout, then I would suggest that it only increments the view by +1 and we also make this function private.

AeonSw4n (changes requested):

AeonSw4n (approved):

PR #647, opened 2023-08-30
Mempool Status and Lock updates, and small refactors

AeonSw4n:


AeonSw4n:
* feature/proof-of-stake
    * **PR #647** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/647" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/647?utm_source=stack-comment).

lib/pos_mempool.go
```diff
@@ -40,7 +39,7 @@ type MempoolIterator interface {
 // by Fee-Time algorithm. More on the Fee-Time algorithm can be found in the documentation of TransactionRegister.
 type PosMempool struct {
 	sync.RWMutex
-	status PosMempoolStatus
+	status *atomic.Int32
```
tholonious, line 42: Why migrate this?

lib/pos_mempool.go
```diff
@@ -164,20 +165,16 @@ func (dmp *PosMempool) Stop() {
 	dmp.txnRegister.Reset()
 	dmp.ledger.Reset()
 
-	dmp.status = PosMempoolStatusNotRunning
+	dmp.status.Store(PosMempoolStatusNotRunning)
 }
 
 func (dmp *PosMempool) IsRunning() bool {
```
tholonious, line 171: Does the `PosMempool.IsRunning()` function have any expected use-cases outside of this class? Similar to the Fast-HotStuff consensus module, it seems like simply exposing the `Start()` and `Stop()` results in the simplest public interface

lib/pos_mempool_persister.go
```diff
@@ -121,32 +127,54 @@ func (mp *MempoolPersister) run() {
 // Stop is used to stop the persister thread and reset the persister state. It will wait for the persister thread to
 // flush the outstanding updateBatch to the database before returning.
 func (mp *MempoolPersister) Stop() error {
-	if mp.status == MempoolPersisterStatusNotRunning {
+	// stopLock ensures that only a single MempolEventExit is enqueued in case someone called Stop concurrently.
+	mp.stopLock.Lock()
```
tholonious, line 131: Why not just hold `mp.Lock()` here?

tholonious (commented):
Rather than introducing additional mutexes here, I'd suggest the following:
- Remove the `PosMempool.IsRunning()` and have the PosMempool functions hold the instance's lock to manually check the status of the mempool
- Reuse the existing top-mutex in the persister rather than introducing a separate stopLock. Refer to https://github.com/deso-protocol/core/pull/640 on how it allows multiple  Stop() operations to be called concurrently and safely.

The fewer mutxes we have in the code, the easier it will be to understand.

lib/pos_mempool_persister.go
AeonSw4n: Ah this one is tricky. I don't like this second lock either. But I couldn't find an alternative that is thread safe on concurrent `Stop()` calls. 

The issue is that `run()` holds a lock when it is processing events. So we could deadlock as we're waiting for the stopGroup to finish.

lib/pos_mempool.go
AeonSw4n: Nah doesn't really have any expected public use-cases. Maybe it should be private. It's cool that you can use it instead of manual status checks and for tests.

tholonious: I'd suggest removing it then. It's valid for the mempool struct's unit tests to have full knowledge of the mempool's inner workings, i.e. to be able to inspect statuses manually.

lib/pos_mempool.go
tholonious: IMO, adding this thread-safe status to enable the `IsRunning()` function results in more effort to understand the concurrency of the struct, without a clear payoff. 

My suggestion would be to defer introducing it until we know we want the `IsRunning()` function at some point in the future.

lib/pos_mempool_persister.go
AeonSw4n: Note: drop StopLock and be okay with the concurrence Stop call undefined behavior.

lib/pos_mempool.go
AeonSw4n: Note: just use int.

lib/pos_mempool.go
```diff
@@ -174,11 +173,11 @@ func (dmp *PosMempool) IsRunning() bool {
...
-		return errors.Wrapf(MempoolErrorNotRunning, "PosMempool.AddTransaction: ")
-	}
+	// First, validate that the transaction is properly formatted according to BalanceModel. We acquire a read lock on
+	// the mempool. This allows multiple goroutines to safely perform transaction validation concurrently. In particular,
+	// transaction signature verification can be parallelized.
+	dmp.RLock()
 
-	// First, validate that the transaction is properly formatted according to BalanceModel.
 	if err := ValidateDeSoTxnSanityBalanceModel(txn, dmp.latestBlockHeight, dmp.params, dmp.globalParams); err != nil {
 		return errors.Wrapf(err, "PosMempool.AddTransaction: Problem validating transaction sanity")
```
tholonious, line 182: What if we return early here? Then `dmp.RUnlock()` below won't get called so the read lock will be held indefinitely

AeonSw4n: Damn, good catch! Fixed in #660

PR #649, opened 2023-08-31
feature/pos_fast_hotstuff_consensus

tholonious:


tholonious:
* feature/proof-of-stake
  * **PR #649** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/649" target="_blank">Graphite</a>  👈
    * **PR #651** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/651" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/649?utm_source=stack-comment).

consensus/event_loop.go
```diff
@@ -0,0 +1,854 @@
...
+		View:           fc.tip.block.GetView(),
+	}
+
+	// Schedule the next crank timer and timeout scheduled tasks
+	fc.resetScheduledTasks()
+
+	return nil
+}
+
+// setSafeBlocks is a helper function that validates the provided blocks, validator sets, and stores them.
```
diamondhands0, line 168: Update comment.

consensus/event_loop.go
```diff
@@ -0,0 +1,854 @@
...
+func (fc *FastHotStuffEventLoop) storeBlocks(tip BlockWithValidators, safeBlocks []BlockWithValidators) error {
+	// Do a basic integrity check on the tip block and validator set
+	if !isProperlyFormedBlock(tip.Block) || !isProperlyFormedValidatorSet(tip.Validators) {
+		return errors.New("Invalid tip block or validator set")
+	}
+
+	// Do a basic integrity check on the blocks and validator sets
+	hasMalformedInput := collections.Any(safeBlocks, func(block BlockWithValidators) bool {
+		return !isProperlyFormedBlock(block.Block) || !isProperlyFormedValidatorSet(block.Validators)
+	})
```
diamondhands0, line 179: Nit: Prefer pulling embedded functions out of function calls like this as follows:

```
	checkMalformedInput := func(block BlockWithValidators) bool {
		return !isProperlyFormedBlock(block.Block) || !isProperlyFormedValidatorSet(block.Validators)
	}

	// Do a basic integrity check on the blocks and validator sets
	hasMalformedInput := collections.Any(safeBlocks, checkMalformedInput)
```

consensus/event_loop.go
```diff
@@ -0,0 +1,854 @@
...
+	}
+
+	// Store the tip block and validator set
+	fc.tip = blockWithValidatorLookup{
+		block:        tip.Block,
+		validatorSet: tip.Validators,
+		validatorLookup: collections.ToMap(tip.Validators, func(validator Validator) string {
+			return validator.GetPublicKey().ToString()
+		}),
+	}
```
diamondhands0, line 193: Nit: Here again I think a little easier to read if you do the following:

```
	validatorToPubkeyString := func(validator Validator) string {
		return validator.GetPublicKey().ToString()
	}

	// Store the tip block and validator set
	fc.tip = blockWithValidatorLookup{
		block:        tip.Block,
		validatorSet: tip.Validators,
		validatorLookup: collections.ToMap(tip.Validators, validatorToPubkeyString),
	}
```

consensus/types_internal.go
```diff
@@ -0,0 +1,197 @@
...
+	return tm.signature
+}
+
+//////////////////////////////////////////////////////////
+// Internal blockWithValidatorLookup type
+//////////////////////////////////////////////////////////
+
+type blockWithValidatorLookup struct {
+	block           Block
+	validatorSet    []Validator          // Ordered slice of validators
```
diamondhands0, line 195: Nit: I would maybe call this validatorList rather than validatorSet. I know we normally talk about the validator set, but it confused me a bit because I thought this was a set struct.

consensus/event_loop.go
diamondhands0: Just realized you actually use that function twice. Even more reason to pull it out.

consensus/event_loop.go
```diff
@@ -0,0 +1,854 @@
...
+
+	// We track the current view here so we know which view to time out on later on.
+	fc.currentView = tip.Block.GetView() + 1
+
+	// Reset QC construction status for the current view
+	fc.hasCrankTimerRunForCurrentView = false
+	fc.hasConstructedQCInCurrentView = false
+
+	// Reset all internal data structures for votes and timeouts
+	fc.votesSeen = make(map[[32]byte]map[string]VoteMessage)
```
diamondhands0, line 72: Nit: Whenever I see `[32]byte` I suspect maybe we should define a type for it. In this case you maybe want something like `VoteHash` or something.

consensus/event_loop.go
```diff
@@ -0,0 +1,854 @@
...
+	fc.timeoutsSeen = make(map[uint64]map[string]TimeoutMessage)
```
diamondhands0, line 73: Nit that also applies to the above. I like to add ByWhatever to map names to indicate what the key is without making the user hover over it for the comment. For example: `votesSeenByVoteHash` or `timeoutsSeenByView`. Not needed, but suggesting for consideration.

consensus/types.go
```diff
@@ -1,53 +1,198 @@
...
+	Validators []Validator
 }
+
+// We want a large buffer for the signal channels to ensure threads don't block when trying to push new
+// signals.
+//
+// TODO: is a size of 100 enough? If we want to bullet-proof this, we could back it by a slice as a
+// secondary buffer. That seems unnecessary since every channel will only have signals pushed by a single
+// producer thread.
+const signalChannelBufferSize = 100
```
diamondhands0, line 119: 100 is pretty small. I think the point of this is just to prevent blowing up memory. If each event is a kilobyte then you could easily do 10,000, which would use 10MB of memory at maximum.

consensus/event_loop.go
```diff
@@ -0,0 +1,854 @@
...
+	}
+
+	// Cancel the timeout task. The server will reschedule it when it advances the view.
+	fc.nextTimeoutTask.Cancel()
+}
+
+// Evict all locally stored votes and timeout messages with stale views. We can safely use the current
+// view to determine what is stale. The consensus mechanism will never construct a block with a view
+// that's lower than its current view. Consider the following:
+// - In the event the event update the chain tip, we will vote for that block and the view it was proposed in
```
diamondhands0, line 741: Nit: typo here

consensus/event_loop.go
```diff
@@ -0,0 +1,854 @@
...
+
+	// Check if we have timed out at for the last n views. If so, we apply exponential
+	// back-off to the timeout base duration.
+	if fc.tip.block.GetView() < fc.currentView-1 {
+		// Note, there is no risk of underflow here because the following is guaranteed:
+		// currentView > tip.block.GetView() + 1.
+		numTimeouts := fc.currentView - fc.tip.block.GetView() - 1
+
+		// Compute the exponential back-off: nextTimeoutDuration * 2^numTimeouts
+		timeoutDuration = fc.timeoutBaseDuration << numTimeouts
```
diamondhands0, line 449: Nit: I generally prefer we avoid bitwise math where possible. It's inflexible, and often the source of errors because people miscalculated.

In the case of a duration, I'm not even 100% sure this is doing what we want since it's an int64. If there's a nice clean function you can use like Pow(2, exponent) or something, prefer that.

consensus/event_loop.go
```diff
@@ -0,0 +1,854 @@
...
+
+	// Check if the event loop is running. If it's not running, then there's nothing
+	// to do here.
+	if fc.status != eventLoopStatusRunning {
+		return
+	}
+
+	// Check for race conditions where the view advanced at the exact moment this task began
+	// or we have already signaled for this view. If so, then there's nothing to do here.
+	if fc.currentView != blockConstructionView {
```
diamondhands0, line 474: Smart. I thought I'd catch a bug related to this race condition but looks like you're on top of it.

consensus/event_loop.go
```diff
@@ -0,0 +1,854 @@
...
+func (fc *FastHotStuffEventLoop) tryConstructVoteQCInCurrentView() *FastHotStuffEvent {
+	// If currentView != tipBlock.View + 1, then we have timed out at some point, and can no longer
+	// construct a block with a QC of votes for the tip block.
+	tipBlock := fc.tip.block
+	if fc.currentView != tipBlock.GetView()+1 {
+		return nil
+	}
+
+	// Fetch the validator set at the tip.
+	validatorSet := fc.tip.validatorSet
```
diamondhands0, line 516: Nit: Note this variable name if you decide to change validatorSet -> validatorList

consensus/event_loop.go
```diff
@@ -0,0 +1,854 @@
...
+	}
+
+	// If we don't have a super-majority vote for the chain tip, then we can't build a QC.
+	if !isSuperMajorityStake(totalVotingStake, totalStake) {
+		return nil
+	}
+
+	// If we reach this point, then we have enough signatures to build a QC for the tip block. Try to
+	// aggregate the signatures. This should never fail.
+	aggregateSignature, err := bls.AggregateSignatures(signatures)
```
diamondhands0, line 561: Hmm... I'm a bit worried that we'll hit an error in here one day and have a lot of trouble tracking it down. If this should really never fail, consider either adding a `glog.Errorf` call with the string `This should never happen` somewhere in the message. Or else change the signature to return error and log it in the caller.

consensus/event_loop.go
```diff
@@ -0,0 +1,854 @@
...
+	if err != nil {
+		return nil
+	}
+
+	// Happy path. Construct the QC and return as an event to signal to the server.
+	return &FastHotStuffEvent{
```
diamondhands0, line 567: Is there any way the view or block tip could advance between the start of this function and this section? I think no because 1) The caller of this function grabs a lock and 2) AdvanceView* grabs a lock always. But just double-checking.

consensus/event_loop.go
```diff
@@ -0,0 +1,854 @@
...
+	// Check if the event loop is running. If it's not running, then there's nothing
+	// to do here.
+	if fc.status != eventLoopStatusRunning {
+		return
+	}
+
+	// Check if the timed out view is stale. If it's stale, then there's nothing to do here.
+	// The view may be stale in the race condition where the view advanced at the exact moment
+	// this task began to execute and wait for the event loop's lock at the top of this function.
+	if fc.currentView != timedOutView {
```
diamondhands0, line 721: Good work. This is a tricky one.

diamondhands0, line 721: Good work. This is a tricky one, and your solution is quite elegant.

consensus/integration_test.go
```diff
@@ -0,0 +1,247 @@
...
+	// as long as 3 out of 4 nodes are online.
+	node1 := newValidatorNode(uint256.NewInt().SetUint64(50), true)  // block proposer
+	node2 := newValidatorNode(uint256.NewInt().SetUint64(50), false) // validator
+	node3 := newValidatorNode(uint256.NewInt().SetUint64(50), false) // validator
+	node4 := newValidatorNode(uint256.NewInt().SetUint64(50), false) // validator
+
+	allNodes := []*validatorNode{node1, node2, node3, node4}
+
+	// Create the genesis block signed by all four nodes.
+	genesisBlock := createDummyBlockWithVoteQC(createDummyBlockHash(), 2, 2, allNodes)
```
diamondhands0, line 27: Maybe dumb question: Why does the genesis block start with view=2 block height=2 rather than (0,0)? I assume it's because some checks fail if you do the latter? Leave a comment to explain either way.

consensus/event_loop.go
```diff
@@ -0,0 +1,854 @@
...
+	totalVotingStake := uint256.NewInt()
+
+	// Track the signatures and signers list for the chain tip
+	signersList := bitset.NewBitset()
+	signatures := []*bls.Signature{}
+
+	// Iterate through the entire validator set and check if each one has voted for the tip block. Track
+	// all voters and their stakes.
+	for ii, validator := range validatorSet {
+		totalStake = uint256.NewInt().Add(totalStake, validator.GetStakeAmount())
```
diamondhands0, line 535: Is this a bug? It looks like you add a validator's stake to the total even if they haven't voted for this block. I think you want to move this line to after you've checked the vote?

consensus/integration_test.go
```diff
@@ -0,0 +1,247 @@
...
+
+	// Validate the resulting chain of blocks stored by all nodes
+	validateAndPrintBlockChain(t, node1, node1.safeBlocks)
+	validateAndPrintBlockChain(t, node2, node2.safeBlocks)
+	validateAndPrintBlockChain(t, node3, node3.safeBlocks)
+	validateAndPrintBlockChain(t, node4, node4.safeBlocks)
+
+	// Verify that the network has produced more blocks after the node 4 stopped
+	require.Greater(t, node3.eventLoop.tip.block.GetView(), node4.eventLoop.tip.block.GetView())
+	require.Greater(t, node3.eventLoop.tip.block.GetHeight(), node4.eventLoop.tip.block.GetHeight())
```
diamondhands0, line 89: This test looks great! I like how you create mock nodes and connect them to each other.

consensus/event_loop_test.go
```diff
@@ -0,0 +1,899 @@
...
+
+	// Stop the event loop
+	fc.Stop()
+
+	// Confirm the event loop status has reverted to initialized
+	require.Equal(t, eventLoopStatusInitialized, fc.status)
+
+	// Confirm that calling fc.Stop() again doesn't panic
+	require.NotPanics(t, fc.Stop)
+}
```
diamondhands0, line 899: I skimmed these unit tests and I really like this. You start by testing that each function behaves correctly under all the edge-cases, then you can build up to the integration tests which I see in the next file.

consensus/integration_test.go
```diff
@@ -0,0 +1,247 @@
...
+
+	// Validate and print resulting chain of blocks stored by all nodes
+	validateAndPrintBlockChain(t, node1, node1.safeBlocks)
+	validateAndPrintBlockChain(t, node2, node2.safeBlocks)
+	validateAndPrintBlockChain(t, node3, node3.safeBlocks)
+
+	// Verify that the network has produced at least one block since it recovered
+	require.Greater(t, node3.eventLoop.tip.block.GetView(), node3TipDuringNetworkHalt.GetView())
+	require.Greater(t, node3.eventLoop.tip.block.GetHeight(), node3TipDuringNetworkHalt.GetHeight())
+}
```
diamondhands0, line 247: This whole framework seems very solid for integration testing. When you combine it with the unit testing you're doing in event_loop_test.go, it feels like we have everything we need to exercise and bullet-proof everything here.

Is there anything you want to add in terms of testing infrastructure beyond this in the short-term?

One thing I thought of is maybe you'd want to be able to "fast-forward" time so that you don't have to wait for the wall clock to hit. But this comes with its own challenges, and it's not obviously worth it because it isn't as "real" as what you're doing here. Still maybe worth looking into whether this is a "thing" that people do in Go integration tests ever.

lib/blockchain.go
```diff
@@ -75,6 +75,13 @@ const (
 	StatusBitcoinHeaderValidateFailed // Deprecated
 )
 
+type CommittedBlockStatus uint8
+
+const (
+	COMMITTED   CommittedBlockStatus = 0
+	UNCOMMITTED CommittedBlockStatus = 1
+)
```
diamondhands0, line 83: As I read through this, I had the following thoughts:
* Why not overload BlockStatus to add a StatusBlockCommitted field?
* If we did this, then we wouldn't need the uncommittedBlockMap. Instead, blocks that are in the blockIndex are implicitly uncommitted if they have StatusBlockValidated but not StatusBlockCommitted.
* If we did the above, the only thing you'd need to get the "block tip" is the uncommitted tip. To get this, I was thinking we could just change the "best chain tip" to be the uncommitted tip. Then if you want the committed tip, you use the blockIndex to wlak back from the "best chain tip" to the first block with StatusBlockCommitted.

Above is the most lightweight way to store blocks in PoS but obviously haven't thought through all the implications.

diamondhands0: Also I don't think you'd have to add anything to the BlockNode structure if you did the above ^

lib/pos_blockchain.go
```diff
@@ -0,0 +1,451 @@
...
+//  1. Determine if we're missing a parent block of this block and any of its parents from the block index.
+//     If so, return the hash of the missing block and add this block to the orphans list.
+//  2. Validate on an incoming block, its header, its block height, the leader, and its QCs (vote or timeout)
+//  3. Store the block in the block index and uncommitted blocks map.
+//  4. Resolves forks within the last two blocks
+//  5. Connect the block to the blockchain's tip
+//  6. Run the commit rule - If applicable, flushes the incoming block's grandparent to the DB
+//  7. Prune in-memory struct holding uncommitted block.
+//  8. Update the currentView to this new block's view + 1
+func (bc *Blockchain) processBlockPoS(desoBlock *MsgDeSoBlock, currentView uint64, verifySignatures bool) (_success bool, _isOrphan bool, _missingBlockHashes []*BlockHash, _err error) {
```
diamondhands0, line 20: @lazynina this looks incomplete. Lmk if you want me to review it now or wait?

lib/server.go
```diff
@@ -2253,14 +2259,16 @@ func (srv *Server) _handleFastHostStuffTimeout(event *consensus.ConsensusEvent)
...
 		srv._handleFastHostStuffVote(event)
-	case consensus.ConsensusEventTypeTimeout:
+	case consensus.FastHotStuffEventTypeTimeout:
 		srv._handleFastHostStuffTimeout(event)
+	case consensus.FastHotStuffEventTypeConstructVoteQC:
+		srv._handleFastHostStuffBlockProposal(event)
+	case consensus.FastHotStuffEventTypeConstructTimeoutQC:
+		srv._handleFastHostStuffEmptyTimeoutBlockProposal(event)
 	}
 }
```
diamondhands0, line 2273: I like this structure a lot. You've managed to isolate most of the fast-hotstuff complexity in a separate lib, where it can be rigorously tested, and this interface for it in server.go is super clean. Basically server.go is the "glue" between the consensus logic and the actual real connections to other nodes. Love it.

consensus/integration_test.go
tholonious: I want to test different stake distributions across nodes and downtime durations for nodes. The current framework basically allows us to test any permutation of nodes, so I want to stress test as much as possible before we make it out to devnet. It would be somewhere on the order of 3-5 more integration tests similar to what already exists.

The "fast-forward" time comment is interesting. It would allow us to test scenarios with extremely consecutive timeouts in the network. I'll look into it.

consensus/event_loop.go
tholonious: Not a bug.

We dynamically sum the entire validator set's stake here. The number doesn't really change so I guess we could compute it once and cache it, but since we iterate over the entire validator set here, we may as well sum it here.

Remember, from the perspective of the FastHotStuffEventLoop, every block height can its own unique validator set. So there's no single "GlobalStake" value. In practice, the validator set will only change for block heights on epoch transitions, but the FastHotStuffEventLoop shouldn't know when that is.

consensus/event_loop.go
tholonious: Yeah, this should never error. I'll add a glog.Errorf

consensus/event_loop.go
tholonious: Your reasoning is correct. Neither the view nor the block tip can shift underneath us here. This function's usage is thread safe.

lib/pos_blockchain.go
tholonious: Marking this as resolved since it's implemented in https://github.com/deso-protocol/core/pull/742

lib/blockchain.go
tholonious: This is a work in progress in https://github.com/deso-protocol/core/pull/742. Marking as resolved since it's implemented there

consensus/integration_test.go
tholonious: The event loop rejects blocks with view=0 because it would mean that the block holds a QC with a view < 1, which should never happen. Rather than have an edge case to handle this, but just side-step it entirely by guaranteeing that blocks in production all have view > 0. When the PoW -> PoS cutover happens, the first transition block will have its view = blockHeight

collections/slice.go
```diff
@@ -1,9 +1,33 @@
...
 	var result []TOutput
 	for _, val := range slice {
 		result = append(result, transformFn(val))
 	}
 	return result
 }
+
+func ToMap[TKey comparable, TValue any](slice []TValue, keyFn func(TValue) TKey) map[TKey]TValue {
+	result := make(map[TKey]TValue)
+	for _, val := range slice {
```
diamondhands0, line 29: Do you need to copy the val to avoid the slice iterator bug? I think you're OK because you're not taking a pointer to the value, but wasn't 100% sure it was safe with generics.

consensus/types.go
```diff
@@ -1,53 +1,221 @@
...
+	TipBlockHash   BlockHash
+	TipBlockHeight uint64
+	View           uint64
+	QC             QuorumCertificate
+	AggregateQC    AggregateQuorumCertificate
+}
+
+// Create an alias type of the 32 bit block hash so that the raw [32]byte type isn't
+// ambiguously repeated in the code base
+type BlockHashValue = [32]byte
```
diamondhands0, line 41: We already have BlockHash aliased to this. Any reason you didn't want to use that here?

diamondhands0: Nevermind I think based on the comment below you want to decouple this type from lib. I just realized that none of this is in lib now.

consensus/types.go
```diff
@@ -1,53 +1,221 @@
...
+	GetHeight() uint64
+	GetView() uint64
+	// This is a hybrid function that returns the QC from the block.
+	// - If the block is a normal block, this returns the QC from validators' votes for the previous block
+	// - If the block contains a timeout QC, this returns the validator high QC aggregated from
+	//   validators' timeout messages
+	// We are able to simplify the GetQC() to this behavior because this QC is only needed to construct
+	// a timeout QC for the next block in the event of a timeout. So, this QC will always be the latest QC
+	// at the current chain's tip that subsequent blocks will build on top of.
+	GetQC() QuorumCertificate
```
diamondhands0, line 125: Makes sense. I guess this lets you always call GetQC() on the tip to get the QC to embed into the next block.

consensus/types.go
```diff
@@ -1,53 +1,221 @@
...
+}
 
-	ConsensusEvents chan *ConsensusEvent
+type BlockWithValidatorList struct {
+	Block Block
+	// The ordered validator list for the next block height after the block. This validator list can be used to
+	// validate votes and timeouts used to build a QC that extends from the block. The validator list must be
+	// sorted in descending order of stake amount with a consistent tie breaking scheme.
+	ValidatorList []Validator
```
diamondhands0, line 133: Almost feel this should be called `NextValidators` or something like that but up to you. The comment makes it clear either way.

consensus/scheduled_task.go
```diff
@@ -0,0 +1,82 @@
...
+// provided param to ensure its attached task is no longer stale, similar to the above.
+func (t *ScheduledTask[TaskParam]) Cancel() {
+	t.lock.Lock()
+	defer t.lock.Unlock()
+
+	if t.timer != nil {
+		t.timer.Stop()
+	}
+
+	t.timer = nil
```
diamondhands0, line 67: Hmm... I think you also want to update the duration here? Otherwise GetDuration() will return a weird value for a canceled task.

consensus/event_loop.go
```diff
@@ -0,0 +1,908 @@
...
+	fc.lock.Lock()
+	defer fc.lock.Unlock()
+
+	// Ensure the event loop is running
+	if fc.status != eventLoopStatusRunning {
+		return errors.New("FastHotStuffEventLoop.ProcessTipBlock: Event loop is not running")
+	}
+
+	// Validate the safe blocks and validator lists, and store them
+	if err := fc.storeBlocks(tip, safeBlocks); err != nil {
```
diamondhands0, line 153: Consider adding an error check before this point to verify that fc.tip.GetView() <= tip.GetView(). As it stands right now, I think you could call ProcessTipBlock with a stale tip and it would accept it, which I don't think is desired behavior.

diamondhands0: Then again maybe you want this function to "reset" everything, irrespective of what the prior state was. If that's the case then it's fine as is.

consensus/event_loop.go
```diff
@@ -0,0 +1,908 @@
...
+			}
+		}
+	}
+
+	// Evict stale timeout messages
+	for view := range fc.timeoutsSeenByView {
+		if isStaleView(fc.currentView, view) {
+			delete(fc.timeoutsSeenByView, view)
+		}
+	}
```
diamondhands0, line 830: This function looks right, but my logic for it was a bit different than the comment. It goes as follows:
* fc.currentView is the view that the next block is going to be proposed on
* That block's QC or AggQC consists of votes/timeouts on the previous block, which has view fc.currentView-1
* That means you can dispose any votes or timeouts with vote.GetView() <> fc.currentView-1

Is the above logic oversimplifying it? If not, I would update the comment.

consensus/event_loop.go
```diff
@@ -0,0 +1,908 @@
...
+		if isEqualBlockHashes(block.block.GetBlockHash(), blockHash) {
+			return true, block.block, block.validatorList, block.validatorLookup
+		}
+	}
+
+	return false, nil, nil, nil
+}
+
+func isStaleView(currentView uint64, testView uint64) bool {
+	return currentView > testView+1
```
diamondhands0, line 907: To me the following is more intuitive. See my comment on evictStale...():
```
testView < currentView - 1
```

consensus/integration_test_types.go
```diff
@@ -0,0 +1,443 @@
...
+		go validator.ProcessVote(vote)
+	}
+}
+
+func (node *validatorNode) handleTimeoutEvent(event *FastHotStuffEvent) {
+	node.lock.Lock()
+	defer node.lock.Unlock()
+
+	// Skip if the timed out view from the event is stale
+	if node.eventLoop.currentView != event.View {
```
diamondhands0, line 361: I almost thought I'd caught a race condition but I think this if statement here covers you...

consensus/event_loop.go
diamondhands0: Sorry that was a dumb question. I think I was confusing totalStake for totalVotingStake...

consensus/integration_test_types.go
```diff
@@ -0,0 +1,443 @@
...
+	block := &block{
+		view:      event.View,
+		blockHash: createDummyBlockHash(),
+		height:    event.TipBlockHeight + 1,
+		qc:        event.QC,
+	}
+
+	// Broadcast the block to all validators.
+	for _, validator := range node.validatorNodes {
+		go validator.ProcessBlock(block)
```
diamondhands0, line 412: I know this is just a test, but if you call ProcessBlcok in a goroutine for the proposer node, then I think you can have a race condition where the following events occur:
1) onTimeoutScheduledTaskExecuted triggers, sent to server.go but not processed yet
2) ProcessValidatorVote is called, which constructs QC and sends to server.go
3) server.go processes QC via handleVoteQCConstructionEvent
4) server.go processes timeout after QC with handleTimeoutEvent

diamondhands0: To avoid the above race condition, you need to make sure that handleVoteQCConstructionEvent increments the view before it returns. Otherwise, a subsequent call to something like handleTimeoutEvent will erroneously conclude that the event is fresh when it's actually stale.

consensus/integration_test_types.go
```diff
@@ -0,0 +1,443 @@
...
+	block := &block{
+		view:        event.View,
+		blockHash:   createDummyBlockHash(),
+		height:      event.TipBlockHeight + 1,
+		aggregateQC: event.AggregateQC,
+	}
+
+	// Broadcast the block to all validators.
+	for _, validator := range node.validatorNodes {
+		go validator.ProcessBlock(block)
```
diamondhands0, line 436: Same feedback as for handleVoteQCConstructionEvent. If you're the node proposing the block, you need to process it before this function returns so that you can guarantee that the view has incremented. Not doing this will cause staleness checks to fail erroneously.

consensus/integration_test_types.go
```diff
@@ -0,0 +1,443 @@
...
+
+	vote := &voteMessage{
+		view:      event.View,
+		blockHash: event.TipBlockHash,
+		publicKey: node.privateKey.PublicKey(),
+		signature: signature,
+	}
+
+	for _, validator := range node.validatorNodes {
+		go validator.ProcessVote(vote)
```
diamondhands0, line 352: I think a goroutine might be OK here because there's no way that the view could increment as a result of processing a vote. But, to be safe, I would call ProcessVote synchronously for the current node. And I would do the same for all the other functions.

consensus/event_loop.go
```diff
@@ -0,0 +1,908 @@
...
+	if vote.GetBlockHash() != fc.tip.block.GetBlockHash() {
+		return nil
+	}
+
+	// Check if we have a super-majority vote for the chain tip.
+	if voteQCEvent := fc.tryConstructVoteQCInCurrentView(); voteQCEvent != nil {
+		// Signal the server that we can construct a QC for the chain tip, and mark that we have
+		// constructed a QC for the current view.
+		fc.hasConstructedQCInCurrentView = true
+		fc.Events <- voteQCEvent
```
diamondhands0, line 340: So one thing to note about this queuing here is that IF we ever hit the limit of 10,000 events on the channel, then I think we're guaranteed to deadlock. This is because one function will be adding to this queue while holding the lock, but no other function can run until the lock is released.

I think to avoid this you have to release the lock before you enqueue. But this is also kindof messy and annoying. Also I think that it won't even work because server.go is synchronous.

The simpler solution is to just call this in a Goroutine. That allows the function to return and release the lock so that other things can dequeue. I think this will cause the "limit" you placed on the channel to be effectively useless, though, since you'll queue up potentially unlimited goroutines. It's probably still better than deadlocking though.

consensus/event_loop.go
```diff
@@ -0,0 +1,908 @@
...
+			timeout.GetPublicKey().ToString(),
+			timeout.GetView(),
+		)
+	}
+
+	// Compute the value sha3-256(timeout.View, timeout.HighQC.View)
+	timeoutSignaturePayload := GetTimeoutSignaturePayload(timeout.GetView(), timeout.GetHighQC().GetView())
+
+	// Verify the vote signature
+	if !isValidSignatureSinglePublicKey(timeout.GetPublicKey(), timeout.GetSignature(), timeoutSignaturePayload[:]) {
```
diamondhands0, line 402: I could have just missed it, but I didn't see you validate the HighQC on the timeout message. I think you need to do this or else someone could attach a fake HighQC with a fake view number.

consensus/event_loop.go
```diff
@@ -0,0 +1,908 @@
...
+
+	// Iterate through all timeouts for the previous view to find the highQC
+	for _, timeout := range timeoutsByValidator {
+		// Check if the high QC from the timeout messages is for a block in our safeBlocks slice. If not,
+		// then we have no knowledge of the block, or the block is not safe to extend from. This should never
+		// happen, but may be possible in the event we receive a timeout message at the same time the block
+		// becomes unsafe to extend from (ex: it's part of an stale reorg). We check for the edge case here to
+		// be 100% safe.
+		isSafeBlock, _, _, validatorLookup := fc.fetchSafeBlockInfo(timeout.GetHighQC().GetBlockHash())
+		if !isSafeBlock {
```
diamondhands0, line 663: It feels like you might benefit from one more message to tell server.go to fetch missing blocks when you encounter them. Something like "MissingBlockEvent" or something, which triggers server.go to request the block from a peer. But I wouldn't add this until later. Maybe just keep in mind that this could be useful for cases where a node missed some data.

consensus/event_loop.go
```diff
@@ -0,0 +1,908 @@
...
+
+	// Check if we have timed out at for the last n views. If so, we apply exponential
+	// back-off to the timeout base duration.
+	if fc.tip.block.GetView() < fc.currentView-1 {
+		// Note, there is no risk of underflow here because the following is guaranteed:
+		// currentView > tip.block.GetView() + 1.
+		numTimeouts := fc.currentView - fc.tip.block.GetView() - 1
+
+		// Compute the exponential back-off: nextTimeoutDuration * 2^numTimeouts
+		timeoutDuration = fc.timeoutBaseDuration * time.Duration(powerOfTwo(numTimeouts))
```
diamondhands0, line 501: This works, but I wonder if we want to impose a "max" timeout duration. I don't think we do, just mentioning.

consensus/event_loop.go
```diff
@@ -0,0 +1,908 @@
...
+		return nil
+	}
+
+	// Check if the vote is for the chain tip. If not, then there's nothing more to do.
+	if vote.GetBlockHash() != fc.tip.block.GetBlockHash() {
+		return nil
+	}
+
+	// Check if we have a super-majority vote for the chain tip.
+	if voteQCEvent := fc.tryConstructVoteQCInCurrentView(); voteQCEvent != nil {
```
diamondhands0, line 336: So I think if you have this here then the crank timer is kindof irrelevant right? Because if you're creating the QC as soon as the votes come in, there's no need to wait for the crank timer.

So here we have a choice:
1. We can either leave this here, in which case the network will run as fast as it possibly can in the steady-state, with votes being shaped into QCs as soon as enough of them have arrived.
2. We can remove this, in which case the crank timer is the one responsible for assembling votes into QCs.

I think in past discussions we have leaned toward (2) because it's more fair to nodes with slow connections to the current leader. But let me know if you'd like to revisit this.

diamondhands0: Nevermind I see you check the crank timer above.

consensus/types_internal.go
```diff
@@ -0,0 +1,197 @@
...
+	"github.com/deso-protocol/core/bls"
+	"github.com/deso-protocol/core/collections/bitset"
+	"github.com/holiman/uint256"
+)
+
+//////////////////////////////////////////////////////////
+// BlockHash interface implementation for testing
+//////////////////////////////////////////////////////////
+
+type blockHash struct {
```
diamondhands0, line 13: After reviewing, I still wasn't sure why we needed these "internal" type defs. Can you just add a comment at the top explaining why they're useful?

consensus/types.go
tholonious: Correct. We want to decouple the this consensus package from lib. But we define the BlockHash as an interface in the consensus package so that the two can interop

consensus/types_internal.go
tholonious: They're just useful for unit testing and integration testing within the consensus module. The module doesn't pull in any types from lib, so it has to define its own barebones structs mostly for tests. I'll explain on the struct definition's comments to say this.

consensus/event_loop.go
tholonious: Correct. We want this function to reset everything regardless of what the previous state was. The consensus rules to decide on the next tip are all implemented in ProcessBlockPoS. We avoid duplicating them here and instead trust the new tip passed by the caller. If ProcessBlockPoS decides a new block is the new tip, we accept it at face value here. If for whatever reason we decide to update the view in ProcessBlockPoS, this consensus module just defers to that reasoning. It's mostly for separation of concerns.

In practice though, the view will never regress.

consensus/event_loop.go
tholonious: The server.go code will do this. When it receives a timeout or vote message that references a block that it hasn't seen before, it requests the block from the peer in-line. This keeps the flow for requesting blocks simple, and couples it with the vote/timeout message receipt

tholonious:
* feature/proof-of-stake
  * **PR #649** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/649?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
    * **PR #832** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/832?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/649?utm_source=stack-comment).

consensus/event_loop.go
tholonious: Seems fine. We can impose an artificial limit of 2^16 (18 hours) so that if the math ever goes haywire, the duration doesn't continue to grow and eventually overflow.

consensus/event_loop.go
tholonious: Correct. That's what the code already does. Here's the description from the comment block above:
```
// The eviction works as follows:
// - Votes: if the next block were to be a regular block with a QC aggregated from votes, then the it must
// satisfy nextBlock.GetView() = tip.block.GetView() + 1, which means that currentView = tip.block.GetView() + 1.
// We can safely evict all votes where vote.GetView() < currentView - 1.
// - Timeouts: if the next block were be an empty block with a timeout QC aggregated from timeout messages,
// then it must satisfy nextBlock.GetView() = timeout.GetView() + 1. We can safely evict all timeout messages with
// timeout.GetView() < currentView - 1.
```

tholonious: This is a simpler explanation than what's in the current comment block. Seems fine to update the comment block to this

consensus/integration_test_types.go
tholonious: That race condition isn't possible. All handlers hold the event loop's mutex when operating on it. When handling a message or an event, all handlers also check if the view on the message/event is state. So, in your example, what would happen is:
1. onTimeoutScheduledTaskExecuted triggers, sent to server.go but not processed yet
2. ProcessValidatorVote is called, which constructs QC and sends to server.go
3. server.go processes QC via handleVoteQCConstructionEvent
4. server.go sees that the timeout has a stale view (has the same view as the QC), and doesn't process the timeout

consensus/integration_test_types.go
tholonious: Same comment as above. All handle.*() functions hold a lock on the event loop when processing incoming messages / events. Once a handler starts running, it is guaranteed to complete before the next one runs

consensus/integration_test_types.go
tholonious: Same comment as above. This is safe because all handle.*() functions hold a lock on the event loop when processing incoming messages / events.

consensus/integration_test.go
tholonious: Just an update here that we have near complete coverage of unit test cases, and a robust sequence of integration tests for the event loop here. At this point, we're best off spending testing effort on the production Server.go.

consensus/event_loop.go
tholonious: Your suggestion is definitely better than having deadlock risk. I've addressed this in https://github.com/deso-protocol/core/pull/836

Although in practice, if the event loop has a backlog of 10K events, then it means that several hours worth of block proposal and timeout events have been queued but not processed by server.go. This would mean that server.go is very broken, and that the node has more issues that need to be resolved in server.go aside from this deadlock

lazynina (approved):

lazynina (approved):

PR #652, opened 2023-09-06
TransactionsConnectStatus and ConnectFailingTransaction

AeonSw4n:


AeonSw4n:
* feature/proof-of-stake
    * **PR #652** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/652" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/652?utm_source=stack-comment).

lib/block_view.go
```diff
@@ -2128,6 +2106,36 @@ func (bav *UtxoView) _connectBasicTransferWithExtraSpend(
 	return totalInput, totalOutput, utxoOpsForTxn, nil
 }
 
+func (bav *UtxoView) _helpVerifySignature(txn *MsgDeSoTxn, blockHeight uint32) error {
```
tholonious, line 2109: `_helpVerifySignature` is a bit ambiguous. The word help implies that this function performs a subset of signature verifications, but not all. Is that correct?

I suggest renaming it to something like `_verifyTxnSignature` for clarity
```suggestion
func (bav *UtxoView) _verifyTxnSignature(txn *MsgDeSoTxn, blockHeight uint32) error {
```

lib/block_view.go
```diff
@@ -3083,6 +3091,30 @@ func (bav *UtxoView) _connectUpdateGlobalParams(
 		}
 	}
 
+	if blockHeight >= bav.Params.ForkHeights.ProofOfStake2ConsensusCutoverBlockHeight {
```
tholonious, line 3094: Note: we need to be able to configure this global param starting with the PoS setup block height
```suggestion
	if blockHeight >= bav.Params.ForkHeights.ProofOfStake1StateSetupBlockHeight {
```

lib/block_view.go
```diff
@@ -3593,6 +3625,97 @@ func (bav *UtxoView) _connectTransaction(txn *MsgDeSoTxn, txHash *BlockHash,
...
+// _connectFailingTransaction is used to process the fee and burn associated with the user submitting a failing transaction.
+// A failing transaction passes formatting validation, yet fails connecting to the UtxoView. This can happen for a number
+// of reasons, such as insufficient DESO balance, wrong public key, etc. With Revolution's Fee-Time block ordering, these
+// failing transactions are included in the blocks and their fees are burned. Initially blocks will include full transactions,
+// and in the future iterations blocks will only include the failing transaction hashes. This is done to prevent spamming
+// attacks on the network with seemingly valid transactions that never commit and just end up occupying the mempool space.
+// In addition, a major part of the effective fees of this transaction is burned with BMF. This makes spam attacks economically
+// disadvantageous. Attacker's funds are burned, to the benefit of everyone else on the network. BMF algorithm also computes
+// a utility fee, which is distributed to the block producer.
+func (bav *UtxoView) _connectFailingTransaction(txn *MsgDeSoTxn, blockHeight uint32, verifySignatures bool) (
```
tholonious, line 3637: What if the transaction correctly connects? Where do you intend to verify that it's not a connecting txn?

lib/network.go
```diff
@@ -1914,6 +1915,15 @@ type MsgDeSoHeader struct {
...
 	ExtraNonce uint64
 
+	// TransactionsConnectStatus is only used for Proof of Stake blocks, starting with
+	// MsgDeSoHeader version 2. For all earlier versions, this field will default to nil.
+	//
+	// This bitset field stores information whether each transaction in the block passes
+	// or fails to connect. The bit at i-th position is set to 1 if the i-th transaction
+	// in the block passes connect, and 0 otherwise. The length of the bitset must match
+	// the number of transactions in the block.
+	TransactionsConnectStatus *bitset.Bitset
```
tholonious, line 1925: Please move the `TransactionsConnectStatus` field out of MsgDeSoHeader and into MsgDeSoBlock. The txn statuses are most relevant to the txns in the block at the time the txns are being connected. 

You can parse the field in MsgDeSoBlock.FromBytes() based on the block header's version. A good example is how the MsgDeSoBlock.BlockProducerInfo field is parsed based on the block header version. The block header's version is used as a proxy for the entire block's version. 

You can follow the same pattern, and check for header version >= 2 here: https://github.com/deso-protocol/core/blob/feature/proof-of-stake/lib/network.go#L2637

lib/network.go
```diff
@@ -2132,6 +2142,12 @@ func (msg *MsgDeSoHeader) EncodeHeaderVersion2(preSignature bool) ([]byte, error
 	// The Nonce and ExtraNonce fields are unused in version 2. We skip them
 	// during both encoding and decoding.
 
+	// TransactionsConnectStatus
```
tholonious, line 2145: Move this encoding to MsgDeSoBlock.EncodeBlockVersion2()

lib/network.go
```diff
@@ -2346,6 +2362,13 @@ func DecodeHeaderVersion2(rr io.Reader) (*MsgDeSoHeader, error) {
 	retHeader.Nonce = 0
 	retHeader.ExtraNonce = 0
 
+	// TransactionsConnectStatus
```
tholonious, line 2365: Move this decoding to MsgDeSoBlock.FromBytes()

lib/block_view.go
tholonious, line 3637: Consider moving this _connectFailingTransaction() function and the computeBMF() function below to a separate PR. The BMF algo is still in discussion, whereas the TransactionConnectStatus field addition is a straighfoward schema change that you get reviewed and merged quickly.

tholonious (commented):
Rather than merging all of this directly into `feature/pos-mempool` (which is feature complete and is ready for DH's review), consider creating a separate feature PR `feature/pos-txn-connect` for this set of upcoming changes. 

This way, all related sets of changes are in their own feature PRs. The mempool will be easy to review for DH as you continue to work on everything fot txn connects

lib/network.go
tholonious, line 1925: Also consider naming this something like `TxnConnectStatusByIndex`, `TxnConnectStatusList`,... so the idea that the status lookup by index is contextualized in the name.

tholonious:
Another thought: should we incorporate the new `TransactionsConnectStatus` field when computing the block hash for a block? Validators vote on (block hash, view) pair. So if the block hash doesn't incorporate the new `TransactionsConnectStatus`, then a validator can doctor the the bitset when propagating blocks to peers, which would cause peers to think the block as invalid when validating it.

lib/block_view.go
AeonSw4n: Hm, yeah we could change the name. But I'm not sure that `_verifyTxnSignature` is better. There's the primary signature verification utility called `_verifySignature`, and we wouldn't want to confuse the two together.

lib/block_view.go
AeonSw4n: This would be a part of the block producer. A transaction will be first connected, as usual. Upon failure, the transaction will be connected with the `_connectFailingTransaction` function.

AeonSw4n:
Closed in favor of #673

PR #644, opened 2023-08-25
Initial boilerplate code for new lockup transactions.

poolcoke:

PR #661, opened 2023-09-11
Draft connect and disconnect logic for CoinLockupTransfer txn type.

poolcoke:


poolcoke:
* feature/lockups
  * **PR #644** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/644" target="_blank">Graphite</a> 
    * **PR #653** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/653" target="_blank">Graphite</a> 
      * **PR #661** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/661" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/661?utm_source=stack-comment).

AeonSw4n:
Do we need the disconnect logic?

poolcoke:
> Random thought: Do we need the disconnect logic here?

Good question... technically if we're full PoS by the time of launch no. This is the first PoS fork though so I think we'll still be using PoW under the hood until the second PoS fork unless I'm missing something. It's probably also good to keep disconnect logic in the event we do need to rewind the chain (e.g. an exploit). Obviously there's other ways around that but I'm not sure it's worth abandoning it without more careful thought.

AeonSw4n:
Okay makes sense! Let's stick with having the disconnect logic.

PR #658, opened 2023-09-11
Implement Fast-HotStuff Timeout Signal

tholonious:

PR #657, opened 2023-09-11
Migrate Fast-HotStuff Event Loop to Use ScheduledTasks

tholonious:
We originally had an event loop as follows:
```
select {
    case <- timeoutSignal:
        // Do something for timeout
    case <- blockProductionSignal
        // Do something
    case <- exitSignal:
        return
}
```

The issue with this is that it's susceptible to race conditions. When the timeoutSignal triggers, it's possible that exactly at that moment, the view has advanced elsewhere. This can result in the signal causing a timeout of a view we didn't intend.

The ScheduledTask pattern captures the exact view we intend to timeout on when scheduling the task, so the above risk is eliminated.

Ex usage of ScheduledTask:
```
nextTimeoutTask.Schedule(
    duration,
    viewToTimeout,
    func(view) {
        // Do something for the timed out view value defined by viewToTimeout
    }
)
```

PR #656, opened 2023-09-11
Implement ScheduledTask Pattern for FastHotStuffEventLoop Timers

tholonious:

PR #654, opened 2023-09-07
Fast HotStuff Timeout Exponential Backoff

tholonious:

PR #663, opened 2023-09-11
Implement Fast-HotStuff Block Construction Signal

tholonious:
The crank timer periodically checks if the event loop is able to construct a block in the current view. If it can, it constructs the QC and signals the server with the result.

### DIFF
--- a/consensus/event_loop.go
+++ b/consensus/event_loop.go
@@ -142,10 +142,10 @@ func (fc *FastHotStuffEventLoop) ProcessTipBlock(tip BlockWithValidators, safeBl
 	// Signal the caller that we can vote for the block. The caller will decide whether to construct and
 	// broadcast the vote.
 	fc.ConsensusEvents <- &ConsensusEvent{
-		EventType:   ConsensusEventTypeVote,
-		BlockHash:   fc.tip.block.GetBlockHash(),
-		BlockHeight: fc.tip.block.GetHeight(),
-		View:        fc.tip.block.GetView(),
+		EventType:      ConsensusEventTypeVote,
+		TipBlockHash:   fc.tip.block.GetBlockHash(),
+		TipBlockHeight: fc.tip.block.GetHeight(),
+		View:           fc.tip.block.GetView(),
 	}
 
 	// Schedule the next block construction and timeout scheduled tasks
@@ -448,10 +448,10 @@ func (fc *FastHotStuffEventLoop) onBlockConstructionScheduledTaskExecuted(blockC
 	if canConstructQC, signersList, signature := fc.tryConstructVoteQCInCurrentView(); canConstructQC {
 		// Signal the server that we can construct a QC for the chain tip
 		fc.ConsensusEvents <- &ConsensusEvent{
-			EventType:   ConsensusEventTypeConstructVoteQC, // The event type
-			View:        fc.currentView,                    // The current view in which we can construct a block
-			BlockHash:   fc.tip.block.GetBlockHash(),       // Block hash for the tip, which we are extending from
-			BlockHeight: fc.tip.block.GetHeight() + 1,      // The next block height
+			EventType:      ConsensusEventTypeConstructVoteQC, // The event type
+			View:           fc.currentView,                    // The current view in which we can construct a block
+			TipBlockHash:   fc.tip.block.GetBlockHash(),       // Block hash for the tip, which we are extending from
+			TipBlockHeight: fc.tip.block.GetHeight(),          // Block height for the tip, which we are extending from
 			QC: &quorumCertificate{
 				blockHash: fc.tip.block.GetBlockHash(), // Block hash for the tip, which we are extending from
 				view:      fc.tip.block.GetView(),      // The view from the tip block. This is always fc.currentView - 1
@@ -461,10 +461,6 @@ func (fc *FastHotStuffEventLoop) onBlockConstructionScheduledTaskExecuted(blockC
 				},
 			},
 		}
-
-		// Cancel the block construction task since we know we can construct a QC in the current view.
-		// It will be rescheduled when we advance view.
-		fc.nextBlockConstructionTask.Cancel()
 
 		return
 	}
@@ -570,9 +566,10 @@ func (fc *FastHotStuffEventLoop) onTimeoutScheduledTaskExecuted(timedOutView uin
 
 	// Signal the server that we are ready to time out
 	fc.ConsensusEvents <- &ConsensusEvent{
-		EventType: ConsensusEventTypeTimeout,   // The timeout event type
-		View:      timedOutView,                // The view we timed out
-		BlockHash: fc.tip.block.GetBlockHash(), // The last block we saw
+		EventType:      ConsensusEventTypeTimeout,   // The timeout event type
+		View:           timedOutView,                // The view we timed out
+		TipBlockHash:   fc.tip.block.GetBlockHash(), // The last block we saw
+		TipBlockHeight: fc.tip.block.GetHeight(),    // The last block we saw
 	}
 
 	// Cancel the timeout task. The server will reschedule it when it advances the view.
--- a/consensus/event_loop.go
+++ b/consensus/event_loop.go
@@ -454,7 +454,7 @@ func (fc *FastHotStuffEventLoop) onBlockConstructionScheduledTaskExecuted(blockC
 			BlockHeight: fc.tip.block.GetHeight() + 1,      // The next block height
 			QC: &quorumCertificate{
 				blockHash: fc.tip.block.GetBlockHash(), // Block hash for the tip, which we are extending from
-				view:      fc.tip.block.GetView(),      // The view from the tip block. This is always currentView - 1
+				view:      fc.tip.block.GetView(),      // The view from the tip block. This is always fc.currentView - 1
 				aggregatedSignature: &aggregatedSignature{
 					signersList: signersList, // The signers list who voted on the tip block
 					signature:   signature,   // Aggregated signature from votes on the tip block
@@ -476,11 +476,11 @@ func (fc *FastHotStuffEventLoop) onBlockConstructionScheduledTaskExecuted(blockC
 }
 
 // tryConstructVoteQCInCurrentView is a helper function that attempts to construct a QC for the tip block
-// so that can be proposed in a block in the current view. The function internally performs all view and vote
+// so that it can be proposed in a block in the current view. The function internally performs all view and vote
 // validations to ensure that the resulting QC is valid. If a QC can be constructed, the function returns
 // the signers list and aggregate signature that can be used to construct the QC.
 //
-// This function must be called while holding the consensus instance's lock held.
+// This function must be called while holding the consensus instance's lock.
 func (fc *FastHotStuffEventLoop) tryConstructVoteQCInCurrentView() (
 	_success bool, // true if and only if we are able to construct a vote QC for the tip block in the current view
 	_signersList *bitset.Bitset, // bitset of signers for the aggregated signature for the tip block
@@ -526,7 +526,7 @@ func (fc *FastHotStuffEventLoop) tryConstructVoteQCInCurrentView() (
 			continue
 		}
 
-		// Track the votes signature, stake, and place in the validator set
+		// Track the vote's signature, stake, and place in the validator set
 		totalVotingStake = uint256.NewInt().Add(totalVotingStake, validator.GetStakeAmount())
 		signersList.Set(ii, true)
 		signatures = append(signatures, vote.GetSignature())
--- a/consensus/event_loop.go
+++ b/consensus/event_loop.go
@@ -3,10 +3,12 @@ package consensus
 import (
 	"time"
 
+	"github.com/holiman/uint256"
 	"github.com/pkg/errors"
 
 	"github.com/deso-protocol/core/bls"
 	"github.com/deso-protocol/core/collections"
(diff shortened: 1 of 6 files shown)


tholonious:
* feature/proof-of-stake
    * **PR #648** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/648" target="_blank">Graphite</a> 
      * **PR #650** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/650" target="_blank">Graphite</a> 
        * **PR #654** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/654" target="_blank">Graphite</a> 
          * **PR #656** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/656" target="_blank">Graphite</a> 
            * **PR #657** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/657" target="_blank">Graphite</a> 
              * **PR #658** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/658" target="_blank">Graphite</a> 
                * **PR #663** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/663" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/663?utm_source=stack-comment).

consensus/event_loop.go
```diff
@@ -393,14 +395,125 @@ func (fc *FastHotStuffEventLoop) resetScheduledTasks() {
...
-	fc.nextBlockConstructionTask.Schedule(fc.blockConstructionInterval, fc.currentView, fc.onBlockConstructionScheduledTask)
+	fc.nextBlockConstructionTask.Schedule(fc.blockConstructionInterval, fc.currentView, fc.onBlockConstructionScheduledTaskExecuted)
 
 	// Schedule the next timeout task. This will run with currentView param.
 	fc.nextTimeoutTask.Schedule(timeoutDuration, fc.currentView, fc.onTimeoutScheduledTaskExecuted)
 }
 
-func (fc *FastHotStuffEventLoop) onBlockConstructionScheduledTask(blockConstructionView uint64) {
-	// TODO
+// When this function is triggered, it means that we have reached out the block construction
```
lazynina, line 404:
```suggestion
// When this function is triggered, it means that we have reached the block construction
```

consensus/event_loop.go
```diff
@@ -393,14 +395,125 @@ func (fc *FastHotStuffEventLoop) resetScheduledTasks() {
...
+		return
+	}
+
+	// TODO: Check if we can construct a timeout QC for the current view, and send the signal to the
+	// server
+
+	return
+}
+
+func (fc *FastHotStuffEventLoop) tryConstructQCInCurrentView() (
```
lazynina, line 453: Let's add a comment that the caller of this function should hold the consensus lock when calling this function as this function does not hold it itself. also a little more comments on the top of the function providing a little overview would be helpful. Also, should we name this `tryConstructVoteQCInCurrentView`? or will this function also be used for timeout qcs?

consensus/event_loop.go
```diff
@@ -393,14 +395,125 @@ func (fc *FastHotStuffEventLoop) resetScheduledTasks() {
...
+	_success bool, // true if and only if we are able to construct a QC for the tip block in the current view
```
lazynina, line 454:
```suggestion
	_success bool, // true if and only if we are able to construct a vote QC for the tip block in the current view
```

consensus/event_loop.go
```diff
@@ -393,14 +395,125 @@ func (fc *FastHotStuffEventLoop) resetScheduledTasks() {
...
+
+	// Compute the chain tip's signature payload.
+	voteSignaturePayload := GetVoteSignaturePayload(tipBlock.GetView(), tipBlock.GetBlockHash())
+
+	// Fetch the validator votes for the chain tip
+	votesByValidator := fc.votesSeen[voteSignaturePayload]
+
+	// Compute the total stake and total stake with votes
+	totalStake := uint256.NewInt()
+	totalStakeWithVotes := uint256.NewInt()
```
lazynina, line 476: maybe we call this `totalVotingStake`?

consensus/event_loop.go
```diff
@@ -393,14 +395,125 @@ func (fc *FastHotStuffEventLoop) resetScheduledTasks() {
...
+		totalStake = uint256.NewInt().Add(totalStake, validator.GetStakeAmount())
+
+		// Skip the validator if it hasn't voted for the the block
+		vote, hasVoted := votesByValidator[validator.GetPublicKey().ToString()]
+		if !hasVoted {
+			continue
+		}
+
+		// Verify the vote signature
+		if !isValidSignatureSinglePublicKey(vote.GetPublicKey(), vote.GetSignature(), voteSignaturePayload[:]) {
```
lazynina, line 494: Would it be more efficient for us to check multiple pub keys against the payload in one go? Just something to noodle on - not sure how much more efficient it is to use `flowCrypto.VerifyBLSSignatureOneMessage` with many pub keys vs. many calls to `flowCrypto.VerifyBLSSignatureOneMessage` with a single pub key.

lazynina (changes requested):

consensus/event_loop.go
tholonious: The current code is O(n) with respect to the number of votes. Every public key has a separate signature in its vote messages. Note, the signatures aren't aggregated. In order to perform a multi-pubkey signature verification, we would need to first aggregate the signatures which would itself be an O(n) operation.

consensus/event_loop.go
```diff
@@ -414,14 +416,135 @@ func (fc *FastHotStuffEventLoop) resetScheduledTasks() {
...
+	if canConstructQC, signersList, signature := fc.tryConstructVoteQCInCurrentView(); canConstructQC {
+		// Signal the server that we can construct a QC for the chain tip
+		fc.ConsensusEvents <- &ConsensusEvent{
+			EventType:   ConsensusEventTypeConstructVoteQC, // The event type
+			View:        fc.currentView,                    // The current view in which we can construct a block
+			BlockHash:   fc.tip.block.GetBlockHash(),       // Block hash for the tip, which we are extending from
+			BlockHeight: fc.tip.block.GetHeight() + 1,      // The next block height
+			QC: &quorumCertificate{
+				blockHash: fc.tip.block.GetBlockHash(), // Block hash for the tip, which we are extending from
+				view:      fc.tip.block.GetView(),      // The view from the tip block. This is always currentView - 1
```
lazynina, line 456: this means `fc.currentView - 1 = fc.tip.block.GetView()` correct?
```suggestion
				view:      fc.tip.block.GetView(),      // The view from the tip block. This is always fc.currentView - 1
```

just helps me better map this in my brain.

consensus/event_loop.go
```diff
@@ -414,14 +416,135 @@ func (fc *FastHotStuffEventLoop) resetScheduledTasks() {
...
+	}
+
+	// TODO: Check if we can construct a timeout QC for the current view, and send the signal to the
+	// server
+
+	return
+}
+
+// tryConstructVoteQCInCurrentView is a helper function that attempts to construct a QC for the tip block
+// so that can be proposed in a block in the current view. The function internally performs all view and vote
```
lazynina, line 478: this comment reads a little funny. 
```suggestion
// so that it can be proposed in a block in the current view. The function internally performs all view and vote
```

consensus/event_loop.go
```diff
@@ -414,14 +416,135 @@ func (fc *FastHotStuffEventLoop) resetScheduledTasks() {
...
+// validations to ensure that the resulting QC is valid. If a QC can be constructed, the function returns
+// the signers list and aggregate signature that can be used to construct the QC.
+//
+// This function must be called while holding the consensus instance's lock held.
```
lazynina, line 482:
```suggestion
// This function must be called while holding the consensus instance's lock.
```

consensus/event_loop.go
lazynina: yeah this is not a today problem.

consensus/event_loop.go
```diff
@@ -414,14 +416,135 @@ func (fc *FastHotStuffEventLoop) resetScheduledTasks() {
...
+		if !hasVoted {
+			continue
+		}
+
+		// Verify the vote signature
+		if !isValidSignatureSinglePublicKey(vote.GetPublicKey(), vote.GetSignature(), voteSignaturePayload[:]) {
+			continue
+		}
+
+		// Track the votes signature, stake, and place in the validator set
```
lazynina, line 528:
```suggestion
		// Track the vote's signature, stake, and place in the validator set
```

consensus/event_loop_test.go
```diff
@@ -595,6 +598,118 @@ func TestResetEventLoopSignal(t *testing.T) {
...
+	// Happy path, there are votes with a super-majority of stake to construct a QC
+	{
+		fc := NewFastHotStuffEventLoop()
+		err := fc.Init(time.Microsecond, time.Hour,
+			BlockWithValidators{block, validatorSet},     // tip
+			[]BlockWithValidators{{block, validatorSet}}, // safeBlocks
+		)
+		require.NoError(t, err)
+
+		// Create a vote from validator 2
```
lazynina, line 672:
```suggestion
		// Create a vote from validator 1
```

lazynina (changes requested):

lazynina (approved):

consensus/event_loop.go
```diff
@@ -415,14 +417,135 @@ func (fc *FastHotStuffEventLoop) resetScheduledTasks() {
...
+
+	// Check if the conditions are met to construct a QC from votes for the chain tip. If so,
+	// we send a signal to the server and cancel the block construction task. The server will
+	// reschedule the task when it advances the view.
+	if canConstructQC, signersList, signature := fc.tryConstructVoteQCInCurrentView(); canConstructQC {
+		// Signal the server that we can construct a QC for the chain tip
+		fc.ConsensusEvents <- &ConsensusEvent{
+			EventType:   ConsensusEventTypeConstructVoteQC, // The event type
+			View:        fc.currentView,                    // The current view in which we can construct a block
+			BlockHash:   fc.tip.block.GetBlockHash(),       // Block hash for the tip, which we are extending from
```
AeonSw4n, line 453: What do you think about renaming:
-  BlockHash -> TipBlockHash
-  BlockHeight -> TipBlockHeight (whilst moving the +1 to the receivers)

This will disambiguate that we're referring to the tip and not the would-be block.

consensus/event_loop.go
```diff
@@ -415,14 +417,135 @@ func (fc *FastHotStuffEventLoop) resetScheduledTasks() {
...
+				aggregatedSignature: &aggregatedSignature{
+					signersList: signersList, // The signers list who voted on the tip block
+					signature:   signature,   // Aggregated signature from votes on the tip block
+				},
+			},
+		}
+
+		// Cancel the block construction task since we know we can construct a QC in the current view.
+		// It will be rescheduled when we advance view.
+		fc.nextBlockConstructionTask.Cancel()
```
AeonSw4n, line 467: Why do we need to cancel here? Wouldn't the interval be canceled automatically since it just triggered this function?

consensus/event_loop.go
```diff
@@ -415,14 +417,135 @@ func (fc *FastHotStuffEventLoop) resetScheduledTasks() {
...
 	// Schedule the next timeout task. This will run with currentView param.
 	fc.nextTimeoutTask.Schedule(timeoutDuration, fc.currentView, fc.onTimeoutScheduledTaskExecuted)
 }
 
-func (fc *FastHotStuffEventLoop) onBlockConstructionScheduledTask(blockConstructionView uint64) {
-	// TODO
+// When this function is triggered, it means that we have reached the block construction
+// time ETA for blockConstructionView. If we have a QC or timeout QC for the view, then we
+// signal the server.
+func (fc *FastHotStuffEventLoop) onBlockConstructionScheduledTaskExecuted(blockConstructionView uint64) {
```
AeonSw4n, line 429: Hmm what if we can't construct a block after the blockConstructionInterval, but then we recieve a couple more votes and now we're able to? Would `onBlockConstructionScheduledTaskExecuted` be called again? Also, should this be gated by a check that we're the leader for this view?

AeonSw4n (changes requested):

consensus/event_loop.go
tholonious: Hmm, I think you're right. There's no chance the task would still be scheduled at this point, so the Cancel() is redundant. I've removed the Cancel()

consensus/event_loop.go
tholonious: Hmmm, that works really well and is less ambiguous for all event types (vote, QC construciton, timeout). I've renamed both fields to your suggestions. Thanks!

consensus/event_loop.go
tholonious: The `onBlockConstructionScheduledTaskExecuted` will get rescheduled on its interval and reruns. See the next PR for the reschedule. 

Re: the leader check, the event loop doesn't really know when the node is leader. This keeps the event loop as simple as possible. The server will know when the node is the leader though. This separation is fine because:
1. The server will only pass vote and timeout messages to the event loop for views when it will be the leader
2. The server will only accept and act on the block construction signal for views when it is the leader

AeonSw4n: Ah yes, I saw you changed it in the next PR. Moved my comments there. And makes sense on the leader check!

AeonSw4n (approved):

PR #664, opened 2023-09-11
Cross-validate View and QC View in Block and Timeout Msgs

tholonious:
The following must always be true:
- For blocks, block.view > block.QC.view
- For timeout messages, timeout.View > timeout.highQC.view

This PR adds validation for the above, and adjusts unit tests so that test data satisfies the above for the happy path.

PR #665, opened 2023-09-12
Implement QC Validation and Super-majority Stake Check

tholonious:


tholonious:
* feature/proof-of-stake
    * **PR #658** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/658" target="_blank">Graphite</a> 
      * **PR #664** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/664" target="_blank">Graphite</a> 
        * **PR #665** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/665" target="_blank">Graphite</a>  👈
        * **PR #663** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/663" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/665?utm_source=stack-comment).

lazynina (approved):

consensus/utils.go
```diff
@@ -6,9 +6,51 @@ import (
...
+
+	// Compute the total stake in the QC and the total stake in the network
+	stakeInQC := uint256.NewInt()
+	totalStake := uint256.NewInt()
+
+	// Fetch the validators in the QC
+	validatorPublicKeysInQC := []*bls.PublicKey{}
+
+	// Fetch the validators in the QC, and compute the sum of stake in the QC and in the network
+	for ii := range validators {
```
AeonSw4n, line 31: Should we also validate that `len(validators) == qc.GetSignersList().Size()`?

consensus/utils.go
```diff
@@ -154,3 +201,38 @@ func isValidSignature(publicKey *bls.PublicKey, signature *bls.Signature, payloa
 	isValid, err := bls.VerifyAggregateSignatureSinglePayload([]*bls.PublicKey{publicKey}, signature, payload)
 	return err == nil && isValid
 }
+
+// This function uses integer math to verify if the provided stake amount represents a
+// super-majority 2f+1 Byzantine Quorum. First we need the following context:
+// - Assume N = total stake in the network
+// - Assume f = safe faulty stake in the network
```
AeonSw4n, line 208: I'm not sure what "safe faulty" means here. Should we just say "faulty stake" or "malicious stake"?

consensus/utils.go
```diff
@@ -154,3 +201,38 @@ func isValidSignature(publicKey *bls.PublicKey, signature *bls.Signature, payloa
...
+// - Assume C = honest stake in the network
+//
+// - We define N = C + f, and N = 3f + 1
+// - We know that f = (N - 1) / 3
+// - We want to determine if we have a BQ where C >= 2f + 1
+//
+// Given the above, we can use derive the following condition as the super-majority QC check:
```
AeonSw4n, line 215: Drop "use".

consensus/utils.go
```diff
@@ -6,9 +6,51 @@ import (
...
+		totalStake.Add(totalStake, validators[ii].GetStakeAmount())
+	}
+
+	// Check if the QC contains a super-majority of stake
+	if !isSuperMajorityStake(stakeInQC, totalStake) {
+		return false
+	}
+
+	// Finally, validate the signature
+	isValidSignature, err := bls.VerifyAggregateSignatureSinglePayload(
```
AeonSw4n, line 45: I'm getting a warning here that `isValidSignature` is already defined. Should we maybe rename it so it doesn't overlap with an existing method?

AeonSw4n (changes requested):

consensus/utils.go
tholonious: Removed the "safe" so it now says "faulty stake"

consensus/utils.go
tholonious: Oof good catch. There was a function with the same name. Fixed now

consensus/utils.go
tholonious: No actually. The Bigint.ToBytes function compresses the byte encoding for a BigInt such if there's a sequence of 0 at higher indices, of the construct, then they are removed from the byte encoding and always assumed to be zero. This way, `len(validators) >= qc.GetSignersList().Size()` for Indices that exist in the validators slice but not in the signers list return a false value.

Ex: if we have a the sequence `true, true, true, false true, false, false, false, false, false`, the binary representation is `0000010111`, which gets compressed to `10111`.  Calling `bitset.Get(8)` returns a value of false for index 9.

AeonSw4n (approved):

PR #671, opened 2023-09-13
PoS Block Producer

AeonSw4n:
**For future reference:**

During Blockchain's ProcessBlockPoS, validate that:
- `HashBitset(MsgDeSoBlock.TxnConnectStatusByIndex).IsEqual( MsgDeSoBlock.Header.TxnConnectStatusByIndexHash)`


AeonSw4n:
* feature/proof-of-stake
    * **PR #671** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/671" target="_blank">Graphite</a>  👈
      * **PR #672** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/672" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/671?utm_source=stack-comment).

lib/block_view.go
```diff
@@ -2077,31 +2078,8 @@ func (bav *UtxoView) _connectBasicTransferWithExtraSpend(
...
-		// and then passed along in ExtraData for evey transaction signed with it.
-		//
-		// We treat block rewards as a special case in that we actually require that they
-		// not have a transaction-level public key and that they not be signed. Doing this
-		// simplifies things operationally for miners because it means they can run their
-		// mining operation without having any private key material on any of the mining
-		// nodes. Block rewards are the only transactions that get a pass on this. They are
-		// also not allowed to have any inputs because they by construction cannot authorize
-		// the spending of any inputs.
-		if txn.TxnMeta.GetTxnType() == TxnTypeBlockReward {
```
lazynina, line 2097: We still need this for verifying old blocks, no?

cmd/run.go
```diff
@@ -83,9 +83,11 @@ func SetupRunFlags(cmd *cobra.Command) {
 		  download historical blocks. Can only be set if HyperSync is true.`)
 
 	// Mempool
-	cmd.PersistentFlags().Uint64("batch-persist-frequency-milliseconds", 30000,
+	cmd.PersistentFlags().Uint64("mempool-backup-time-millis", 30000,
 		"The frequency in milliseconds with which the mempool will persist its state to disk. "+
 			"The default value is 30 seconds, or 30,000 milliseconds.")
+	cmd.PersistentFlags().Uint64("max-mempool-pos-size-bytes", 3<<30,
```
diamondhands0, line 89: Don't use this << notation. It's hard to read and error-prone. It's better to just write out the number.

lib/block_view.go
```diff
@@ -3593,6 +3648,139 @@ func (bav *UtxoView) _connectTransaction(txn *MsgDeSoTxn, txHash *BlockHash,
...
+
+// _connectFailingTransaction is used to process the fee and burn associated with the user submitting a failing transaction.
+// A failing transaction passes formatting validation, yet fails connecting to the UtxoView. This can happen for a number
+// of reasons, such as insufficient DESO balance, wrong public key, etc. With Revolution's Fee-Time block ordering, these
+// failing transactions are included in the blocks and their fees are burned. Initially blocks will include full transactions,
+// and in the future iterations blocks will only include the failing transaction hashes. This is done to prevent spamming
+// attacks on the network with seemingly valid transactions that never commit and just end up occupying the block space.
+// In addition, a major part of the effective fees of this transaction is burned with BMF. This makes spam attacks economically
+// disadvantageous. Attacker's funds are burned, to the benefit of everyone else on the network. BMF algorithm also computes
+// a utility fee, which is distributed to the block producer.
```
diamondhands0, line 3691: Don't put future speculation directly in comments. If you want, it can be added as a TODO underneath so that it's clear that we may not get to it.

lib/block_view.go
```diff
@@ -3593,6 +3648,139 @@ func (bav *UtxoView) _connectTransaction(txn *MsgDeSoTxn, txHash *BlockHash,
...
+// utility fee to the block producer. This is so that block producers have no advantage over other network participants
+// in accruing DESO. The utility fee that block producers get from transaction fees is just enough so that it's economically
+// advantageous to include as many transactions as possible in blocks. Reliably doing so for an extended period of time
+// could accumulate to some DESO from transaction fees.
+// The utility fee is computed by taking a binary logarithm of the fee, and the remainder is burned. BMF(fee) computes:
+//
+//	burnFee := fee - log_2(fee), utilityFee := log_2(fee).
+func computeBMF(fee uint64) (_burnFee uint64, _utilityFee uint64) {
+	// Compute the utility fee as log_2(fee). We can find it by taking the bit length of fee.
+	utilityFee := uint64(bits.Len64(fee))
```
diamondhands0, line 3773: I wouldn't be clever here. It's better to just use BigFloatLog or a standard function because 1) it's easier to read/verify and 2) it's easier to change it in the future.

lib/block_view.go
```diff
@@ -3593,6 +3648,139 @@ func (bav *UtxoView) _connectTransaction(txn *MsgDeSoTxn, txHash *BlockHash,
...
+	failingTransactionFee := uint256.NewInt().SetUint64(txn.TxnFeeNanos)
+	basisPointsAsUint256 := uint256.NewInt().SetUint64(10000)
+
+	effectiveFeeU256 := failingTransactionRate.Mul(failingTransactionRate, failingTransactionFee)
+	effectiveFeeU256.Div(effectiveFeeU256, basisPointsAsUint256)
+	// We should never overflow on the effective fee, since FailingTransactionBMFRateBasisPoints is <= 10000.
+	// But if for some magical reason we do, we set the effective fee to the max uint64. We don't error, and
+	// instead let _spendBalance handle the overflow.
+	if !effectiveFeeU256.IsUint64() {
+		effectiveFeeU256.SetUint64(math.MaxUint64)
```
diamondhands0, line 3731: We should probably error here?

lib/constants.go
```diff
@@ -707,8 +716,8 @@ type DeSoParams struct {
 	// The maximum aggregate number of bytes of transactions included in the PoS mempool.
```
diamondhands0, line 716: These fields were stored in flags. If that's the case they shouldn't be in constants.go. constants.go is for consensus-level constants, not flag constants.

lib/network.go
```diff
@@ -1914,6 +1915,14 @@ type MsgDeSoHeader struct {
 	// event that ASICs become powerful enough to have birthday problems in the future.
 	ExtraNonce uint64
 
+	// TransactionsConnectStatus is only used for Proof of Stake blocks, starting with
+	// MsgDeSoHeader version 2. For all earlier versions, this field will default to nil.
+	//
+	// The hash of the TxnConnectStatusByIndex field in MsgDeSoBlock. It is stored to ensure
+	// that the TxnConnectStatusByIndex is part of the header hash, which is signed by the
+	// proposer. The full index is stored in the block to offload space complexity.
+	TxnConnectStatusByIndexHash *BlockHash
```
diamondhands0, line 1924: I thought we said we'd add it to the existing merkle root rather than adding another field to the header for this. What made us change?

lib/pos_block_producer.go
```diff
@@ -0,0 +1,206 @@
...
+	if err != nil {
+		return nil, errors.Wrapf(err, "PosBlockProducer.CreateBlockTemplate: Problem creating block without header")
+	}
+
+	// Fill out what we can in the block header. This function won't fill out any of the consensus QC fields.
+	block.Header.Version = HeaderVersion2
+	block.Header.PrevBlockHash = latestBlockView.TipHash
+
+	// Compute the merkle root for the block now that all of the transactions have been added.
+	merkleRoot, _, err := ComputeMerkleRoot(block.Txns)
```
diamondhands0, line 92: I would add the extra info you want to add to the merkle root this ComputeMerkleRoot function with a new argument that is allowed to be nil, and delete the new merkle root field you're adding. Also you neglected to set the new merkle root field here looks like, but again I would just delete it.

lib/pos_mempool.go
```diff
@@ -54,18 +80,24 @@ type PosMempool struct {
...
+	// facilitating a "replace by higher fee" feature. This feature gives users the ability to replace their existing
+	// mempool transaction with a new transaction having the same nonce but higher fee.
+	nonceTracker *NonceTracker
 
 	// latestBlockView is used to check if a transaction is valid before being added to the mempool. The latestBlockView
 	// checks if the transaction has a valid signature and if the transaction's sender has enough funds to cover the fee.
 	// The latestBlockView should be updated whenever a new block is added to the blockchain via UpdateLatestBlock.
+	// PosMempool only needs read-access to the block view. It isn't necessary to copy the block view before passing it
+	// to the mempool.
 	latestBlockView *UtxoView
```
diamondhands0, line 101: If it's an expectation that this will be a read-only field I recommend putting readonly in the name so nobody screws it up in the future.

lib/pos_mempool.go
```diff
@@ -98,177 +131,226 @@ func NewPosMempoolIterator(it *FeeTimeIterator) *PosMempoolIterator {
...
 
 	// Create the persister
-	dmp.persister = NewMempoolPersister(dmp.db, int(dmp.params.MempoolBackupTimeMilliseconds))
-
-	// Start the persister and retrieve transactions from the database.
-	dmp.persister.Start()
-	err = dmp.loadPersistedTransactions()
-	if err != nil {
-		return errors.Wrapf(err, "PosMempool.Start: Problem loading persisted transactions")
+	if !mp.inMemoryOnly {
```
diamondhands0, line 171: Can you consolidate this if with the one above?

lib/block_view_types.go
```diff
@@ -3805,6 +3805,11 @@ type GlobalParamsEntry struct {
 	// then the first bucket will be [1000, 1099], the second bucket will be [1100, 1209], the third bucket will
 	// be [1210, 1330], etc.
 	FeeBucketRateMultiplierBasisPoints uint64
+
+	// FailingTransactionBMFRateBasisPoints is the factor of the transaction fee that is used for the computation
+	// BMF. The value is expressed in basis points. For example a value of 2500 means that 25% of the fee will be
+	// failing transaction fee will be used in the BMF algorithm.
+	FailingTransactionBMFRateBasisPoints uint64
```
diamondhands0, line 3812: Can we rename this to FailingTransactionBMFMultiplierBasisPoints? Rate makes me think of "fee rate" and it was confusing until I read the code.

diamondhands0: Just to clarify why I don't like "rate": Rate is usually something "per" something else. Like "per kilobyte." Meanwhile, multiplier is just multiplied by something else, which is what this is.

lib/block_view.go
AeonSw4n: I like it being lightweight like this. SpendBalance should handle all its inputs, and error here

lib/constants.go
AeonSw4n: Oh, you're right!

lib/block_view_types.go
AeonSw4n: Yeah, I like Multiplier more too

PR #672, opened 2023-09-13
PoS Block Producer: TxnConnectStatusByIndex

AeonSw4n:
**TODO:**

During Blockchain's ProcessBlockPoS, validate that:
- `MsgDeSoBlock.TxnConnectStatusByIndex.Size() == len(MsgDeSoBlock.Txns)`
- `HashBitset(MsgDeSoBlock.TxnConnectStatusByIndex).IsEqual( MsgDeSoBlock.Header.TxnConnectStatusByIndexHash)`


lib/network.go
```diff
@@ -2132,6 +2141,12 @@ func (msg *MsgDeSoHeader) EncodeHeaderVersion2(preSignature bool) ([]byte, error
 	// The Nonce and ExtraNonce fields are unused in version 2. We skip them
 	// during both encoding and decoding.
 
+	// TxnConnectStatusByIndexHash
+	if msg.TxnConnectStatusByIndexHash == nil {
```
lazynina, line 2145: Does this need to be gated on some migration?

lib/network.go
```diff
@@ -2346,6 +2361,13 @@ func DecodeHeaderVersion2(rr io.Reader) (*MsgDeSoHeader, error) {
 	retHeader.Nonce = 0
 	retHeader.ExtraNonce = 0
 
+	// TxnConnectStatusByIndexHash
+	retHeader.TxnConnectStatusByIndexHash = &BlockHash{}
```
lazynina, line 2365: do we need some sort of new migration or header version to handle this?

lib/network.go
```diff
@@ -2721,6 +2760,14 @@ func (msg *MsgDeSoBlock) FromBytes(data []byte) error {
 		}
 	}
 
+	// Version 2 blocks have a TxnStatusConnectedIndex attached to them.
+	if ret.Header.Version == HeaderVersion2 {
```
lazynina, line 2764: okay I see - this is where the header migration is handled

lib/network.go
tholonious: No migration needed here. The existence of the filed is gated by the header version value = 2

lib/network.go
tholonious: No migration needed here. The header version above defines whether the field exists or not

lib/network_test.go
```diff
@@ -189,38 +196,38 @@ func TestHeaderConversionAndReadWriteMessage(t *testing.T) {
...
-		assert.NoError(err)
+		require.NoError(err)
 
 		testHdr := NewMessage(MsgTypeHeader)
 		err = testHdr.FromBytes(data)
-		assert.NoError(err)
+		require.NoError(err)
 
-		assert.Equal(expectedBlockHeader, testHdr)
+		require.True(assert.Equal(expectedBlockHeader, testHdr))
```
tholonious, line 205: Does this work?
```suggestion
		require.Equal(expectedBlockHeader, testHdr)
```

lib/network_test.go
```diff
@@ -189,38 +196,38 @@ func TestHeaderConversionAndReadWriteMessage(t *testing.T) {
 	// versions we want to test.
 	for _, expectedBlockHeader := range expectedBlockHeadersToTest {
 		data, err := expectedBlockHeader.ToBytes(false)
-		assert.NoError(err)
+		require.NoError(err)
```
tholonious, line 199: Why not pass in the testing instance `t` to the NoError function? Same comment for all of uses of require below
```suggestion
		require.NoError(t, err)
```

tholonious (changes requested):
Code LGTM over all. The unit tests failed in BuildKite though: https://buildkite.com/deso/core/builds/3507#018aae2a-e1b3-470c-876b-0940ba91191e

Can you take a look at the failure?

Not sure why the BK status occasionally doesn't show up in Github. I'll look into it separately.

tholonious (approved):
LGTM! Feel free to merge this and follow up on the TODOs with a separate PR. The bitset size and Bitset hash equivalence check are nice utilities, but don't preclude this code form shipping. 🚀

PR #673, opened 2023-09-13
PoS Block Producer: _connectFailingTransaction

AeonSw4n:

PR #676, opened 2023-09-15
PoS Block Producer: Mempool replace with higher fee

AeonSw4n:


AeonSw4n:
* feature/proof-of-stake
    * **PR #671** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/671" target="_blank">Graphite</a> 
      * **PR #672** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/672" target="_blank">Graphite</a> 
        * **PR #673** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/673" target="_blank">Graphite</a> 
          * **PR #676** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/676" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/676?utm_source=stack-comment).

lib/pos_mempool_nonce_tracker.go
```diff
@@ -0,0 +1,52 @@
+package lib
+
+// PosMempoolNonceTracker is a helper struct that is used to track (public key, nonce) pairs in pos mempool.
+// It is used to facilitate the "replace by higher fee" feature, which allows users to resubmit transactions.
+// If a user submits a transaction with an existing DeSoNonce.PartialID (among this user's transactions in the mempool),
+// then the new transaction will replace the old one if the new transaction has a higher fee.
+type PosMempoolNonceTracker struct {
+	// nonceMap indexes mempool transactions by (public key, nonce) pairs.
+	nonceMap map[PublicKey]map[uint64]*MempoolTx
```
lazynina, line 9: Nonces are actually PKID + Uint64. We should use PKIDs so we can continue support for swap identity transactions. Additionally, I'd recommend we use the DeSoNonce struct as the map key instead of having nested maps, unless there's a use case for looking up all the nonces for a given public key that I'm unaware of.
```suggestion
	nonceMap map[DeSoNonce]*MempoolTx
```

lib/pos_mempool_nonce_tracker.go
```diff
@@ -0,0 +1,52 @@
...
+
+func NewPosMempoolNonceTracker() *PosMempoolNonceTracker {
+	return &PosMempoolNonceTracker{
+		nonceMap: make(map[PublicKey]map[uint64]*MempoolTx),
+	}
+}
+
+// GetTxnByPublicKeyNonce returns the transaction with the given public key and nonce pair.
+func (pmnt *PosMempoolNonceTracker) GetTxnByPublicKeyNonce(pk PublicKey, nonce uint64) *MempoolTx {
+	innerMap, ok := pmnt.nonceMap[pk]
```
lazynina, line 20: Do we need a read lock here?

lib/pos_mempool_nonce_tracker.go
```diff
@@ -0,0 +1,52 @@
...
+	txn, ok := innerMap[nonce]
+	if !ok {
+		return nil
+	}
+	return txn
+}
+
+// RemoveTxnByPublicKeyNonce removes a (pk, nonce) pair from the nonce tracker.
+func (pmnt *PosMempoolNonceTracker) RemoveTxnByPublicKeyNonce(pk PublicKey, nonce uint64) {
+	innerMap, ok := pmnt.nonceMap[pk]
```
lazynina, line 33: We should introduce a mutex to control reads and writes on the map.

lazynina (changes requested):

lib/pos_mempool_nonce_tracker.go
lazynina: This suggestion isn't correct. We can create a new struct that has a transactor PKID (or public key, still need to sort this piece out) plus the DeSoNonce so we can avoid the nested map.

tholonious: Just clarifying Nina's comments:

Nonces are made up of the tuple (ExpirationBlockHeight, PartialID). You'll need the map lookup keys to contain both of these values. As Nina suggested, the `DeSoNonce` is the right struct to use to represent the nonce, not uint64.

On using PublicKey vs PKID to identify the user, I don't have a strong opinion on it. Which every key we use, should be consistent across the whole of the mempool though. I.e. if we use PKID, then the ledge should use PKID as well.

lib/pos_mempool_nonce_tracker.go
tholonious: IMO, it's fine to define this tracker with no internal mutex. As long as the parent PosMempool struct wraps the usage of the tracker with its own mutex, it's all kosher and simple to read.

AeonSw4n:
Change PKID to PublicKey

tholonious (approved):

PR #685, opened 2023-09-21
Pass In All Safe Blocks to Fast-HotStuff Event Loop

tholonious:
The event loop needs to have the full chain of blocks from the committed tip to the uncommitted tip, since they can all be extended from when building the next block. The validator set should also be expected to change at any of those blocks, since the chain may span an epoch transitions.

This PR updates the event loop's Init and ProcessSafeBlocks functions so that they take in the above chain of blocks, and the validator set at each block height.

### DIFF
--- a/consensus/event_loop.go
+++ b/consensus/event_loop.go
@@ -60,7 +60,7 @@ func (fc *FastHotStuffEventLoop) Init(
 	}
 
 	// We track the current view here so we know which view to time out on later on.
-	fc.currentView = safeBlocks[len(safeBlocks)-1].Block.GetView() + 1
+	fc.currentView = tip.Block.GetView() + 1
 
 	// Reset all internal data structures for votes and timeouts
 	fc.votesSeen = make(map[[32]byte]map[string]VoteMessage)
--- a/consensus/event_loop.go
+++ b/consensus/event_loop.go
@@ -22,7 +22,8 @@ func NewFastHotStuffEventLoop() *FastHotStuffEventLoop {
 // params:
 //   - blockConstructionInterval: block construction duration must be > 0
 //   - timeoutBaseDuration: timeout base duration must be > 0
-//   - tip: the current tip of the blockchain, with the validator set at that block height
+//   - tip: the current tip of the blockchain, with the validator set at that block height. This may
+//     be a committed or uncommitted block.
 //   - safeBlocks: an unordered slice of blocks including the committed tip, the uncommitted tip,
 //     all ancestors of the uncommitted tip that are safe to extend from, and all blocks from forks
 //     that are safe to extend from. This function does not validate the collection of blocks. It
@@ -110,7 +111,7 @@ func (fc *FastHotStuffEventLoop) AdvanceView() (uint64, error) {
 // timers.
 //
 // Expected params:
-//   - tip: the current tip of the blockchain, with the validator set at that block height
+//   - tip: the current uncommitted tip of the blockchain, with the validator set at that block height
 //   - safeBlocks: an unordered slice of blocks including the committed tip, the uncommitted tip,
 //     all ancestors of the uncommitted tip that are safe to extend from, and all blocks from forks
 //     that are safe to extend from. This function does not validate the collection of blocks. It
--- a/consensus/event_loop.go
+++ b/consensus/event_loop.go
@@ -6,6 +6,7 @@ import (
 	"github.com/pkg/errors"
 
 	"github.com/deso-protocol/core/bls"
+	"github.com/deso-protocol/core/collections"
 )
 
 func NewFastHotStuffEventLoop() *FastHotStuffEventLoop {
@@ -21,18 +22,19 @@ func NewFastHotStuffEventLoop() *FastHotStuffEventLoop {
 // params:
 //   - blockConstructionInterval: block construction duration must be > 0
 //   - timeoutBaseDuration: timeout base duration must be > 0
-//   - chainTip: the input block must have a valid block hash, block height, view, and QC
-//   - validators: the validators must be sorted in decreasing order of stake, with a
-//     consistent tie breaking scheme. The validator set is expected to be valid for
-//     validating votes and timeouts for the next block height.
+//   - tip: the current tip of the blockchain, with the validator set at that block height
+//   - safeBlocks: an unordered slice of blocks including the committed tip, the uncommitted tip,
+//     all ancestors of the uncommitted tip that are safe to extend from, and all blocks from forks
+//     that are safe to extend from. This function does not validate the collection of blocks. It
+//     expects caller to know and decide what blocks are safe to extend from.
 //
-// Given the above, This function updates the chain tip internally, and re-initializes all internal
-// data structures that are used to track incoming votes and timeout messages for QC construction.
+// Given the above, This function updates the tip internally, stores the safe blocks, and re-initializes
+// all internal data structures that are used to track incoming votes and timeout messages for QC construction.
 func (fc *FastHotStuffEventLoop) Init(
 	blockConstructionInterval time.Duration,
 	timeoutBaseDuration time.Duration,
-	chainTip Block,
-	validators []Validator,
+	tip BlockWithValidators,
+	safeBlocks []BlockWithValidators,
 ) error {
 	// Grab the consensus instance's lock
 	fc.lock.Lock()
@@ -51,20 +53,13 @@ func (fc *FastHotStuffEventLoop) Init(
 		return errors.New("FastHotStuffEventLoop.Init: Timeout base duration must be > 0")
 	}
 
-	// Validate the integrity of the block
-	if !isProperlyFormedBlock(chainTip) {
-		return errors.New("FastHotStuffEventLoop.Init: Invalid block")
+	// Validate the safe blocks and validator sets, and store them
+	if err := fc.storeBlocks(tip, safeBlocks); err != nil {
+		return errors.Wrap(err, "FastHotStuffEventLoop.Init: ")
 	}
 
-	// Validate the integrity of the validator set
-	if !isProperlyFormedValidatorSet(validators) {
-		return errors.New("FastHotStuffEventLoop.Init: Invalid validator set")
-	}
-
-	// Update the latest safe block and validator set
-	fc.chainTip = chainTip
-	fc.currentView = chainTip.GetView() + 1
-	fc.validatorsAtChainTip = validators
+	// We track the current view here so we know which view to time out on later on.
+	fc.currentView = safeBlocks[len(safeBlocks)-1].Block.GetView() + 1
 
 	// Reset all internal data structures for votes and timeouts
 	fc.votesSeen = make(map[[32]byte]map[string]VoteMessage)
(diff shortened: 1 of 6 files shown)


tholonious:
* feature/proof-of-stake
    * **PR #665** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/665" target="_blank">Graphite</a> 
      * **PR #685** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/685" target="_blank">Graphite</a>  👈
      * **PR #663** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/663" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/685?utm_source=stack-comment).

consensus/event_loop.go
```diff
@@ -109,41 +111,47 @@ func (fc *FastHotStuffEventLoop) AdvanceView() (uint64, error) {
...
+// of stake, with a consistent tie breaking scheme.
+func (fc *FastHotStuffEventLoop) ProcessSafeBlocks(safeBlocks []BlockWithValidatorSet) error {
 	// Grab the consensus instance's lock
 	fc.lock.Lock()
 	defer fc.lock.Unlock()
 
 	// Ensure the consensus instance is running
 	if fc.status != consensusStatusRunning {
-		return errors.New("FastHotStuffEventLoop.ProcessSafeBlock: Consensus instance is not running")
+		return errors.New("FastHotStuffEventLoop.ProcessTipBlock: Consensus instance is not running")
```
lazynina, line 134: Should this be `ProcessSafeBlocks`?

lazynina (approved):

consensus/event_loop.go
tholonious: Good catch. Fixed now

consensus/event_loop.go
```diff
@@ -521,6 +550,23 @@ func (fc *FastHotStuffEventLoop) hasTimedOutForView(publicKey *bls.PublicKey, vi
...
+func (fc *FastHotStuffEventLoop) getBlockAndValidatorSetByHash(blockHash BlockHash) (
+	bool, Block, []Validator, map[string]Validator,
+) {
+	// A linear search here is fine. The safeBlocks slice is expected to be extremely small as it represents the
+	// number of uncommitted blocks in the blockchain. During steady stake, it will have a size of 3 blocks
+	// (one committed, two uncommitted). In the worse case, where the network has an unlucky series of
+	// timeout -> block -> timeout -> block,... it can still be expected to have < 10 blocks.
+	blockHashValue := blockHash.GetValue()
+	for _, block := range fc.safeBlocks {
+		if block.block.GetBlockHash().GetValue() == blockHashValue {
```
lazynina, line 562: typically we like to use `bytes.Equal` to compare byte slices, but since these are `[32]byte` types we can't do that. Maybe we define a function to help with this. I can't quite remember if there was reason that doing a comparison with `==` would result in unexpected outcomes. Alternatively we add a `Eq` function to the interface definition of `BlockHash`, which the existing BlockHash struct already implements.

consensus/event_loop.go
```diff
@@ -521,6 +550,23 @@ func (fc *FastHotStuffEventLoop) hasTimedOutForView(publicKey *bls.PublicKey, vi
...
 
+func (fc *FastHotStuffEventLoop) getBlockAndValidatorSetByHash(blockHash BlockHash) (
+	bool, Block, []Validator, map[string]Validator,
+) {
+	// A linear search here is fine. The safeBlocks slice is expected to be extremely small as it represents the
+	// number of uncommitted blocks in the blockchain. During steady stake, it will have a size of 3 blocks
+	// (one committed, two uncommitted). In the worse case, where the network has an unlucky series of
+	// timeout -> block -> timeout -> block,... it can still be expected to have < 10 blocks.
+	blockHashValue := blockHash.GetValue()
+	for _, block := range fc.safeBlocks {
```
lazynina, line 561: Could we use your `collections.Any` here?

consensus/types.go
```diff
@@ -144,16 +152,17 @@ type FastHotStuffEventLoop struct {
 	nextBlockConstructionTask *ScheduledTask[uint64]
 	nextTimeoutTask           *ScheduledTask[uint64]
 
-	// The latest block accepted by the caller. We only keep track of the latest safe block here because
-	// it's the block we vote on, and construct a vote QC for.
-	chainTip Block
+	// Stores all safe blocks that we can
```
lazynina, line 155: This comment seems out of place?

lazynina (changes requested):

consensus/types.go
tholonious: Good catch

consensus/event_loop.go
tholonious: We want to extract the block and validator set at this block height, so a simple loop is straightfoward here. Not ideal to use collections.Any since that's only meant to return true/false based on a condition.

consensus/event_loop.go
tholonious: You sure that using the `==` equality checks for doesn't work for arrays? Note, these aren't slices. Slices are reference types so it's ambiguous what `==` does underneath.

These are fixed length byte arrays, which means they're value types where the length (32) is part of the type definition. The `==` comparison performs a value comparison. I was confirmed to confirm this via https://go.dev/play/.

consensus/event_loop.go
```diff
@@ -21,18 +22,19 @@ func NewFastHotStuffEventLoop() *FastHotStuffEventLoop {
 // params:
 //   - blockConstructionInterval: block construction duration must be > 0
 //   - timeoutBaseDuration: timeout base duration must be > 0
-//   - chainTip: the input block must have a valid block hash, block height, view, and QC
-//   - validators: the validators must be sorted in decreasing order of stake, with a
-//     consistent tie breaking scheme. The validator set is expected to be valid for
-//     validating votes and timeouts for the next block height.
+//   - tip: the current tip of the blockchain, with the validator set at that block height
```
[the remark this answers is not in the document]
tholonious: Ah good question. It's whatever the tip is in blockchain.go. It'll almost always be the uncommitted tip. But it could be the committed tip in the edge case during the PoW -> PoS cutover. Either way, this code is agnostic to the status of the block. I've added a comment to say the block may be either committed or uncommitted.

consensus/event_loop.go
lazynina: Yeah you right. I think the issue was slices which as you say have ambiguous lengths.

consensus/event_loop.go
```diff
@@ -109,41 +105,34 @@ func (fc *FastHotStuffEventLoop) AdvanceView() (uint64, error) {
...
-func (fc *FastHotStuffEventLoop) ProcessSafeBlock(block Block, validators []Validator) error {
+// timers.
+//
+// Expected params:
+//   - tip: the current uncommitted tip of the blockchain, with the validator set at that block height
+//   - safeBlocks: an unordered slice of blocks including the committed tip, the uncommitted tip,
+//     all ancestors of the uncommitted tip that are safe to extend from, and all blocks from forks
+//     that are safe to extend from. This function does not validate the collection of blocks. It
+//     expects the caller to know and decide what blocks are safe to extend from.
+func (fc *FastHotStuffEventLoop) ProcessTipBlock(tip BlockWithValidators, safeBlocks []BlockWithValidators) error {
```
AeonSw4n, line 119: Seems like a more accurate name for this function would be `ProcessFork` or something alike. Since it's basically what we're doing here.

consensus/event_loop.go
```diff
@@ -163,6 +152,47 @@ func (fc *FastHotStuffEventLoop) ProcessSafeBlock(block Block, validators []Vali
...
+	// Store the blocks and validator sets
+	fc.safeBlocks = collections.Transform(safeBlocks, func(block BlockWithValidators) blockWithValidatorLookup {
+		return blockWithValidatorLookup{
+			block:        block.Block,
+			validatorSet: block.Validators,
+			validatorLookup: collections.ToMap(block.Validators, func(validator Validator) string {
+				return validator.GetPublicKey().ToString()
+			}),
+		}
+	})
```
AeonSw4n, line 191: Cool utils functions!

consensus/types_internal.go
```diff
@@ -18,6 +18,10 @@ func (bh *blockHash) GetValue() [32]byte {
 	return bh.value
 }
 
+func (bh *blockHash) Eq(other BlockHash) bool {
+	return bh.value == other.GetValue()
```
AeonSw4n, line 22: Why not bytes.Equal?

AeonSw4n (approved):

consensus/event_loop.go
tholonious: This function would be called whenever the tip changes. Which could happen due to a fork, or when the previous tip is extended. Can't say if any of these are any more representative of what it does: ProcessFork, ProcessNewTip,...

FWIW, this function has had many names in the past (UpdateChainTip, ProcessSafeBlock, ProcessTipBlock,...). But they all did the same thing 😄. So I'm more inclined to just run with the current name for now

AeonSw4n: Haha you're right we keep changing it. Alright let's roll with ProcessTipBlock

consensus/event_loop.go
```diff
@@ -83,9 +79,9 @@ func (fc *FastHotStuffEventLoop) Init(
 	return nil
 }
 
-// AdvanceView is called when the chain tip has not changed but the consensus instance has signaled a
-// timeout, and can advance to the next view. This function resets the timeout scheduled task and block
-// production scheduled task for the next view.
+// AdvanceView is called when the tip has not changed but the consensus instance has timed out. This
+// function advances the view and resets the timeout scheduled task and block production scheduled
+// tasks.
 func (fc *FastHotStuffEventLoop) AdvanceView() (uint64, error) {
```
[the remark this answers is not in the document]
tholonious: Sounds good. I'll rename it

consensus/event_loop.go
```diff
@@ -51,20 +54,13 @@ func (fc *FastHotStuffEventLoop) Init(
...
-	if !isProperlyFormedValidatorSet(validators) {
-		return errors.New("FastHotStuffEventLoop.Init: Invalid validator set")
-	}
-
-	// Update the latest safe block and validator set
-	fc.chainTip = chainTip
-	fc.currentView = chainTip.GetView() + 1
-	fc.validatorsAtChainTip = validators
+	// We track the current view here so we know which view to time out on later on.
+	fc.currentView = safeBlocks[len(safeBlocks)-1].Block.GetView() + 1
```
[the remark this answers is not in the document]
tholonious: Great catch! The slice was ordered in my original implementation, but I dropped that property later on. This should be `tip.Block.GetView()`

consensus/event_loop.go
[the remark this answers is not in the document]
tholonious: On second thought, this rename will lead to merge conflicts for PRs later in the stack. I'll follow up with a separate PR for the rename.

consensus/types_internal.go
tholonious: Ended up deleting this function entirely since it's not used. But you're right that bytes.Equal(blockHash[:]) is the right solution. I'll use that later in the stack

consensus/event_loop.go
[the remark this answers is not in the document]
tholonious: Renamed it in this PR: https://github.com/deso-protocol/core/pull/699

PR #677, opened 2023-09-20
introduce-committed-status-to-block-node

lazynina:
- Defines an enum for CommittedBlockStatus with two values, COMMITTED and UNCOMMITTED
- Adds CommittedStatus to BlockNode struct
- Renames NewBlockNode to NewPoWBlockNode, this function sets CommittedStatus to COMMITTED when called. The function signature does not change.
- Create NewPoSBlockNode function with the same parameters as NewPoWBlockNode PLUS CommittedStatus

PR #686, opened 2023-09-21
process-pos-block-skeleton

lazynina:


lazynina:
* feature/proof-of-stake
    * **PR #677** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/677" target="_blank">Graphite</a> 
      * **PR #686** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/686" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/686?utm_source=stack-comment).

lib/blockchain.go
```diff
@@ -2641,9 +2641,170 @@ func (bc *Blockchain) processBlockPoW(desoBlock *MsgDeSoBlock, verifySignatures
 // 4. Connect the block to the blockchain's tip
 // 5. If applicable, flush the incoming block's grandparent to the DB
 // 6. Notify the block proposer, pacemaker, and voting logic that the incoming block has been accepted
-func (bc *Blockchain) processBlockPoS(desoBlock *MsgDeSoBlock, verifySignatures bool) (_isMainChain bool, _isOrphan bool, err error) {
+func (bc *Blockchain) processBlockPoS(desoBlock *MsgDeSoBlock, verifySignatures bool) (_success bool, _isOrphan bool, _blockHashToRequest *BlockHash, _err error) {
```
tholonious, line 2644: Went back and forth on this. Should we just make the `_blockHashToRequest` return value a slice of BlockHashes? Seems more flexible and a bit more future proof of we do. 

The node implementation to request the blocks won't change really.

tholonious (commented):
The Fast-HotStuff reference implementation's `handleBlockFromPeer` function is a good reference for a skeleton for this function: https://github.com/deso-protocol/hotstuff_pseudocode/blob/6409b51c3a9a953b383e90619076887e9cebf38d/fast_hotstuff_bls.go#L573

Everything in the `handleBlockFromPeer` function aside from the safeVote predicate need happen in this function too.

lib/blockchain.go
```diff
@@ -2639,9 +2639,170 @@ func (bc *Blockchain) processBlockPoW(desoBlock *MsgDeSoBlock, verifySignatures
...
 	// TODO: Implement me
-	return false, false, fmt.Errorf("ProcessBlockPoS: Not implemented yet")
+	// 1. Surface Level validation fo the block
```
AeonSw4n, line 2644: Suggestion. Maybe we could simplify this blockchain.go a little by consolidating the sanity checks into one `validateDeSoBlockPos()`, or something alike. For pos mempool I made a special file called `validation_deso_txn` where I confined all the validation logic for MsgDeSoTxn. We could do the same for MsgDeSoBlock with a file like `validation_deso_block`. Usually these sanity checks only require a couple of inputs like desoparams or globalparams, so it feels like they these can be made static functions. In the case of mempool, I think the code ended up much simpler, and txn validation was easily portable to other places like UtxoView. What do you think?

lazynina: I've moved it to a new file and packaged it all together into one calling function. Having a bunch of little functions allows for easier unit testing and for breaking the work into smaller chunks.

Unfortunately we still need access to the blockchain struct in order to perform these validations as we are comparing attributes of the block to the current state of the blockchain, so reuse won't be that valuable unfortunately. but certainly feels nice to have them in a separate file.

lib/blockchain.go
```diff
@@ -482,6 +482,13 @@ type Blockchain struct {
 	bestHeaderChain    []*BlockNode
 	bestHeaderChainMap map[BlockHash]*BlockNode
 
+	// Track the highest QC we've seen.
+	highestQC *QuorumCertificate
+
+	// Tracks all uncommitted blocks in memory
+	uncommittedBlocks    []*MsgDeSoBlock
```
tholonious, line 489: Do we need slice here given we have the uncommitedBlocksMap below?

lib/blockchain.go
```diff
@@ -482,6 +482,13 @@ type Blockchain struct {
...
+	uncommittedBlocksMap map[BlockHash]*MsgDeSoBlock
```
tholonious, line 490: What's interesting about this map is that you could use it to store not just the uncommitted blocks within the best chain, but also uncommitted blocks from possible re-orgs as well. 

It would be the equivalent of the `safeBlocks` map in the reference implementation. So we could pass in these blocks into the `ProcessTipBlock(tip, safeBlocks)` function of the event loop so it can properly construct timeout QCs that result in reorgs. 

Just a thought

lib/blockchain.go
```diff
@@ -482,6 +482,13 @@ type Blockchain struct {
 	bestHeaderChain    []*BlockNode
 	bestHeaderChainMap map[BlockHash]*BlockNode
 
+	// Track the highest QC we've seen.
+	highestQC *QuorumCertificate
```
tholonious, line 486: Do we need to store this separately? Alternatively, you could dynamically return by having a getter function that returns: `bestChain[len(bestChain) - 1].Header.GetQC()`.

If we can avoid storing it as a duplicate here, it would be ideal. But NBD either way

lib/blockchain.go
```diff
@@ -2633,15 +2640,172 @@ func (bc *Blockchain) processBlockPoW(desoBlock *MsgDeSoBlock, verifySignatures
...
-// 5. If applicable, flush the incoming block's grandparent to the DB
-// 6. Notify the block proposer, pacemaker, and voting logic that the incoming block has been accepted
-func (bc *Blockchain) processBlockPoS(desoBlock *MsgDeSoBlock, verifySignatures bool) (_isMainChain bool, _isOrphan bool, err error) {
+//  1. Validate on an incoming block, its header, its block height, the leader, and its QCs (vote or timeout)
+//  2. Store the block in the block index
+//  3. Determine if we're missing a parent block of this block and any of its parents from the block index.
+//     If so, return the hash of the missing block.
+//  4. Resolves forks within the last two blocks
+//  5. Connect the block to the blockchain's tip
+//  6. Update in-memory struct holding uncommitted block. @sofonias - should this happen AFTER we run the commit rule?
```
tholonious, line 2649: Hmm, seems like it'll be easier to do this after the commit rule

You definitely need to evict some blocks from the uncommitted block map, and potentially add the incoming block. It might be easier to do this after you've run the commit rule. You'll know best once you play around with it

lib/blockchain.go
```diff
@@ -2633,15 +2640,172 @@ func (bc *Blockchain) processBlockPoW(desoBlock *MsgDeSoBlock, verifySignatures
...
+//  7. Run the commit rule - If applicable, flushes the incoming block's grandparent to the DB
+//  8. Update the currentView to this new block's view + 1
+//  9. Notify the block proposer, pacemaker, and voting logic that the incoming block has been accepted by returning success.
```
tholonious, line 2652: No need to have ProcessBlockPoS do this. The server._handleBlock function would pass the status to the Fast-HotStuff event loop
```suggestion
```

lib/validate_deso_block_pos.go
tholonious, line 1: Maybe rename this to pos_block_validation.go?

lib/blockchain.go
```diff
@@ -2633,15 +2640,172 @@ func (bc *Blockchain) processBlockPoW(desoBlock *MsgDeSoBlock, verifySignatures
 }
 
 // processBlockPoS runs the Fast-Hotstuff block connect and commit rule as follows:
```
tholonious, line 2642: What do you think of moving the `ProcessBlockPoS` function to a new `pos_blockchain.go` file? Seems like a tossup since it splits a single struct's implementation into two files, but the `blockchain.go` file is already massive. 

Up to you

lib/blockchain.go
```diff
@@ -2633,15 +2640,172 @@ func (bc *Blockchain) processBlockPoW(desoBlock *MsgDeSoBlock, verifySignatures
...
+		return false, false, nil, err
+	}
+
+	// 7. Commit grandparent if possible.
+	if err = bc.commitGrandparents(desoBlock); err != nil {
+		return false, false, nil, err
+	}
+
+	// 8. Update current view to block's view + 1
+	if err = bc.updateCurrentView(desoBlock); err != nil {
```
tholonious, line 2740: Should it be possible for this to error? Ideally we can avoid an error here so we can't end up in a weird state after the above changes

lib/blockchain.go
```diff
@@ -2633,15 +2640,172 @@ func (bc *Blockchain) processBlockPoW(desoBlock *MsgDeSoBlock, verifySignatures
...
+
+	// 6. Update in-memory struct holding uncommitted blocks.
+	if err = bc.updateUncommittedBlocks(desoBlock); err != nil {
+		return false, false, nil, err
+	}
+
+	// 7. Commit grandparent if possible.
+	if err = bc.commitGrandparents(desoBlock); err != nil {
+		return false, false, nil, err
+	}
```
tholonious, line 2737: Worth thinking about how to order these such that an intermediate failure in one of them doesn't leave the node in a weird state.

tholonious (approved):

lib/blockchain.go
lazynina: Yeah this makes sense. We don't want our highestQC to be from a possible fork, so just using the best chain is sufficient.

lib/blockchain.go
lazynina: I think the proper order should be:
1. Add the block to the best chain. If this fails for any reason, we can't commit grandparents and we can't update uncommitted block ever, so it's a good one to handle first.
2. Commit grandparents. We'll do this second as we need to have the new block on the best chain before we can run the commit rule. 
3. Updating uncommitted blocks should probably be done in two pieces as I think we need to do the following:

- Add this new block to the uncommittedBlocksMap, this is trivial
- Pruning the uncommittedBlocksMaps. This we only need to do if we hit the commit rule OR the reorg rule.
- Neither of these should return an error.

lib/blockchain.go
lazynina: I'm moving all of the logic to `pos_blockchain.go` (including the validations) so it can all live in one easy place.

PR #690, opened 2023-09-22
Minor Cleanup for Encoder Test and Balance Model Height Setters

tholonious:
There's an interplay between between the encoder determinism testing and unit tests that set the balance model heights. This PR cleans up 4 things:
- for nested tests, it runs the the inner tests as sub-tests so that their cleanups are run immediately after the subtest completes
- it updates the test encoder setup to a unit function instead of a unit test in itself. This prevents the original setup and shutdown from being run when not needed
- it removes duplicate calls to setBalanceModelBlockHeights within the same test, so that fewer t.Cleanups are scheduled for it
- when cleaning up DeSoMiner and DeSoBlockProducer, it adds a time.Sleep after the two instances been stopped to avoid race conditions

The above don't eliminate the race conditions in CI, but reduce the chances significantly enough point that they're no longer disruptive

### DIFF
--- a/lib/block_view_access_group_members_test.go
+++ b/lib/block_view_access_group_members_test.go
@@ -51,10 +51,10 @@ func (data *accessGroupMembersTestData) GetInputType() transactionTestInputType 
 func TestBalanceModelAccessGroupMembers(t *testing.T) {
 	setBalanceModelBlockHeights(t)
 
-	TestAccessGroupMembersAdd(t)
-	TestAccessGroupMembersUpdate(t)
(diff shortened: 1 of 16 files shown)


tholonious:
* feature/proof-of-stake
  * **PR #690** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/690" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/690?utm_source=stack-comment).

lazynina (approved):

PR #692, opened 2023-09-23
diamondhands lockups review 001

diamondhands0:


lib/block_view_types.go
```diff
@@ -919,6 +937,28 @@ type UtxoOperation struct {
...
 	// prior to a unstake or unlock stake txn.
 	PrevLockedStakeEntries []*LockedStakeEntry
+
+	//
+	// Coin Lockup fields
+	//
+
+	// PrevLockedBalanceEntry is the previous LockedBalanceEntry prior
+	// to a DAO coin lockup.
+	PrevLockedBalanceEntry *LockedBalanceEntry
```
diamondhands0, line 947: No action necessary here, but Proof of Stake should make disconnects no longer valid. If Sofonias is at a point where this is the case, it's possible we can make this the first txn that doesn't need disconnects, and therefore that doesn't need all this prev crap.

Otherwise, it will be the last txn we'll ever write that has this...

lib/db_utils.go
```diff
@@ -559,7 +559,21 @@ type DBPrefixes struct {
...
+	// corresponding unlock timestamp appended to sort by timestamp.
+	// Prefix, <HodlerPKID [33]byte>, <ProfilePKID [33]byte>, <UnlockTimestampUnixNanoSecs int64> -> <LockedBalanceEntry>
+	PrefixLockedBalanceEntryByHODLerPKIDProfilePKIDExpirationTimestampNanoSecs []byte `prefix_id:"[91]" is_state:"true"`
+
+	// PrefixLockupYieldCurvePointByProfilePKIDAndDurationNanoSecs:
+	// Retrieves a LockupYieldCurvePoint.
+	// The structure of the key enables quick lookups for a (ProfilePKID, Duration) pair as well
+	// as quick construction of yield curve plots over time.
+	// Prefix, <ProfilePKID [33]byte>, <LockupDurationNanoSecs int64> -> <LockupYieldCurvePoint>
+	PrefixLockupYieldCurvePointByProfilePKIDAndDurationNanoSecs []byte `prefix_id:"[92]" is_state:"true"`
```
diamondhands0, line 574: These indexes seem great. It's always nice when a single index suffices for an entry, rather than needing a bunch of indexes for various lookups.

lib/db_utils.go
```diff
@@ -10661,6 +10681,409 @@ func DBDeletePostAssociationWithTxn(txn *badger.Txn, snap *Snapshot, association
...
+	if len(lockedBalanceEntry.HODLerPKID) != btcec.PubKeyBytesLenCompressed {
+		return fmt.Errorf("DbPutLockedBalanceEntryMappingsWithTxn: HODLer PKID "+
+			"length %d != %d", len(lockedBalanceEntry.HODLerPKID), btcec.PubKeyBytesLenCompressed)
+	}
+	if len(lockedBalanceEntry.ProfilePKID) != btcec.PubKeyBytesLenCompressed {
+		return fmt.Errorf("DbPutLockedBalanceEntryMappingsWithTxn: Profile PKID "+
+			"length %d != %d", len(lockedBalanceEntry.ProfilePKID), btcec.PubKeyBytesLenCompressed)
+	}
+
+	if err := DBSetWithTxn(txn, snap, _dbKeyForLockedBalanceEntry(lockedBalanceEntry),
```
diamondhands0, line 10718: Good job using DBSetWithTxn. I was worried you would use txn.Set() which causes bugs because it doesn't update the snapshot properly for hypersync.

lib/db_utils.go
```diff
@@ -10661,6 +10681,409 @@ func DBDeletePostAssociationWithTxn(txn *badger.Txn, snap *Snapshot, association
...
+func DBGetLockedBalanceEntryForHODLerPKIDProfilePKIDExpirationTimestampNanoSecsWithTxn(txn *badger.Txn, snap *Snapshot,
+	hodlerPKID *PKID, profilePKID *PKID, expirationTimestamp int64) *LockedBalanceEntry {
+
+	key := _dbKeyForLockedBalanceEntry(LockedBalanceEntry{
+		HODLerPKID:                  hodlerPKID,
+		ProfilePKID:                 profilePKID,
+		ExpirationTimestampNanoSecs: expirationTimestamp,
+	})
+	lockedBalanceEntryBytes, err := DBGetWithTxn(txn, snap, key)
+	if err != nil {
```
diamondhands0, line 10766: I found it a bit odd that we return something here when it's missing. Two comments here:
1. I'm not really sure what the point of this is if we're going to return the thing they passed in anyway basically.
2. If you definitely want this functionality, then normally what we do is check that err is "Not Found" or whatever the "Not Found" enum is. Otherwise, you run the risk of swallowing more serious errors. Maybe you can return nil if it's an error you don't recognize, and check it on the caller?

lib/db_utils.go
```diff
@@ -10661,6 +10681,409 @@ func DBDeletePostAssociationWithTxn(txn *badger.Txn, snap *Snapshot, association
...
+	currentTimestampUnixNanoSecs int64,
+) ([]*LockedBalanceEntry, error) {
+	// Retrieve all LockedBalanceEntries from db matching hodlerPKID, profilePKID, and
+	// ExpirationTimestampNanoSecs <= currentTimestampUnixNanoSecs.
+	// NOTE: While ideally we would start with <prefix, HODLerPKID, ProfilePKID> and
+	//       seek till <prefix, HODLerPKID, ProfilePKID, currentTimestampUnixNanoSecs>,
+	//       Badger does not support this functionality as the ValidForPrefix() function
+	//       stops when a mismatched prefix occurs, not a "lexicographically less than" prefix.
+	//       For this reason, we start with <prefix, HODLerPKID, ProfilePKID, currentTimestampUnixNanoSecs>
+	//       and iterate backwards while we're valid for the prefix <prefix, HODLerPKID, ProfilePKID>.
```
diamondhands0, line 10811: This is an excellent comment. Love to see it.

lib/db_utils.go
```diff
@@ -10661,6 +10681,409 @@ func DBDeletePostAssociationWithTxn(txn *badger.Txn, snap *Snapshot, association
...
+		ExpirationTimestampNanoSecs: currentTimestampUnixNanoSecs,
+	})
+
+	// Valid for prefix <prefix, HODLerPKID, ProfilePKID>
+	prefixKey := DBPrefixKeyForLockedBalanceEntryByHODLerPKIDandProfilePKID(&LockedBalanceEntry{
+		HODLerPKID:  hodlerPKID,
+		ProfilePKID: profilePKID,
+	})
+
+	// Create an iterator. We set the iterator to reverse in o
```
diamondhands0, line 10831: Typo

lib/db_utils.go
```diff
@@ -10661,6 +10681,409 @@ func DBDeletePostAssociationWithTxn(txn *badger.Txn, snap *Snapshot, association
...
+
+		// Convert LockedBalanceEntryBytes to LockedBalanceEntry.
+		rr := bytes.NewReader(lockedBalanceEntryBytes)
+		lockedBalanceEntry, err := DecodeDeSoEncoder(&LockedBalanceEntry{}, rr)
+		if err != nil {
+			return nil, errors.Wrapf(err, "DBGetUnlockableLockedBalanceEntriesWithTxn: "+
+				"error decoding LockedBalanceEntry: ")
+		}
+
+		if lockedBalanceEntry.ExpirationTimestampNanoSecs < currentTimestampUnixNanoSecs {
```
diamondhands0, line 10857: * Shouldn't it always be the case that ExpirationTimestampNanoSecs is less than currentTimestampUnixNanosSecs, since that's how you constructed the seek? If yes, add a comment explaining that this is a sanity check. If not, explain in a comment why this if check is needed.
* Separate question: It looks like this will ignore things when they're EQUAL to currentTimestampUnixNanoSecs. It's an edge-case, but just wanted to highlight it. Maybe mention in a comment that this is intended.

lib/db_utils.go
```diff
@@ -10661,6 +10681,409 @@ func DBDeletePostAssociationWithTxn(txn *badger.Txn, snap *Snapshot, association
...
+	if len(iterLeftKey) < len(Prefixes.PrefixLockupYieldCurvePointByProfilePKIDAndDurationNanoSecs) {
+		return nil
+	}
+	if !bytes.Equal(iterLeftKey[:len(Prefixes.PrefixLockupYieldCurvePointByProfilePKIDAndDurationNanoSecs)],
+		Prefixes.PrefixLockupYieldCurvePointByProfilePKIDAndDurationNanoSecs) {
+		return nil
+	}
+
+	// Fetch the LockupYieldCurvePoint beneath the key.
+	leftLockupYieldCurvePointBytes, err := DBGetWithTxn(txn, snap, iterLeftKey)
```
diamondhands0, line 11009: I think there's a subtle bug here. The seek could return a yield curve point belonging to another user. Example:
* duration1: user1 curve point 0
* duration2: user1 curve point 1
* <== Imagine duration3 is here ==>
* duration4: user2 curve point 0

In this case, if you were to seek for duration3 and user2 I think you would end up getting a value for user1 and not properly filter it out.

lib/db_utils.go
```diff
@@ -10661,6 +10681,409 @@ func DBDeletePostAssociationWithTxn(txn *badger.Txn, snap *Snapshot, association
...
+	key := _dbKeyForLockupYieldCurvePoint(LockupYieldCurvePoint{
+		ProfilePKID:            profilePKID,
+		LockupDurationNanoSecs: lockupDurationNanoSecs,
+	})
+
+	// Seek left of the yield curve point.
+	iterRightOpts := badger.DefaultIteratorOptions
+	iterRight := txn.NewIterator(iterRightOpts)
+	iterRight.Seek(key)
+	iterRightKey := iterRight.Item().Key()
```
diamondhands0, line 11032: I'm not sure this does what we want. Badger always seeks to a key "less than or equal" to the one you're seeking. So if you have [1, 2, 3] and you seek for 2.5, for example, it will return 2. But the value that's right of 2.5 is actually 3.

I could be wrong but just noodle on it and make sure. Also check the left seeker to verify this as well. But I think the left seeker works properly by default because "less than or equal" does what we want.

lib/db_utils.go
```diff
@@ -10661,6 +10681,409 @@ func DBDeletePostAssociationWithTxn(txn *badger.Txn, snap *Snapshot, association
...
+	if len(iterRightKey) < len(Prefixes.PrefixLockupYieldCurvePointByProfilePKIDAndDurationNanoSecs) {
+		return nil
+	}
+	if !bytes.Equal(iterRightKey[:len(Prefixes.PrefixLockupYieldCurvePointByProfilePKIDAndDurationNanoSecs)],
+		Prefixes.PrefixLockupYieldCurvePointByProfilePKIDAndDurationNanoSecs) {
+		return nil
+	}
+
+	// Fetch the LockupYieldCurvePoint beneath the key.
+	rightLockupYieldCurvePointBytes, err := DBGetWithTxn(txn, snap, iterRightKey)
```
diamondhands0, line 11046: The same bug I mentioned for the left seeker applies here. It could return a point for another user.

lib/block_view.go
```diff
@@ -520,6 +532,29 @@ func (bav *UtxoView) CopyUtxoView() (*UtxoView, error) {
 		newView.TransactorNonceMapKeyToTransactorNonceEntry[entryKey] = &newEntry
 	}
 
+	// Copy the LockedBalanceEntries
+	newView.LockedBalanceEntryKeyToLockedBalanceEntry = make(map[LockedBalanceEntryKey]*LockedBalanceEntry,
```
diamondhands0, line 536: I think you properly copied the keys and values. Just make extra sure you don't have this bug in this code:
* https://go.dev/blog/loopvar-preview

lib/block_view.go
```diff
@@ -3443,6 +3490,17 @@ func (bav *UtxoView) _connectTransaction(txn *MsgDeSoTxn, txHash *BlockHash,
 	case TxnTypeUnjailValidator:
 		totalInput, totalOutput, utxoOpsForTxn, err = bav._connectUnjailValidator(txn, txHash, blockHeight, verifySignatures)
 
+	case TxnTypeCoinLockup:
+		// TODO: Once merged, update blockTimestamp parameter
+		totalInput, totalOutput, utxoOpsForTxn, err = bav._connectCoinLockup(txn, txHash, blockHeight, 0, verifySignatures)
```
diamondhands0, line 3495: Normally when I do something like set an argument to zero I like to add a FIXME comment so that I make sure I don't miss it. Normally I treat "FIXME" as "DO NOT MERGE." Ideally we would have a hook that prevents merging as long as a FIXME is in your changes, but we're not that advanced right now...

lib/block_view_lockups.go
```diff
@@ -0,0 +1,1784 @@
...
+	isDeleted                   bool
+}
+
+type LockedBalanceEntryKey struct {
+	HODLerPKID                      PKID
+	ProfilePKID                     PKID
+	ExpirationTimestampUnixNanoSecs int64
+}
+
+func (lockedBalanceEntry *LockedBalanceEntry) Copy() *LockedBalanceEntry {
```
diamondhands0, line 34: Do you need this? If not, delete. If yes, make sure you use it rather than doing *lockedBalanceEntry, eg in the _delete function.

diamondhands0: Also wouldn't Decode(Encode())  be a better way of copying? Maybe ther are other places where we could do this too.

lib/block_view_lockups.go
```diff
@@ -0,0 +1,1784 @@
...
+	}
+
+	// Add a mapping for the LockedBalanceEntry in the view.
+	bav.LockedBalanceEntryKeyToLockedBalanceEntry[lockedBalanceEntry.ToMapKey()] = lockedBalanceEntry
+}
+
+func (bav *UtxoView) _deleteLockedBalanceEntry(lockedBalanceEntry *LockedBalanceEntry) {
+	// Create a tombstone entry.
+	tombstoneLockedBalanceEntry := *lockedBalanceEntry
+	tombstoneLockedBalanceEntry.isDeleted = true
```
diamondhands0, line 124: Better to use Copy() here?

lib/block_view_lockups.go
```diff
@@ -0,0 +1,1784 @@
...
+	}
+
+	// Then, pull unlockable LockedBalanceEntries from the UtxoView.
+	var unlockableLockedBalanceEntries []*LockedBalanceEntry
+	for _, lockedBalanceEntry := range bav.LockedBalanceEntryKeyToLockedBalanceEntry {
+		// Filter to matching LockedBalanceEntries.
+		if !lockedBalanceEntry.HODLerPKID.Eq(hodlerPKID) ||
+			!lockedBalanceEntry.ProfilePKID.Eq(profilePKID) ||
+			lockedBalanceEntry.ExpirationTimestampNanoSecs > currentTimestampNanoSecs ||
+			lockedBalanceEntry.BalanceBaseUnits.IsZero() ||
```
diamondhands0, line 192: Hmm... isn't it possible that the balance could be zero? If this is possible, then we would end up never garbage-collecting a zero entry, resulting in bugs. Consider removing the IsZero() check here.

lib/block_view_lockups.go
```diff
@@ -0,0 +1,1784 @@
...
+	isDeleted                 bool
+}
+
+type LockupYieldCurvePointKey struct {
+	ProfilePKID            *PKID
+	LockupDurationNanoSecs int64
+}
+
+func (lockupYieldCurvePoint *LockupYieldCurvePoint) Copy() *LockupYieldCurvePoint {
+	return &LockupYieldCurvePoint{
```
diamondhands0, line 224: Same commnets as for previous Copy()

lib/block_view_lockups.go
```diff
@@ -0,0 +1,1784 @@
...
+	var rightLockupPoint *LockupYieldCurvePoint
+
+	// Check the view for yield curve points.
+	if _, pointsInView := bav.PKIDToLockupYieldCurvePointKeyToLockupYieldCurvePoints[*profilePKID]; pointsInView {
+		for _, lockupYieldCurvePoint := range bav.PKIDToLockupYieldCurvePointKeyToLockupYieldCurvePoints[*profilePKID] {
+			// Check for nil pointer cases.
+			if lockupYieldCurvePoint.LockupDurationNanoSecs < lockupDuration && leftLockupPoint == nil {
+				leftLockupPoint = lockupYieldCurvePoint
+			}
+			if lockupYieldCurvePoint.LockupDurationNanoSecs >= lockupDuration && rightLockupPoint == nil {
```
diamondhands0, line 357: You have a very subtle behavior here where left is "strictly less" than the lockupDuration while right is "greater than or equal" to the lockupDuration. This is fine, but you need to make 100% sure that the Badger lookups mirror this behavior precisely or else there will be subtle bugs.

lib/block_view_lockups.go
```diff
@@ -0,0 +1,1784 @@
...
+
+			// Check if the point is "more left" than the current left point.
+			if lockupYieldCurvePoint.LockupDurationNanoSecs < lockupDuration &&
+				lockupYieldCurvePoint.LockupDurationNanoSecs > leftLockupPoint.LockupDurationNanoSecs {
+				leftLockupPoint = lockupYieldCurvePoint.Copy()
+			}
+
+			// Check if the point is "more right" than the current right point.
+			if lockupYieldCurvePoint.LockupDurationNanoSecs >= lockupDuration &&
+				lockupYieldCurvePoint.LockupDurationNanoSecs < leftLockupPoint.LockupDurationNanoSecs {
```
diamondhands0, line 369: I think this is a bug. I think you meant `< rightLockupPoint.LockupDurationNanoSecs`

lib/block_view_lockups.go
```diff
@@ -0,0 +1,1784 @@
...
+func (bav *UtxoView) GetLocalYieldCurvePoints(profilePKID *PKID, lockupDuration int64) (
+	_leftLockupPoint *LockupYieldCurvePoint, _rightLockupPoint *LockupYieldCurvePoint) {
+	var leftLockupPoint *LockupYieldCurvePoint
+	var rightLockupPoint *LockupYieldCurvePoint
+
+	// Check the view for yield curve points.
+	if _, pointsInView := bav.PKIDToLockupYieldCurvePointKeyToLockupYieldCurvePoints[*profilePKID]; pointsInView {
+		for _, lockupYieldCurvePoint := range bav.PKIDToLockupYieldCurvePointKeyToLockupYieldCurvePoints[*profilePKID] {
+			// Check for nil pointer cases.
+			if lockupYieldCurvePoint.LockupDurationNanoSecs < lockupDuration && leftLockupPoint == nil {
```
diamondhands0, line 354: You aren't checking isDeleted in this loop, which I think will lead to subtle bugs.

lib/block_view_lockups.go
```diff
@@ -0,0 +1,1784 @@
...
+		leftDBLockupPoint.LockupDurationNanoSecs < lockupDuration &&
+		leftDBLockupPoint.LockupDurationNanoSecs > leftLockupPoint.LockupDurationNanoSecs {
+		leftLockupPoint = leftDBLockupPoint
+	}
+	if rightDBLockupPoint != nil &&
+		rightDBLockupPoint.ProfilePKID.Eq(profilePKID) &&
+		rightDBLockupPoint.LockupDurationNanoSecs >= lockupDuration &&
+		rightDBLockupPoint.LockupDurationNanoSecs < rightLockupPoint.LockupDurationNanoSecs {
+		rightLockupPoint = rightDBLockupPoint
+	}
```
diamondhands0, line 401: Whenever you're using something from the db, you generally want to update the view with it. I think here you want to call a _set on the left/right points that you find.

But I think there's another issue I'll put in a secondary comment.

lib/block_view_lockups.go
```diff
@@ -0,0 +1,1784 @@
...
+
+	return leftLockupPoint, rightLockupPoint
```
diamondhands0, line 403: There is a weird edge-case here where something could be in the view marked as isDeleted. If this is the case, then you need to actually do multiple things to avoid getting screwed by it:
1. You need to skip it when iterating over the view, which I didn't see above
2. You need to skip it when you fetch from the db. This is actually harder than it seems because you don't have a way of getting the *next* leftmost/rightmost point while *skipping* a point that's in the view.

Given all of the complexity associated with this left/right stuff, I had an alternative proposal for you. Let's just do the inefficient thing of fetching *all* the points from the db every time, and get rid of this left/right stuff in Badger. The new code would look as follows:
* When you call _get, you read *all* the lockup points from the db for a particular PKID and set them on the view if they're not already there.
* When you need leftmost/rightmost, you call the above _get to load up the view with *everything* in badger, and then iterate over the view like you have above (but just be careful to check for isDeleted)

The above is much simpler, and avoids having to write skipping logic in badger. If it ever becomes too inefficient, we can always replace it with something better.

diamondhands0: Btw in case it isn't clear, the above change would keep the exact schema you have for the view where it's `map[PKID]map[LockupPointKey]` or whatever.

lib/block_view_lockups.go
```diff
@@ -0,0 +1,1784 @@
...
+	// Construct UtxoOps in the event this transaction is reverted.
+	return 0, 0, utxoOpsForTxn, nil
+}
+
+func CalculateLockupYield(
+	principal *uint256.Int,
+	apyYieldBasisPoints *uint256.Int,
+	durationNanoSecs *uint256.Int,
+) (*uint256.Int, error) {
+	// Note: We could compute either simple of compounding interest. While compounding interest is ideal from an
```
diamondhands0, line 893: Fix typos.

lib/block_view_lockups.go
```diff
@@ -0,0 +1,1784 @@
...
+	//       application perspective, it becomes incredibly difficult to implement from a numerical perspective.
+	//       This is because compound interest requires fractional exponents rather for computing the yield.
+	//       Determining overflow and preventing excessive money-printers becomes tricky in the compound interest case.
+	//       For this reason, we opt to use simple interest.
+	//
+	// Simple interest formula:
```
diamondhands0, line 899: I think Sofonias figured out compound interest for PoS. I think it's also fine to ship this the way it is though. It's arguably easier to reason about this way because you just get X% for locking up for Y duration, similar to how bonds work.

lib/block_view_lockups.go
```diff
@@ -0,0 +1,1784 @@
...
+		//       nothing else is put in its place.
+		bav._deleteLockupYieldCurvePoint(&LockupYieldCurvePoint{
+			ProfilePKID:            profilePKID,
+			LockupDurationNanoSecs: txMeta.LockupYieldDurationNanoSecs,
+		})
+	}
+
+	// Check if we're updating transfer restriction.
+	var prevLockupTransferRestriction TransferRestrictionStatus
+	if txMeta.NewLockupTransferRestrictions {
```
diamondhands0, line 1109: Maybe don't allow this for DeSo?

lib/block_view_lockups.go
```diff
@@ -0,0 +1,1784 @@
...
+		return 0, 0, nil, errors.Wrapf(RuleErrorCoinLockupTransferInsufficientBalance,
+			"_connectCoinLockupTransfer")
+	}
+
+	// Credit the sender's balance entry.
+	senderLockedBalanceEntry.BalanceBaseUnits = *uint256.NewInt().Sub(
+		&senderLockedBalanceEntry.BalanceBaseUnits, txMeta.LockedCoinsToTransferBaseUnits)
+
+	// Fetch the recipient's balance entry.
+	receiverLockedBalanceEntry := bav.GetLockedBalanceEntryForHODLerPKIDProfilePKIDExpirationTimestampNanoSecs(
```
diamondhands0, line 1349: This is tricky. It works because the db function returns an empty balance entry constructed for you as a convenience. Leave a comment here. Or, honestly, I think it would make it more clear if the GetLocked...() function returned nil and you constructed it here explicitly.

lib/block_view_lockups.go
```diff
@@ -0,0 +1,1784 @@
...
+		}
+	}
+	if !remainingUnlockBalance.IsZero() {
+		return 0, 0, nil, errors.Wrapf(RuleErrorCoinUnlockInsufficientUnlockableCoins,
+			"_connectCoinUnlock")
+	}
+
+	// Update the LockedBalanceEntries.
+	for _, lockedBalanceEntry := range newLockedBalanceEntries {
+		bav._setLockedBalanceEntry(lockedBalanceEntry)
```
diamondhands0, line 1567: I think this is a bug in that it will set a bunch of LockedBalanceEntrys in the db with zero balance. Instead, you may want to _delete newBalanceEntrys if they're zero otherwise set them.

lib/block_view_lockups.go
```diff
@@ -0,0 +1,1784 @@
...
+		// Delete the existing mappings in the db for this LockedBalanceEntry.
+		// They will be re-added if the corresponding entry in memory has isDeleted=false.
+		if err := DbDeleteLockedBalanceEntryWithTxn(txn, bav.Snapshot, *lockedBalanceEntry); err != nil {
+			return errors.Wrapf(
+				err, "_flushLockedBalanceEntriesToDbWithTxn: Problem deleting mappings "+
+					"for LockedBalanceEntry: %v", &lockedBalanceEntryKey)
+		}
+	}
+	for _, lockedBalanceEntry := range bav.LockedBalanceEntryKeyToLockedBalanceEntry {
+		if lockedBalanceEntry.isDeleted || lockedBalanceEntry.BalanceBaseUnits.IsZero() {
```
diamondhands0, line 1729: I see you're skipping based on IsZero() here. I think this will work, but we tend to prefer not doing implicit deletions like this because it can lead to unexpected and hard-to-track bugs. Instead, I would try and make it so that your code marks things with zero balance with isDeleted=true before it gets here.

lib/block_view_lockups.go
```diff
@@ -0,0 +1,1784 @@
...
+		hodlerPKID, profilePKID, txMeta.UnlockTimestampNanoSecs)
+	if lockedBalanceEntry == nil || lockedBalanceEntry.isDeleted {
+		lockedBalanceEntry = &LockedBalanceEntry{
+			HODLerPKID:                  hodlerPKID,
+			ProfilePKID:                 profilePKID,
+			ExpirationTimestampNanoSecs: txMeta.UnlockTimestampNanoSecs,
+			BalanceBaseUnits:            *uint256.NewInt(),
+		}
+	}
+	previousLockedBalanceEntry := *lockedBalanceEntry
```
diamondhands0, line 858: I don't see a bug with this right now, but it makes me a little uncomfortable that we allow the user to modify an existing LockedBalanceEntry. It works here because you only allow them to ADD to it. I think this is OK. It would be a little safer to just disallow modifying existing entries, thus forcing them to create a new one that's like 1 nanosecond in the future. But if you can just inspect this really carefully to make sure there are no bugs with the modify approach, I think we can keep it the way it is.

lib/block_view_lockups.go
```diff
@@ -0,0 +1,1784 @@
...
+
+	// Try connecting the basic transfer without considering transaction metadata.
+	_, _, utxoOpsForBasicTransfer, err := bav._connectBasicTransfer(txn, txHash, blockHeight, verifySignatures)
+	if err != nil {
+		return 0, 0, nil, errors.Wrapf(err, "_connectCoinLockup")
+	}
+	utxoOpsForTxn = append(utxoOpsForTxn, utxoOpsForBasicTransfer...)
+
+	// Grab the txn metadata.
+	txMeta := txn.TxnMeta.(*CoinLockupMetadata)
```
diamondhands0, line 682: So something a little bit tricky is that we have global transfer restrictions for a coin that are supposed to disallow people from sending coins to each other before we're ready. I think you need to check that "non-lockup" transfer restriction in this function. It should be pretty easy though. You just have to call a function like IsValidTransfer or something that Isaac wrote a while back.

lib/block_view_lockups.go
```diff
@@ -0,0 +1,1784 @@
...
+		txnYieldBasisPoints = rightYieldCurvePoint.LockupYieldAPYBasisPoints
+		txnYieldEarningDurationNanoSecs = rightYieldCurvePoint.LockupDurationNanoSecs
+	}
+
+	// Convert variables to a consistent uint256 representation. This is to use them in SafeUint256 math.
+	txnYieldBasisPoints256 := uint256.NewInt().SetUint64(txnYieldBasisPoints)
+	txnYieldEarningDurationNanoSecs256 := uint256.NewInt().SetUint64(uint64(txnYieldEarningDurationNanoSecs))
+
+	// Compute the yield associated with this operation, checking to ensure there's no overflow.
+	yieldFromTxn, err :=
```
diamondhands0, line 818: Something a little annoying that we may want to put a little juice into noodling on is how we can allow people to both stake their DESO *and* to leverage this lockup functionality.

For example, staking your DESO earns you yield based on how much we want to pay validators - let's call this the validator yield. But defining yield with these functions would enable a *secondary* yield that you can actually only access if you *don't* stake your DESO. Let's call this secondary yield the lockup yield.

I kindof wish we could find a clever way to combine these. Maybe you have to FIRST stake your DESO with these functions, and *then* you can allocate that LOCKED deso to PoS. This would allow you to earn BOTH the PoS yield and the P yield. But would require coordination with Sofonias and stuff.

I think what we can do is add this to our list of shit for PoS as a "nice to have." If we don't get to it, which we probably won't, then we just ship the "yield unification" change later on after PoS has fully stabilized. It will likely manifest as a secondary way to stake to PoS with locked DESO.

lib/block_view_lockups.go
```diff
@@ -0,0 +1,1784 @@
...
+		CalculateLockupYield(txMeta.LockupAmountBaseUnits, txnYieldBasisPoints256, txnYieldEarningDurationNanoSecs256)
+	if err != nil {
+		return 0, 0, nil, errors.Wrap(err, "_connectCoinLockup")
+	}
+
+	// We check that the minted yield does not cause an overflow in the transactor's balance.
+	// In the case of DeSo being locked up, we must check that the resulting amount is less than 2**64.
+	if uint256.NewInt().Sub(MaxUint256, yieldFromTxn).Lt(transactorBalanceNanos256) {
```
diamondhands0, line 826: Something I was thinking is that maybe we can delete a bunch of Isaac's old lockup code. Specifically, it seems like we can use your new LockedBalanceEntrys to delete:
* PrefixLockedStakeByValidatorAndStakerAndLockedAt and all its associated stuff
* _connectUnlockStake and all its helpers

The only issue I see with this is that Isaac's version relies on "epoch number" whereas this relies on timestamp. But maybe that's ok in the sense that we can "fudge" our unlocks using time rather than epochs, eg setting the unlock as "3 hours from now" rather than "3 epochs from the current epoch." This would introduce a risk that people could unlock a little sooner than they're supposed to. But I don't think this would be a real issue. And it would cause a simplification of the code that could be very nice.

You may need Sofonias's help with this. We can also merge this and then do that consolidation as a follow-up btw. That may be a better/simpler way to do it. You test that all of this is working as intended and then do the consolidation.

lib/block_view_lockups.go
```diff
@@ -0,0 +1,1784 @@
...
+//
+// CoinLockupTransfer Transaction Logic
+//
+
+func (bav *UtxoView) _connectCoinLockupTransfer(
+	txn *MsgDeSoTxn,
+	txHash *BlockHash,
+	blockHeight uint32,
+	verifySignatures bool,
+) (_totalInput uint64,
```
diamondhands0, line 1258: Meta comment after reviewing this function. Initially it was such a strange thing to me. On the one hand, there's no real reason why this function shouldn't exist. It's kindof like baking liquid staking into the protocol level. But on the other hand, the whole point of lockups is that you CANT transfer them!

After noodling on it, I like this function. It effectively implements the "liquid bonding" concept that we talked about in Revolution, whereby locked coins can be sent back and forth similar to a "bond." I also like that it's optional, meaning the creator can turn it off if they really want lockup to mean no movement.

lib/block_view_lockups.go
```diff
@@ -0,0 +1,1784 @@
...
+
+//
+// CoinUnlock Transaction Logic
+//
+
+func (bav *UtxoView) _connectCoinUnlock(
+	txn *MsgDeSoTxn, txHash *BlockHash,
+	blockHeight uint32, blockTimestamp int64,
+	verifySignatures bool) (_totalInput uint64, _totalOutput uint64, _utxoOps []*UtxoOperation, _err error) {
+	var utxoOpsForTxn []*UtxoOperation
```
diamondhands0, line 1452: Meta comment after reviewing: Do we need to require a balance to be passed here? I can imagine it's just going to be a little annoying to implement where first you have to FETCH how much you want to unlock and THEN you can unlock it. It's also inviting floating point issues where maybe the UI is off by a single base unit and then this function fails.

If there's a good reason to require them to specify how much they want to unlock then I'm in favor of it. But it seemed better to either get rid of it or, perhaps better, to ignore it if they set the value to zero or something like that. The latter gives you best of both worlds. If you want to unlock a specific amount you can, otherwise it will just unlock everything that's available.

lib/db_utils.go
diamondhands0: Revisiting this at the end after completing my review. I still think it makes everything easier to read and less error-prone to not return a default value here.

lib/block_view_lockups.go
```diff
@@ -0,0 +1,1784 @@
...
+	ProfilePKID                 *PKID
+	ExpirationTimestampNanoSecs int64
+	BalanceBaseUnits            uint256.Int
+	isDeleted                   bool
+}
+
+type LockedBalanceEntryKey struct {
+	HODLerPKID                      PKID
+	ProfilePKID                     PKID
+	ExpirationTimestampUnixNanoSecs int64
```
diamondhands0, line 31: Sometimes you refer to it as Expiration sometimes as Unlock. I'm not sure that's needed vs sticking with Unlock for all.

lib/block_view_types.go
poolcoke: Piotr brought this up as well about a week ago. Are there potential benefits to keeping disconnects around? I was thinking possible for debugging, testing, or emergency reversals. Not sure if any of those are worthwhile concerns.

lib/db_utils.go
poolcoke: Yes you're definitely right. An earlier version would filter for that but it appears that code got lost somewhere along the refactor.

lib/db_utils.go
poolcoke: This is definitely going to be tested, but I think it's fine per the Badger documentation here: https://pkg.go.dev/github.com/dgraph-io/badger#section-readme

It states, "Just like forward iteration goes to the first key which is equal or greater than the SEEK key, reverse iteration goes to the first key which is equal or lesser than the SEEK key. Therefore, SEEK key would not be part of the results."

I'll follow up on this.

lib/block_view_lockups.go
poolcoke: I have it added as it's a pattern we put in other places (e.g. BalanceEntry). I doesn't seem to be critical or apart of any interface, but I also don't think it's harmful to keep.

lib/block_view_lockups.go
poolcoke: Currently we garbage collect on the flush. Not sure if this is what we want to do, but it's something we did with BalanceEntry as well. It's pretty deep in the DB utils where we actually garbage collect the zero balances.

lib/block_view_lockups.go
poolcoke: Ah yes subtle error.

lib/block_view_lockups.go
poolcoke: Agree. Definitely would be a nasty bug.

lib/block_view_lockups.go
poolcoke: While I agree, I think it's a bit annoying given that we have users specify the lockup expiration in their lockup transaction. Having to put a lockup 1 nanosecond in the future would be a pretty bad UX, especially if we were supporting something like locked coin distributions.

lib/block_view_lockups.go
poolcoke: I agree, one of the big reasons we need transfers though is vested token distributions. For instance, a creator wants to distribute locked tokens that their supporters can unlock in a year.

lib/block_view_lockups.go
poolcoke: Agree this should be cleaned up. I'll move everything to "UnlockTimestamp" for consistency.

PR #693, opened 2023-09-24
diamondhands pos mempool review 2023 09 23

diamondhands0:


cmd/config.go
```diff
@@ -42,6 +42,9 @@ type Config struct {
 	DisableEncoderMigrations  bool
 	HypersyncMaxQueueSize     uint32
 
+	// Mempool
+	MempoolBackupTimeMilliseconds uint64
```
diamondhands0, line 46: I would call this an "interval" because it's how often you back it up: `MempoolBackupIntervalMillis`

cmd/run.go
```diff
@@ -82,6 +82,11 @@ func SetupRunFlags(cmd *cobra.Command) {
 		- hypersync: Will sync by downloading historical state, and will NOT
 		  download historical blocks. Can only be set if HyperSync is true.`)
 
+	// Mempool
+	cmd.PersistentFlags().Uint64("batch-persist-frequency-milliseconds", 30000,
```
diamondhands0, line 86: Why not call this the same as the argument? `mempool-backup-interval-milliseconds`

lib/pos_mempool.go
```diff
@@ -0,0 +1,490 @@
...
+	Next() bool
+	Value() (*MsgDeSoTxn, bool)
+	Initialized() bool
+}
+
+// PosMempool is used by the node to keep track of uncommitted transactions. The main responsibilities of the PosMempool
+// include addition/removal of transactions, back up of transaction to database, and retrieval of transactions ordered
+// by Fee-Time algorithm. More on the Fee-Time algorithm can be found in the documentation of TransactionRegister.
+type PosMempool struct {
+	sync.RWMutex
```
diamondhands0, line 42: Avoid the use of embedding in Go. Better to call this something like `mtx sync.RWMutex`

Reasons to avoid embedding:
* A lot of devs aren't used to it. Even I actually had to do a double-take when I saw it thinking "is this a new feature?"
* When you see PosMempool.something in various places, it makes it more annoying to figure out where it came from because you can't just look at the struct and see it.
* If the definition of the thing you're embedding changes, it can have unexpected behaviors, adding and removing fields from your struct randomly.

lib/pos_mempool.go
```diff
@@ -0,0 +1,490 @@
...
+	dmp.Lock()
+	defer dmp.Unlock()
+
+	if dmp.IsRunning() {
+		return nil
+	}
+
+	// Setup the database.
+	mempoolDirectory := filepath.Join(dmp.dir, "mempool")
+	opts := DefaultBadgerOptions(mempoolDirectory)
```
diamondhands0, line 128: I'm wondering if we want to deal with the memory requirements of opening a new badger. But I think if we can keep it on the DEFAULT options it's pretty low right? How much memory does a default badger instance use if you're regularly writing a gig to it?

lib/pos_mempool.go
```diff
@@ -0,0 +1,490 @@
...
+	}
+
+	return nil
+}
+
+func (dmp *PosMempool) validateTransaction(txn *MsgDeSoTxn) error {
+	dmp.RLock()
+	defer dmp.RUnlock()
+
+	if err := CheckTransactionSanity(txn, uint32(dmp.latestBlockHeight), dmp.params); err != nil {
```
diamondhands0, line 238: I'm not sure CheckTransactonSanity is doing anything essential anymore. Maybe we can delete it?

lib/validation_deso_txn.go
```diff
@@ -0,0 +1,166 @@
...
+		if totalOutNanos > MaxNanos {
+			return errors.Wrapf(RuleErrorTotalOutputExceedsMax, "ValidateDeSoTxnFormatBalanceModel: Total output "+
+				"amount %d exceeds max %d", totalOutNanos, MaxNanos)
+		}
+	}
+	return nil
+}
+
+// ValidateDeSoTxnMinimalNetworkFee validates that the transaction is above the network's minimal fee.
+func ValidateDeSoTxnMinimalNetworkFee(txn *MsgDeSoTxn, globalParams *GlobalParamsEntry) error {
```
diamondhands0, line 155: With PoS we're not going to have a minimal network fee anymore. Instead, the fee can be ZERO. This will probably be the case for some time (until blocks start getting filled). So I would just remove this check entirely.

Instead, node mempools will be responsible for putting higher-fee txns at the front of the queue and dropping spam as needed.

lib/pos_transaction_register.go
```diff
@@ -0,0 +1,655 @@
...
+func (tr *TransactionRegister) removeTransactionNoLock(txn *MempoolTx) error {
+	if txn == nil || txn.Hash == nil {
+		return fmt.Errorf("TransactionRegister.RemoveTransaction: Transaction or transaction hash is nil")
+	}
+
+	if _, ok := tr.txnMembership[*txn.Hash]; !ok {
+		return nil
+	}
+
+	if tr.totalTxnsSizeBytes < txn.TxSizeBytes {
```
diamondhands0, line 144: A bit of an odd check. How could a transaction we're removing ever exceed the size of the whole transaction register? If it's a sanity-check, document that with a comment just so it's clear the reason why we're checking this.

Also if you want a sanity-check, what might be better is periodically checking that the sum of TxSizeBytes in tr.txnMembership is equal to tr.totalTxnsSizeBytes somehow. But not sure this is worth the performance hit, since you will have to lock it while you check this. Maybe if you architected a "once per minute" check or something.

lib/pos_transaction_register.go
```diff
@@ -0,0 +1,655 @@
...
+		return fmt.Errorf("TransactionRegister.RemoveTransaction: Transaction with transaction hash %v size %v "+
+			"exceeds total mempool size %v", txn.Hash.String(), txn.TxSizeBytes, tr.totalTxnsSizeBytes)
+	}
+
+	// Determine the min fee of the bucket based on the transaction's fee rate.
+	bucketMinFeeNanosPerKb, _ := computeFeeTimeBucketRangeFromFeeNanosPerKB(txn.FeePerKB,
+		tr.minimumNetworkFeeNanosPerKB, tr.feeBucketRateMultiplierBasisPoints)
+	// Remove the transaction from the bucket.
+	if bucket, exists := tr.feeTimeBucketsByMinFeeMap[bucketMinFeeNanosPerKb]; exists {
```
diamondhands0, line 153: If it doesn't exist thats' actually an error it seems like. Maybe check that case and error?

lib/pos_transaction_register.go
```diff
@@ -0,0 +1,655 @@
...
+	if txn == nil || txn.Hash == nil {
+		return
+	}
+
+	if !tb.txnMembership.Includes(*txn.Hash) {
+		return
+	}
+
+	tb.txnsSet.Remove(txn)
+	tb.txnMembership.Remove(*txn.Hash)
```
diamondhands0, line 511: Why do we track txns with two sets? Why not just `map[BlockHash]*MempoolTx`

lib/pos_mempool.go
```diff
@@ -0,0 +1,490 @@
...
+		}
+	}
+
+	// Add the transaction to the mempool and then prune if needed.
+	if err := dmp.addTransactionNoLock(mempoolTx, true); err != nil {
+		return errors.Wrapf(err, "PosMempool.AddTransaction: Problem adding transaction to mempool")
+	}
+
+	// Add the transaction to the nonce tracker.
+	dmp.nonceTracker.AddTxnByPublicKeyNonce(*pk, *txn.TxnNonce, mempoolTx)
```
diamondhands0, line 225: Why isn't the nonceTracker update just a part of removeTransactionNoLock() and addTransactionNoLock() like the other things? Seems like you could do the following:
```
1. Keep the check above where you remove an existing txn if it has a matching nonce. This will remove it from nonceTracker.

2. Add the new txn with addTransactionNoLock, which will add the new txn's nonce to the nonceTracker.
```

lib/pos_mempool.go
```diff
@@ -0,0 +1,490 @@
...
+	// We set the persistToDb flag to false so that persister doesn't try to save the transactions.
+	for _, txn := range txns {
+		if err := dmp.addTransactionNoLock(txn, false); err != nil {
+			glog.Errorf("PosMempool.Start: Problem adding transaction with hash (%v) from persister: %v",
+				txn.Hash, err)
+			continue
+		}
+		// If the transaction was successfully added, also include it in the nonce tracker.
+		pk := NewPublicKey(txn.Tx.PublicKey)
+		dmp.nonceTracker.AddTxnByPublicKeyNonce(*pk, *txn.Tx.TxnNonce, txn)
```
diamondhands0, line 310: If nonceTracker was a part of addTransactionNoLock, you could skip this second portion.

lib/pos_transaction_register.go
```diff
@@ -0,0 +1,655 @@
...
+	if minPrunedBytes == 0 {
+		return nil, nil
+	}
+
+	prunedBytes := uint64(0)
+	prunedTxns := []*MempoolTx{}
+
+	// Find the FeeTime bucket at the end of the Set. It'll have the smallest fee among the buckets in the register.
+	it := tr.feeTimeBucketSet.Iterator()
+	it.End()
```
diamondhands0, line 309: This it.End() thing is a little tricky. Add a comment explaining that you need to do it because you're doing a reverse iteration. Also add the same comment below.

Also why not something more like the following:
```
// We start with a call to it.End() to go *one past* the last item. This is required when iterating backward.
it := tr.feeTimeBucketSet.Iterator() && 
for it.End(); it.Prev(); {

}
```

lib/pos_mempool.go
```diff
@@ -0,0 +1,490 @@
...
+		return
+	}
+
+	dmp.globalParams = globalParams
+	mempoolTxns := dmp.txnRegister.GetFeeTimeTransactions()
+	newRegister := NewTransactionRegister(dmp.globalParams)
+	removedTxnHashes := []string{}
+
+	for _, mempoolTx := range mempoolTxns {
+		if err := newRegister.AddTransaction(mempoolTx); err == nil {
```
diamondhands0, line 473: You forgot to update the nonceTracker properly in this function :)

lib/pos_mempool_persister.go
```diff
@@ -0,0 +1,261 @@
...
+
+	mp.updateBatch = nil
+
+	return nil
+}
+
+// GetPersistedTransactions is used to retrieve all transactions from the database. It will return an error if the persister
+// is not currently running or if there was an issue retrieving the transactions.
+func (mp *MempoolPersister) GetPersistedTransactions() ([]*MempoolTx, error) {
+	mp.Lock()
```
diamondhands0, line 215: Can this be RLock()?

lib/pos_mempool_transaction.go
```diff
@@ -0,0 +1,124 @@
...
+	"time"
+)
+
+// MempoolTx contains a transaction along with additional metadata like the
+// fee and time added.
+type MempoolTx struct {
+	Tx *MsgDeSoTxn
+
+	// TxMeta is the transaction metadata
+	TxMeta *TransactionMetadata
```
diamondhands0, line 16: You're not setting this anywhere and it appears unused.

lib/constants.go
```diff
@@ -699,10 +699,25 @@ type DeSoParams struct {
 	// before they are jailed.
 	DefaultJailInactiveValidatorGracePeriodEpochs uint64
 
+	// DefaultFeeBucketRateMultiplierBasisPoints is the rate of growth of the fee bucket ranges. The multiplier is given
+	// as basis points. For example a value of 1000 means that the fee bucket ranges will grow by 10% each time.
+	DefaultFeeBucketRateMultiplierBasisPoints uint64
+
+	// DefaultFailingTransactionBMFRateBasisPoints is the default rate of the failing transaction fee, in basis points,
+	// that is used to in BMF calculations. E.g. a value of 2500 means that 25% of the failing transaction's fee is used
```
diamondhands0, line 707: Fix typos

lib/constants.go
```diff
@@ -699,10 +699,25 @@ type DeSoParams struct {
...
+	// in BMF calculations.
+	DefaultFailingTransactionBMFRateBasisPoints uint64
+
 	ForkHeights ForkHeights
 
 	EncoderMigrationHeights     *EncoderMigrationHeights
 	EncoderMigrationHeightsList []*MigrationHeight
+
+	// The maximum aggregate number of bytes of transactions included in the PoS mempool.
+	MaxMempoolPosSizeBytes uint64
```
diamondhands0, line 717: Part of me wants to start this off as a flag, but I think it's fine to leave it as part of the constants for now. Let's just keep in mind that this is a requirement we'll be foisting on all node operators.

diamondhands0: On second thought let's start this off as a flag so that we can tune it. Once we feel the value we've selected should apply across the network we can hard fork it into a param.

lib/constants.go
```diff
@@ -699,10 +699,25 @@ type DeSoParams struct {
...
+
+	// MempoolBackupTimeMilliseconds is the frequency with which pos mempool persists transactions to storage.
+	MempoolBackupTimeMilliseconds uint64
```
diamondhands0, line 720: This should definitely be a flag not a ParamUpdater param.

lib/constants.go
```diff
@@ -1103,9 +1118,18 @@ var DeSoMainnetParams = DeSoParams{
...
+	DefaultFeeBucketRateMultiplierBasisPoints: uint64(1000),
+
+	// The rate of the failing transaction's fee used in BMF calculations.
+	DefaultFailingTransactionBMFRateBasisPoints: uint64(2500),
+
 	ForkHeights:                 MainnetForkHeights,
 	EncoderMigrationHeights:     GetEncoderMigrationHeights(&MainnetForkHeights),
 	EncoderMigrationHeightsList: GetEncoderMigrationHeightsList(&MainnetForkHeights),
+
+	MaxMempoolPosSizeBytes:        3 << 30, // 3Gb
```
diamondhands0, line 1131: Let's make these flags. See previous comment.

lib/db_entry.go
```diff
@@ -0,0 +1,75 @@
...
+
+import "bytes"
+
+// -------------------------------------------------------------------------------------
+// DBEntry
+// -------------------------------------------------------------------------------------
+
+// DBEntry is used to represent a database record. It's more convenient than passing
+// <key, value> everywhere.
+type DBEntry struct {
```
diamondhands0, line 11: Didn't we have this defined somewhere already like for Hypersync or something? I swear I've seen it before.

lib/mempool.go
```diff
@@ -61,38 +61,6 @@ var (
 	LowFeeTxLimitBytesPerTenMinutes = 150000 // Allow 150KB per minute in low-fee txns.
 )
```
diamondhands0, line 62: Maybe we should rename this file to legacy_mempool.go or something so it's clear that it will all be deleted soon.

lib/network.go
```diff
@@ -3809,6 +3856,30 @@ func (msg *MsgDeSoTxn) UnmarshalJSON(data []byte) error {
 	return nil
 }
```
diamondhands0, line 3858: Should this be in the mempool file?

lib/network.go
```diff
@@ -1914,6 +1915,14 @@ type MsgDeSoHeader struct {
 	// event that ASICs become powerful enough to have birthday problems in the future.
 	ExtraNonce uint64
 
+	// TransactionsConnectStatus is only used for Proof of Stake blocks, starting with
+	// MsgDeSoHeader version 2. For all earlier versions, this field will default to nil.
+	//
+	// The hash of the TxnConnectStatusByIndex field in MsgDeSoBlock. It is stored to ensure
+	// that the TxnConnectStatusByIndex is part of the header hash, which is signed by the
+	// proposer. The full index is stored in the block to offload space complexity.
+	TxnConnectStatusByIndexHash *BlockHash
```
diamondhands0, line 1924: I get what you're trying to do here. But could we maybe integrate it into the existing TransactionMerkleRoot somehow? What I mean is maybe we can just change how we compute TransactionMerkleRoot when we cut over to PoS so that it includes a hash of the bitset you've added. It's kindof the point of the TransactionMerkleRoot root that it "rolls up" all the hashes you care about, which seems to include this one.

Maybe you can add bitsetHash as an optional argument to the function that computes the TransactionMerkleRoot. If you have a bitset, then it adds its hash to the merkle. Otherwise, it's the "old way" of doing the TransactionMerkleRoot.

lib/pos_block_producer.go
```diff
@@ -0,0 +1,145 @@
+package lib
```
diamondhands0, line 1: I was told to hold off on reviewing this file so going to wait.

lib/pos_mempool_manager.go
```diff
@@ -0,0 +1,50 @@
+package lib
```
diamondhands0, line 1: Holding off on reviewing this file since it relies on the block producer.

lib/pos_mempool.go
AeonSw4n: Hm, I disagree with this one. I think embeddings make sense when they simplify the syntax and meaningfully augment the functionality to the type. In the case of mutex, imo embedding is the preferred path, since Lock/Unlock functionality is common enough to meaningfully augment PosMempool type.

In terms of the prevelance of the embedding pattern, I actually started using the mutex embedding after seeing it in a bunch of blockchain repos. So I think the readers of our code will be comfortable with this embedding.

lib/pos_mempool.go
AeonSw4n: I'm a little worried to take it out. It's doing some sanity-checking to transaction inputs that seem not entirely obsolete. I think it's a "safe bet" to keep it.

lib/validation_deso_txn.go
AeonSw4n: This minimum fee is also for backwards-compatibility. It's not meant to invalidate any meaningful transactions, rather set a really small lowerbound for the possible transaction fees. Many things are easier when the fee rate can't be 0, and having no minimal fee forces us to gate against this edge-case elsewhere. For instance Fee buckets won't work without non-zero starting fee. If we set the bucket starting fee to 1 nano, then it'll never grow, so it needs to be at least 1/growthRate, otherwise the growth will be <1 which may not round up.

The minimum fee in question would be like 1000 nanos, really small and noone would ever hit it. I think it's easiest to keep this one as it is.

lib/pos_transaction_register.go
AeonSw4n: Ah yes this is a sanity check, comment added. It should never get triggered UNLESS somehow the referenced transaction changes. But we assume the MempoolTx type to be indempotent, so this sanity-check is redundant.

In terms of diagnostics, I think this would be handled by the struct that manages the mempool. I'm thinking a `mempool_manager.go` file would house this struct, and be responsible for giving other compoenents access to the mempool as well as doing things like diagnostics. The mempool_manager will run its own goroutine loop. 

I already started working on the architecture design enabling such mempool_manager struct to exist. You'll see it soon :)

lib/pos_transaction_register.go
AeonSw4n: We need the `txnsSet` for Fee-Time ordering of transactions. When it comes to `txnMembership`, I actually started with map but we decied to use the Set during review. The map and the Set are very similar. In fact Set is just a wrapper around a map. I think it's slightly preferable to use the Set since it comes with more functionality.

lib/pos_transaction_register.go
AeonSw4n: Added the comment. Eh, I tried using the iterator inline in the "for" but it end up looking more complicated.

lib/pos_mempool_transaction.go
AeonSw4n: Backwards-compatibility 😢

lib/constants.go
AeonSw4n: It should still be in DeSo params though, no? Just not in GlobalParams.

lib/constants.go
AeonSw4n: Yup, it's a flag now. It's still in DeSoParams tho, as opposed to being passed as part of NewMempool() signature. It seems cleaner this way

lib/db_entry.go
AeonSw4n: Haha it was, I just moved it around.

lib/mempool.go
AeonSw4n: Hah I'm down, purge! File renamed

lib/network.go
AeonSw4n: This should be deleted lol. It's unused. Better to do it in a separate PR tho.

lib/network.go
AeonSw4n: We discussed this with @tholonious and it felt like modifying the existing MerkleRoot utils to cater to hashing the new index will be tricky from the standpoint of backwards-compatibility and separation of concerns. 

A separate hash in the header adds just 32 bytes, which seems like an acceptable downside.

lib/pos_block_producer.go
AeonSw4n: The time has come

lib/pos_mempool_manager.go
AeonSw4n: The time has come

lib/validation_deso_txn.go
tholonious: To simplify the exponential fee bucketing math, we decided some time back to have a non-zero minimal network fee. In practice, we'll set the minimum fee to 1 nano so that the fee bucketing doesn't need any addition edge case handling to work with a near-zero starting fee. Given we can have a 1 nano min fee, I don't believe a zero fee provides much more value.

Marking this as resolved based on the above.

lib/network.go
tholonious: Just adding here that we resolved this after the review in favor of sticking with the current field. We have a clear upgrade path for merging this into the merkle root when we introduce a new header version.

tholonious:
Closing PR since comments have been addressed in https://github.com/deso-protocol/core/pull/578

PR #698, opened 2023-09-27
Generate random bytes for test block hash

tholonious:


tholonious:
* feature/proof-of-stake
    * **PR #698** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/698" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/698?utm_source=stack-comment).

consensus/utils_test.go
```diff
@@ -390,12 +391,16 @@ func createDummyBLSPrivateKey() *bls.PrivateKey {
...
-		},
+		value: byteArray,
 	}
 }
+
+func generateRandomBytes(numBytes int) []byte {
+	randomBytes := make([]byte, numBytes)
+	rand.Read(randomBytes)
+	return randomBytes
+}
```
AeonSw4n, line 406: I personally don't like using unseeded randomness because you can't repro. What I found more efficient is picking a random number and initializing rand like this:
```
	seed := int64(993)
	rand := rand.New(rand.NewSource(seed))
```
This way if you have an error it'll be easier to debug. Also it's convenient to define randomness in the parent test and pass it down when needed. So if you really want to stress-test your code you could iterate through seeds on the parent test.

AeonSw4n (approved):

consensus/utils_test.go
tholonious: Not a priority here given how simple the usage of createBlockHash will be for tests. What I need here is a dead simple self-contained function that returns pseudo-random block hashs without repeating values. Reproducibility of the generated value across test executions isn't important.

PR #705, opened 2023-10-02
Validate Block Integrity

lazynina:


lazynina:
* feature/proof-of-stake
    * **PR #686** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/686" target="_blank">Graphite</a> 
      * **PR #705** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/705" target="_blank">Graphite</a>  👈
        * **PR #706** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/706" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/705?utm_source=stack-comment).

lib/pos_blockchain.go
```diff
@@ -134,8 +135,48 @@ func (bc *Blockchain) validateDeSoBlockPoS(desoBlock *MsgDeSoBlock) error {
...
-	return errors.New("IMPLEMENT ME")
+	// First make sure we have a non-nil header
+	if desoBlock.Header == nil {
+		return RuleErrorNilBlockHeader
+	}
+
+	// Timestamp validation
+	// What's the criteria? Is it just that it's not in the future?
+	// And that it's not too far in the past? Too far being older than the current best chain tip?
+	if desoBlock.Header.TstampNanoSecs < bc.GetBestChainTip().Header.TstampNanoSecs {
```
tholonious, line 146: A block can't have a timestamp that precedes its parent's timestamp. The parent may or may not be the current chain tip, so you need to explicitly look up the parent here.

If the parent doesn't exist, then the validation should fail here.
```suggestion
	if desoBlock.Header.TstampNanoSecs < bc.getParentBlock[desoBlock.Header.GetHash())] {
```

lib/pos_blockchain.go
```diff
@@ -134,8 +135,48 @@ func (bc *Blockchain) validateDeSoBlockPoS(desoBlock *MsgDeSoBlock) error {
 // that the timestamp is valid, that the version of the header is valid,
 // and other general integrity checks (such as not malformed).
 func (bc *Blockchain) validateBlockGeneral(desoBlock *MsgDeSoBlock) error {
```
tholonious, line 137: Maybe rename this `validateBlockIntegrity`? If I understand correctly, this function is intended to validate that hte block is proper formed. Seems like this name suggestion could be more representative of what it does.

This isn't a strong opinion fwiw

lib/pos_blockchain.go
```diff
@@ -134,8 +135,48 @@ func (bc *Blockchain) validateDeSoBlockPoS(desoBlock *MsgDeSoBlock) error {
...
+		return RuleErrorNilBlockHeader
+	}
+
+	// Timestamp validation
+	// What's the criteria? Is it just that it's not in the future?
+	// And that it's not too far in the past? Too far being older than the current best chain tip?
+	if desoBlock.Header.TstampNanoSecs < bc.GetBestChainTip().Header.TstampNanoSecs {
+		return RuleErrorPoSBlockTstampNanoSecsTooOld
+	}
+	if desoBlock.Header.TstampNanoSecs > uint64(time.Now().UnixNano()) {
```
tholonious, line 149: The current block can have a timestamp that's up to 10 minutes beyond the node's current timestamp. The "10 minute" threshold may be either configured DeSo param or an on-chain GlobalParam. Up to you which one to go with. Whatever is simpler seems fine.

```suggestion
	if desoBlock.Header.TstampNanoSecs > uint64(time.Now().UnixNano().Add(10 * time.Minute)) {
```

lib/pos_blockchain.go
```diff
@@ -134,8 +135,48 @@ func (bc *Blockchain) validateDeSoBlockPoS(desoBlock *MsgDeSoBlock) error {
...
+
+	if !isVoteQCEmpty && len(desoBlock.Txns) == 0 {
+		return RuleErrorVoteQCWithoutTransactions
+	}
+
+	if !isTimeoutQCEmpty && len(desoBlock.Txns) != 0 {
+		return RuleErrorTimeoutQCWithTransactions
+	}
+
+	// TODO: What other checks do we need to do here?
```
tholonious, line 178: Some other validations that come to mind:
- Validate the txn merkle root
- Validate that the proposer's keys aren't nil/empty
- Validate that the proposer's random seed hash is properly formed
- Validate that the QCs are properly formed as well

tholonious (changes requested):

lib/pos_blockchain.go
lazynina: I'll add it to the deso params and leave a TODO that we may want to move to global params for now.

consensus/event_loop_test.go
```diff
@@ -691,7 +691,7 @@ func TestVoteQCConstructionSignal(t *testing.T) {
 		select {
 		case blockConstructionSignal = <-fc.Events:
 			// Do nothing
-		case <-time.After(10 * time.Millisecond):
+		case <-time.After(100 * time.Millisecond):
```
tholonious, line 694: This change is part of a PR lower down in the stack. Remember to run `gt rs --restack` locally and resubmit this PR before merging

lib/block_view_types.go
```diff
@@ -3796,6 +3796,9 @@ type GlobalParamsEntry struct {
 	// allow a validator to be inactive for (neither voting nor proposing
 	// blocks) before they are jailed.
 	JailInactiveValidatorGracePeriodEpochs uint64
+
+	//
+
```
tholonious, line 3801:
```suggestion
```

lib/pos_blockchain.go
```diff
@@ -130,12 +144,97 @@ func (bc *Blockchain) validateDeSoBlockPoS(desoBlock *MsgDeSoBlock) error {
...
+	}
+
+	if !isTimeoutQCEmpty && !isVoteQCEmpty {
+		return RuleErrorBothTimeoutAndVoteQC
+	}
+
+	if !isVoteQCEmpty && len(desoBlock.Txns) == 0 {
+		return RuleErrorVoteQCWithoutTransactions
+	}
+
```
tholonious, line 198: Seems like it's fine if we have a regular empty block. For example, if network activity is low and there are no txns, then the block proposer will still need to construct blocks at its regular cadence, but with no txns.
```suggestion
```

lib/pos_blockchain.go
```diff
@@ -130,12 +144,97 @@ func (bc *Blockchain) validateDeSoBlockPoS(desoBlock *MsgDeSoBlock) error {
...
+		}
+		if !merkleRoot.IsEqual(computedMerkleRoot) {
+			return RuleErrorInvalidMerkleRoot
+		}
+	} else {
+		// TODO: do we want to ensure a nil merkle root if we have a timeout QC?
+		if merkleRoot != nil {
+			return RuleErrorTimeoutQCWithMerkleRoot
+		}
+	}
```
tholonious, line 234: Should this explicitly check if the block has one or more txns? See my comment above on why it may be fine to have a block with 0 txns.

tholonious (changes requested):

tholonious (approved):

PR #708, opened 2023-10-02
ln/validate-ancestors-exist

lazynina:

PR #707, opened 2023-10-02
implement add block to block index and get uncommitted tip view

lazynina:

PR #706, opened 2023-10-02
add validateBlockHeight function and tests

lazynina:

PR #713, opened 2023-10-03
ln/validate-block-view-pos

lazynina:


lazynina:
* feature/proof-of-stake
    * **PR #712** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/712" target="_blank">Graphite</a> 
      * **PR #705** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/705" target="_blank">Graphite</a> 
        * **PR #706** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/706" target="_blank">Graphite</a> 
          * **PR #707** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/707" target="_blank">Graphite</a> 
            * **PR #708** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/708" target="_blank">Graphite</a> 
              * **PR #713** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/713" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/713?utm_source=stack-comment).

lib/pos_blockchain.go
```diff
@@ -202,8 +202,33 @@ func (bc *Blockchain) validateBlockHeight(desoBlock *MsgDeSoBlock) error {
...
+			return RuleErrorPoSBlockViewEarlierThanCommittedBlock
+		}
+		return nil
+	}
+	// If the chain tip is uncommitted, we need to check that the view is greater than the
+	// uncommitted block's view.
+	if desoBlock.Header.ProposedInView <= chainTip.Header.ProposedInView {
+		return RuleErrorPoSBlockViewEarlierThanUncommittedBlock
+	}
+	return nil
```
tholonious, line 231: I think you need to compare the block's view against its parent's view. If the block has a regular QC, the view needs to be the parent's view + 1. If it's a timeout QC, then the view must be strictly greater than the parent's view + 1.

The block may be part of a fork that doesn't extend from the uncommitted chain tip, so we need to validate the view somehow in that scenario. LMK if I just misunderstood this PR and you plan on doing this separately elsewhere.

lib/pos_blockchain.go
```diff
@@ -262,15 +262,44 @@ func (bc *Blockchain) validateBlockHeight(desoBlock *MsgDeSoBlock) error {
...
+	committedBlock, _ := bc.getHighestCommittedBlock()
+	if committedBlock == nil {
+		// This is an edge case we'll never hit in practice since all the PoW blocks
+		// are committed.
+		return errors.New("validateBlockView: No committed blocks found")
+	}
+	// Validate that the view is greater than the latest committed block view.
+	if desoBlock.Header.ProposedInView <= committedBlock.Header.ProposedInView {
+		return RuleErrorPoSBlockViewEarlierThanCommittedBlock
+	}
```
tholonious, line 281: What's the intended purpose of comparing the block's view against the committed tip's? It's a bit redundant since we'll later on need to validate the block's view against the node's current view, which is a cheaper validation.

lib/pos_blockchain.go
```diff
@@ -262,15 +262,44 @@ func (bc *Blockchain) validateBlockHeight(desoBlock *MsgDeSoBlock) error {
...
+		if desoBlock.Header.ProposedInView != parentBlock.Header.ProposedInView+1 {
+			return RuleErrorPoSVoteBlockViewNotOneGreaterThanParent
+		}
+	} else {
+		// If our current block has a timeout QC, then we need to validate that the
+		// view is strictly greater than the latest uncommitted block's view.
+		if desoBlock.Header.ProposedInView <= parentBlock.Header.ProposedInView {
+			return RuleErrorPoSTimeoutBlockViewNotGreaterThanParent
+		}
+	}
```
tholonious, line 301: These are great sanity checks for the blocks view. But they're not the full set of checks we need to perform on the view later down in the ProcessBlockPoS flow. Later on, you'll need to compare the block's view against the node's current view, and to perform the safeVote predicate (https://github.com/deso-protocol/hotstuff_pseudocode/blob/main/fast_hotstuff_bls.go#L430).

IMO, it'll be cleaner to move this code into the `validateBlockIntegrity` function so it's clear these are just sanity checks.

tholonious (changes requested):

lib/pos_blockchain.go
lazynina: Nixed this validation.

lib/pos_blockchain.go
lazynina: Will introduce the rest of those validations in a separate PR. All these validation functions are called from `validateDeSoBlockPoS` and wanted to keep each validation function a reasonable size. We can refactor later if we want and rejigger the tests, but I'd prefer to keep them broken up for now.

lib/pos_blockchain.go
```diff
@@ -257,15 +257,34 @@ func (bc *Blockchain) validateBlockHeight(desoBlock *MsgDeSoBlock) error {
...
+	// If our current block has a vote QC, then we need to validate that the
+	// view is exactly one greater than the latest uncommitted block.
+	if desoBlock.Header.ValidatorsTimeoutAggregateQC.isEmpty() {
+		if desoBlock.Header.ProposedInView != parentBlock.Header.ProposedInView+1 {
+			return RuleErrorPoSVoteBlockViewNotOneGreaterThanParent
+		}
+	} else {
+		// If our current block has a timeout QC, then we need to validate that the
+		// view is strictly greater than the latest uncommitted block's view.
```
tholonious, line 282:
```suggestion
		// view is strictly greater than the parent block's view
```

lib/pos_blockchain.go
```diff
@@ -257,15 +257,34 @@ func (bc *Blockchain) validateBlockHeight(desoBlock *MsgDeSoBlock) error {
...
-	return errors.New("IMPLEMENT ME")
+	// Validate that the view is greater than the latest uncommitted block.
+	parentBlock, exists := bc.blockIndex[*desoBlock.Header.PrevBlockHash]
+	if !exists {
+		// Note: this should never happen as we only call this function after
+		// we've validated that all ancestors exist in the block index.
+		return RuleErrorMissingParentBlock
+	}
+	// If our current block has a vote QC, then we need to validate that the
+	// view is exactly one greater than the latest uncommitted block.
```
tholonious, line 275:
```suggestion
	// view is exactly one greater than the parent's block
```

tholonious (approved):

PR #724, opened 2023-10-11
Use Explicit Flag To Track When Crank Timer Has Elapsed

tholonious:

PR #725, opened 2023-10-11
Add Byte Comparison Utility for BlockHash type

tholonious:


tholonious:
* feature/proof-of-stake
    * **PR #724** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/724" target="_blank">Graphite</a> 
      * **PR #725** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/725" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/725?utm_source=stack-comment).

consensus/utils.go
```diff
@@ -252,3 +253,10 @@ func isSuperMajorityStake(stake *uint256.Int, totalStake *uint256.Int) bool {
 	// Check if 3Cq - 2N - 1 >= 0
 	return superMajorityConditionSum.Sign() >= 0
 }
+
+func isEqualBlockHashes(hash1 BlockHash, hash2 BlockHash) bool {
```
lazynina, line 257: we use `.Eq` for the lib implementation of block hash - would nice if they were the same. Not a deal breaker though.

tholonious: We can't really reuse the .Eq function from the lib package here. The lib package needs to import the consensus package, and not the other way around. We could alternatively update hte .Eq function in lib to use this function. I can return to it later on when renaming these interface types and fields during cleanup later.

lazynina (approved):

PR #729, opened 2023-10-11
add ValidateBlockLeader logic

lazynina:

PR #728, opened 2023-10-11
ln/add-seek-over-leader-schedule

lazynina:

PR #727, opened 2023-10-11
add block to best chain

lazynina:

PR #730, opened 2023-10-11
ln/validate-ancestors-exist

lazynina:


lazynina:
* feature/pos-process-block
        * **PR #730** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/730" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/730?utm_source=stack-comment).

lib/pos_blockchain.go
```diff
@@ -337,15 +333,29 @@ func (bc *Blockchain) validateTimeoutQC(desoBlock *MsgDeSoBlock, validatorSet []
...
+func (bc *Blockchain) validateAncestorsExist(parentHash BlockHash) *BlockHash {
 	// Notes: starting from the block passed in, we'll look for the parent in the block index.
 	// 1. If the parent does not appear in the block index, we'll return the parent's hash.
 	// 2. If the parent exists in the block index AND is in the best chain, we can safely assume
 	//    that all ancestors exist in the block index.
 	// 3. If the parent exists in the block index but is not in the best chain, we repeat from
 	//    step 1 with the parent as the block passed in.
-	// TODO: Implement me
-	return nil, errors.New("IMPLEMENT ME")
+
```
lazynina, line 343: Do we need to check orphans?

lib/pos_blockchain.go
```diff
@@ -337,15 +333,29 @@ func (bc *Blockchain) validateTimeoutQC(desoBlock *MsgDeSoBlock, validatorSet []
...
+		return &parentHash
+	}
+
+	// if parent exists in block index AND is in best chain, return nil. We can
+	// safely assume that all ancestor exist then.
+	_, bestChainExists := bc.bestChainMap[parentHash]
+	if bestChainExists {
+		return nil
+	}
+	return bc.validateAncestorsExist(*parentBlockNode.Header.PrevBlockHash)
```
tholonious, line 358: Why do we check if the parent is in the best chain? The parent's existence in the block index would guarantee that all of the block's ancestors are known, no?

lazynina: I think I wrote this before we concretified the rules around the block index. I'll update this since we won't be putting anything in the block index unless we have its parent. If it's parent isn't in the block index, it's an orphan and we can figure out how to proceed those in a separate PR

lib/pos_blockchain.go
```diff
@@ -363,18 +360,15 @@ func (bc *Blockchain) validateTimeoutQC(desoBlock *MsgDeSoBlock, validatorSet []
...
-	return nil, errors.New("IMPLEMENT ME")
+// validateAncestorsExist checks the parent exists in the block index. If it does not,
+// return the parent's hash. Else, return nil.
+func (bc *Blockchain) validateAncestorsExist(parentHash BlockHash) *BlockHash {
+	// check for parent in block index
+	// if parent does not exist in block index, return parent's hash
+	if _, exists := bc.blockIndex[parentHash]; !exists {
+		return &parentHash
+	}
+	return nil
```
tholonious, line 371: Just a nit. It's a bit strange to return the input block hash if it doesn't exist. You may as well make this function return a bool that signifies whether the the ancestors exist or not. This way, the naming and intended use of the function become dead simple.

```suggestion
// hasKnownAncestors checks the parent exists in the block index. It returns true if all ancestors
// of the provided block hash exist in the block index. It returns false otherwise.
func (bc *Blockchain) hasKnownAncestors(blockHash BlockHash) bool {
        _, exists := bc.blockIndex[blockHash]
        return exists
```

tholonious (changes requested):

lib/pos_blockchain.go
tholonious: I don't think so. Just checking if the block's parent exists in the block index is sufficient. But you can always iterate if needed

tholonious (approved):

PR #732, opened 2023-10-12
Validate QCs

lazynina:


lazynina:
* feature/pos-process-block
        * **PR #730** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/730" target="_blank">Graphite</a> 
          * **PR #732** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/732" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/732?utm_source=stack-comment).

consensus/utils.go
```diff
@@ -20,19 +20,53 @@ func IsValidSuperMajorityQuorumCertificate(qc QuorumCertificate, validators []Va
...
+		aggQC.GetAggregatedSignature().GetSignature(),
+		signedPayloads,
+	)
+	if err != nil || !isValidSignature {
+		return false
+	}
+	return true
+}
+
+func checkSuperMajorityStakeForAggregatedSignature(aggSig AggregatedSignature, validators []Validator) (bool, []*bls.PublicKey) {
```
tholonious, line 59: Simplify this to just pass in the signers list and validator set. This way it's clearer from the function signature alone exactly what the function does
```suggestion
func isSuperMajorityStakeSignersList(signersList *bitset.Bitset, validators []Validator) (bool, []*bls.PublicKey) {
```

consensus/utils.go
```diff
@@ -20,19 +20,53 @@ func IsValidSuperMajorityQuorumCertificate(qc QuorumCertificate, validators []Va
 	// Compute the signature that validators in the QC would have signed
 	signaturePayload := GetVoteSignaturePayload(qc.GetView(), qc.GetBlockHash())
 
+	isValidSuperMajority, validatorPublicKeysInQC := checkSuperMajorityStakeForAggregatedSignature(qc.GetAggregatedSignature(), validators)
```
tholonious, line 23: nit: the result is always valid. The bool just tells us whether or not there is a supermajority
```suggestion
	hasSuperMajorityStake, signerPublicKeys := checkSuperMajorityStakeForAggregatedSignature(qc.GetAggregatedSignature(), validators)
```

consensus/utils.go
```diff
@@ -20,19 +20,53 @@ func IsValidSuperMajorityQuorumCertificate(qc QuorumCertificate, validators []Va
...
+
+	return isValidSignatureManyPublicKeys(validatorPublicKeysInQC, qc.GetAggregatedSignature().GetSignature(), signaturePayload[:])
+}
+
+func IsValidSuperMajorityAggregateQuorumCertificate(aggQC AggregateQuorumCertificate, validators []Validator) bool {
+	if !isProperlyFormedAggregateQC(aggQC) || !isProperlyFormedValidatorSet(validators) {
+		return false
+	}
+
+	isValidSuperMajority, signerPublicKeys := checkSuperMajorityStakeForAggregatedSignature(aggQC.GetAggregatedSignature(), validators)
```
tholonious, line 36:
```suggestion
	hasSuperMajorityStake, signerPublicKeys := checkSuperMajorityStakeForAggregatedSignature(aggQC.GetAggregatedSignature(), validators)
```

consensus/utils.go
```diff
@@ -20,19 +20,53 @@ func IsValidSuperMajorityQuorumCertificate(qc QuorumCertificate, validators []Va
...
+	if !isValidSuperMajority {
+		return false
+	}
+
+	signedPayloads := [][]byte{}
```
tholonious, line 41: Can you add a comment here on why we do the payload aggregation this way? There's a nice comment in pos_network.go that you can refer to on how the highQC views are stored and how the payloads are signed: https://github.com/deso-protocol/core/blob/feature/proof-of-stake/lib/pos_network.go#L481

lib/pos_blockchain.go
```diff
@@ -472,6 +488,48 @@ func (bc *Blockchain) GetUncommittedTipView() (*UtxoView, error) {
 	return utxoView, nil
 }
 
+// BuildUtxoViewToBlockParent builds a UtxoView to the block provided. It does this by
+// identifying all uncommitted ancestors of this block and then connecting those blocks.
+func (bc *Blockchain) BuildUtxoViewToBlockParent(parentHash BlockHash) (*UtxoView, error) {
```
tholonious, line 493: nit: this function shouldn't internally care about why it's being called. If we define this function as simple building a UtxoView given a blockHash, then it can become a general purpose function that we can reuse. Ex: you can reuse it to do the heavy lifting for `bc.getUncommittedTipView()`
```suggestion
func (bc *Blockchain) buildUtxoViewForBlock(blockHash BlockHash) (*UtxoView, error) {
```

tholonious, line 493: As an alternative name, wdyt of this? In other places, we've started to transition the naming for "UtxoView" to "BlockView". I have no strong opinion here

```suggestion
func (bc *Blockchain) buildBlockView(blockHash BlockHash) (*UtxoView, error) {
```

lib/pos_blockchain.go
```diff
@@ -472,6 +488,48 @@ func (bc *Blockchain) GetUncommittedTipView() (*UtxoView, error) {
...
+	uncommittedAncestors := []*BlockNode{}
+	currentBlock := bc.blockIndex[parentHash]
+	if currentBlock == nil {
+		return nil, errors.Errorf("BuildUtxoViewToBlockParent: Block %v not found in block index", parentHash)
+	}
+	for currentBlock.CommittedStatus == UNCOMMITTED {
```
tholonious, line 499: What if the input block is committed but it is not the committed tip? We wouldn't be able to build a UtxoView for it. It'll be safest to perform that check here and error in that case. This is definitely one function I think is worth bullet proofing, by assuming that the caller may unknowingly pass in a bad input

tholonious (changes requested):

lib/pos_blockchain.go
lazynina: Block view is a little confusing to me - it could sometimes refer to `block.Header.ProposedInView`. So for now, I think it's cleanest to still refer to these as `UtxoView` structs for now

lib/pos_blockchain.go
```diff
@@ -348,8 +355,17 @@ func (bc *Blockchain) validateBlockLeader(desoBlock *MsgDeSoBlock) error {
 // validateQC validates that the QC of this block is valid, meaning a super majority
 // of the validator set has voted (or timed out). Assumes ValidatorEntry list is sorted.
 func (bc *Blockchain) validateQC(desoBlock *MsgDeSoBlock, validatorSet []*ValidatorEntry) error {
```
tholonious, line 357: nit: rename boolean functions to use an `isXYZ` or `hasXYZ` naming pattern. It's also good to pass in the bare minimum data needed to perform the validation. In this case, we only need to pass in the QC and validator set. All good little things that make the function more self-documenting based on the signature alone.
```suggestion
func (bc *Blockchain) isValidQC(qc *QuorumCertificate, validatorSet []*ValidatorEntry) error {
```

tholonious (approved):

lazynina:
### Merge activity

* **Oct 17, 7:42 PM**: @lazynina started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/732).

PR #736, opened 2023-10-12
Implement should reorg

lazynina:

PR #735, opened 2023-10-12
Update error formatting in pos process block

lazynina:

PR #734, opened 2023-10-12
ln/check-parent-block-committed-status-in-integrity-checks

lazynina:

PR #733, opened 2023-10-12
Handle uncaught errors in fromBytes functions in pos_network.go

lazynina:

PR #737, opened 2023-10-13
Handle reorg

lazynina:


lazynina:
* feature/pos-process-block
        * **PR #730** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/730" target="_blank">Graphite</a> 
          * **PR #732** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/732" target="_blank">Graphite</a> 
            * **PR #733** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/733" target="_blank">Graphite</a> 
              * **PR #734** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/734" target="_blank">Graphite</a> 
                * **PR #735** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/735" target="_blank">Graphite</a> 
                  * **PR #736** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/736" target="_blank">Graphite</a> 
                    * **PR #737** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/737" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/737?utm_source=stack-comment).

lib/pos_blockchain.go
```diff
@@ -409,12 +409,42 @@ func (bc *Blockchain) shouldReorg(desoBlock *MsgDeSoBlock, currentView uint64) b
...
 // handleReorg handles a reorg to the block provided. It does not check whether or not we should
 // perform a reorg, so this should be called after shouldReorg. It will do the following:
-// 1. Update the bestChain and bestChainMap by removing blocks that are not uncommitted ancestor of this block.
+// 1. Update the bestChain and bestChainMap by removing blocks that are uncommitted and are not ancestors of this block.
 // 2. Update the bestChain and bestChainMap by adding blocks that are uncommitted ancestors of this block.
 // Note: addBlockToBestChain will be called after this to handle adding THIS block to the best chain.
 func (bc *Blockchain) handleReorg(desoBlock *MsgDeSoBlock) error {
-	// TODO: Implement me.
-	return errors.New("IMPLEMENT ME")
+	// For simplicity, we remove all uncommitted blocks and then re-add them.
```
tholonious, line 416:
```suggestion
	// Check if we should perform a reorg here
	if !bc.shouldReorg(desoBlock, currentView) {
		return false, nil
	}
	
	// For simplicity, we remove all uncommitted blocks and then re-add them.
```

lib/pos_blockchain.go
```diff
@@ -409,12 +409,42 @@ func (bc *Blockchain) shouldReorg(desoBlock *MsgDeSoBlock, currentView uint64) b
...
+	isCommittedTip := false
+	currentBlock := bc.blockIndex[*desoBlock.Header.PrevBlockHash]
+	for !isCommittedTip {
+		if currentBlock.Hash.IsEqual(highestCommittedBlock.Hash) {
+			isCommittedTip = true
+			break
+		}
+		ancestors = append(ancestors, currentBlock)
+		currentBlock = bc.blockIndex[*currentBlock.Header.PrevBlockHash]
+	}
```
tholonious, line 443: Just to make the reorg algorithm a bit safe, can you find the list of ordered block nodes that connect to the committed tip *before* removing blocks from the bestChain above? 

This way, if the lookup fails for any reason, the function can return without any side effects to the blockchain data structure

lib/pos_blockchain.go
```diff
@@ -409,12 +409,42 @@ func (bc *Blockchain) shouldReorg(desoBlock *MsgDeSoBlock, currentView uint64) b
...
+		delete(bc.bestChainMap, *bc.bestChain[ii].Hash)
+	}
+	// Shorten best chain back to committed tip.
+	bc.bestChain = bc.bestChain[:idx+1]
+
+	// Find the ordered list of block nodes that connects the incoming block to the committed tip.
+	ancestors := []*BlockNode{}
+	isCommittedTip := false
+	currentBlock := bc.blockIndex[*desoBlock.Header.PrevBlockHash]
+	for !isCommittedTip {
```
tholonious, line 436: I know you confirm that the incoming block is a descendant of the committed tip outside of this function. But the assumption that we'll reach the committed tip is a bit scary. Maybe add a TODO here for we can revisit this later so we can reduce the risk of this assumption.

The risk here is that, if the assumption ever breaks, then this can loop may end up iterating through the entire blockchain.

tholonious (changes requested):

tholonious (approved):

lib/pos_blockchain.go
```diff
@@ -408,20 +406,50 @@ func (bc *Blockchain) addBlockToBlockIndex(desoBlock *MsgDeSoBlock) error {
...
+	if !bc.shouldReorg(desoBlock, currentView) {
+		return false, nil
+	}
+	// For simplicity, we remove all uncommitted blocks and then re-add them.
+	highestCommittedBlock, idx := bc.getHighestCommittedBlock()
+	if highestCommittedBlock == nil {
+		// This is an edge case we'll never hit in practice since all the PoW blocks
+		// are committed.
+		return false, errors.New("tryReorgToNewTip: No committed blocks found")
+	}
```
tholonious, line 438: We don't need this check anymore. This check was only needed when this function previously had a loop that traced lineage back to the committed tip.

lib/pos_blockchain.go
```diff
@@ -408,20 +406,50 @@ func (bc *Blockchain) addBlockToBlockIndex(desoBlock *MsgDeSoBlock) error {
...
+	for ii := idx + 1; ii < len(bc.bestChain); ii++ {
+		delete(bc.bestChainMap, *bc.bestChain[ii].Hash)
+	}
+	// Shorten best chain back to committed tip.
+	bc.bestChain = bc.bestChain[:idx+1]
+
+	for ii := len(ancestorsToCommittedTip) - 1; ii >= 0; ii-- {
+		bc.addBlockToBestChain(ancestorsToCommittedTip[ii])
+	}
+	return true, nil
```
tholonious, line 452: Do you want to add the incoming block to the `uncommittedBlocks` map as well? We may as well perform all changes to that map from one place. It seems natural to perform them here since the best chain is also being updated

lib/pos_blockchain.go
lazynina: hmm we still need to get the index of the committed tip so we can easily delete all the uncommitted blocks that come after it from `bestChain` and `bestChainMap`

tholonious (approved):

lazynina:
### Merge activity

* **Oct 17, 7:42 PM**: @lazynina started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/737).

PR #740, opened 2023-10-18
Draft - CommitBlock testing

lazynina:

PR #742, opened 2023-10-19
Feature/pos process block

lazynina:


lazynina:
  * **PR #742** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/742" target="_blank">Graphite</a>  👈
    * **PR #740** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/740" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/742?utm_source=stack-comment).

collections/slice.go
```diff
@@ -31,3 +31,9 @@ func ToMap[TKey comparable, TValue any](slice []TValue, keyFn func(TValue) TKey)
 	}
 	return result
 }
+
+func Reverse[T any](input []T) {
+	for ii, jj := 0, len(input)-1; ii < jj; ii, jj = ii+1, jj-1 {
+		input[ii], input[jj] = input[jj], input[ii]
+	}
+}
```
tholonious, line 39: Always safer to create a new output slice than to have side effects on the input. The caller may not expect their input to be mutated internally. The memory overhead of performing the copy is fine since we'll rarely use this on large slices

```suggestion
func Reverse[T any](input []T) []T {
	output := make([]T, len(input)
	for ii := 0, len(input)-1; ii = ii+1 {
		output[len(input) - 1 - ii] = input[ii]
	}
	return output
}
```

lib/pos_blockchain.go
```diff
@@ -0,0 +1,1134 @@
...
+//  1. Determine if we're missing the parent block of this block.
+//     If so, return the hash of the missing block and add this block to the orphans list.
+//  2. Validate the incoming block, its header, its block height, the leader, and its QCs (vote or timeout)
+//  3. Store the block in the block index and save to DB.
+//  4. try to apply the incoming block as the tip (performing reorgs as necessary). If it can't be applied, we exit here.
+//  5. Run the commit rule - If applicable, flushes the incoming block's grandparent to the DB
+func (bc *Blockchain) processBlockPoS(block *MsgDeSoBlock, currentView uint64, verifySignatures bool) (_success bool, _isOrphan bool, _missingBlockHashes []*BlockHash, _err error) {
+	// If we can't hash the block, we can never store in the block index and we should throw it out immediately.
+	if _, err := block.Hash(); err != nil {
+		return false, false, nil, errors.New("processBlockPoS: Problem hashing block")
```
diamondhands0, line 24: I'd include the error using errors.Wrapf(err, ...)

lib/pos_blockchain.go
```diff
@@ -0,0 +1,1134 @@
...
+	}
+	// Get all the blocks between the current block and the committed tip. If the block
+	// is an orphan, then we store it after performing basic validations.
+	// If the block extends from any committed block other than the committed tip,
+	// then we throw it away.
+	lineageFromCommittedTip, err := bc.getLineageFromCommittedTip(block)
```
diamondhands0, line 30: How is this not the same as the following block?
```
	lineageFromCommittedTip, err := bc.getLineageFromCommittedTip(block)
	if err == RuleErrorMissingAncestorBlock {
		// In this case, the block is an orphan that does not extend from any blocks
		// on our best chain. Try to process the orphan by running basic validations.
		// If it passes basic integrity checks, we'll store it with the hope that we
		// will eventually get a parent that connects to our best chain.
		missingBlockHashes := []*BlockHash{block.Header.PrevBlockHash}
		// FIXME: I sketeched some of these steps but not all...
		// Step 0: I think we still need to do some basic validation on this block, like
		// verifying that it's signed by the leader for example to prevent spamming.
		// I didn't do that.
		// Step 1: Create a new BlockNode for this block with status STORED.
		// Step 2: Add it to the blockIndexByHash and store it in Badger. This is handled by addBlockToBlockIndex
		return false, true, missingBlockHashes, bc.processOrphanBlockPoS(block)
	}
	if err != nil {
		return false, false, nil, errors.Wrap(err, "processBlockPoS: Problem getting lineage from committed tip: UNEXPECTED ERROR!!!")
	}
```

lazynina: Because I forgot to add a call to `storeValidateFailedBlockInBlockIndex`!! Good catch. If we hit any of the specified errors, we know that we'll never be able to validate this block, so we can go ahead and mark it as ValidateFailed.

lib/pos_blockchain.go
```diff
@@ -0,0 +1,1134 @@
...
+	}
+
+	// Header validation
+	if block.Header.Version != HeaderVersion2 {
+		return RuleErrorInvalidPoSBlockHeaderVersion
+	}
+
+	// Malformed block checks
+	// All blocks must have at least one txn
+	if len(block.Txns) == 0 {
```
diamondhands0, line 362: I'm not sure if we need a block reward in PoS. I can't remember where we landed after discussing it, but right now validators are rewarded based on an APY and based on fees from txns. The APY is split between stakers and the validator they've delegated to based on the validator's commission. So what would we need a block reward for? Just as extra incentive for the leader I guess? Not crazy.

If we choose to keep the block reward let's make it adjustable via ParamUpdater, including setting it to zero if it's no longer needed in the future.

lib/pos_blockchain.go
```diff
@@ -0,0 +1,1134 @@
...
+	isVoteQCEmpty := block.Header.ValidatorsVoteQC.isEmpty()
+	if isTimeoutQCEmpty && isVoteQCEmpty {
+		return RuleErrorNoTimeoutOrVoteQC
+	}
+
+	if !isTimeoutQCEmpty && !isVoteQCEmpty {
+		return RuleErrorBothTimeoutAndVoteQC
+	}
+
+	if len(block.Txns) == 0 {
```
diamondhands0, line 391: Looks like you already checked this above.

lazynina: yep - will delete.

lib/pos_blockchain.go
lazynina: @tholonious @AeonSw4n - we discussed it and were going to keep the block reward to reward the leader I believe.

lib/pos_blockchain.go
```diff
@@ -0,0 +1,1134 @@
...
+	computedMerkleRoot, _, err := ComputeMerkleRoot(block.Txns)
+	if err != nil {
+		return errors.Wrapf(err, "isProperlyFormedBlockPoS: Problem computing merkle root")
+	}
+	if !merkleRoot.IsEqual(computedMerkleRoot) {
+		return RuleErrorInvalidMerkleRoot
+	}
+
+	// TODO: What other checks do we need to do here?
+	return nil
```
diamondhands0, line 426: Some things from our previous discussion:
* Check that ProposerPublicKey corresponds to a valid validator based on our next-hour snapshot.
* Check that the QC is signed by 2/3rds majority of stake.
* The RandomSeedHash is determined by signing the previous RandomSeedHash with the validator's key. So that can/should be checked here and in other parts of the validation.
* You can also check that the ProposerVotePartialSignature is valid.

lazynina: - ProposerPublicKey check for non-orphans is handled in `hasValidBlockProposerPoS`. This validation for orphans is a work in progress as there's some slightly different logic we need for orphans.
- QC is signed by 2/3rds majority is handled in `isValidPoSQuorumCertificate`. This validation for orphans is a work in progress as there's some slightly different logic we need for orphans.
- Didn't know that about the random seed hash, will add.
- yep will add a proposer vote partial sig validation

lib/pos_blockchain.go
```diff
@@ -0,0 +1,1134 @@
...
+		// 	set in PutBlockWithTxn. Block rewards are part of the state, and they should be identical to the ones
+		// 	we've fetched during Hypersync. Is there an edge-case where for some reason they're not identical? Or
+		// 	somehow ancestral records get corrupted?
+		if storeFullBlock {
+			if innerErr := PutBlockWithTxn(txn, bc.snapshot, block); innerErr != nil {
+				return errors.Wrapf(innerErr, "upsertBlockAndBlockNodeToDB: Problem calling PutBlock")
+			}
+		}
+
+		// TODO: if storeFullBlock = false, then we should probably remove the block from the DB? This can
```
diamondhands0, line 745: I don't think we need to worry too much about this case. It should be extremely rare for a block to pass our store-check without eventually being valid. Specifically, I think it'll be really hard for a block to get a valid QC without being actually valid.

lib/pos_blockchain.go
```diff
@@ -0,0 +1,1134 @@
...
+			glog.V(2).Infof("upsertBlockAndBlockNodeToDB: Preparing snapshot flush")
+		}
+		// TODO: Do we want to write the full block once.
+		// Store the new block in the db under the
+		//   <blockHash> -> <serialized block>
+		// index.
+		// TODO: In the archival mode, we'll be setting ancestral entries for the block reward. Note that it is
+		// 	set in PutBlockWithTxn. Block rewards are part of the state, and they should be identical to the ones
+		// 	we've fetched during Hypersync. Is there an edge-case where for some reason they're not identical? Or
+		// 	somehow ancestral records get corrupted?
```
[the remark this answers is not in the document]
lazynina: This is the same comment that is in `ProcessBlockPoW` so was leaving it here in case we need to address it here. @AeonSw4n left the original comment in blockchain.go when doing hypersync I believe, so maybe he has more context.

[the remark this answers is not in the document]
diamondhands0: I'm not sure if this is what you're referring to, but HyperSync could get messed up if you consider stored blocks as a part of your state. In particular, you can have:
- node A saw block B1 and stored it
- node B did not see block B1 and therefore didn't store it

The solution to the above is to only include data corresponding to the main chain in HyperSync. If you also include data from stored or invalid blocks, then the state that one is hypersyncing can differ. This makes me think we should probably remove the PrepareAncestralRecordsFlush from everywhere EXCEPT for where we COMMIT the block.

lib/pos_blockchain.go
```diff
@@ -0,0 +1,1134 @@
...
+func (bc *Blockchain) validateAndIndexBlockPoS(block *MsgDeSoBlock) (*BlockNode, error) {
+	blockHash, err := block.Header.Hash()
+	if err != nil {
+		return nil, errors.Wrapf(err, "validateAndIndexBlockPoS: Problem hashing block %v", block)
+	}
+
+	// Base case - Check if the block is validated or validate failed. If so, we can return early.
+	blockNode, exists := bc.blockIndexByHash[*blockHash]
+	if exists && (blockNode.IsValidateFailed() || blockNode.IsValidated()) {
+		return blockNode, nil
```
diamondhands0, line 164: Hmm don't you want to return an error if the block is invalid?

lib/pos_blockchain.go
```diff
@@ -0,0 +1,1134 @@
...
+	// this block as ValidateFailed. If the parent is not ValidateFailed, we ONLY store the block and move on.
+	// We don't want to store it as ValidateFailed because we don't know if it's actually invalid.
+	if parentBlockNode.IsValidateFailed() {
+		return bc.storeValidateFailedBlockWithWrappedError(block, errors.New("parent block is ValidateFailed"))
+	}
+
+	// If the parent block still has a Stored status, it means that we weren't able to validate it
+	// despite trying. The current block will also be stored as a Stored block.
+	if !parentBlockNode.IsValidated() {
+		return bc.storeValidateFailedBlockWithWrappedError(block, errors.New("parent block is neither Validated nor ValidateFailed"))
```
diamondhands0, line 185: I assume this is because we post-process child blocks after we validate.

lib/pos_blockchain.go
```diff
@@ -0,0 +1,1134 @@
...
+	validatorsByStake, err := utxoView.GetAllSnapshotValidatorSetEntriesByStake()
+	if err != nil {
+		// This should never happen. If the parent is validated and extends from the tip, then we should
+		// be able to fetch the validator set at its block height for it. This failure can only happen due
+		// to transient badger issues. We return the block node as-is as a best effort thing.
+		return blockNode, nil
+	}
+
+	// Validate the block's QC. If it's invalid, we store it as ValidateFailed.
+	if err = bc.isValidPoSQuorumCertificate(block, validatorsByStake); err != nil {
```
diamondhands0, line 224: I'm skipping over this function. Lmk if you want me to look at it very carefully.

lib/pos_blockchain.go
lazynina: No - This function only returns an error if we hit an unexpected error such as an error reading from badger or something. The caller is returned the block node and can decide what to do from there. If the caller receives an error, it means that something unexpected happened and we don't know if the block is truly invalid or just had some weird intermittent error.

lib/pos_blockchain.go
lazynina: Yes please look at it. This is the 2/3s supermajority QC check which is pretty critical.

lib/pos_blockchain.go
```diff
@@ -0,0 +1,1134 @@
...
+	// The number of blocks that have been added to the chain since the start of the epoch is block.Height - currentEpoch.InitialHeight.
+	// The difference between these two numbers is the number of timeouts that have occurred in this epoch.
+	// For each timeout, we need to go to the next leader in the schedule.
+	// If we have more timeouts than leaders in the schedule, we start from the top of the schedule again,
+	// which is why we take the modulo of the length of the leader schedule.
+	// A quick example: If we have 3 leaders in the schedule and the epoch started at height 10 and view 11,
+	// and the current block is at height 15 and view 17, then the number of timeouts that have occurred is
+	// (17 - 11) - (15 - 10) = 1. This means this block should be proposed by the 2nd leader in the schedule,
+	// which is at index 1.
+	leaderIdxUint64 := (viewDiff - heightDiff) % uint64(len(leaders))
```
diamondhands0, line 531: This timeout leader schedule calculation makes sense to me.

lib/pos_blockchain.go
lazynina: that's correct - although I *think* I should update this line to simply call `storeBlockInBlockIndex` instead of marking it as validate failed. Although I'm not 100% sure how we'd ever hit this case since `validatePreviouslyIndexedBlockPoS` will either mark the parent block validate or validate failed OR return an error (which gets handled above)

lib/pos_blockchain.go
```diff
@@ -0,0 +1,1134 @@
...
+	if leaderIdxUint64 > math.MaxUint16 {
+		return false, nil
+	}
+	leaderIdx := uint16(leaderIdxUint64)
+	leaderEntry, err := utxoView.GetSnapshotLeaderScheduleValidator(leaderIdx)
```
diamondhands0, line 536: So I notice the utxoView is set for the PrevBlock, but it looks like you're trying to use it to get the leader for the *current* block. I think that works as long as the function is returning the leader schedule for the *next* block, which I think it is. Otherwise, this could get messed up on epoch boundaries where the leader schedule is changing. Just double-check it.

lazynina: When we connect the parent block, if it's the last block in the epoch, the end of epoch hook runs and the CurrentEpochEntry is updated so we have the correct leader schedule & validator set for the next block.

lib/pos_blockchain.go
```diff
@@ -0,0 +1,1134 @@
...
+	// and can just add this block to the best chain.
+	chainTip := bc.GetBestChainTip()
+	if chainTip.Hash.IsEqual(blockNode.Header.PrevBlockHash) {
+		bc.addBlockToBestChain(blockNode)
+		return true, nil
+	}
+	// Check if we should perform a reorg here.
+	// If we shouldn't reorg AND the incoming block doesn't extend the chain tip, we know that
+	// the incoming block will not get applied as the new tip.
+	if !bc.shouldReorg(blockNode, currentView) {
```
diamondhands0, line 783: This shouldReorg thing was confusing me a bit. Consider the following examples:
* `bestChain = A(view=1) -> B(view=2, uncommitted) -> C(view=3, uncommitted)` and now we have block `C'(view=3, uncommitted)`. I think that we would always reorg to C' in this case, which I think is fine.
* But what if we have `bestChain = A(view=1)` and we have `B(view=2, uncommitted)` and `C(view=3, uncommitted)` such that B and C aren't in the bestChain yet. I think this function would still count that as a reorg even though we're just extending the bestChain by two.

I think the latter scenario is still fine, it's just a little weird. But I also think it probably can't happen in practice the way we call this. I'm still reviewing.

lib/pos_blockchain.go
```diff
@@ -0,0 +1,1134 @@
...
+// functions have validated that this block is not extending from a committed block
+// that is not the latest committed block, so there is no need to validate that here.
+func (bc *Blockchain) shouldReorg(blockNode *BlockNode, currentView uint64) bool {
+	chainTip := bc.GetBestChainTip()
+	// If this block extends from the chain tip, there's no need to reorg.
+	if chainTip.Hash.IsEqual(blockNode.Header.PrevBlockHash) {
+		return false
+	}
+	// If the block is proposed in a view less than the current view, there's no need to reorg.
+	return blockNode.Header.ProposedInView >= currentView
```
diamondhands0, line 822: I haven't reviewed the hardcore consensus rules that thoroughly yet. But it looks like this rule makes it so that we'd reorg in the following scenario:
* Suppose `bestChain = A(view=1, committed) -> B(view=2, uncommitted) -> C(view=3, uncommitted)` and now block `A(view=1, committed) -> D(view=4, uncommitted)` comes in, such that D extends from A but with a higher view. In this case I think we would reorg from A->B->C to A->D.

I think the above is all the right thing to do. But just double-check it.

lazynina: That's correct and that's also my understanding of the reorg logic for fast hotstuff. @AeonSw4n @tholonious - this is correct ya?

lib/pos_blockchain.go
lazynina: I don't think we could have a situation where we didn't add B and C to the BestChain.

lib/pos_blockchain.go
```diff
@@ -0,0 +1,1134 @@
...
+		// ability to roll it back in the future.
+		if innerErr := PutUtxoOperationsForBlockWithTxn(txn, bc.snapshot, uint64(blockNode.Height), blockNode.Hash, utxoOpsForBlock); innerErr != nil {
+			return errors.Wrapf(innerErr, "ProcessBlock: Problem writing utxo operations to db on simple add to tip")
+		}
+		return nil
+	})
+	if err != nil {
+		return errors.Wrapf(err, "commitBlockPoS: Problem putting block in db: ")
+	}
+	if err = utxoView.FlushToDb(uint64(blockNode.Height)); err != nil {
```
diamondhands0, line 959: Hmm I think this has to be in the previous badger txn and use utxoView.FlushToDbWithTxn(). If you don't do this, you run the risk of the db getting corrupted during a restart. It would also be a reeeally hard to find bug...

diamondhands0: Put another way, everything we write to the DB when we commit a block has to be atomic. The only way to do that is to commit it all with the same badger txn.

lib/pos_blockchain.go
diamondhands0: Confirmed you can't connect C without B already being in bestChain. This is because of the way we process orphans at the end of processBlockPoS.

lib/pos_blockchain.go
diamondhands0: Awesome sounds good. Yeah I figured this function is telling you what the *next* leader is going to be.

lib/pos_blockchain.go
lazynina: So the REALLY annoying thing is that `utxoView.FlushToDB` calls `PrepareAncestralRecordsFlush`, and having nested calls to `PrepareAncestralRecordsFlush` raises an error. So do I not call the prepare ancestral records flush stuff at the top here? or how should it work? Trying to do stuff similar to what's in ProcessBlockPoW

lazynina: discussed offline - will write a flush to db func that doesn't call PrepareAncestralRecords flush itself.

lazynina: resolved in #800

lib/pos_blockchain.go
lazynina: resolved in #800

lib/pos_blockchain.go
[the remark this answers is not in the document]
lazynina: So PutBlockWithTxn puts block rewards into an index that has `is_state` set to true. So this makes me think that what you're saying is correct. Block rewards from blocks that don't make it into the main chain shouldn't be included. Let's move this discussion offline.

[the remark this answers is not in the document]
lazynina: resolved in #802

lib/pos_blockchain.go
```diff
@@ -0,0 +1,1331 @@
...
+func (bc *Blockchain) getOrCreateBlockNodeFromBlockIndex(block *MsgDeSoBlock) (*BlockNode, error) {
+	hash, err := block.Header.Hash()
+	if err != nil {
+		return nil, errors.Wrapf(err, "getOrCreateBlockNodeFromBlockIndex: Problem hashing block %v", block)
+	}
+	blockNode := bc.blockIndexByHash[*hash]
+	if blockNode != nil {
+		return blockNode, nil
+	}
+	prevBlockNode := bc.blockIndexByHash[*block.Header.PrevBlockHash]
```
diamondhands0, line 855: I want to point out here that prevBlockNode would be nil in the case where you fetch blocks in a suboptimal order. For example:
* Imagine the chain is A -> B -> C
* The node currently has A only, and now receives C
* We would create a BlockNode for C and add it to our index with Parent = nil

I'm mentioning the above because we need to later FIX C after we receive B. I'm still early in the code, but making a note to make sure we're doing this.

lib/pos_blockchain.go
```diff
@@ -0,0 +1,1331 @@
...
+	// If the block extends from any committed block other than the committed tip,
+	// then we throw it away.
+	lineageFromCommittedTip, err := bc.getLineageFromCommittedTip(block)
+	if err == RuleErrorDoesNotExtendCommittedTip ||
+		err == RuleErrorParentBlockHasViewGreaterOrEqualToChildBlock ||
+		err == RuleErrorParentBlockHeightNotSequentialWithChildBlockHeight ||
+		err == RuleErrorAncestorBlockValidationFailed {
+		// In this case, the block extends a committed block that is NOT the tip
+		// block. We will never accept this block, so mark it as ValidateFailed.
+		if _, innerErr := bc.storeValidateFailedBlockInBlockIndex(block); innerErr != nil {
```
diamondhands0, line 39: I read through this function and I noticed that we are storing a BlockNode whenever we get here, but not storing the block itself. I think this optimization is clever, but it's still dangerous because someone could just send us millions of garbage blocks and cause us to OOM.

Instead, we need a hard and fast rule: We should not store anything in memory OR on disk until we've done some anti-spam validation. I'm still reading through, but my gut would be that we need to do an additional QC check or something harder to pass before we store an invalid block for later.

lib/pos_blockchain.go
```diff
@@ -0,0 +1,1331 @@
...
+		}
+		return false, false, nil, err
+	}
+	if err == RuleErrorMissingAncestorBlock {
+		// In this case, the block is an orphan that does not extend from any blocks
+		// on our best chain. Try to process the orphan by running basic validations.
+		// If it passes basic integrity checks, we'll store it with the hope that we
+		// will eventually get a parent that connects to our best chain.
+		missingBlockHashes := []*BlockHash{block.Header.PrevBlockHash}
+		// FIXME: I sketched some of these steps but not all...
```
diamondhands0, line 51: Don't forget to clean up the comments.

lib/pos_blockchain.go
```diff
@@ -0,0 +1,1331 @@
...
+	isVoteQCEmpty := block.Header.ValidatorsVoteQC.isEmpty()
+	if isTimeoutQCEmpty && isVoteQCEmpty {
+		return RuleErrorNoTimeoutOrVoteQC
+	}
+
+	if !isTimeoutQCEmpty && !isVoteQCEmpty {
+		return RuleErrorBothTimeoutAndVoteQC
+	}
+
+	if block.Txns[0].TxnMeta.GetTxnType() != TxnTypeBlockReward {
```
diamondhands0, line 567: Add the following todo:

TODO: Do we need block rewards in PoS? Validators are rewarded based on commissions on APY from stake, making this field obsolete.

lib/pos_blockchain.go
```diff
@@ -0,0 +1,1331 @@
...
+		var epochEntrySnapshotAtEpochNumber uint64
+		epochEntrySnapshotAtEpochNumber, err = utxoView.ComputeSnapshotEpochNumberForEpoch(epochEntry.EpochNumber)
+		if err != nil {
+			return errors.Wrapf(err, "processOrphanBlockPoS: Problem getting snapshot at epoch number for poch entry at epoch #%d", epochEntry.EpochNumber)
+		}
+		// Okay now that we've gotten the SnapshotAtEpochNumber for the prev/next epoch, we can make sure that the proposer
+		// of the block is within the set of potential block proposers for the prev/next epoch based on
+		// the VotingPublicKey.
+		// First, we get the snapshot validator entry based on the BLS public key in the header.
+		snapshotBLSPublicKeyPKIDEntry, err := utxoView.GetSnapshotValidatorBLSPublicKeyPKIDPairEntry(block.Header.ProposerVotingPublicKey, epochEntrySnapshotAtEpochNumber)
```
diamondhands0, line 179: nit: Line length.

lib/pos_blockchain.go
```diff
@@ -0,0 +1,1331 @@
...
+	if err = bc.isValidPoSQuorumCertificate(block, validatorsByStake); err != nil {
+		// If we hit an error, we know that the QC is invalid and we'll never accept this block,
+		// so we mark it as validate failed.
+		if _, innerErr := bc.storeValidateFailedBlockInBlockIndex(block); innerErr != nil {
+			return errors.Wrapf(innerErr, "processOrphanBlockPoS: Problem adding validate failed block to block index: %v", err)
+		}
+		return nil
+	}
+	// Add to blockIndexByHash with status STORED only as we are not sure if it's valid yet.
+	_, err = bc.storeBlockInBlockIndex(block)
```
diamondhands0, line 269: OK, so as I read through this function I saw a lot of storeValidateFailedBlockInBlockIndex, which again I think is dangerous because it means someone can spam us really hard with invalid blocks. But I think it's a relatively easy fix. Below is a proposal:
* I think we shouldn't store anything until we've done the QC check. On the bright side, it looks like we can do the QC without having to do anything else.
* Given the above, I recommend we break all of the code above storeBlockInBlockIndex into its own function called something like `validateBlockQC` or something like that. When you do this, replace all instances of storeValidateFailedBlockInBlockIndex with returning an error.
* Once you've done that, I would call validateBlockQC at the tippy top of processBlockPoS, effectively as an anti-spam measure. I would do it before you even check the lineage.
* After you've called validateBlockQC, you're now safe to store everything thing you want thereafter because passing the QC check is hard. So you can store the invalid block if it fails the lineage check in processBlockPoS, for example.

lib/pos_blockchain.go
```diff
@@ -0,0 +1,1331 @@
...
+//  2. Validate the incoming block, its header, its block height, the leader, and its QCs (vote or timeout)
+//  3. Store the block in the block index and save to DB.
+//  4. try to apply the incoming block as the tip (performing reorgs as necessary). If it can't be applied, we exit here.
+//  5. Run the commit rule - If applicable, flushes the incoming block's grandparent to the DB
+func (bc *Blockchain) processBlockPoS(block *MsgDeSoBlock, currentView uint64, verifySignatures bool) (
+	_success bool, _isOrphan bool, _missingBlockHashes []*BlockHash, _err error) {
+	// If we can't hash the block, we can never store in the block index and we should throw it out immediately.
+	if _, err := block.Hash(); err != nil {
+		return false, false, nil, errors.Wrapf(err, "processBlockPoS: Problem hashing block")
+	}
```
diamondhands0, line 27: In another comment I mention breaking the block sanity check and the QC check out into something like `validateBlockQC()`. If you do that, I would put that function right here, and I would return an error without storing anything if it fails. This would allow the caller to disconnect the node that sent the messed up block without opening us to spam possibillity. It would also allow you to store invalid blocks in later sections of the code, like in the next if statement, which is useful.

lib/pos_blockchain.go
diamondhands0: Ah sorry just saw this reply. Ok if we keep it can we make it based on a GlobalParam instead of on supply.go?

lib/pos_blockchain.go
```diff
@@ -0,0 +1,1331 @@
...
+		return nil, errors.Wrapf(err, "validateAndIndexBlockPoS: Problem adding block to block index: %v", err)
+	}
+
+	// Check if the block is properly formed and passes all basic validations.
+	if err = bc.isValidBlockPoS(block); err != nil {
+		return bc.storeValidateFailedBlockWithWrappedError(block, err)
+	}
+
+	// Validate the block's random seed signature
+	isValidRandomSeedSignature, err := bc.hasValidProposerRandomSeedSignaturePoS(block)
```
diamondhands0, line 348: Don't do this yet, but I'm just flagging that processing signatures is probably going to make historical block processing a lot slower. We don't need to fix it yet but put a TODO here, and wherever we do a signature check to indicate that we may want to flag this behind a `verifySignatures` check.

lib/pos_blockchain.go
```diff
@@ -0,0 +1,1331 @@
...
+	})
+
+	// If we fail to connect the block, then it means the block is invalid. We should store it as ValidateFailed.
+	if _, err = utxoView.ConnectBlock(block, txHashes, true, nil, block.Header.Height); err != nil {
+		// If it doesn't connect, we want to mark it as ValidateFailed.
+		return bc.storeValidateFailedBlockWithWrappedError(block, err)
+	}
+
+	// We can now add this block to the block index since we have performed all basic validations.
+	blockNode, err = bc.storeValidatedBlockInBlockIndex(block)
```
diamondhands0, line 424: So at this point you store the block and mark it as validated. But now suppose the node shuts down right here and reboots. Wouldn't this result in an inconsistent state?

I think you need to mark the block as validated in the same txn where you update the tip. Otherwise, you can get weird edge cases where a new block comes in and you think you've processed this one but you haven't actually.

Thaat would mean that you should just store it here but not mark it as validated.

diamondhands0: Hey sorry I realized that this is OK because of the changes we're making with PoS. Namely: We don't update badger with all the connected transactions until we COMMIT, which means that it's correct to mark the block as validated here and store it.

lib/pos_blockchain.go
```diff
@@ -0,0 +1,1331 @@
...
+		}
+	}
+	return nil
+}
+
+// canCommitGrandparent determines if the grandparent of the current block can be committed.
+// The grandparent can be committed if there exists a direct parent-child relationship
+// between the grandparent and parent of the new block, meaning the grandparent and parent
+// are proposed in consecutive views, and the "parent" is an ancestor of the incoming block (not necessarily consecutive views).
+// Additionally, the grandparent must not already be committed.
```
diamondhands0, line 1074: Would be nice to add a reference to Revolution PoS doc here so people understand where this rule comes from.

diamondhands0: Also this is definitely the commit rule I remember. But something I was thinking about is what happens if you have a chain like the following, where the x'es mean "timeout occurred" for that view:
```
A -> x -> x -> B -> x -> x -> C -> x -> x -> D -> x -> x -> E
```

I think you could have an infinitely long chain of this where blocks would just never commit. Is that true? If so, this feels wrong. It feels like we should be able to commit if there are three blocks of QC built up, even if there are timeouts between them. I would ask Piotr about it, as he went the deepest on this but I can also dig into the stuff if you can't find the answer.

The reason why this is important btw is we want to be able to tell exchanges that they can consider a certain number of blocks as committed. And it's kindof annoying if, in theory, it could take an infinite number of blocks to commit.

lib/pos_blockchain.go
```diff
@@ -0,0 +1,1331 @@
...
+		if err = bc.runCommitRuleOnBestChain(); err != nil {
+			return false, false, nil, errors.Wrap(err, "processBlockPoS: error running commit rule: ")
+		}
+	}
+
+	// Now that we've processed this block, we check for any blocks that were previously
+	// stored as orphans, which are children of this block. We can  process them now.
+	blockNodesAtNextHeight := bc.blockIndexByHeight[uint64(blockNode.Height)+1]
+	for _, blockNodeAtNextHeight := range blockNodesAtNextHeight {
+		if blockNodeAtNextHeight.Header.PrevBlockHash.IsEqual(blockNode.Hash) && blockNodeAtNextHeight.IsStored() && !blockNodeAtNextHeight.IsValidated() {
```
diamondhands0, line 94: Nit: Line length.

lib/blockchain.go
```diff
@@ -65,18 +66,43 @@ const (
...
+	StatusBlockStored         = 1 << 3 // Stored means that the block has been added to the block index and stored in the DB.
+	StatusBlockValidated      = 1 << 4 // Validated means that the block has passed validations and is eligible to be part of the best chain.
+	StatusBlockValidateFailed = 1 << 5 // Validate Failed means that the block did not pass validations and will never be part of the best chain.
 
-	StatusBitcoinHeaderValidated      // Deprecated
-	StatusBitcoinHeaderValidateFailed // Deprecated
+	StatusBitcoinHeaderValidated      = 1 << 6 // Deprecated
+	StatusBitcoinHeaderValidateFailed = 1 << 7 // Deprecated
+
+	StatusBlockCommitted = 1 << 8 // Committed means that the block has been committed to the blockchain according to the Fast HotStuff commit rule. Only set on blocks after the cutover for PoS
```
diamondhands0, line 80: Add a NEXT_TAG = 9 at the bottom.

lib/blockchain.go
```diff
@@ -558,26 +628,27 @@ func (bc *Blockchain) _initChain() error {
...
-		tipNode := bc.blockIndex[*bestBlockHash]
+		tipNode := bc.blockIndexByHash[*bestBlockHash]
 		if tipNode == nil {
 			return fmt.Errorf("_initChain(block): Best hash (%#v) not found in block index", bestBlockHash)
 		}
 
 		// Walk back from the best node to the genesis block and store them all
 		// in bestChain.
-		bc.bestChain, err = GetBestChain(tipNode, bc.blockIndex)
+		bc.bestChain, err = GetBestChain(tipNode, bc.blockIndexByHash)
```
diamondhands0, line 651: So the uncommitted blocks that are after the committed tip and validated aren't added to best chain? How do they get added?

Seems like maybe you could add them here now that you have blockIndexByHeight.

lib/block_view_validator.go
```diff
@@ -1635,6 +1802,16 @@ func (bav *UtxoView) IsValidRegisterAsValidatorMetadata(
 		}
 	}
 
+	// Error if VotingPublicKey is already taken.
+	validatorBLSPublicKeyPKIDPairEntry, err := bav.GetBLSPublicKeyPKIDPairEntry(metadata.VotingPublicKey)
```
diamondhands0, line 1806: I'm not going throught his BLS public key manipulation too closely. Just test it thoroughly to make sure it's all updating correctly across epoch boundaries especially.

lib/pos_blockchain.go
lazynina: This is broken even in PoW. There are a handful of cases where Parent is not set. The safe way to access a parent is to grab the parent's hash from the Header and look it up from the block index.

lib/blockchain.go
lazynina: Once we have a new block come in, the process block PoS logic automatically determines where to reorg to. We can add some extra logic here to update the best chain on restart, but this is a minor improvement as we'll auto re-org when we get the first new block.

lib/pos_blockchain.go
lazynina: Discussed offline - we'll still have block reward txns, but they'll only included fees (BMF)

tholonious (approved):

PR #753, opened 2023-10-26
ProcessBlockPoS Updates

diamondhands0:


lib/pos_blockchain.go
```diff
@@ -375,51 +539,41 @@ func (bc *Blockchain) getLineageFromCommittedTip(desoBlock *MsgDeSoBlock) ([]*Bl
...
-	if chainTip.Hash.IsEqual(newTip.Header.PrevBlockHash) {
-		bc.addBlockToBestChain(newTipBlockNode)
-		return true, nil
-	}
-	// Check if we should perform a reorg here.
-	// If we shouldn't reorg AND the incoming block doesn't extend the chain tip, we know that
-	// the incoming block will not get applied as the new tip.
-	if !bc.shouldReorg(newTip, currentView) {
-		return false, nil
-	}
```
lazynina, line 415: Why are we getting rid of this shouldReorg call? This is critical to fast hotstuff if I'm not mistaken.

lib/pos_blockchain.go
```diff
@@ -534,25 +659,38 @@ func (bc *Blockchain) canCommitGrandparent(currentBlock *BlockNode) (_grandparen
...
 	// Connect a view up to the parent of the block we are committing.
 	utxoView, err := bc.getUtxoViewAtBlockHash(*block.Header.PrevBlockHash)
 	if err != nil {
 		return errors.Wrapf(err, "runCommitRuleOnBestChain: Problem initializing UtxoView: ")
 	}
 	// Get the full uncommitted block from the uncommitted blocks map
-	grandParentBlock, exists := bc.uncommittedBlocksMap[*blockHash]
+	// FIXME: This is using blockHash when it should use a parent hash or something.
+	// It was like this when I got here.
+	grandParentBlockNode, exists := bc.blockIndex[*blockHash]
```
lazynina, line 686: This is just a typo. We call "commitBlock" on the grandparent block node if the commit rule tells us to. I just forgot to update the name of the variable.

PR #755, opened 2023-10-26
Replace CommittedStatus with existing BlockStatus field

lazynina:


lazynina:
    * **PR #755** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/755" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/755?utm_source=stack-comment).

lib/blockchain.go
```diff
@@ -323,28 +324,19 @@ func NewPoWBlockNode(
...
 // NewPoSBlockNode is a new helper function to create a block node
 // as we need to control the value of the CommittedStatus field.
-func NewPoSBlockNode(
-	parent *BlockNode,
-	hash *BlockHash,
-	height uint32,
-	header *MsgDeSoHeader,
-	status BlockStatus,
-	committedStatus CommittedBlockStatus) *BlockNode {
+func NewPoSBlockNode(parent *BlockNode, hash *BlockHash, height uint32, header *MsgDeSoHeader, status BlockStatus) *BlockNode {
```
tholonious, line 332: Do we need a separate `NewPoSBlockNode` function now? Seems like we can reuse the original `NewBlockNode` function that's currently just used for PoW now

lib/blockchain.go
```diff
@@ -63,24 +63,35 @@ const (
...
 
-	StatusBlockProcessed
-	StatusBlockStored
-	StatusBlockValidated
-	StatusBlockValidateFailed
+	StatusBlockProcessed      = 1 << 2
+	StatusBlockStored         = 1 << 3
+	StatusBlockValidated      = 1 << 4
+	StatusBlockValidateFailed = 1 << 5
+	StatusBlockCommitted      = 1 << 8
```
tholonious, line 73: Can you add descriptions for what each of these statuses means, and when each one is used? It'll make future maintainers' lives much easier

lib/pos_blockchain.go
```diff
@@ -364,7 +364,7 @@ func (bc *Blockchain) getLineageFromCommittedTip(desoBlock *MsgDeSoBlock) ([]*Bl
 		if currentBlock.Hash.IsEqual(highestCommittedBlock.Hash) {
 			break
 		}
-		if currentBlock.CommittedStatus == COMMITTED {
+		if IsBlockCommitted(currentBlock) {
```
tholonious, line 367: nit: maybe differentiate naming between blocks (of type MsgDeSoBlock) from block nodes (of type BlockNode). In this case, we can make it a bit more obvious that the committed check is just checking the status on the node rather than checking if the block itself is committed in the chain through a lookup underneath
```suggestion
		if currentBlockNode.IsCommitted() {
```

lib/blockchain.go
```diff
@@ -63,24 +63,35 @@ const (
...
 
-	StatusBitcoinHeaderValidated      // Deprecated
-	StatusBitcoinHeaderValidateFailed // Deprecated
+	StatusBitcoinHeaderValidated      = 1 << 6 // Deprecated
+	StatusBitcoinHeaderValidateFailed = 1 << 7 // Deprecated
 )
 
-type CommittedBlockStatus uint8
+// A block is stored if it has been added to the blockIndex and stored in the DB.
+func IsBlockStored(blockNode *BlockNode) bool {
```
tholonious, line 80:
```suggestion
func (blockNode *BlockNode) IsStored() bool {
```

lib/blockchain.go
```diff
@@ -63,24 +63,35 @@ const (
...
+	return blockNode.Status&StatusBlockStored != 0
+}
 
-const (
-	COMMITTED   CommittedBlockStatus = 0
-	UNCOMMITTED CommittedBlockStatus = 1
-)
+// A block is validated if it has passed all validations. A block that is validated is
+// generally always stored first.
+func IsBlockValidated(blockNode *BlockNode) bool {
```
tholonious, line 86:
```suggestion
func (blockNode *BlockNode) IsValidated() bool {
```

lib/blockchain.go
```diff
@@ -63,24 +63,35 @@ const (
...
+	return blockNode.Status&StatusBlockValidated != 0
+}
+
+// A block is committed if it has passed all validations and it has been committed to
+// the blockchain according to the Fast HotStuff commit rule.
+func IsBlockCommitted(blockNode *BlockNode) bool {
```
tholonious, line 92:
```suggestion
func (blockNode *BlockNode) IsCommitted() bool {
```

tholonious (changes requested):

lib/pos_blockchain.go
lazynina: Will address naming conventions in a separate PR

tholonious (approved):

lib/blockchain.go
```diff
@@ -63,24 +63,35 @@ const (
...
 
-	StatusBlockProcessed
-	StatusBlockStored
-	StatusBlockValidated
-	StatusBlockValidateFailed
+	StatusBlockProcessed      = 1 << 2 // Process means that the block is not an orphan and has been processed. This helps prevent us from reprocessing a block that we've already attempted to validate and add to the block index.
+	StatusBlockStored         = 1 << 3 // Stored means that the block has been added to the block index and stored in the DB.
+	StatusBlockValidated      = 1 << 4 // Validated means that the block has passed validations and is eligible to be part of the best chain.
+	StatusBlockValidateFailed = 1 << 5 // Validate Failed means that the block did not pass validations and will never be part of the best chain.
+	StatusBlockCommitted      = 1 << 8 // Committed means that the block has been committed to the blockchain according to the Fast HotStuff commit rule. Only set on blocks after the cutover for PoS
```
AeonSw4n, line 73: Wouldn't it be better if StatusBlockCommitted was placed after the deprecated fields so the values only increase?

AeonSw4n (approved):
Nice, I like it this way. Clean.

lib/blockchain.go
lazynina: I did it this way because that's how @diamondhands0 did it in his PR. I have no preference whatsoever.

tholonious: Equally easy to read both ways IMO. Marking this as resolved

PR #756, opened 2023-10-26
Remove uncommitted blocks map

lazynina:

PR #758, opened 2023-10-27
modify skeleton based on DH feedback

lazynina:


lazynina:
    * **PR #755** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/755" target="_blank">Graphite</a> 
      * **PR #756** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/756" target="_blank">Graphite</a> 
        * **PR #758** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/758" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/758?utm_source=stack-comment).

lib/pos_blockchain.go
```diff
@@ -18,45 +18,108 @@ import (
...
+	}
 	if err == RuleErrorMissingAncestorBlock {
+		// In this case, the block is an orphan that does not connect to any blocks
+		// on our best chain. In this case we'll store it with the hope that we
+		// will eventually get a parent that connects to our best chain.
 		missingBlockHashes := []*BlockHash{desoBlock.Header.PrevBlockHash}
-		blockHash, err := desoBlock.Header.Hash()
-		// If we fail to get the block hash, this block isn't valid at all, so we
-		// don't need to worry about adding it to the orphan list or block index.
+		// FIXME: I sketeched some of these steps but not all...
```
AeonSw4n, line 43: Not sure if you added a check for this already. For TxnConnectStatusByIndex we should add the following block validation step:

```
HashBitset(MsgDeSoBlock.TxnConnectStatusByIndex).IsEqual( 
MsgDeSoBlock.Header.TxnConnectStatusByIndexHash)
```

lazynina: I'll add this in. We didn't have it in the MsgDeSoBlock when I was writing all the validations. I'll add this in another PR in the stack.

lazynina: @AeonSw4n - this is addressed in #774 . Please let me know if you have other feedback on this PR.

lib/pos_blockchain.go
```diff
@@ -18,45 +18,116 @@ import (
...
+	// We know the utxoView will be valid because we check that all ancestor blocks have been validated.
 	utxoView, err := bc.getUtxoViewAtBlockHash(*desoBlock.Header.PrevBlockHash)
 	if err != nil {
 		return false, false, nil, errors.Wrap(err, "processBlockPoS: Problem initializing UtxoView: ")
 	}
+	// Connect this block to the UtxoView.
+	txHashes := collections.Transform(desoBlock.Txns, func(txn *MsgDeSoTxn) *BlockHash {
+		return txn.Hash()
+	})
+	_, err = utxoView.ConnectBlock(desoBlock, txHashes, true, bc.eventManager, desoBlock.Header.Height)
```
tholonious, line 120: Do we really want to pass through the bc.eventManager here? I'm not sure what the downstream implications are for event listeners that have subscribed to it

lib/pos_blockchain.go
```diff
@@ -66,6 +137,7 @@ func (bc *Blockchain) processBlockPoS(desoBlock *MsgDeSoBlock, currentView uint6
 	//     set for the block, we can at least store it locally.
 	// 2. Validate QC
 	if err = bc.validateQC(desoBlock, validatorsByStake); err != nil {
+		// TODO: If this fails, do we want to store the block as Stored & ValidateFailed?
```
tholonious, line 140: Yes, if the block fails QC validation, then it fits the criteria of ValidateFailed. We want to store it as such because we'll know next time we see this block, or see a child of this block that we don't want to try to process it again.

lib/pos_blockchain.go
```diff
@@ -74,12 +146,17 @@ func (bc *Blockchain) processBlockPoS(desoBlock *MsgDeSoBlock, currentView uint6
...
+		return false, false, nil, errors.New("processBlockPoS: Block not found in block index after adding it. " +
+			"This should never happen.")
+	}
 
 	// 4. Try to apply the incoming block as the new tip. This function will
 	// first perform any required reorgs and then determine if the incoming block
 	// extends the chain tip. If it does, it will apply the block to the best chain
 	// and appliedNewTip will be true and we can continue to running the commit rule.
-	appliedNewTip, err := bc.tryApplyNewTip(desoBlock, currentView, lineageFromCommittedTip)
+	appliedNewTip, err := bc.tryApplyNewTip(newBlockNode, currentView, lineageFromCommittedTip)
```
tholonious, line 159: Is this correct? Assuming that the current block's parent is validated and connected from the above processBlock(block.Parent) call, then this tryApplyNewTip just adds the current block to the tip of the chain. It's not possible to reorg here because the recursive calls to processBlock(...) perform their reorgs internally.

Maybe I'm missing something?

lib/pos_blockchain.go
```diff
@@ -468,8 +545,10 @@ func (bc *Blockchain) storeCommittedBlockInBlockIndex(block *MsgDeSoBlock) error
 	return bc.upsertBlockAndBlockNodeToDB(block, blockNode)
 }
 
-// upsertBlockAndBlockNodeToDB adds the block to the block index with the given status. It also writes the block to the block
-// index in badger.
+// TODO: Do we want to allow ONLY writing to the HeightHashToNodeInfo prefix? This would allow us to skip
+// storing blocks that failed validation.
+// upsertBlockToDBBlockIndex writes the BlockNode to the blockIndex in badger and writes the full block
```
tholonious, line 550:
```suggestion
// upsertBlockAndBlockNodeToDB writes the BlockNode to the blockIndex in badger and writes the full block
```

lib/pos_blockchain.go
```diff
@@ -507,28 +587,43 @@ func (bc *Blockchain) upsertBlockAndBlockNodeToDB(desoBlock *MsgDeSoBlock, newBl
 	return nil
 }
 
+func (bc *Blockchain) putBestHash(hash *BlockHash) error {
```
tholonious, line 590: Nit: rename this for clarity. I realize that the existing naming uses the "put best hash" term, but this seems like a decent place to break that so the call sites for this function are clearer
```suggestion
func (bc *Blockchain) storeTipHashInDB(hash *BlockHash) error {
```

tholonious (changes requested):

lib/pos_blockchain.go
```diff
@@ -18,45 +18,116 @@ import (
...
+		// on our best chain. In this case we'll store it with the hope that we
+		// will eventually get a parent that connects to our best chain.
 		missingBlockHashes := []*BlockHash{desoBlock.Header.PrevBlockHash}
-		blockHash, err := desoBlock.Header.Hash()
-		// If we fail to get the block hash, this block isn't valid at all, so we
-		// don't need to worry about adding it to the orphan list or block index.
+		// FIXME: I sketeched some of these steps but not all...
+		// Step 0: I think we still need to do some basic validation on this block, like
+		// verifying that it's signed by the leader for example to prevent spamming.
+		// I didn't do that.
```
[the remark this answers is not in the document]
lazynina: I plan on adding them in a separate PR as I'm not 100% sure about what all the checks we want to perform yet are. basically everything we can check w/o requiring access to the parent.

lib/pos_blockchain.go
```diff
@@ -18,45 +18,116 @@ import (
...
+	// Note that if any other ancestor is not validated, calling processBlockPoS on the parent will
+	// end up calling processBlockPoS on all of its ancestors that aren't validated yet.
+	parentBlockNode, exists := bc.blockIndex[*desoBlock.Header.PrevBlockHash]
+	if !exists {
+		return false, false, nil, errors.New("processBlockPoS: Parent block not found in block index. This should never happen.")
+	}
+	if parentBlockNode.IsValidateFailed() {
+		// We'll never add this to the best chain because its parent is invalid.
+		// If its parent is ValidateFailed, we can update the block index with the block status ValidateFailed
+		if addToBlockIndexErr := bc.storeValidateFailedBlockInBlockIndex(desoBlock); addToBlockIndexErr != nil {
```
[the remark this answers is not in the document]
lazynina: oh this is a copy paste mistake. Above we need to distinguish between an outer and inner error so we can construct a proper error message to return

lib/pos_blockchain.go
```diff
@@ -18,45 +18,116 @@ import (
...
 
-	// 2. Start with all sanity checks of the block.
+	// At this point, we know that all the ancestors of this block have been successfully
+	// validated. That means we can now proceed with attempting to validate THIS block.
+
+	// Do all the sanity checks on the block.
 	// TODO: Check if err is for view > latest committed block view and <= latest uncommitted block.
 	// If so, we need to perform the rest of the validations and then add to our block index.
 	if err = bc.validateDeSoBlockPoS(desoBlock); err != nil {
+		// TODO: If validations fail, do we want to even bother storing the block? Are there any validations that
```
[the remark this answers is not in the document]
lazynina: I'll update `upsertBlockAndBlockNodeToDB` to take in an additional param `storeFullBlock` which determines whether or not we write the full block w/ `PutBlockWithTxn`

lib/pos_blockchain.go
lazynina: You're right. we shouldn't since this is a throw away view.

lib/pos_blockchain.go
lazynina: It's possible that the parent blocks will not satisfy the `shouldReorg` condition 
`blockNode.Header.ProposedInView >= currentView`, in which case we will not reorg.

lazynina: I'm concerned about the recursive process parent calls above tho. Does it break fast hotstuff rules in any way? Does the reorg condition need to be updated to require that the ProposedInView of this block must exceed the current uncommitted tip's ProposedInView?

tholonious: This is resolved by the recursive validateAndIndexBlockPoS function in the downstream PRs.

tholonious (approved):

PR #767, opened 2023-10-30
Address DH's feedback on FastHotStuffEventLoop structure

tholonious:
Changes in this PR:
- Rename "ValidatorSet" to "ValidatorList" to signify that it's a list of pre-sorted validators
- Rename the "votesSeen" map to "votesSeenByBlockHash", and "timeoutsSeen" to "timeoutsSeenByView" so the map keys are clearer
- Remove bit shifting math used to compute exponential backoff for timeouts. The math is now replaced by a clearer to understand "powerOfTwo()" arithmetic computation
- Remove anonymous functions previously used in collections.Transform(...), and instead make them named utility functions so they can be reused
- Fix typos

### DIFF
--- a/collections/slice.go
+++ b/collections/slice.go
@@ -1,5 +1,9 @@
 package collections
 
+func All[T any](slice []T, predicate func(T) bool) bool {
+	return !Any(slice, predicate)
+}
+
 func Any[T any](slice []T, predicate func(T) bool) bool {
 	for _, val := range slice {
 		if predicate(val) {
--- a/collections/slice.go
+++ b/collections/slice.go
@@ -1,7 +1,10 @@
 package collections
 
 func All[T any](slice []T, predicate func(T) bool) bool {
-	return !Any(slice, predicate)
+	negatedPredicate := func(val T) bool {
+		return !predicate(val)
+	}
+	return !Any(slice, negatedPredicate)
 }
 
 func Any[T any](slice []T, predicate func(T) bool) bool {
--- a/consensus/event_loop.go
+++ b/consensus/event_loop.go
@@ -441,7 +441,7 @@ func (fc *FastHotStuffEventLoop) resetScheduledTasks() {
 		numTimeouts := fc.currentView - fc.tip.block.GetView() - 1
 
 		// Compute the exponential back-off: nextTimeoutDuration * 2^numTimeouts
-		timeoutDuration = fc.timeoutBaseDuration << numTimeouts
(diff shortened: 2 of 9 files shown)


tholonious:
* feature/proof-of-stake
    * **PR #767** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/767" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/767?utm_source=stack-comment).

collections/slice.go
```diff
@@ -1,5 +1,9 @@
 package collections
 
+func All[T any](slice []T, predicate func(T) bool) bool {
+	return !Any(slice, predicate)
```
lazynina, line 4: Is this right? This would mean that the predicate is false for every item in the slice. 
```suggestion
	for _, val := range slice {
		if !predicate(val) {
			return false
		}
	}
	return true
```

consensus/event_loop.go
```diff
@@ -165,41 +166,35 @@ func (fc *FastHotStuffEventLoop) ProcessTipBlock(tip BlockWithValidators, safeBl
...
 }
 
-// setSafeBlocks is a helper function that validates the provided blocks, validator sets, and stores them.
+// storeBlocks is a helper function that validates the provided blocks, validator lists, and stores them.
 // It must be called while holding the event loop's lock.
-func (fc *FastHotStuffEventLoop) storeBlocks(tip BlockWithValidators, safeBlocks []BlockWithValidators) error {
-	// Do a basic integrity check on the tip block and validator set
-	if !isProperlyFormedBlock(tip.Block) || !isProperlyFormedValidatorSet(tip.Validators) {
-		return errors.New("Invalid tip block or validator set")
+func (fc *FastHotStuffEventLoop) storeBlocks(tip BlockWithValidatorList, safeBlocks []BlockWithValidatorList) error {
```
lazynina, line 171: Is this tip the uncommitted tip? should we make that explicit?

lazynina (changes requested):

consensus/event_loop.go
tholonious: It could be the committed or uncommitted tip (depending on the state of the chain). Just whatever the the last block in the chain is, this is it.

lazynina (approved):

PR #768, opened 2023-10-31
PoS Validator Connect/Disconnect Flow 

AeonSw4n:


AeonSw4n:
  * **PR #768** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/768" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/768?utm_source=stack-comment).

lib/connection_manager.go
```diff
@@ -220,18 +234,18 @@ func (cmgr *ConnectionManager) subFromGroupKey(na *wire.NetAddress) {
...
 			break
 		}
 
-		if cmgr.connectedOutboundAddrs[addrmgr.NetAddressKey(addr.NetAddress())] {
+		// Lock the address map since multiple threads will be trying to read
+		// and modify it at the same time.
+		cmgr.mtxConnectedOutboundAddrs.RLock()
+		ok := cmgr.connectedOutboundAddrs[addrmgr.NetAddressKey(addr.NetAddress())]
+		cmgr.mtxConnectedOutboundAddrs.RUnlock()
+		if ok {
```
tholonious, line 248: What has changed about the threading model that we need to make things thread-safe in this way?

lib/connection_manager.go
```diff
@@ -156,15 +162,19 @@ func NewConnectionManager(
 		//newestBlock: _newestBlock,
 
 		// Initialize the peer data structures.
-		outboundConnIPGroups:   make(map[string]int),
-		persistentPeers:        make(map[uint64]*Peer),
-		outboundPeers:          make(map[uint64]*Peer),
-		inboundPeers:           make(map[uint64]*Peer),
-		connectedOutboundAddrs: make(map[string]bool),
+		outboundConnIPGroups:       make(map[string]int),
```
tholonious, line 165: It's worth adding comments here to explain what the key is for each of these maps. I'm assuming it's PeerID -> Peer, and PeerID -> OutboundConnectionAttempt,... A comment will clarify for future readers

tholonious (changes requested):

lib/bls_keystore.go
```diff
@@ -0,0 +1,147 @@
+package lib
+
+import (
+	"github.com/deso-protocol/core/bls"
+	"github.com/deso-protocol/core/consensus"
+	"github.com/pkg/errors"
+)
+
+// BLSSigner is a wrapper for the bls.PrivateKey type, which abstracts away the private key
```
lazynina, line 9: this file seems like it belongs in the bls package

lib/server.go
```diff
@@ -1131,6 +1165,8 @@ func (srv *Server) _handleSnapshot(pp *Peer, msg *MsgDeSoSnapshotData) {
 		"<%v>, Last entry: <%v>), (number of entries: %v), metadata (%v), and isEmpty (%v), from Peer %v",
 		msg.SnapshotChunk[0].Key, msg.SnapshotChunk[len(msg.SnapshotChunk)-1].Key, len(msg.SnapshotChunk),
 		msg.SnapshotMetadata, msg.SnapshotChunk[0].IsEmpty(), pp)))
+	// Free up a slot in the operationQueueSemaphore, now that a chunk has been processed.
+	srv.snapshot.FreeOperationQueueSemaphore()
```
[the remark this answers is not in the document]
AeonSw4n: The previous hack approach was causing a deadlock when I run integration tests

lib/server.go
```diff
@@ -690,16 +718,22 @@ func (srv *Server) GetSnapshot(pp *Peer) {
 	}
 	// If operationQueueSemaphore is full, we are already storing too many chunks in memory. Block the thread while
 	// we wait for the queue to clear up.
-	srv.snapshot.operationQueueSemaphore <- struct{}{}
-	// Now send a message to the peer to fetch the snapshot chunk.
-	pp.AddDeSoMessage(&MsgDeSoGetSnapshot{
-		SnapshotStartKey: lastReceivedKey,
-	}, false)
+	go func() {
```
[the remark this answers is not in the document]
AeonSw4n: This corresponds to my other modification of the `operationQueueSemaphore`. Without this change, we run into a deadlock in integration tests.

cmd/node.go
```diff
@@ -27,12 +27,13 @@ import (
...
-	Params   *lib.DeSoParams
-	Config   *Config
-	Postgres *lib.Postgres
+	Server    *lib.Server
+	ChainDB   *badger.DB
+	TXIndex   *lib.TXIndex
+	Params    *lib.DeSoParams
+	Config    *Config
+	Postgres  *lib.Postgres
+	Listeners []net.Listener
```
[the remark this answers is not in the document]
AeonSw4n: It's useful to know node's listeners during integration testing.

lib/connection_manager.go
AeonSw4n: This was a bug. The mutex is meant for `connectedOutboundAddrs`

lib/connection_manager.go
AeonSw4n: We got a comment on this in the ConnectionManager struct definition. Should we make it more verbose?

lib/bls_keystore.go
AeonSw4n: So the idea with the BLSSigner is that it's a context-specific wrapper around the BLS utils. The BLSigner enumerates all the possible BLS signatures a DeSo node can make. Maybe we should consider renaming the struct/file to make it more protocol-specific. Like DeSoBLSSigner or something like that.

lib/connection_manager.go
AeonSw4n: For outboundConnectionAttempts it's actually attemptId -> OutboundConnectionAttempt. I added a comment to clarify this.

lib/bls_keystore.go
tholonious: +1 to @AeonSw4n's point. This struct is a protocol aware signer as opposed to a general purpose BLS utility. IMO, it's best placed in the lib package.

lazynina: okay cool!

lib/connection_manager.go
```diff
@@ -181,17 +195,21 @@ func (cmgr *ConnectionManager) GetAddrManager() *addrmgr.AddrManager {
...
 }
 
+func (cmgr *ConnectionManager) SetTargetOutboundPeers(numPeers uint32) {
+	cmgr.targetOutboundPeers = numPeers
+}
+
 // Check if the address passed shares a group with any addresses already in our
 // data structures.
-func (cmgr *ConnectionManager) isRedundantGroupKey(na *wire.NetAddress) bool {
+func (cmgr *ConnectionManager) IsFromRedundantOutboundIPAddress(na *wire.NetAddress) bool {
```
lazynina, line 204: Does this function need to be public? Just wondering.

lib/connection_manager.go
```diff
@@ -386,133 +324,79 @@ func IPToNetAddr(ipStr string, addrMgr *addrmgr.AddrManager, params *DeSoParams)
...
+	defer cmgr.mtxConnectedOutboundAddrs.RUnlock()
+	return cmgr.connectedOutboundAddrs[addrmgr.NetAddressKey(netAddr)]
+}
 
-		// If this is an outbound peer, create an outbound connection.
-		if isOutbound {
-			conn = cmgr._getOutboundConn(persistentAddr)
-		}
+func (cmgr *ConnectionManager) IsAttemptedOutboundIpAddress(netAddr *wire.NetAddress) bool {
+	return cmgr.attemptedOutboundAddrs[addrmgr.NetAddressKey(netAddr)]
```
lazynina, line 336: Do we need a mutex around attemptedOutboundAddrs or nah?

lib/connection_manager.go
```diff
@@ -700,10 +586,35 @@ func (cmgr *ConnectionManager) addPeer(pp *Peer) {
...
+}
+
+func (cmgr *ConnectionManager) SendMessage(msg DeSoMessage, peerId uint64) error {
+	if peer, ok := cmgr.connectedPeers[peerId]; ok {
+		glog.V(1).Infof("SendMessage: Sending message %v to peer %d", msg.GetMsgType().String(), peerId)
+		peer.AddDeSoMessage(msg, false)
+	} else {
+		return fmt.Errorf("SendMessage: Peer with ID %d not found", peerId)
+	}
+	return nil
```
lazynina, line 599:
```suggestion
	if peer, ok := cmgr.connectedPeers[peerId]; ok {
		glog.V(1).Infof("SendMessage: Sending message %v to peer %d", msg.GetMsgType().String(), peerId)
		peer.AddDeSoMessage(msg, false)
		return nil
	}
	return fmt.Errorf("SendMessage: Peer with ID %d not found", peerId)
```

lazynina: OR
```suggestion
	if peer, ok := cmgr.connectedPeers[peerId]; !ok {
		return fmt.Errorf("SendMessage: Peer with ID %d not found", peerId)
	}
	glog.V(1).Infof("SendMessage: Sending message %v to peer %d", msg.GetMsgType().String(), peerId)
	peer.AddDeSoMessage(msg, false)
	return nil
```

lib/pos_connection.go
```diff
@@ -0,0 +1,150 @@
...
+		attemptId:    oca.attemptId,
+		address:      oca.netAddr,
+		connection:   nil,
+		isPersistent: oca.isPersistent,
+		failed:       true,
+	}
+}
+
+func (oca *OutboundConnectionAttempt) Stop() {
+	if oca.terminated {
```
tholonious, line 122: Do you want to add a mutex around this check? If two threads call the Stop() function at the same time, then the one of them will panic when trying to close the already closed exitChan

tholonious (changes requested):

lib/connection_manager.go
AeonSw4n: Yup, we do. Good call.

lib/connection_manager.go
AeonSw4n: Ah yea good observation, I actually simplified it in a later PR. Moved it to this PR.

lib/connection_manager.go
```diff
@@ -386,159 +311,97 @@ func IPToNetAddr(ipStr string, addrMgr *addrmgr.AddrManager, params *DeSoParams)
...
+}
 
-		// We connected to the peer and it passed its version negotiation.
-		// Handle the next steps in the main loop.
-		cmgr.newPeerChan <- peer
+// _dialOutboundConnection is the internal method that spawns and initiates an OutboundConnectionAttempt, which handles the
+// connection attempt logic. It returns the attemptId of the attempt that was created.
+func (cmgr *ConnectionManager) _dialOutboundConnection(addr *wire.NetAddress, attemptId uint64, isPersistent bool) (_attemptId uint64) {
+	connectionAttempt := NewOutboundConnectionAttempt(attemptId, addr, isPersistent,
+		cmgr.params.DialTimeout, cmgr.outboundConnectionChan)
```
tholonious, line 362: You'll need a mutex around this map write too, no?

lib/network_connection.go
```diff
@@ -0,0 +1,212 @@
...
+
+	exitChan chan bool
+	status   outboundConnectionAttemptStatus
+}
+
+type outboundConnectionAttemptStatus int
+
+const (
+	outboundConnectionAttemptUnknown    outboundConnectionAttemptStatus = 0
+	outboundConnectionAttemptStarted    outboundConnectionAttemptStatus = 1
```
tholonious, line 98: nit: these are more standard const names that match the terminology we use elsewhere
```suggestion
	outboundConnectionAttemptInitialized    outboundConnectionAttemptStatus = 0
	outboundConnectionAttemptRunning      outboundConnectionAttemptStatus = 1
```

tholonious (commented):
LGTM. Just two tiny comments

PR #775, opened 2023-11-02
recursive validate func w/o side effects on chain other than adding to block index

lazynina:


lazynina:
    * **PR #758** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/758" target="_blank">Graphite</a> 
      * **PR #775** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/775" target="_blank">Graphite</a>  👈
      * **PR #761** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/761" target="_blank">Graphite</a> 
        * **PR #762** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/762" target="_blank">Graphite</a> 
          * **PR #774** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/774" target="_blank">Graphite</a> 
            * **PR #759** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/759" target="_blank">Graphite</a> 
              * **PR #765** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/765" target="_blank">Graphite</a> 
                * **PR #766** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/766" target="_blank">Graphite</a> 
                  * **PR #771** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/771" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/775?utm_source=stack-comment).

lib/pos_blockchain.go
```diff
@@ -57,66 +57,145 @@ func (bc *Blockchain) processBlockPoS(block *MsgDeSoBlock, currentView uint64, v
...
 
-	// At this point, we know that all the ancestors of this block have been successfully
-	// validated. That means we can now proceed with attempting to validate THIS block.
-
-	// Do all the sanity checks on the block.
-	// TODO: Check if err is for view > latest committed block view and <= latest uncommitted block.
-	// If so, we need to perform the rest of the validations and then add to our block index.
-	if err = bc.validateDeSoBlockPoS(block); err != nil {
-		// TODO: If validations fail, do we want to even bother storing the block? Are there any validations that
+	if err := bc.validateDeSoBlockPoS(block); err != nil {
```
tholonious, line 156: I realize you're still brainstorming names from the block validations. Consider calling this something like `isValidPoSBlock()` or `isValidBlockPoS()` so that it's clear the function just checks if the block is correct without any other side effects
```suggestion
	if err := bc.isValidPoSBlock(block); err != nil {
```

lib/pos_blockchain.go
```diff
@@ -129,42 +208,17 @@ func (bc *Blockchain) processBlockPoS(block *MsgDeSoBlock, currentView uint64, v
 	// 2. Validate QC
 	if err = bc.validateQC(block, validatorsByStake); err != nil {
```
tholonious, line 209: Similar naming suggestion here since the QC validation doesn't have side effects
```suggestion
	if err = bc.isValidPoSQuorumCertificate(block, validatorsByStake); err != nil {
```

tholonious (changes requested):

lib/pos_blockchain.go
```diff
@@ -57,66 +51,179 @@ func (bc *Blockchain) processBlockPoS(block *MsgDeSoBlock, currentView uint64, v
...
-			return false, false, nil, errors.Wrapf(err, "processBlockPoS: Problem getting parent block %v", parentBlockNode.Hash)
-		}
-		_, _, _, err = bc.processBlockPoS(blk, currentView, verifySignatures)
-		if err != nil {
-			return false, false, nil, errors.Wrapf(err, "processBlockPoS: Problem processing parent block %v", parentBlockNode.Hash)
+			// If we don't have the parent, we need to fetch it from the network.
+			// We won't handle that in here, so we should just store the block and return.
+			blockNode, storeInBlockIndexErr = bc.storeBlockInBlockIndex(block)
+			if storeInBlockIndexErr != nil {
+				return blockNode, errors.Wrapf(err, "isValidPoSBlock: Problem adding block to block index: %v", storeInBlockIndexErr)
```
tholonious, line 154: The algorithm failed here too. The resulting blockNode isn't trustworthy, so we shouldn't return it
```suggestion
				return nil, errors.Wrapf(err, "isValidPoSBlock: Problem adding block to block index: %v", storeInBlockIndexErr)
```

lib/pos_blockchain.go
```diff
@@ -57,66 +51,179 @@ func (bc *Blockchain) processBlockPoS(block *MsgDeSoBlock, currentView uint64, v
...
+			}
+			return blockNode, errors.Wrapf(err, "isValidPoSBlock: Problem getting parent block %v", parentBlockNode.Hash)
```
tholonious, line 156: Do we want to omit the error here? We were able to successfully run the validation algorithm on the block and store it. It just happens to be that we don't have the parent in the DB, so we can't validate things any further
```suggestion
			return blockNode, nil
```

lib/pos_blockchain.go
```diff
@@ -57,66 +51,179 @@ func (bc *Blockchain) processBlockPoS(block *MsgDeSoBlock, currentView uint64, v
...
-		if err = bc.storeValidateFailedBlockInBlockIndex(block); err != nil {
-			return false, false, nil, errors.Wrap(err, "processBlockPoS: Problem adding validate failed block to block index")
+		blockNode, err = bc.storeValidateFailedBlockInBlockIndex(block)
+		if err != nil {
+			return blockNode, errors.Wrap(err, "isValidPoSBlock: Problem adding validate failed block to block index")
 		}
-		return false, false, nil, errors.New("processBlockPoS: block's parent failed validation: ")
+		return blockNode, errors.New("isValidPoSBlock: block's parent failed validation: ")
 	}
 	if !parentBlockNode.IsValidated() {
```
tholonious, line 146: IMO, you can delete this block to simplify the nested checks below. Rather than checking the parentBlockNode's status manually in the blockIndex, you can fetch the parent block from DB and call `isValidBlock(parentBlock)` on it directly. This eliminates the need for checking the parent block's status twice.

Granted, it is less efficient than simply performing an in-memory check of the block index first, but Badger is quite good at caching recently used data in memory that I'm sure in practice, the performance difference is negligible. I'd bias towards starting with the more concise (less nested) solution and optimizing later if needed.

tholonious (changes requested):

tholonious (approved):

PR #778, opened 2023-11-02
Instantiate FastHotStuffEventLoop Within ConsensusController

tholonious:

PR #779, opened 2023-11-02
Handle And Broadcast Local Timeout Signal

tholonious:


tholonious:
  * **PR #778** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/778" target="_blank">Graphite</a> 
    * **PR #779** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/779" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/779?utm_source=stack-comment).

lib/pos_consensus_controller.go
```diff
@@ -62,13 +65,82 @@ func (cc *ConsensusController) HandleFastHostStuffVote(event *consensus.FastHotS
...
+		return errors.Errorf("HandleFastHostStuffTimeout: Error advancing view on timeout: %v", err)
+	}
+
+	// Construct the timeout message
+	timeoutMsg := NewMessage(MsgTypeValidatorTimeout).(*MsgDeSoValidatorTimeout)
+	timeoutMsg.MsgVersion = MsgValidatorTimeoutVersion0
+	timeoutMsg.TimedOutView = event.View
+	// TODO: Somehow we need to know the validator's ECDSA public key. Fill this out once the
+	// mapping between BLS and ECDSA keys is implemented and available in the consensus module.
+	// timeoutMsg.PublicKey = <ECDSA public key here>
```
tholonious, line 114: @AeonSw4n, @lazynina we need the validator's ECDSA public key available in this controller for timeout and vote construction. We'll similarly need it for block construction. A few of options in how we can do it:
- We pull it from the on-chain state by constructing a UtxoView at the current tip to fetch the current validator set's ValidatorEntries
- We maintain a cached version of the validator set in this controller for each block height. We need to compute the validator set for each safe block anyway when processing each incoming block
- The validator's ECDSA public key is configured along with the BLS signer metadata and passed along into the controller

My preference would be option 2. LMK if you have other thoughts

lazynina (changes requested):
LGTM - just some nits on comments.

lib/pos_consensus_controller.go
tholonious: Sounds good. I'll follow up with another PR for it

lazynina (approved):

PR #777, opened 2023-11-02
Process Block e2e testing w/ recursive validate

lazynina:

PR #776, opened 2023-11-02
commit rule testing

lazynina:

PR #780, opened 2023-11-02
sa/recursive-validation-func-pr-review-feedback

tholonious:


tholonious:
    * **PR #758** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/758" target="_blank">Graphite</a> 
      * **PR #775** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/775" target="_blank">Graphite</a> 
        * **PR #780** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/780" target="_blank">Graphite</a>  👈
          * **PR #777** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/777" target="_blank">Graphite</a> 
      * **PR #761** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/761" target="_blank">Graphite</a> 
        * **PR #762** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/762" target="_blank">Graphite</a> 
          * **PR #774** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/774" target="_blank">Graphite</a> 
            * **PR #759** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/759" target="_blank">Graphite</a> 
              * **PR #765** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/765" target="_blank">Graphite</a> 
                * **PR #766** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/766" target="_blank">Graphite</a> 
                  * **PR #771** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/771" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/780?utm_source=stack-comment).

lib/pos_blockchain.go
```diff
@@ -100,160 +101,165 @@ func (bc *Blockchain) processBlockPoS(block *MsgDeSoBlock, currentView uint64, v
...
-			}
-			return blockNode, errors.Wrapf(err, "isValidPoSBlock: Problem processing parent block %v", parentBlockNode.Hash)
+		blockNode, innerErr := bc.storeBlockInBlockIndex(block)
+		if innerErr != nil {
+			return nil, errors.Wrapf(err, "validateAndIndexBlockPoS: Problem adding block to block index: %v", innerErr)
 		}
+		return blockNode, nil
+	}
+
+	// At this point, we know the parent block is validation. We can now perform all other validations
```
lazynina, line 151:
```suggestion
	// At this point, we know the parent block is validated. We can now perform all other validations
```

lib/pos_blockchain.go
```diff
@@ -100,160 +101,165 @@ func (bc *Blockchain) processBlockPoS(block *MsgDeSoBlock, currentView uint64, v
...
-		blockNode, storeInBlockIndexErr = bc.storeBlockInBlockIndex(block)
-		if storeInBlockIndexErr != nil {
-			return blockNode, errors.Wrapf(err, "isValidPoSBlock: Problem adding block to block index: %v", storeInBlockIndexErr)
-		}
-		return blockNode, errors.Wrap(err, "isValidPoSBlock: Problem getting validator set: ")
-	}
-	// TODO(diamondhands): I don't think the TODO below makes sense given the changes but take a look.
-	// TODO: If block belongs in the next epoch or it's from an epoch far ahead in the future, we may not be able to validate its QC at all.
-	// the validator set for that epoch may be entirely different.
+	// TODO: If block belongs in the next epoch or it's from an epoch far ahead in the future, we may not be able to
```
lazynina, line 185: I think we want to connect the block before we check the QC. This way the utxo view can use the `CurrentEpochEntry` for "this block" to determine if the QC is valid. I think that why DH moved the connect block to happen here.

lib/pos_blockchain.go
```diff
@@ -100,160 +101,165 @@ func (bc *Blockchain) processBlockPoS(block *MsgDeSoBlock, currentView uint64, v
...
-		if storeInBlockIndexErr != nil {
-			return blockNode, errors.Wrapf(err, "isValidPoSBlock: Problem adding block to block index: %v", storeInBlockIndexErr)
+		// Here we want to at least store this block since we don't know if it's invalid.
+		// Hitting an error here should really never happen.
+		blockNode, innerErr := bc.storeBlockInBlockIndex(block)
+		if innerErr != nil {
+			return nil, errors.Wrapf(err, "validateAndIndexBlockPoS: Problem adding block to block index: %v", innerErr)
 		}
-		return blockNode, errors.Wrap(err, "processBlockPoS: Problem initializing UtxoView: ")
+		return blockNode, nil
```
lazynina, line 177: Since we store the block on line 153, we don't need to store it again here. I was thinking the store block on line 153 is basically saying "Okay we have a parent that is validated, we should at least store this block now". This also has the nice property of allowing us to avoid calling `storeBlockInBlockIndex` every time we encounter an error that doesn't result in the block being marked as Validate Failed.

lib/pos_blockchain.go
```diff
@@ -100,160 +101,165 @@ func (bc *Blockchain) processBlockPoS(block *MsgDeSoBlock, currentView uint64, v
...
 	// 2. Validate QC
+	validatorsByStake, err := utxoView.GetAllSnapshotValidatorSetEntriesByStake()
+	if err != nil {
+		// Here we want to at least store this block since we don't know if it's invalid.
+		// Hitting an error here should really never happen.
+		blockNode, innerErr := bc.storeBlockInBlockIndex(block)
+		if innerErr != nil {
+			return nil, errors.Wrapf(err, "validateAndIndexBlockPoS: Problem adding block to block index: %v", innerErr)
+		}
+		return blockNode, nil
```
lazynina, line 201: We'll be able to get rid of this since we go ahead and store the block on line 153
```suggestion
		return blockNode, nil
```

lib/pos_blockchain.go
```diff
@@ -646,6 +667,10 @@ func (bc *Blockchain) upsertBlockAndBlockNodeToDB(block *MsgDeSoBlock, blockNode
 			}
 		}
 
+		// TODO: if storeFullBlock = false, then we should probably remove the block from the DB? This can
+		// happen if we had a block stored in the DB but then determined that it would have failed validation.
+		// We would need to evict the block from the DB in that case.
+
```
lazynina, line 673: I think this is a later optimization we can make. I *think* it's fine, but I'd rather deal with this performance enhancement later.

lazynina (changes requested):
I also need to go through `isValidBlockPoS` and each validate function it calls to make sure ALL errors result in us marking the block as validate failed as opposed to some other error (transient badger issue, etc.)

lib/pos_blockchain.go
tholonious: Makes sense. Definitely a disk space optimization for the future

lib/pos_blockchain.go
```diff
@@ -100,160 +101,165 @@ func (bc *Blockchain) processBlockPoS(block *MsgDeSoBlock, currentView uint64, v
...
-		if err != nil {
-			return blockNode, errors.Wrap(err, "isValidPoSBlock: Problem adding block to block index: ")
-		}
-		return blockNode, errors.New("isValidPoSBlock: Parent block not found in block index. This should never happen.")
+
+	// Run the validation for the parent and update the block index with the parent's status. We first
+	// check if the parent has a cached status. If so, we use the cached status. Otherwise, we run
+	// the full validation algorithm on it, then index and an use the result.
+	parentBlockNode, err := bc.validateCachedBlockPoS(block.Header.PrevBlockHash)
+	if err != nil {
```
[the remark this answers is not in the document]
tholonious: If there's an error, it means the validation algorithm failed. In that case, it means things are running abnormally or in some unexpected state. I'd bias towards not storing the block because to minimize the side effects from this abnormal execution of the validation algo.

That's not a strongly held opinion though

lib/pos_blockchain.go
tholonious: I updated it to store the block with validate failed status. In this case, if the UtxoView can't build correctly for the block, then it means the block is bad.

lib/pos_blockchain.go
tholonious: Makes sense! If the UtxoView can't be built then there's no way we'd be able to validate the QC anyway. Good catch!

tholonious: We need to use the previous block height's validator set to validate an incoming block. The order we perform the QC validation and then the UtxoView construction is correct

lib/pos_blockchain.go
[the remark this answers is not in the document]
lazynina: Seems fine either way to me. Basically the caller of ProcessBlockPoS will need to request that we try to process this block again at a future time.

lib/pos_blockchain.go
```diff
@@ -100,160 +101,158 @@ func (bc *Blockchain) processBlockPoS(block *MsgDeSoBlock, currentView uint64, v
...
 	//   - Add utility to UtxoView to fetch the validator set given an arbitrary block height. If we can't fetch the
 	//     validator set for the block, then we reject it (even if it later turns out to be a valid block)
 	//   - Add block to the block index before QC validation such that even if we aren't able to fetch the validator
 	//     set for the block, we can at least store it locally.
 	// 2. Validate QC
+	validatorsByStake, err := utxoView.GetAllSnapshotValidatorSetEntriesByStake()
+	if err != nil {
+		// This should never happen. If the parent is validated and extends from the tip, then we should
+		// be able to fetch the validator set at its block height for it. This failure can only happen due
+		// to transient or badger issues. We return the block node as-is as a best effort thing.
```
lazynina, line 188:
```suggestion
		// to transient badger issues. We return the block node as-is as a best effort thing.
```

PR #782, opened 2023-11-03
DeSo Tokens Lockups

poolcoke:


lib/block_producer.go
```diff
@@ -225,9 +225,7 @@ func (desoBlockProducer *DeSoBlockProducer) _getBlockTemplate(publicKey []byte)
 			if err != nil {
 				return nil, nil, nil, errors.Wrapf(err, "Error copying UtxoView: ")
```
diamondhands0, line 226: Go through and cut all long lines down.

diamondhands0: Won't this error if you pass 0 as the timestamp instead of what you expect to actually put in the block as the timestamp?

lib/block_view.go
```diff
@@ -520,6 +532,29 @@ func (bav *UtxoView) CopyUtxoView() (*UtxoView, error) {
...
+	newView.LockedBalanceEntryKeyToLockedBalanceEntry = make(map[LockedBalanceEntryKey]*LockedBalanceEntry,
+		len(bav.LockedBalanceEntryKeyToLockedBalanceEntry))
+	for entryKey, entry := range bav.LockedBalanceEntryKeyToLockedBalanceEntry {
+		newEntry := *entry
+		newView.LockedBalanceEntryKeyToLockedBalanceEntry[entryKey] = &newEntry
+	}
+
+	// Copy the LockupYieldCurvePoints
+	newView.PKIDToLockupYieldCurvePointKeyToLockupYieldCurvePoints =
+		make(map[PKID]map[LockupYieldCurvePointKey]*LockupYieldCurvePoint, len(bav.PKIDToLockupYieldCurvePointKeyToLockupYieldCurvePoints))
```
diamondhands0, line 545: Shorten lines like these. Your editor should have a line in it that tells you what the cutoff is. When lines are long like this it makes reviewing the code annoying for others.

lib/block_view_access_group_members_test.go
```diff
@@ -1764,14 +1764,7 @@ func TestAccessGroupMembersTxnWithDerivedKey(t *testing.T) {
...
 		// Connect txn.
-		utxoOps, _, _, _, err := utxoView.ConnectTransaction(
-			txn,
-			txn.Hash(),
-			getTxnSize(*txn),
-			testMeta.savedHeight,
-			true,
-			false,
-		)
+		utxoOps, _, _, _, err := utxoView.ConnectTransaction(txn, txn.Hash(), getTxnSize(*txn), testMeta.savedHeight, 0, true, false)
```
diamondhands0, line 1767: Fix long lines here too please.

lib/block_view_access_group_test.go
```diff
@@ -613,14 +613,7 @@ func TestAccessGroupTxnWithDerivedKey(t *testing.T) {
...
 		// Connect txn.
-		utxoOps, _, _, _, err := utxoView.ConnectTransaction(
-			txn,
-			txn.Hash(),
-			getTxnSize(*txn),
-			testMeta.savedHeight,
-			true,
-			false,
-		)
+		utxoOps, _, _, _, err := utxoView.ConnectTransaction(txn, txn.Hash(), getTxnSize(*txn), testMeta.savedHeight, 0, true, false)
```
diamondhands0, line 616: Fix long lines everywhere.

lib/block_view_lockups.go
```diff
@@ -0,0 +1,2178 @@
...
+		LockupDurationNanoSecs: lockupYieldCurvePoint.LockupDurationNanoSecs,
+	}
+}
+
+// DeSoEncoder Interface Implementation for LockupYieldCurvePoint
+
+func (lockupYieldCurvePoint *LockupYieldCurvePoint) RawEncodeWithoutMetadata(blockHeight uint64, skipMetadata ...bool) []byte {
+	var data []byte
+	data = append(data, EncodeToBytes(blockHeight, lockupYieldCurvePoint.ProfilePKID, skipMetadata...)...)
+	data = append(data, UintToBuf(uint64(lockupYieldCurvePoint.LockupDurationNanoSecs))...)
```
diamondhands0, line 250: Idk what happens if you call uint64(negativeInt64). Does it just work or does it panic?

lib/block_view_lockups.go
```diff
@@ -0,0 +1,2178 @@
...
+	if len(dbYieldCurvePoints) > 0 {
+		// Check if there's a yield curve in the view for the associated profile.
+		if _, mapInView := bav.PKIDToLockupYieldCurvePointKeyToLockupYieldCurvePoints[*profilePKID]; !mapInView {
+			bav.PKIDToLockupYieldCurvePointKeyToLockupYieldCurvePoints[*profilePKID] =
+				make(map[LockupYieldCurvePointKey]*LockupYieldCurvePoint)
+		}
+
+		// Check if any of the points needs to be cached in the view.
+		for _, yieldCurvePoint := range dbYieldCurvePoints {
+			if _, pointInView :=
```
diamondhands0, line 371: When a lookup is really long like this, break it into two lines. Like this:
```
_, pointInView := bav.PKIDToLockupYieldCurvePointKeyToLockupYieldCurvePoints[*profilePKID][yieldCurvePoint.ToMapKey()]
if !pointInView {
    bav._setLockupYieldCurvePoint(yieldCurvePoint)
}
```

lib/block_view_lockups.go
```diff
@@ -0,0 +1,2178 @@
...
+	if _, pointsInView := bav.PKIDToLockupYieldCurvePointKeyToLockupYieldCurvePoints[*profilePKID]; pointsInView {
+		for _, lockupYieldCurvePoint := range bav.PKIDToLockupYieldCurvePointKeyToLockupYieldCurvePoints[*profilePKID] {
+			// Ensure the point is not deleted.
+			if lockupYieldCurvePoint.isDeleted {
+				continue
+			}
+
+			// Check for nil pointer cases.
+			if lockupYieldCurvePoint.LockupDurationNanoSecs < lockupDuration && leftLockupPoint == nil {
+				leftLockupPoint = lockupYieldCurvePoint
```
diamondhands0, line 388: Why do you call Copy() below but not here? Seems like this could be a bug.

lib/block_view_lockups.go
```diff
@@ -0,0 +1,2178 @@
...
+
+			// Check if the point is "more right" than the current right point.
+			if lockupYieldCurvePoint.LockupDurationNanoSecs >= lockupDuration &&
+				lockupYieldCurvePoint.LockupDurationNanoSecs < rightLockupPoint.LockupDurationNanoSecs {
+				rightLockupPoint = lockupYieldCurvePoint.Copy()
+			}
+		}
+	}
+
+	return leftLockupPoint, rightLockupPoint, nil
```
diamondhands0, line 408: This function looks good now.

lib/block_view_lockups.go
```diff
@@ -0,0 +1,2178 @@
...
+	//    The point (LockupYieldDurationNanoSecs, LockupYieldAPYBasisPoints)
+	//    is added to the profile's yield curve. If a point with the same duration already exists
+	//    on the profile's yield curve, it will be updated with the new yield.
+	//    Note if LockupYieldDurationNanoSecs=0, nothing is modified or added at t=0.
+	// Assuming RemoveYieldCurvePoint is true:
+	//    The point (LockupYieldDurationNanoSecs, XXX) is removed from the profile's yield curve.
+	//    Note that LockupYieldAPYBasisPoints is ignored in this transaction.
+	//
+	// By setting LockupYieldDurationNanoSecs to zero, the yield curve attached to the profile
+	// is left unmodified. In any UpdateDAOCoinLockupParams transaction looking to modify only
```
diamondhands0, line 485: Fix instances where you have DAOCoin.

lib/block_view_lockups.go
```diff
@@ -0,0 +1,2178 @@
...
+	LockupYieldDurationNanoSecs int64
+	LockupYieldAPYBasisPoints   uint64
+	RemoveYieldCurvePoint       bool
+
+	// When NewLockupTransferRestrictions is set true, the TransferRestrictionStatus specified
+	// in the transaction is updated in the transactor's profile for locked coins.
+	// Any subsequent transfers utilizing the transactor's locked coins are validated against
+	// the updated locked transfer restriction status.
+	NewLockupTransferRestrictions   bool
+	LockupTransferRestrictionStatus TransferRestrictionStatus
```
diamondhands0, line 496: So just so I understand: When it's Unrestricted, that means I can technically transfer locked balances to other users. Is that roughly correct?

lib/block_producer.go
diamondhands0: After reading the _connect functions, I think you definitely need to set the actual block timestamp here FYI.

lib/block_view_lockups.go
```diff
@@ -0,0 +1,2178 @@
...
+	var transactorBalanceNanos256 *uint256.Int
+	var prevTransactorBalanceEntry *BalanceEntry
+	var prevCoinEntry *CoinEntry
+	if profilePKID.IsZeroPKID() {
+		// NOTE: When spending balances, we need to check for immature block rewards. Since we don't have
+		//       the block rewards yet for the current block, we subtract one from the current block height
+		//        when spending balances.
+
+		// Check the DeSo balance of the user.
+		transactorBalanceNanos, err := bav.GetSpendableDeSoBalanceNanosForPublicKey(txn.PublicKey, blockHeight-1)
```
diamondhands0, line 750: Hmm... Do we do this in other places? I thought you don't need to subtract 1 for this to work but I can't remember. Just triple check that this is the same behavior as everywhere else, or that there's a good reason why it's different.

lib/block_view_lockups.go
```diff
@@ -0,0 +1,2178 @@
...
+		if txMeta.LockupAmountBaseUnits.Gt(transactorBalanceNanos256) {
+			return 0, 0, nil,
+				errors.Wrap(RuleErrorCoinLockupInsufficientCoins, "_connectCoinLockup")
+		}
+
+		// We store the previous transactor balance entry in the event we need to revert the transaction.
+		prevTransactorBalanceEntry = transactorBalanceEntry.Copy()
+
+		// Spend the transactor's DAO coin balance.
+		transactorBalanceEntry.BalanceNanos =
```
diamondhands0, line 791: I think it's a good idea to Copy() it before you modify it and set it. It just prevents weird stuff from happening in general.

diamondhands0: Eh you copied the Prev one so it's probably fine.

lib/block_view_lockups.go
```diff
@@ -0,0 +1,2178 @@
...
+		bav._setDAOCoinBalanceEntryMappings(transactorBalanceEntry)
+
+		// Create a copy of the associated CoinEntry in the event we must roll back the transaction.
+		prevCoinEntry = profileEntry.DAOCoinEntry.Copy()
+
+		// Update CoinsInCirculation and NumberOfHolders associated with the DAO coin balance.
+		profileEntry.DAOCoinEntry.CoinsInCirculationNanos = *uint256.NewInt().Sub(
+			&profileEntry.DAOCoinEntry.CoinsInCirculationNanos,
+			txMeta.LockupAmountBaseUnits)
+		if transactorBalanceEntry.BalanceNanos.IsZero() {
```
diamondhands0, line 802: I think you'd be extra safe if you add `&& !prevTransactorBalanceEntry.IsZero()`. This covers an edge case where the balance you fetched was zero. It can't ever happen because you check that you're locking up >0 coins, but still good to be extra safe.

lib/block_view_lockups.go
```diff
@@ -0,0 +1,2178 @@
...
+	// The transactor earns yield only for the lockup duration specified by the left yield curve point but will
+	// be unable to unlock the coins until the transaction specified lockup duration expires.
+	txnYieldBasisPoints := uint64(0)
+	txnYieldEarningDurationNanoSecs := int64(0)
+	if leftYieldCurvePoint != nil &&
+		leftYieldCurvePoint.LockupDurationNanoSecs < lockupDurationNanoSeconds {
+		txnYieldBasisPoints = leftYieldCurvePoint.LockupYieldAPYBasisPoints
+		txnYieldEarningDurationNanoSecs = leftYieldCurvePoint.LockupDurationNanoSecs
+	}
+	if rightYieldCurvePoint != nil &&
```
diamondhands0, line 830: You could have avoided this second edge-case if you made the left point <= and the right point strictly >. No need to fix it just saying.

lib/block_view_lockups.go
```diff
@@ -0,0 +1,2178 @@
...
+	prevSenderLockedBalanceEntry := senderLockedBalanceEntry.Copy()
+
+	// Check that the sender's balance entry has sufficient balance.
+	if txMeta.LockedCoinsToTransferBaseUnits.Gt(&senderLockedBalanceEntry.BalanceBaseUnits) {
+		return 0, 0, nil, errors.Wrapf(RuleErrorCoinLockupTransferInsufficientBalance,
+			"_connectCoinLockupTransfer")
+	}
+
+	// Credit the sender's balance entry.
+	senderLockedBalanceEntry.BalanceBaseUnits = *uint256.NewInt().Sub(
```
diamondhands0, line 1428: I think you mean "Debit the sender's balance entry" in the comment.

lib/block_view_lockups.go
```diff
@@ -0,0 +1,2178 @@
...
+		hodlerPKID, profilePKID, blockTimestamp)
+	if err != nil {
+		return 0, 0, nil, errors.Wrapf(err, "_connectCoinUnlock")
+	}
+	if len(unlockableLockedBalanceEntries) == 0 {
+		return 0, 0, nil, errors.Wrapf(RuleErrorCoinUnlockNoUnlockableCoinsFound,
+			"_connectCoinUnlock")
+	}
+
+	// Unlock coins until the amount specified by the transaction is deducted.
```
diamondhands0, line 1665: What is this comment saying? It looks like you unlock coins until you run out of balance entries, not until an amount is deducted.

lib/block_view_lockups.go
```diff
@@ -0,0 +1,2178 @@
...
+		prevCoinEntry = profileEntry.DAOCoinEntry.Copy()
+		newCoinsInCirculationNanos, err := SafeUint256().Add(
+			&profileEntry.DAOCoinEntry.CoinsInCirculationNanos,
+			unlockedBalance)
+		if err != nil {
+			return 0, 0, nil, errors.Wrapf(RuleErrorCoinUnlockCausesCoinsInCirculationOverflow,
+				"_connectCoinUnlock")
+		}
+		profileEntry.DAOCoinEntry.CoinsInCirculationNanos = *newCoinsInCirculationNanos
+		if prevTransactorBalanceEntry.BalanceNanos.IsZero() {
```
diamondhands0, line 1725: Is it possible that the previous balance was zero but you ended up unlocking nothing and so you increment this erroneously? If you just check that the AFTER balance is >0 in this if you're good.

lib/db_utils.go
```diff
@@ -10661,6 +10680,483 @@ func DBDeletePostAssociationWithTxn(txn *badger.Txn, snap *Snapshot, association
...
+	rr := bytes.NewReader(leftLockupYieldCurvePointBytes)
+	DecodeFromBytes(leftLockupYieldCurvePointObj, rr)
+	if !leftLockupYieldCurvePointObj.ProfilePKID.Eq(profilePKID) {
+		return nil
+	}
+
+	return leftLockupYieldCurvePointObj
+}
+
+func DBGetRightLockupYieldCurvePointWithTxn(txn *badger.Txn, snap *Snapshot, profilePKID *PKID,
```
diamondhands0, line 11090: It's hard to be 100% sure all this is working so I'm just going to assume you tested it all extremely thoroughly.

lib/legacy_mempool.go
```diff
@@ -839,17 +839,15 @@ func (mp *DeSoMempool) addTransaction(
 
 	// Add it to the universal view. We assume the txn was already added to the
 	// backup view.
-	_, _, _, _, err = mp.universalUtxoView._connectTransaction(mempoolTx.Tx, mempoolTx.Hash, int64(mempoolTx.TxSizeBytes), height,
-		false /*verifySignatures*/, false /*ignoreUtxos*/)
+	_, _, _, _, err = mp.universalUtxoView._connectTransaction(mempoolTx.Tx, mempoolTx.Hash, int64(mempoolTx.TxSizeBytes), height, 0, false, false)
```
diamondhands0, line 842: Here I suspect that using 0 as the value may cause issues versus, for example, using the current timestamp.

lib/network.go
```diff
@@ -6578,6 +6665,30 @@ func (tsl *TransactionSpendingLimit) FromBytes(blockHeight uint64, rr *bytes.Rea
 
 	// StakeLimitMap, UnstakeLimitMap, and UnlockStakeLimitMap, gated by the encoder migration.
 	if MigrationTriggered(blockHeight, ProofOfStake1StateSetupMigration) {
+		// LockupLimitMap
```
diamondhands0, line 6668: Not reviewing this thoroughly. Make sure it's tested.

lib/block_view_lockups.go
```diff
@@ -0,0 +1,2178 @@
...
+	default:
+		return LockupLimitScopeTypeUndefined
+	}
+}
+
+func (bav *UtxoView) _checkLockupTxnSpendingLimitAndUpdateDerivedKey(
+	derivedKeyEntry DerivedKeyEntry,
+	profilePublicKey *PublicKey,
+	lockupOperation LockupLimitOperation,
+) (DerivedKeyEntry, error) {
```
diamondhands0, line 2114: I didn't review any of these spending limit checkers too thoroughly. Just make sure you've tested them thoroughly.

lib/legacy_mempool.go
diamondhands0: Is there any downside to using the current timestamp here rather than zero? I didn't see any. I don't think we've determined how the cutover to the new mempool is going to happen yet, and using the current timestamp will likely make your testing easier.

lib/block_view_lockups.go
diamondhands0: Ugh I looked around and it seems like we call it with both blockHeight and blockHeight-1, and it's not clear which is actually the one we want. I think blockHeight (without the -1) is the more robust check because it will exclude block rewards from the tip. But I'm not 100% sure. @lazynina do you know if there's a good reason to call GetSpendableDeSoBalanceNanosForPublicKey with blockHeight-1 in UtxoView?

lib/block_view_lockups.go
```diff
@@ -0,0 +1,2179 @@
...
+	utxoOpsForTxn = append(utxoOpsForTxn, utxoOpsForBasicTransfer...)
+
+	// Grab the txn metadata.
+	txMeta := txn.TxnMeta.(*UpdateCoinLockupParamsMetadata)
+
+	// Get the profilePKID from the transactor public key.
+	var profilePKID *PKID
+	_, updaterIsParamUpdater := GetParamUpdaterPublicKeys(blockHeight, bav.Params)[MakePkMapKey(txn.PublicKey)]
+	if updaterIsParamUpdater {
+		profilePKID = ZeroPKID.NewPKID()
```
lazynina, line 1109: so any param updater cannot have a coin that has any lock up params? This seems okay, but we should make this explicit here with a comment.

lib/block_view_lockups.go
```diff
@@ -0,0 +1,2179 @@
...
+			return 0, 0, nil,
+				errors.Wrap(RuleErrorCoinLockupTransferOnNonExistentProfile, "_connectCoinLockupTransfer")
+		}
+	}
+
+	// Fetch PKIDs for the recipient, sender, and profile.
+	var senderPKID *PKID
+	if _, senderIsParamUpdater :=
+		GetParamUpdaterPublicKeys(blockHeight, bav.Params)[MakePkMapKey(txn.PublicKey)]; senderIsParamUpdater {
+		senderPKID = ZeroPKID.NewPKID()
```
lazynina, line 1389: wait why do we do this? This seems odd. If the sender is a param updater, we set the sender to the zero PKID? huh? Do you mean to be doing this for ProfilePublicKey? This seems like a bug.

lib/block_view_lockups.go
```diff
@@ -0,0 +1,2179 @@
...
+	}
+
+	// Check if transfers are limited to profile owner only.
+	if transferRestrictionStatus == TransferRestrictionStatusProfileOwnerOnly && !profilePKID.Eq(senderPKID) {
+		return 0, 0, nil,
+			errors.Wrapf(RuleErrorCoinLockupTransferRestrictedToProfileOwner, "_connectCoinLockupTransfer")
+	}
+
+	// Check if the transfers are limited to DAO members only.
+	// Here, a "DAO member" is anyone who holds either unlocked or locked DAO coins associated with the profile.
```
lazynina, line 1464: Do we need to update DAO Coin transfer logic as well?

lazynina (changes requested):

lib/block_view_lockups.go
```diff
@@ -0,0 +1,2193 @@
...
+	return &CoinUnlockMetadata{}
+}
+
+//
+// CoinLockup Transaction Logic
+//
+
+func (bav *UtxoView) _connectCoinLockup(
+	txn *MsgDeSoTxn, txHash *BlockHash,
+	blockHeight uint32, blockTimestamp int64,
```
tholonious, line 670: Always include the units of measurement in variable names like this
```suggestion
	blockHeight uint32, blockTimestampNanoSecs int64,
```

lib/block_view.go
```diff
@@ -3257,23 +3356,21 @@ func (bav *UtxoView) ValidateDiamondsAndGetNumDeSoNanos(
...
 }
 
-func (bav *UtxoView) ConnectTransaction(txn *MsgDeSoTxn, txHash *BlockHash,
-	txnSizeBytes int64,
-	blockHeight uint32, verifySignatures bool, ignoreUtxos bool) (
-	_utxoOps []*UtxoOperation, _totalInput uint64, _totalOutput uint64,
-	_fees uint64, _err error) {
+func (bav *UtxoView) ConnectTransaction(
+	txn *MsgDeSoTxn, txHash *BlockHash, txnSizeBytes int64,
+	blockHeight uint32, blockTimestamp int64, verifySignatures bool,
```
tholonious, line 3361: Always include the units of measurement in variable/parameter names. The ambiguity can lead to bugs down the line if someone misinterprets the units
```suggestion
	blockHeight uint32, blockTimestampNanoSecs int64, verifySignatures bool,
```

lib/block_view_lockups.go
```diff
@@ -0,0 +1,2193 @@
...
+	return nil
+}
+
+//
+// CoinUnlock Transaction Logic
+//
+
+func (bav *UtxoView) _connectCoinUnlock(
+	txn *MsgDeSoTxn, txHash *BlockHash,
+	blockHeight uint32, blockTimestamp int64,
```
tholonious, line 1606:
```suggestion
	blockHeight uint32, blockTimestampNanoSecs int64,
```

lib/pos_block_producer.go
```diff
@@ -108,7 +108,8 @@ func (pbp *PosBlockProducer) createBlockTemplate(latestBlockView *UtxoView, newB
 
 // createBlockWithoutHeader is a helper function used by createBlockTemplate. It constructs a partially filled out
 // block with Fee-Time ordered transactions. The returned block all its contents filled, except for the header.
-func (pbp *PosBlockProducer) createBlockWithoutHeader(latestBlockView *UtxoView, newBlockHeight uint64) (BlockTemplate, error) {
+func (pbp *PosBlockProducer) createBlockWithoutHeader(latestBlockView *UtxoView,
+	newBlockHeight uint64, newBlockTimestamp uint64) (BlockTemplate, error) {
```
tholonious, line 112:
```suggestion
	newBlockHeight uint64, newBlockTimestampNanoSecs uint64) (BlockTemplate, error) {
```

lib/pos_block_producer.go
```diff
@@ -139,7 +141,8 @@ func (pbp *PosBlockProducer) createBlockWithoutHeader(latestBlockView *UtxoView,
 }
 
 // getBlockTransactions is used to retrieve fee-time ordered transactions from the mempool.
-func (pbp *PosBlockProducer) getBlockTransactions(latestBlockView *UtxoView, newBlockHeight uint64, maxBlockSizeBytes uint64) (
+func (pbp *PosBlockProducer) getBlockTransactions(latestBlockView *UtxoView, newBlockHeight uint64,
+	newBlockTimestamp uint64, maxBlockSizeBytes uint64) (
```
tholonious, line 145:
```suggestion
	newBlockTimestampNanoSecs uint64, maxBlockSizeBytes uint64) (
```

lib/block_view_lockups_test.go
```diff
@@ -0,0 +1,3333 @@
...
+	"github.com/btcsuite/btcd/btcec"
+	"github.com/dgraph-io/badger/v3"
+	"github.com/holiman/uint256"
+	"github.com/pkg/errors"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+func TestForkHeight(t *testing.T) {
```
tholonious, line 13: nit: ambiguous name
```suggestion
func TestCoinLockupForkHeight(t *testing.T) {
```

lib/db_utils.go
```diff
@@ -559,7 +559,21 @@ type DBPrefixes struct {
...
+	PrefixLockedBalanceEntryByHODLerPKIDProfilePKIDUnlockTimestampNanoSecs []byte `prefix_id:"[91]" is_state:"true"`
+
+	// PrefixLockupYieldCurvePointByProfilePKIDAndDurationNanoSecs:
+	// Retrieves a LockupYieldCurvePoint.
+	// The structure of the key enables quick lookups for a (ProfilePKID, Duration) pair as well
+	// as quick construction of yield curve plots over time.
+	// Prefix, <ProfilePKID [33]byte>, <LockupDurationNanoSecs int64> -> <LockupYieldCurvePoint>
+	PrefixLockupYieldCurvePointByProfilePKIDAndDurationNanoSecs []byte `prefix_id:"[92]" is_state:"true"`
+
+	// NEXT_TAG: 93
```
tholonious, line 576: The recently merged state syncer code adds new prefixes that precede these ones. Once you merge the state syncer code into your feature branch, you'll need to update the prefix values 91 and 92 to 93 and 94 respectively. The next available tag will be 95

lib/db_utils.go
```diff
@@ -800,6 +814,12 @@ func StatePrefixToDeSoEncoder(prefix []byte) (_isEncoder bool, _encoder DeSoEnco
 	} else if bytes.Equal(prefix, Prefixes.PrefixSnapshotStakeToRewardByValidatorAndStaker) {
 		// prefix_id:"[90]"
 		return true, &StakeEntry{}
+	} else if bytes.Equal(prefix, Prefixes.PrefixLockedBalanceEntryByHODLerPKIDProfilePKIDUnlockTimestampNanoSecs) {
+		// prefix_id:"[91]"
+		return true, &LockedBalanceEntry{}
+	} else if bytes.Equal(prefix, Prefixes.PrefixLockupYieldCurvePointByProfilePKIDAndDurationNanoSecs) {
+		// prefix_id:"[92]"
+		return true, &LockupYieldCurvePoint{}
```
tholonious, line 822: Same comment as above. You'll need to update these prefix values from 91 & 92 to 93 & 94

tholonious (changes requested):

lazynina (approved):
Unresolved some previous conversations that appear to not be addressed, but overall looks good. The conversations are all minor.

lib/db_utils.go
```diff
@@ -10661,6 +10680,483 @@ func DBDeletePostAssociationWithTxn(txn *badger.Txn, snap *Snapshot, association
...
+		lockedBalanceEntry, err := DecodeDeSoEncoder(&LockedBalanceEntry{}, rr)
+		if err != nil {
+			return nil, errors.Wrapf(err, "DBGetUnlockableLockedBalanceEntriesWithTxn: "+
+				"error decoding LockedBalanceEntry: ")
+		}
+
+		// This check is redundant. It's included to be extra safe only unlockable locked balance entries are included.
+		if lockedBalanceEntry.UnlockTimestampNanoSecs < currentTimestampUnixNanoSecs {
+			lockedBalanceEntries = append(lockedBalanceEntries, lockedBalanceEntry)
+		}
```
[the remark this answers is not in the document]
lazynina: Works for me.

lib/db_utils.go
```diff
@@ -10661,6 +10680,483 @@ func DBDeletePostAssociationWithTxn(txn *badger.Txn, snap *Snapshot, association
...
+	// Create an iterator.
+	opts := badger.DefaultIteratorOptions
+	iterator := txn.NewIterator(opts)
+	defer iterator.Close()
+
+	// Store matching LockupYieldCurvePoints to return.
+	var lockupYieldCurvePoints []*LockupYieldCurvePoint
+
+	// Loop.
+	for iterator.Seek(startKey); iterator.ValidForPrefix(validKey); iterator.Next() {
```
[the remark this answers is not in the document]
lazynina: works for me.

tholonious (approved):

PR #786, opened 2023-11-03
add get safe blocks function

lazynina:

PR #785, opened 2023-11-03
Add blockIndexByHeight to Blockchain struct, rename blockIndex to blockIndexByHash

lazynina:

PR #784, opened 2023-11-03
check validate failed status, proposed in view and height when getting lineage to determine validity

lazynina:

PR #783, opened 2023-11-03
refactor validate block leader so we can not special case rule errors

lazynina:

PR #787, opened 2023-11-03
try to validate orphans after running commit rule and before returning

lazynina:


lazynina:
    * **PR #783** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/783" target="_blank">Graphite</a> 
      * **PR #784** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/784" target="_blank">Graphite</a> 
          * **PR #777** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/777" target="_blank">Graphite</a> 
            * **PR #785** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/785" target="_blank">Graphite</a> 
              * **PR #786** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/786" target="_blank">Graphite</a> 
                * **PR #787** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/787" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/787?utm_source=stack-comment).

lazynina:
    * **PR #783** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/783" target="_blank">Graphite</a> 
      * **PR #784** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/784" target="_blank">Graphite</a> 
          * **PR #777** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/777" target="_blank">Graphite</a> 
            * **PR #785** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/785" target="_blank">Graphite</a> 
              * **PR #786** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/786" target="_blank">Graphite</a> 
                * **PR #787** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/787" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/787?utm_source=stack-comment).

lib/pos_blockchain.go
```diff
@@ -76,17 +76,33 @@ func (bc *Blockchain) processBlockPoS(block *MsgDeSoBlock, currentView uint64, v
 		return false, false, nil, errors.Wrap(err, "processBlockPoS: Problem applying new tip: ")
 	}
 
-	// If the incoming block is not applied as the new tip, we return early.
-	if !appliedNewTip {
-		return false, false, nil, nil
+	// 5. Commit grandparent if possible. Only need to do this if we applied a new tip.
+	if appliedNewTip {
+		if err = bc.runCommitRuleOnBestChain(); err != nil {
+			return false, false, nil, errors.Wrap(err, "processBlockPoS: error committing grandparents: ")
```
tholonious, line 82: nit: technically this can commit any number of blocks that are ancestors of the grandparent block
```suggestion
			return false, false, nil, errors.Wrap(err, "processBlockPoS: error running commit rule: ")
```

lib/pos_blockchain.go
```diff
@@ -76,17 +76,33 @@ func (bc *Blockchain) processBlockPoS(block *MsgDeSoBlock, currentView uint64, v
...
+		}
 	}
 
-	// 5. Commit grandparent if possible.
-	if err = bc.runCommitRuleOnBestChain(); err != nil {
-		return false, false, nil, errors.Wrap(err, "processBlockPoS: error committing grandparents: ")
+	// Now that we've processed this block, we check for any orphans that
+	// are children of this block and try to process them.
```
tholonious, line 87:
```suggestion
	// Now that we've processed this block, we check for any blocks that were previously
	// stored as orphans, which are children of this block. We can  process them now.
```

lib/pos_blockchain.go
```diff
@@ -76,17 +76,33 @@ func (bc *Blockchain) processBlockPoS(block *MsgDeSoBlock, currentView uint64, v
...
+	orphansAtNextHeight := bc.blockIndexByHeight[blockNode.Height+1]
+	for _, orphan := range orphansAtNextHeight {
+		if orphan.Header.PrevBlockHash.IsEqual(blockNode.Hash) && orphan.IsStored() && !orphan.IsValidated() {
+			// TODO: how do we want to handle failures when validating orphans.
```
tholonious, line 91: Seems fine to log the failure and move on. Processing orphans at this point is a best effort thing.

lib/pos_blockchain.go
```diff
@@ -76,17 +76,33 @@ func (bc *Blockchain) processBlockPoS(block *MsgDeSoBlock, currentView uint64, v
...
+			var orphanBlock *MsgDeSoBlock
+			orphanBlock, err = GetBlock(orphan.Hash, bc.db, bc.snapshot)
+			if err != nil {
+				glog.Errorf("processBlockPoS: Problem getting orphan block %v", orphan.Hash)
+				continue
+			}
+			if _, err = bc.validateAndIndexBlockPoS(orphanBlock); err != nil {
```
tholonious, line 98: Do we want to call processBlockPoS here instead? The goal is to extend the chain as far as possible by aggressively processing previous orphans when possible
```suggestion
			if _, err = bc.processBlockPoS(orphanBlock, currentView); err != nil {
```

lazynina: Interesting, I thought we wanted to prevent side effects like this, but I'm open to it. Need to think of any unintended consequences here. For example, what if there are multiple blocks at the next height with this block as its parent? I guess future calls to processBlockPoS will resolve the reorgs for us.

tholonious: I don't have a good answer, and would be fine wither approach here. Maybe DH has a stronger opinion here.

In either case, future reorgs triggered by the longest chain would resolve things. That said, it seems somewhat worthwhile here to do the best effort thing and see how far we can extend the chain as far as possible by aggressively and naively processing orphans. This could be especially useful if we download blocks out of order from peers, which can definitely happen.

lazynina: I think it makes sense to call `processBlockPoS` then. I've updated the PR to do so.

lib/pos_blockchain.go
```diff
@@ -76,17 +77,32 @@ func (bc *Blockchain) processBlockPoS(block *MsgDeSoBlock, currentView uint64, v
...
+	blockNodesAtNextHeight := bc.blockIndexByHeight[uint64(blockNode.Height)+1]
+	for _, blockNodeAtNextHeight := range blockNodesAtNextHeight {
+		if blockNodeAtNextHeight.Header.PrevBlockHash.IsEqual(blockNode.Hash) && blockNodeAtNextHeight.IsStored() && !blockNodeAtNextHeight.IsValidated() {
+			var orphanBlock *MsgDeSoBlock
+			orphanBlock, err = GetBlock(blockNodeAtNextHeight.Hash, bc.db, bc.snapshot)
+			if err != nil {
+				glog.Errorf("processBlockPoS: Problem getting orphan block %v", blockNodeAtNextHeight.Hash)
+				continue
+			}
+			if _, _, _, err = bc.processBlockPoS(orphanBlock, currentView, verifySignatures); err != nil {
```
tholonious, line 98: I think you still need to capture whether one of the recursive processBlockPoS calls applied the new tip, no? If one of the recursive calls applied a new tip, then it means the current execution of processBlockPoS resulted in a new tip.

And we need to reflect that via the `appliedNewTip` return value at the bottom of this function. I'd pretty define the `appliedNewTip` return value to mean: "we processed this block and as a result, the tip for the best chain has changed to some new tip"

lazynina: oooh good call. will update.

tholonious (approved):

PR #800, opened 2023-11-07
Address DH feedback

lazynina:

PR #802, opened 2023-11-08
All Block rewards mature immediately - no need to write block rewards to index

lazynina:
The block rewards index is also part of the state synced by hypersync, so this also solves the issue of storing blocks that are not part of the best chain.

### DIFF
--- a/lib/blockchain_test.go
+++ b/lib/blockchain_test.go
@@ -5,6 +5,7 @@ import (
 	"flag"
 	"fmt"
 	"log"
+	"math"
 	"math/big"
 	"math/rand"
 	"os"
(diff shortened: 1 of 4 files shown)


lazynina:
    * **PR #800** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/800" target="_blank">Graphite</a> 
      * **PR #802** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/802" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/802?utm_source=stack-comment).

lib/blockchain_test.go
```diff
@@ -451,7 +452,7 @@ func _getBalanceWithView(t *testing.T, chain *Blockchain, utxoView *UtxoView, pk
 
 func TestBalanceModelBlockTests(t *testing.T) {
 	setBalanceModelBlockHeights(t)
-
+	DeSoTestnetParams.ForkHeights.ProofOfStake2ConsensusCutoverBlockHeight = math.MaxUint32
```
tholonious, line 455: Does it make sense to move this block height update into `setBalanceModelBlockHeights` so it can be reverted in the cleanup? It seemed like part of the cause for the sporadic CI failures were the place where we set fork heights before a test, but never reverted them.

lazynina: the issue is that we started setting the `ProofOfStake2ConsensusCutoverBlockHeight` in `setBalanceModelBlockHeights` and this test assumes we're still using PoW blocks. I think it's cleaner to just manually set the fork height to a max uint32 here instead of creating a separate `setPoSBlockHeights` helper function. Lmk what you think. I can add a comment above that line saying 

```
This test assumes we're using PoW blocks and thus we need to set the PoS cutover fork height to some distant future height
```

tholonious: Sounds good to me

tholonious (approved):

PR #803, opened 2023-11-08
Enhanced orphan validations

lazynina:

PR #804, opened 2023-11-08
Implement hasValidProposerPartialSig to validate the proposer vote partial signature in the header of a block

lazynina:


lazynina:
    * **PR #800** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/800" target="_blank">Graphite</a> 
      * **PR #802** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/802" target="_blank">Graphite</a> 
        * **PR #803** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/803" target="_blank">Graphite</a> 
          * **PR #804** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/804" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/804?utm_source=stack-comment).

lib/pos_blockchain.go
```diff
@@ -589,6 +633,46 @@ func (bc *Blockchain) hasValidBlockViewPoS(block *MsgDeSoBlock) error {
...
+	votingPublicKey := block.Header.ProposerVotingPublicKey
+	proposerPublicKey := block.Header.ProposerPublicKey
+	proposerPartialSig := block.Header.ProposerVotePartialSignature
+	// If the proposer partial sig is nil, we can't validate it. That's an error.
+	if proposerPartialSig.IsEmpty() {
+		return false, nil
+	}
+	// Get the snapshot validator entry for the proposer.
+	proposerPKID := bav.GetPKIDForPublicKey(proposerPublicKey.ToBytes()).PKID
+	snapshotValidatorEntry, err := bav.GetSnapshotValidatorSetEntryByPKIDAtSnapshotAtEpochNumber(proposerPKID, snapshotAtEpochNumber)
```
tholonious, line 646: Can you filter this to just the leader schedule for the epoch instead? I assume you're intended to make sure the proposer is actually an eligible leader for the epoch. Rough steps to perform this check are:
1. Fetch the entire snapshot leader schedule for the epoch, and find the leader with the above PKID
2. Fetch the snapshot validator corresponding to the leader's PKID
3. Validate the signature

tholonious (changes requested):

lib/pos_blockchain.go
lazynina: This is already caught in the leader check in the function `hasValidBlockProposerPoS`, so this would be duplicative. Basically this function just checks that the `ProposerPublicKey`, `ProposerVotingPublicKey`, and `ProposerVotePartialSignature` are correct. By not forcing this function to validate the leader, we're able to use this check in validating orphans from the previous & next epoch. When we go to process a "former orphan", we'll do the proper leader check and it'll be caught there.

tholonious: We discussed offline that the leader check is performed elsewhere. Marking this as resolved.

lib/pos_blockchain.go
```diff
@@ -589,6 +633,46 @@ func (bc *Blockchain) hasValidBlockViewPoS(block *MsgDeSoBlock) error {
...
+	if err != nil {
+		return false, errors.Wrapf(err, "hasValidProposerPartialSignaturePoS: Problem getting snapshot validator entry")
+	}
+	// If the snapshot validator entry is nil or deleted, we didn't snapshot
+	// the validator at this epoch, so we will never accept this block.
+	if snapshotValidatorEntry == nil || snapshotValidatorEntry.isDeleted {
+		return false, nil
+	}
+	// If the voting public key from the block's header doesn't match the
```
tholonious, line 655:
```suggestion
	// If the public key from the block's header doesn't match the snapshotted
	// public key from the validator entry, then it means that the validator's
	// PKID <-> public key has changed. We cannot validate this block
	if !snapshotValidatorEntry.PublicKey.Eq(proposerPublicKey) {
		return false, errors.Wrapf(err, "hasValidProposerPartialSignaturePoS: validator public key has changed)
	}

	// If the voting public key from the block's header doesn't match the
```

lib/pos_snapshot_entries.go
```diff
@@ -256,7 +259,7 @@ func (bav *UtxoView) GetSnapshotValidatorSetEntryByPKID(pkid *PKID) (*ValidatorE
 	if err != nil {
 		return nil, errors.Wrapf(
 			err,
-			"GetSnapshotValidatorSetEntryByPKID: problem retrieving ValidatorEntry from db: ",
+			"GetCurrentSnapshotValidatorSetEntryByPKID: problem retrieving ValidatorEntry from db: ",
```
tholonious, line 262:
```suggestion
			"GetSnapshotValidatorSetEntryByPKIDAtEpochNumber: problem retrieving ValidatorEntry from db: ",
```

lib/pos_blockchain.go
```diff
@@ -322,6 +339,21 @@ func (bc *Blockchain) validateAndIndexBlockPoS(block *MsgDeSoBlock) (*BlockNode,
 		// We return the block node as-is as a best effort thing.
 		return blockNode, nil
 	}
+	currentEpochEntry, err := utxoView.GetCurrentEpochEntry()
```
tholonious, line 342: Assuming I haven't missed anything, do we need to first verify here that the block belongs in the current epoch, before validating the partial signature? Even though `validateAndIndexBlockPoS` may currently only be called for blocks in the current epoch, it may be worth a check here again to bullet proof things.

lazynina: yes - if the current block is not in the current epoch after successfully constructing the view based on the parent block, then we'll never accept this block.

lib/pos_blockchain.go
```diff
@@ -322,6 +339,26 @@ func (bc *Blockchain) validateAndIndexBlockPoS(block *MsgDeSoBlock) (*BlockNode,
...
+	if err != nil {
+		return nil, errors.Wrap(err, "validateAndIndexBlockPoS: Problem getting current epoch entry")
+	}
+	// If after constructing a UtxoView based on the parent block, we find that the current block is not in
+	// the current epoch, this should result in a validate failed state.
+	if !currentEpochEntry.ContainsBlockHeight(block.Header.Height) {
+		return bc.storeValidateFailedBlockWithWrappedError(block, errors.New("block is not in current epoch"))
```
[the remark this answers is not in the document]
tholonious: Resolved offline. The block height value in the block's header is known to be invalid if we hit the case

tholonious (approved):

PR #815, opened 2023-11-14
Clean up vote and timeout message construction for broadcast

tholonious:

PR #811, opened 2023-11-14
Handle Block Construction Signal and Broadcast

tholonious:

PR #816, opened 2023-11-14
Implement E2E Block Processing Between Blockchain and FastHotStuffEventLoop

tholonious:


tholonious:
  * **PR #815** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/815?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #816** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/816?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
    * **PR #811** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/811?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/816?utm_source=stack-comment).

lib/pos_consensus_controller.go
```diff
@@ -211,3 +211,109 @@ func (cc *ConsensusController) HandleHeader(pp *Peer, msg *MsgDeSoHeader) {
 func (cc *ConsensusController) HandleBlock(pp *Peer, msg *MsgDeSoBlock) {
 	// TODO
 }
+
+// processNewTipBlock tries to apply a new tip block to both the Blockchain and FastHotStuffEventLoop data
+// structures. It guarantees
```
lazynina, line 216: Finish comment? also try to start the comment w/ the function name like `tryProcessBlockAsNewTip tries to apply...`

lib/pos_consensus_controller.go
```diff
@@ -211,3 +211,109 @@ func (cc *ConsensusController) HandleHeader(pp *Peer, msg *MsgDeSoHeader) {
...
+	}
+
+	// Build a UtxoView at the new tip
+	utxoViewAtTipBlock, err := cc.blockchain.getUtxoViewAtBlockHash(*tipBlockHash)
+	if err != nil {
+		return nil, errors.Errorf("HandleFastHostStuffBlockProposal: Error fetching UtxoView for tip block: %v", err)
+	}
+
+	// Fetch the validator set at the new tip
+	validatorSetAtTipBlock, err := utxoViewAtTipBlock.GetFullSnapshotValidatorSet()
```
lazynina, line 256: hmm do we want the validator set from before connecting the block? In `pos_blockchain.go`, we connect to the parent block and then pull the validator set before connecting the incoming block. I'm a little concerned about the edge case where we have something like the below :

```
| ------- epoch n ------- | ------ epoch n + 1 ------ |
                       ... x-1 -> x -> x + 1
```
Where x+1 is the first block in the next epoch. When we get the utxo view at X, GetCurrentEpochEntry will return epoch n + 1.

lib/pos_consensus_controller.go
```diff
@@ -211,3 +211,109 @@ func (cc *ConsensusController) HandleHeader(pp *Peer, msg *MsgDeSoHeader) {
...
+
+	// Fetch the validator sets for the safe blocks
+	safeBlocksWithValidators := make([]consensus.BlockWithValidatorList, len(safeBlocks))
+	for ii, safeBlock := range safeBlocks {
+		safeBlockHash, err := safeBlock.Hash()
+		if err != nil {
+			return nil, errors.Errorf("HandleFastHostStuffBlockProposal: Error hashing safe block: %v", err)
+		}
+
+		utxoViewAtSafeBlock, err := cc.blockchain.getUtxoViewAtBlockHash(*safeBlockHash)
```
lazynina, line 281: I think we may be able to avoid constructing a new utxo view for each safeBlockHash as this is probably a relatively expensive operation. I think we can compute the epoch given the block height, which then gives us the snapshot at epoch and then a vanilla UtxoView w/o any connected blocks should be able to fetch the validator set for us. The advantage of this is that we only need a single utxo view and we'll have the validator set cached for each safe block. We shouldn't have more than 2-3 different validator sets and we'll basically be fetching the same information for each block (just depends on the epoch its in really).

lib/pos_consensus_controller.go
```diff
@@ -211,3 +211,109 @@ func (cc *ConsensusController) HandleHeader(pp *Peer, msg *MsgDeSoHeader) {
...
+// We stub them out here to unblock consensus work.
+////////////////////////////////////////////////////////////////////////////////
+
+func (bc *Blockchain) getUtxoViewAtBlockHash(blockHash BlockHash) (*UtxoView, error) {
+	return nil, errors.New("getUtxoViewAtBlockHash: replace me with a real implementation later")
+}
+
+func (bc *Blockchain) getSafeBlocks() ([]*MsgDeSoHeader, error) {
+	return nil, errors.New("getSafeBlocks: replace me with a real implementation later")
+}
```
lazynina, line 319: ah yes, if you want the names we have in `feature/pos-process-block`, you can use the following signatures:

```
func (bc *Blockchain) GetSafeBlocks() ([]*BlockNode, error) {
```
Looks like you have the correct signature for `getUtxoViewAtBlockHash`

lib/pos_snapshot_entries.go
```diff
@@ -257,6 +257,20 @@ func (bav *UtxoView) GetSnapshotValidatorSetEntryByPKID(pkid *PKID) (*ValidatorE
...
+		return nil, errors.Wrapf(err, "GetFullSnapshotValidatorSet: problem retrieving SnapshotGlobalParamsEntry: ")
+	}
+
+	validatorSet, err := bav.GetSnapshotValidatorSetByStakeAmount(snapshotGlobalParamsEntry.ValidatorSetMaxNumValidators)
+	if err != nil {
+		return nil, errors.Wrapf(err, "GetFullSnapshotValidatorSet: problem retrieving ValidatorSetAtTipBlock: ")
+	}
+
+	return validatorSet, nil
+}
```
lazynina, line 272: I have an identical function defined in my stack. We'll just want to use one or the other. Mine is named `GetAllSnapshotValidatorSetEntriesByStake` for reference.

lazynina (changes requested):

lib/pos_consensus_controller.go
tholonious: Discussed offline. We'll keep the function signature for this as `func (bc *Blockchain) getSafeBlocks() ([]*MsgDeSoHeader, error)` so that the Blockchain struct doesn't expose BlockNodes where they're not necessary

lib/pos_consensus_controller.go
tholonious: No, we want the validator set after connecting the block. The way to think about the validator set is "this is the validator set we use to validate the next block that extends from this safe block".

lib/pos_snapshot_entries.go
tholonious: I'll delete this function once we merge with your feature branch.

lib/pos_consensus_controller.go
lazynina: ahhhh okay. Maybe we can clarify that in a comment somewhere? or fix the naming a bit? Not sure what makes it most clear.

lib/pos_consensus_controller.go
tholonious: This is a great optimization! I've added an in-memory cache for this.

lib/pos_consensus_controller.go
tholonious: I added comments for this where we fetch the validator sets for the tip block and the safe blocks

lib/pos_consensus_controller.go
```diff
@@ -212,6 +212,82 @@ func (cc *ConsensusController) HandleBlock(pp *Peer, msg *MsgDeSoBlock) {
 	// TODO
 }
 
+// processNewTipBlock tries to apply a new tip block to both the Blockchain and FastHotStuffEventLoop data
```
lazynina, line 215:
```suggestion
// tryProcessBlockAsNewTip tries to apply a new tip block to both the Blockchain and FastHotStuffEventLoop data
```
Starting comments that are immediately above function names allows goland to show this description as a preview when highlighting the function elsewhere in the codebase.

lazynina (approved):
Just one nit on the comment

PR #819, opened 2023-11-15
Add ConsensusController Utility to Fetch Validator Set For Blocks

tholonious:


tholonious:
  * **PR #819** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/819?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/819?utm_source=stack-comment).

lib/pos_consensus_controller.go
```diff
@@ -211,3 +211,63 @@ func (cc *ConsensusController) HandleHeader(pp *Peer, msg *MsgDeSoHeader) {
...
+	for ii, block := range blocks {
+		blockHash, err := block.Hash()
+		if err != nil {
+			return nil, errors.Errorf("error computing block hash: %v", err)
+		}
+
+		utxoViewAtBlock, err := cc.blockchain.getUtxoViewAtBlockHash(*blockHash)
+		if err != nil {
+			return nil, errors.Errorf("error fetching UtxoView for block: %v", err)
+		}
```
lazynina, line 231: So this function gets us half way there, but I think what we want to do is create a single utxo view struct without connecting any uncommitted blocks. We should have all the snapshot data available in the database that we want and connecting parents only really helps us figure out "current epoch". 

What I think we want to do is the following:
1. create a single vanilla utxo view `NewUtxoView(cc.blockchain.db, cc.blockchain.params, nil, cc.blockchain.snapshot, nil)` or maybe not nil for event manager, but that's up to you.
2. `currentEpochEntry, err := utxoView.GetCurrentEpochEntry()` 
3. For each block,
-  if `currentEpochEntry.ContainsBlockHeight(block.Height)`, use current epoch to compute snapshotAtEpochNumber
-  if not , we know must be in the next epoch. We can simulate the epoch using a function I wrote, basically just ripped out of the rollover logic,  called `computeNextEpochEntry`. Confirm block is in the next epoch and lookup snapshotAtEpochNumber for that computed epoch.
- Look up all validators using `GetAllSnapshotValidatorSetEntriesByStakeAtEpochNumber` (all in my pos-process-block stack)

lazynina (changes requested):

lib/pos_consensus_controller.go
```diff
@@ -211,3 +212,125 @@ func (cc *ConsensusController) HandleHeader(pp *Peer, msg *MsgDeSoHeader) {
...
+	}
+
+	// Create a map to cache the validator set entries by epoch number. Two blocks in the same epoch will have
+	// the same validator set, so we can use an in-memory cache to optimize the validator set lookup for them.
+	validatorSetEntriesBySnapshotEpochNumber := make(map[uint64][]*ValidatorEntry)
+
+	// Create a UtxoView for the committed tip block. We will use this to fetch the validator set for the
+	// all of the safe blocks.
+	//
+	// TODO: Do we need to pass a pointer to the Postgres object here?
```
lazynina, line 238: We won't need to do this actually!
```suggestion
```

lazynina (approved):

PR #831, opened 2023-11-20
Feature/end of epoch utxo ops

lazynina:


lazynina:
* feature/proof-of-stake
  * **PR #831** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/831?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
    * **PR #837** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/837?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/831?utm_source=stack-comment).

tholonious (approved):
Approving since this looks good on my side

lib/pos_staking_rewards.go
```diff
@@ -177,39 +182,51 @@ func (bav *UtxoView) computeStakerRewardAndValidatorCommission(
...
+			PrevStakeEntries:     []*StakeEntry{stakeEntry.Copy()},
+			PrevValidatorEntry:   validatorEntry.Copy(),
+			StakeAmountNanosDiff: rewardNanos,
+		}
 		stakeEntry.StakeAmountNanos = uint256.NewInt().Add(stakeEntry.StakeAmountNanos, uint256.NewInt().SetUint64(rewardNanos))
 		bav._setStakeEntryMappings(stakeEntry)
-
-		return nil
+		validatorEntry.TotalStakeAmountNanos = uint256.NewInt().Add(validatorEntry.TotalStakeAmountNanos, uint256.NewInt().SetUint64(rewardNanos))
+		bav._setValidatorEntryMappings(validatorEntry)
```
diamondhands0, line 213: Oh interesting. So it looks like we had a bug before whereby we weren't updating the TotalStakeAmountNanos on a validator when restaking right?

lazynina: That's correct! Found it while I was in here.

lazynina: And this reminds me that we also need to update the global total stake amount nanos value

diamondhands0: Can we add a sanity-check that runs at the end of every Nth RunEpochCompleteHook and either logs an error or panics if these values differ? Ideally the N value could be adjusted by ParamUpdater so that we can start it =1 then make it bigger to increase efficiency.

diamondhands0: Honestly even if it just ran every time to start, that would probably help us catch bugs, and we can remove it manually if it ends up bottlenecking us.

lib/block_view.go
```diff
@@ -1628,27 +1628,49 @@ func (bav *UtxoView) DisconnectBlock(
...
+				}
+			case OperationTypeStakeDistribution:
+				// We don't allow stake distribution utxo operations unless it's the end of an epoch.
+				if !isLastBlockInEpoch {
+					return fmt.Errorf("DisconnectBlock: Found add balance operation in block %d that is not the end of an epoch", desoBlock.Header.Height)
+				}
+				if len(utxoOp.PrevStakeEntries) != 1 {
+					return fmt.Errorf("DisconnectBlock: Expected exactly one prev stake entry for stake distribution op")
+				}
+				bav._setStakeEntryMappings(utxoOp.PrevStakeEntries[0])
```
diamondhands0, line 1672: I think you're not resetting the validator amount nanos or the global staked amount nanos here :)

lazynina: We no longer have a global staked amount nanos tracker actually! adding validator reset in another PR.

lib/pos_staking_rewards.go
tholonious: @diamondhands0 @lazynina we no longer have a global state amount nanos Badger index. So, we don't need to and wouldn't be able to run the cross-validation suggested here.

PR #847, opened 2023-12-04
Add an interface exposing relevant mempool functions for backend consumption

lazynina:


lazynina:
* feature/proof-of-stake
  * **PR #847** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/847?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/847?utm_source=stack-comment).

lib/server.go
```diff
@@ -418,6 +433,7 @@ func NewServer(
 		snapshot:                     _snapshot,
 		nodeMessageChannel:           _nodeMessageChan,
 		forceChecksum:                _forceChecksum,
+		fastHotStuffEventLoop:        consensus.NewFastHotStuffEventLoop(),
```
tholonious, line 436: I'll resolve this on in another PR. For this PR, you can remove this though
```suggestion
```

lib/server.go
```diff
@@ -217,8 +218,22 @@ func (srv *Server) GetBlockchain() *Blockchain {
 	return srv.blockchain
 }
 
+type BackendMempool interface {
```
tholonious, line 221: Now that we've seen the new interface expressed in the code, it's worth just merging this interface with the existing `Mempool` interface. You can have the existing DeSoMempool struct just stub out the interface methods that we don't want/need it to support. The stubbed out methods can return a "Not Implemented" error

lib/pos_mempool.go
```diff
@@ -580,3 +580,51 @@ func (mp *PosMempool) UpdateGlobalParams(globalParams *GlobalParamsEntry) {
...
+	if !mp.IsRunning() {
+		return nil
+	}
+	return mp.getTransactionsNoLock()
+}
+
+func (mp *PosMempool) IsTransactionInPool(txHash *BlockHash) bool {
+	// TODO: Do I need to hold a wider lock or nah?
+	mp.txnRegister.RLock()
+	defer mp.txnRegister.RUnlock()
```
tholonious, line 617: It's fine to hold the broader mempool lock since it's just a read lock. This way, we don't have to worry about edge cases where other data structures get updated but not the transaction register

```suggestion
	mp.RLock()
	defer mp.RUnlock()
```

tholonious (changes requested):

lib/server.go
lazynina: Deal!!

tholonious (approved):

PR #857, opened 2023-12-07
PoS RemoteNode and RemoteNodeId

AeonSw4n:


lib/pos_remote_node.go
```diff
@@ -0,0 +1,456 @@
...
+	}
+	persistentStr := "PERSISTENT"
+	if !peer.isPersistent {
+		persistentStr = "NON-PERSISTENT"
+	}
+	logStr := fmt.Sprintf("SUCCESS version negotiation for (%s) (%s) peer (%v).", inboundStr, persistentStr, peer)
+	glog.V(1).Info(logStr)
+}
+
+func (hm *HandshakeMetadata) NegotiatedProtocolVersion() ProtocolVersionType {
```
tholonious, line 432:
```suggestion
func (hm *HandshakeMetadata) GetNegotiatedProtocolVersion() ProtocolVersionType {
```

lib/server.go
```diff
@@ -1588,11 +1598,11 @@ func (srv *Server) _startSync() {
 
 }
 
-func (srv *Server) _handleNewPeer(pp *Peer) {
+func (srv *Server) _handleHandshakePeer(pp *Peer) {
```
tholonious, line 1601:
```suggestion
func (srv *Server) _handlePeerHandshakeComplete(pp *Peer) {
```

lib/server.go
```diff
@@ -1588,11 +1598,11 @@ func (srv *Server) _startSync() {
...
 	isSyncCandidate := pp.IsSyncCandidate()
 	isSyncing := srv.blockchain.isSyncing()
 	chainState := srv.blockchain.chainState()
-	glog.V(1).Infof("Server._handleNewPeer: Processing NewPeer: (%v); IsSyncCandidate(%v), syncPeerIsNil=(%v), IsSyncing=(%v), ChainState=(%v)",
+	glog.V(1).Infof("Server._handleHandshakePeer: Processing NewPeer: (%v); IsSyncCandidate(%v), syncPeerIsNil=(%v), IsSyncing=(%v), ChainState=(%v)",
```
tholonious, line 1605:
```suggestion
	glog.V(1).Infof("Server._handlePeerHandshakeComplete: Processing NewPeer: (%v); IsSyncCandidate(%v), syncPeerIsNil=(%v), IsSyncing=(%v), ChainState=(%v)",
```

lib/pos_remote_node.go
```diff
@@ -0,0 +1,456 @@
...
+func (rn *RemoteNode) GetId() RemoteNodeId {
+	return rn.id
+}
+
+func (rn *RemoteNode) GetPeerId() uint64 {
+	peerId, _ := rn.id.GetIds()
+	return peerId
+}
+
+func (rn *RemoteNode) CreateOutboundConnection(netAddr *wire.NetAddress) {
```
tholonious, line 128: Is this a dial? If so, please rename for clarity. It's also worth adding a comment block on top of the function signature
```suggestion
func (rn *RemoteNode) DialOutboundConnection(netAddr *wire.NetAddress) {
```

lib/pos_remote_node.go
```diff
@@ -0,0 +1,456 @@
...
+		return
+	}
+
+	attemptId := rn.srv.CreateOutboundConnection(netAddr)
+	id := NewRemoteNodeAttemptedId(attemptId)
+	rn.SetId(id)
+	rn.connectionStatus = RNConnectionStatus_Attempted
+}
+
+func (rn *RemoteNode) CreatePersistentOutboundConnection(netAddr *wire.NetAddress) {
```
tholonious, line 139: Same thing. Is this a dial?
```suggestion
func (rn *RemoteNode) DialPersistentOutboundConnection(netAddr *wire.NetAddress) {
```

tholonious (changes requested):

lib/pos_remote_node.go
```diff
@@ -0,0 +1,456 @@
...
+	if rn.connectionStatus != RNConnectionStatus_Connected {
+		return
+	}
+
+	if err := rn.srv.SendMessage(desoMsg, rn.GetPeerId()); err != nil {
+		glog.Errorf("sendMessage: Problem sending message to peer (id= %d): %v", rn.peer.ID, err)
+	}
+}
+
+func (rn *RemoteNode) InitiateHandshake(reserveNonce func(versionNonce uint64)) error {
```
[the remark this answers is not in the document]
AeonSw4n: Great suggestion! I couldn't figure out a good way to simplify this, and the nonce is the solution I was looking for

lib/pos_remote_node.go
```diff
@@ -0,0 +1,456 @@
...
+
+func (rn *RemoteNode) IsInbound() bool {
+	return rn.connectionType == RNConnectionType_Inboound
+}
+
+func (rn *RemoteNode) IsOutbound() bool {
+	return rn.connectionType == RNConnectionType_Outbound
+}
+
+func (rn *RemoteNode) SetId(id RemoteNodeId) {
```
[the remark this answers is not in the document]
AeonSw4n: I made it private, since it'll really only be accessed inside of RemoteNode. The RemoteNode id can change depending on the lifecycle of the connection. Attempted connection will have (peerId, attemptId) = (0, y) while a formed connection will have (peerId, attemptId) = (x, 0) or (x, y), which will be adjusted over time.

lib/pos_remote_node.go
```diff
@@ -0,0 +1,456 @@
...
+	rn.id = id
+}
+
+func (rn *RemoteNode) GetId() RemoteNodeId {
```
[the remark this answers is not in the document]
AeonSw4n: It's used in the ConnectionController or the new RemoteNodeManager

lib/pos_remote_node.go
```diff
@@ -0,0 +1,456 @@
...
+
+//type RemoteNodeId struct {
+//	PeerId    uint64
+//	AttemptId uint64
+//}
+
+type RemoteNode struct {
+	// Should we have ID here?
+	peer             *Peer
+	id               RemoteNodeId
```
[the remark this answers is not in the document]
AeonSw4n: We'll use the RemoteNodeId as the key while indexing RemoteNodes in the RemoteNodeManager (old RemoteNodeIndexer & Manager). The problem with using Peer as the source of truth is that the Peer is not yet instantiated when indexing attempted connections. Instead, the RemoteNodeId allows us to index RemoteNodes with formed and unformed connections with a single type. 

Also it seems like having RemoteNodeId in the RemoteNode vs using Peer's id makes it easier to handle reconnects for persistent peers. In such a case you can keep the same RemoteNode and discard the Peer, then set the new Peer's id to the one prior to the reconnect.

You raised a good point though. Let me think about this more as I bulletproof the code with tests.

lib/pos_remote_node.go
```diff
@@ -0,0 +1,456 @@
...
+func (rn *RemoteNode) SetId(id RemoteNodeId) {
+	rn.id = id
+}
+
+func (rn *RemoteNode) GetId() RemoteNodeId {
+	return rn.id
+}
+
+func (rn *RemoteNode) GetPeerId() uint64 {
+	peerId, _ := rn.id.GetIds()
```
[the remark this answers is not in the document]
AeonSw4n: Check my response to your previous comment on this.

lib/pos_remote_node_id.go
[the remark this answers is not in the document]
AeonSw4n: Not sure if you saw the RemoteNodeIndexer when you wrote this comment. Do you still prefer to separate attemptId and peerId from one another?

lib/pos_remote_node.go
```diff
@@ -0,0 +1,456 @@
...
+	// Should we have ID here?
+	peer             *Peer
+	id               RemoteNodeId
+	connectionType   RNConnectionType
+	connectionStatus RNConnectionStatus
+
+	params *DeSoParams
+
+	bc  *Blockchain
+	srv *Server
```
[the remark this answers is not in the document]
AeonSw4n: There is one place where we need to rely on the Server, which is to signal the PeerHandshakeComplete message. If we used ConnectionManager for this, it would basically just call on Server to signal this message anyway. Besides this one, I purged every other Server reference.

lib/pos_remote_node.go
```diff
@@ -0,0 +1,456 @@
...
+func (rn *RemoteNode) ConnectInboundPeer(conn net.Conn, na *wire.NetAddress) error {
+	peer := rn.srv.ConnectPeer(conn, na, RemoteNodeIdNoAttempt, false, false)
+	if peer == nil {
+		return errors.Errorf("ConnectInboundPeer: Problem connecting peer (%s)", conn.RemoteAddr().String())
+	}
+	rn.SetConnectedPeer(peer)
+	return nil
+}
+
+func (rn *RemoteNode) ConnectOutboundPeer(conn net.Conn, na *wire.NetAddress, attemptId uint64, isOutbound bool, isPersistent bool) error {
```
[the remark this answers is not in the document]
AeonSw4n: Seems like we could rename this, do you have any suggestions? The purpose of this function is to create a peer after a successful connection is formed.

lib/remote_node.go
```diff
@@ -0,0 +1,468 @@
...
+
+const (
+	RemoteNodeConnectionStatus_NotConnected RemoteNodeConnectionStatus = 0
+	RemoteNodeConnectionStatus_Connected    RemoteNodeConnectionStatus = 1
+	RemoteNodeConnectionStatus_Attempted    RemoteNodeConnectionStatus = 2
+	RemoteNodeConnectionStatus_Terminated   RemoteNodeConnectionStatus = 3
+)
+
+type RemoteNode struct {
+	// Should we have ID here?
```
lazynina, line 33: Is the ID you're referring to here different than the `id` field on line 35?

lib/remote_node.go
```diff
@@ -0,0 +1,468 @@
...
+	connectionType   RemoteNodeConnectionType
+	connectionStatus RemoteNodeConnectionStatus
+
+	params *DeSoParams
+
+	srv  *Server
+	bc   *Blockchain
+	cmgr *ConnectionManager
+
+	minTxFeeRateNanosPerKB uint64
```
lazynina, line 45: What does this `minTxFeeRateNanosPerKB` field do? How does it work?

lib/remote_node.go
```diff
@@ -0,0 +1,468 @@
...
+	hyperSync              bool
```
lazynina, line 46: Do we also want to know if it's archival or not?

lib/remote_node.go
```diff
@@ -0,0 +1,468 @@
...
+	if peer == nil {
+		return errors.Errorf("ConnectInboundPeer: Problem connecting peer (%s)", conn.RemoteAddr().String())
+	}
+	rn.SetConnectedPeer(peer)
+	return nil
+}
+
+// ConnectOutboundPeer connects a peer once a successful outbound connection has been established.
+func (rn *RemoteNode) ConnectOutboundPeer(conn net.Conn, na *wire.NetAddress, attemptId uint64, isOutbound bool, isPersistent bool) error {
+	peer := rn.cmgr.ConnectPeer(conn, na, attemptId, isOutbound, isPersistent)
```
lazynina, line 166: should we update `ConnectPeer` to actually return us an error? Can do this in a separate workstream, but seems like we're just guessing why a connection failed and may lead to headaches later on.

lib/remote_node.go
```diff
@@ -0,0 +1,468 @@
...
+
+	// Record the nonce of this version message before we send it so we can
+	// detect self connections and so we can validate that the peer actually
+	// controls the IP she's supposedly communicating to us from.
+	vMeta := rn.GetHandshakeMetadata()
+	vMeta.versionNoncesSent = nonce
+	//uint64(RandInt64(math.MaxInt64))
+
+	if err := rn.cmgr.SendMessage(verMsg, rn.peer.ID); err != nil {
+		glog.Errorf("sendVersionMessage: Problem sending version message to peer (id= %d): %v", rn.peer.ID, err)
```
lazynina, line 229: Should we return this error instead of glogging it? Is there ever a scenario in which the caller would want to handle this error?

lib/remote_node.go
```diff
@@ -0,0 +1,468 @@
...
+	if !rn.peer.IsOutbound() {
+		// Respond to the version message if this is an inbound peer.
+		rn.sendVersionMessage(responseNonce)
+	}
+	// After sending and receiving a compatible version, complete the
+	// negotiation by sending and receiving a verack message.
+	verackTimeExpected := time.Now().Add(rn.params.VersionNegotiationTimeout)
+	vMeta.verackTimeExpected = &verackTimeExpected
+	if err := rn.sendVerack(); err != nil {
+		glog.Errorf("RemoteNode.HandleVersionMessage: Problem sending verack message to peer (id= %d): %v", rn.peer.ID, err)
```
lazynina, line 319: I generally prefer to return errors and require the caller to handle them instead of glogging them. It just becomes tricky to track down bugs later on.

lib/remote_node.go
```diff
@@ -0,0 +1,468 @@
...
+	return nil
+}
+
+func (rn *RemoteNode) newVerackMessage() (*MsgDeSoVerack, error) {
+	verack := NewMessage(MsgTypeVerack).(*MsgDeSoVerack)
+	vMeta := rn.GetHandshakeMetadata()
+
+	// Include the nonce we received in the peer's version message
+	verack.NonceReceived = vMeta.versionNoncesReceived
+	if vMeta.negotiatedProtocolVersion == ProtocolVersion2 {
```
lazynina, line 344: not a issue for this PR but do we want `negotiatedProtocolVersion` to actually represent the release tag of core that node is running? This may be useful for future forks.

lib/remote_node.go
```diff
@@ -0,0 +1,468 @@
...
+	nonceSent := vMeta.versionNoncesSent
+
+	if vMeta.verackTimeExpected != nil && vMeta.verackTimeExpected.Before(time.Now()) {
+		glog.V(1).Infof("RemoteNode.HandleVerackMessage: Requesting PeerDisconnect for id: (%v) "+
+			"verack timeout. Time expected: %v, now: %v", rn.peer.ID, vMeta.verackTimeExpected.UnixMicro(), time.Now().UnixMicro())
+		rn.cmgr.CloseConnection(rn.peer.ID)
+		return
+	}
+	// If the verack message has a nonce unseen for us, then request peer disconnect.
+	if vrkMsg.NonceReceived != nonceSent && vrkMsg.NonceReceived != nonceReceived {
```
lazynina, line 370: This condition is confusing to me, but basically the `NonceReceived` must be either the nonceReceived or the nonceSent? or do we need to be checking `vrkMsg.NonceSent != nonceSent` and `vrk.NonceReceived != nonceReceived`?

lib/remote_node.go
```diff
@@ -0,0 +1,468 @@
...
+			return
+		}
+
+		ok, err := BLSVerifyPoSValidatorHandshake(vrkMsg.NonceSent, vrkMsg.NonceReceived, vrkMsg.TstampMicro,
+			vrkMsg.Signature, vrkMsg.PublicKey)
+		if err != nil || !ok {
+			glog.V(1).Infof("RemoteNode.HandleVerackMessage: Requesting PeerDisconnect for id: (%v) "+
+				"verack signature verification failed: %v", rn.peer.ID, err)
+			rn.cmgr.CloseConnection(rn.peer.ID)
+			return
```
lazynina, line 420: should we be returning an error?

lib/server.go
```diff
@@ -545,6 +545,8 @@ func NewServer(
 		_miner = nil
 	}
 
+	srv.consensusController = NewConsensusController(_params, _chain, nil, _blsKeystore.GetSigner())
```
lazynina, line 548: Let's fix these issues so that CI can run.

lib/remote_node.go
AeonSw4n: I've addressed this now. The RemoteNode now has a singular Id, and the OutboundConnectionAttempt & Peer will inherit this Id.

lib/remote_node.go
AeonSw4n: Doesn't really do anything. We pass it around in version msg exchange, but we never quite use it.

lib/remote_node.go
AeonSw4n: Actually, it looks like `cmgr.ConnectPeer` should never fail so I just turned `ConnectOutboundPeer` and `ConnectInboundPeer` into voids.

lib/remote_node.go
AeonSw4n: We indeed do lol.

lib/remote_node.go
AeonSw4n: Hmm, well maybe it's still worthwhile for these functions to return an error if RemoteNode `connectionStatus` is incorrect.

lib/remote_node.go
AeonSw4n: Should be fixed now.

lib/remote_node.go
AeonSw4n: Yeah, I think this is a good idea!

lib/remote_node.go
```diff
@@ -0,0 +1,599 @@
...
+	// We use an int64 instead of a uint64 for convenience.
+	ver.TstampSecs = time.Now().Unix()
+
+	ver.Nonce = nonce
+	ver.UserAgent = rn.params.UserAgent
+
+	// When a node asks you for what height you have, you should reply with the height of the latest actual block you
+	// have. This makes it so that peers who have up-to-date headers but missing blocks won't be considered for initial
+	// block download.
+	ver.StartBlockHeight = uint32(rn.bc.BlockTip().Header.Height)
```
tholonious, line 343: Rather than reading the current height from the blockchain, see if you can pass the current height into the InitiateHandshake and HandleVersionMessage functions instead

lib/remote_node.go
```diff
@@ -0,0 +1,599 @@
...
+	ver.MinFeeRateNanosPerKB = rn.minTxFeeRateNanosPerKB
+
+	return ver
+}
+
+// HandleVersionMessage is called upon receiving a version message from the RemoteNode's peer. The peer may be the one
+// initiating the handshake, in which case, we should respond with our own version message. To do this, we pass the
+// responseNonce to this function, which we will use in our response version message.
+func (rn *RemoteNode) HandleVersionMessage(verMsg *MsgDeSoVersion, responseNonce uint64) error {
+	if rn.connectionStatus != RemoteNodeStatus_Connected {
```
tholonious, line 355: What happens if the peer sends messages out of order, or if they send multiple version messages? It'll be good to make sure the following are safe for the handshake process:
1. We receive the peer's version message before we send a version message
2. The peer sends multiple version messages
3. The peer sends a verack message before a version message

lib/remote_node.go
```diff
@@ -0,0 +1,599 @@
...
+	peer *Peer
+	// The id is the unique identifier of this RemoteNode. For outbound connections, the id will be the same as the
+	// attemptId of the OutboundConnectionAttempt, and the subsequent id of the outbound peer. For inbound connections,
+	// the id will be the same as the inbound peer's id.
+	id               RemoteNodeId
+	connectionStatus RemoteNodeStatus
+
+	params *DeSoParams
+	srv    *Server
+	bc     *Blockchain
```
tholonious, line 63: Try removing the blockchain field from RemoteNode. Seems like it's only used to extract the chain's current height when sending the version message. The height can be passed as a param into InitiateHandshake and HandleVersionMessage by the caller instead. This will make sure the RemoteNode has less knowledge of the chain
```suggestion
```

lib/remote_node.go
```diff
@@ -0,0 +1,599 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("RemoteNode.newVerackMessage: Problem signing verack message: %v", err)
+		}
+	}
+	return verack, nil
+}
+
+// HandleVerackMessage handles a verack message received from the peer.
+func (rn *RemoteNode) HandleVerackMessage(vrkMsg *MsgDeSoVerack) error {
+	if rn.connectionStatus != RemoteNodeStatus_Connected {
```
tholonious, line 469: Similar question as above. Make sure the handshake is safe if the peer sends does any of the following:
1. Sends a verack message before a version message
2. Sends a verack message before we send our verack message

tholonious (changes requested):

cmd/config.go
```diff
@@ -17,6 +17,7 @@ type Config struct {
 	TXIndex              bool
 	Regtest              bool
 	PostgresURI          string
+	PosValidatorSeed     string
```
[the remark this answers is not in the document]
lazynina: Seconded!

cmd/node.go
```diff
@@ -202,13 +202,15 @@ func (node *Node) Start(exitChannels ...*chan struct{}) {
 	// Setup eventManager
 	eventManager := lib.NewEventManager()
 
+	blsKeystore, err := lib.NewBLSKeystore(node.Config.PosValidatorSeed)
```
lazynina, line 205: We need to handle the error here! I'm assuming we hit an error if the seed is invalid.
```suggestion
	blsKeystore, err := lib.NewBLSKeystore(node.Config.PosValidatorSeed)
	if err != nil {
		panic(err)
	}
```

lib/remote_node.go
lazynina: agreed

lib/remote_node.go
```diff
@@ -0,0 +1,599 @@
...
+
+	params *DeSoParams
+	srv    *Server
+	bc     *Blockchain
+	cmgr   *ConnectionManager
+
+	// minTxFeeRateNanosPerKB is the minimum transaction fee rate in nanos per KB that our node will accept.
+	minTxFeeRateNanosPerKB uint64
+	hyperSync              bool
+	archivalMode           bool
```
lazynina, line 69: Should we use the `NodeSyncType` type instead here?

lib/remote_node.go
```diff
@@ -0,0 +1,599 @@
...
+	// versionNonceSent is the nonce sent in the Version message to the peer.
+	versionNonceSent uint64
+	// versionNonceReceived is the nonce received in the Version message from the peer.
+	versionNonceReceived uint64
+	// userAgent is a meta level label that can be used to analyze the network.
+	userAgent string
+	// ServiceFlag is a bitfield that indicates the services supported by the peer.
+	ServiceFlag ServiceFlag
+	// StartingBlockHeight is the block height of the peer's block tip during the Version exchange.
+	StartingBlockHeight uint32
```
lazynina, line 98: We sometimes use uint64s instead. It's kinda annoying. we should probably make this a uint64 to be forward compatible as I think eventually we'll want to move all block height representations to be a uint64.

lib/remote_node.go
```diff
@@ -0,0 +1,599 @@
...
+		params:                 params,
+		minTxFeeRateNanosPerKB: minTxFeeRateNanosPerKB,
+		hyperSync:              hyperSync,
+		archivalMode:           archivalMode,
+		posValidator:           posValidator,
+	}
+}
+
+// setStatusValidated sets the connection status of the remote node to validated.
+func (rn *RemoteNode) setStatusValidated() {
```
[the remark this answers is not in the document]
lazynina: @tholonious - it should be `setStatusHandshakeCompleted`

lib/remote_node.go
```diff
@@ -0,0 +1,599 @@
...
+func (rn *RemoteNode) getHandshakeMetadata() *HandshakeMetadata {
+	if rn.handshakeMetadata == nil {
+		rn.handshakeMetadata = &HandshakeMetadata{}
+	}
+	return rn.handshakeMetadata
+}
+
+func (rn *RemoteNode) GetNegotiatedProtocolVersion() ProtocolVersionType {
+	meta := rn.getHandshakeMetadata()
+	return meta.negotiatedProtocolVersion
```
lazynina, line 173:
```suggestion
	return rn.getHandshakeMetadata().negotiatedProtocolVersion
```

lib/remote_node.go
```diff
@@ -0,0 +1,599 @@
...
+}
+
+func (rn *RemoteNode) GetValidatorPublicKey() *bls.PublicKey {
+	meta := rn.getHandshakeMetadata()
+	return meta.validatorPublicKey
```
lazynina, line 178:
```suggestion
	return rn.getHandshakeMetadata().validatorPublicKey
```

lib/remote_node.go
```diff
@@ -0,0 +1,599 @@
...
+	defer rn.mtx.Unlock()
+
+	rn.cmgr.DialPersistentOutboundConnection(netAddr, rn.GetId().ToUint64())
+	rn.setStatusAttempted()
+	return nil
+}
+
+// ConnectInboundPeer connects a peer once a successful inbound connection has been established.
+func (rn *RemoteNode) ConnectInboundPeer(conn net.Conn, na *wire.NetAddress) error {
+	if rn.connectionStatus != RemoteNodeStatus_NotConnected {
```
lazynina, line 231: I've seen this check a few times. Should we create a function on `connectionStatus` called `IsConnected`?

lib/remote_node.go
```diff
@@ -0,0 +1,599 @@
...
+}
+
+// setStatusTerminated sets the connection status of the remote node to terminated.
+func (rn *RemoteNode) setStatusTerminated() {
+	rn.connectionStatus = RemoteNodeStatus_Terminated
+}
+
+// setStatusAttempted sets the connection status of the remote node to attempted.
+func (rn *RemoteNode) setStatusAttempted() {
+	rn.connectionStatus = RemoteNodeStatus_Attempted
```
lazynina, line 153: Should these `setStatus` functions validate the state transition? For example, should `setStatusAttempted` complain if the status is `RemoteNodeStatus_Terminated`?

lib/remote_node.go
lazynina: seconded!

lib/remote_node.go
```diff
@@ -0,0 +1,599 @@
...
+
+	// Record the services the peer is advertising.
+	vMeta.ServiceFlag = verMsg.Services
+
+	// Record the tstamp sent by the peer and calculate the time offset.
+	timeConnected := time.Unix(verMsg.TstampSecs, 0)
+	vMeta.timeConnected = &timeConnected
+	currentTime := time.Now().Unix()
+	if currentTime > verMsg.TstampSecs {
+		vMeta.timeOffsetSecs = currentTime - verMsg.TstampSecs
```
lazynina, line 390: timeOffsetSecs is an int64, so why not allow it to represent both positive and negative values? This way we know who is "ahead" or "behind". If we do not need this information, then maybe `timeOffsetSecs` should be a uint64.

lazynina (changes requested):

cmd/config.go
```diff
@@ -103,6 +104,7 @@ func LoadConfig() *Config {
 	config.TXIndex = viper.GetBool("txindex")
 	config.Regtest = viper.GetBool("regtest")
 	config.PostgresURI = viper.GetString("postgres-uri")
+	config.PosValidatorSeed = viper.GetString("pos-validator-seed")
```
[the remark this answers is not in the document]
AeonSw4n: I think it should stay as seed. It's not obvious to me if our BIP39 implementation works for BLS keys.

cmd/config.go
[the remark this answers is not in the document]
AeonSw4n: Check my comment re `viper.GetSting()`

cmd/node.go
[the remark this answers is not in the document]
AeonSw4n: Yeaa will fix. I changed it in ConnectionController PR.

cmd/config.go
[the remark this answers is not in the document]
tholonious: Leaving it as PosValidatorSeed seems fine since we're not sure 100% yet whether or not BIP39 seed phrase -> BLS priv key derivation works. So the seed may need to be something else. It's fine to leave it the name as-is until that's disambiguated

lib/remote_node.go
AeonSw4n: I think it ends up cleaner if the caller should make this decision

lib/remote_node.go
AeonSw4n: Addressed

lib/remote_node.go
```diff
@@ -0,0 +1,648 @@
...
+	rn.mtx.Lock()
+	defer rn.mtx.Unlock()
+
+	rn.cmgr.DialPersistentOutboundConnection(netAddr, rn.GetId().ToUint64())
+	rn.setStatusAttempted()
+	return nil
+}
+
+// ConnectInboundPeer connects a peer once a successful inbound connection has been established.
+func (rn *RemoteNode) ConnectInboundPeer(conn net.Conn, na *wire.NetAddress) error {
```
tholonious, line 261: The comment for this function is helpful. Just a nit on the naming. Since the connection has already been established, it's actually being attached to the RemoteNode, rather than a new peer connection being formed
```suggestion
// AttachInboundConnection creates an inbound peer once a successful inbound connection has been established.
func (rn *RemoteNode) AttachInboundConnection(conn net.Conn, na *wire.NetAddress) error {
```

lib/remote_node.go
```diff
@@ -0,0 +1,648 @@
...
+	defer rn.mtx.Unlock()
+
+	id := rn.GetId().ToUint64()
+	rn.peer = rn.cmgr.ConnectPeer(id, conn, na, false, false)
+	rn.setStatusConnected()
+	return nil
+}
+
+// ConnectOutboundPeer connects a peer once a successful outbound connection has been established.
+func (rn *RemoteNode) ConnectOutboundPeer(conn net.Conn, na *wire.NetAddress, isPersistent bool) error {
```
tholonious, line 276:
```suggestion
// AttachOutboundConnection creates an outbound a peer once a successful outbound connection has been established.
func (rn *RemoteNode) AttachOutboundConnection(conn net.Conn, na *wire.NetAddress, isPersistent bool) error {
```

lib/remote_node.go
```diff
@@ -0,0 +1,648 @@
...
+// initiating the handshake, in which case, we should respond with our own version message. To do this, we pass the
+// responseNonce to this function, which we will use in our response version message.
+func (rn *RemoteNode) HandleVersionMessage(verMsg *MsgDeSoVersion, responseNonce uint64) error {
+	rn.mtx.Lock()
+	defer rn.mtx.Unlock()
+
+	if rn.connectionStatus != RemoteNodeStatus_Connected {
+		return fmt.Errorf("HandleVersionMessage: RemoteNode is not connected")
+	}
+	if rn.getHandshakeStage() != HandshakeStage_NotStarted && rn.getHandshakeStage() != HandshakeStage_VersionSent {
```
tholonious, line 393: This should also be an OR, no? 
```suggestion
	if rn.getHandshakeStage() != HandshakeStage_NotStarted || rn.getHandshakeStage() != HandshakeStage_VersionSent {
```

lib/remote_node.go
```diff
@@ -0,0 +1,648 @@
...
+	case RemoteNodeStatus_Connected, RemoteNodeStatus_HandshakeCompleted:
+		rn.cmgr.CloseConnection(id)
+	}
+	rn.setStatusTerminated()
+}
+
+func (rn *RemoteNode) SendMessage(desoMsg DeSoMessage) error {
+	if rn.connectionStatus != RemoteNodeStatus_Connected && rn.connectionStatus != RemoteNodeStatus_HandshakeCompleted {
+		return fmt.Errorf("SendMessage: Remote node is not connected")
+	}
```
tholonious, line 308: I'm a bit confused about the intended use of the `SendMessage` function.

Is the function intended to be called from outside of the RemoteNode struct? If so, you'll need the remote node to have completed the handshake. 

However, you call `SendMessage` for sending Version and Verack messages, in which case, you'll need to be able send the messages regardless of the handshake status of the remote node.

If my understanding above is correct, it seems best to use this `SendMessage` function for external uses only. And to have a private `sendMessage` function for internal uses
```suggestion
func (rn *RemoteNode) SendMessage(desoMsg DeSoMessage) error {
	if rn.connectionStatus != RemoteNodeStatus_HandshakeCompleted() {
		return fmt.Errorf("SendMessage: Remote node is not connected")
	}
	
	return rn.sendMessage(desoMsg)
}

func (rn *RemoteNode) sendMessage(desoMsg DeSoMessage) error {
```

lib/remote_node.go
```diff
@@ -0,0 +1,648 @@
...
+// that is passed in as an argument.
+func (rn *RemoteNode) sendVersionMessage(nonce uint64) error {
+	verMsg := rn.newVersionMessage(nonce)
+
+	// Record the nonce of this version message before we send it so we can
+	// detect self connections and so we can validate that the peer actually
+	// controls the IP she's supposedly communicating to us from.
+	rn.handshakeMetadata.versionNonceSent = nonce
+
+	if err := rn.SendMessage(verMsg); err != nil {
```
tholonious, line 350:
```suggestion
	if err := rn.sendMessage(verMsg); err != nil {
```

lib/remote_node.go
```diff
@@ -0,0 +1,648 @@
...
+}
+
+// sendVerack constructs and sends a verack message to the peer.
+func (rn *RemoteNode) sendVerack() error {
+	verackMsg, err := rn.newVerackMessage()
+	if err != nil {
+		return err
+	}
+
+	if err := rn.SendMessage(verackMsg); err != nil {
```
tholonious, line 470:
```suggestion
	if err := rn.sendMessage(verackMsg); err != nil {
```

lib/remote_node.go
```diff
@@ -0,0 +1,648 @@
...
+	RemoteNodeStatus_Connected          RemoteNodeStatus = 1
+	RemoteNodeStatus_HandshakeCompleted RemoteNodeStatus = 2
+	RemoteNodeStatus_Attempted          RemoteNodeStatus = 3
+	RemoteNodeStatus_Terminated         RemoteNodeStatus = 4
+)
+
+type HandshakeStage uint8
+
+const (
+	HandshakeStage_NotStarted  HandshakeStage = 0
```
tholonious, line 29: Nicely done collapsing the statuses to the fewest possible!

lib/remote_node.go
AeonSw4n: Hmm, I think it should be AND. We should continue this function if NotStarted || VersionSent, and error if the negation is true. So the De Morgan is !NotStarted && !VersionSent.

lib/remote_node.go
AeonSw4n: Nice simplification!

PR #830, opened 2023-11-20
feature/pos-networking-and-syncing

tholonious:

PR #793, opened 2023-11-06
PoS Validator Management

AeonSw4n:

PR #858, opened 2023-12-07
PoS RemoteNodeIndexer

AeonSw4n:


AeonSw4n:
* feature/proof-of-stake
  * **PR #830** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/830?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #768** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/768?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #857** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/857?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
        * **PR #858** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/858?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
      * **PR #792** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/792?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
        * **PR #793** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/793?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/858?utm_source=stack-comment).

lib/pos_remote_node_indexer.go
```diff
@@ -0,0 +1,231 @@
...
+	return rni.NonValidatorOutboundIndex
+}
+
+func (rni *RemoteNodeIndexer) GetNonValidatorInboundIndex() RemoteNodeIndexInterface[RemoteNodeId] {
+	return rni.NonValidatorInboundIndex
+}
+
+func (rni *RemoteNodeIndexer) GetNonValidatorAttemptedIndex() RemoteNodeIndexInterface[RemoteNodeId] {
+	return rni.NonValidatorAttemptedIndex
+}
```
tholonious, line 70: This comment hinges on my suggestion above to merge (NonValidatorOutboundIndex, NonValidatorInboundIndex, and NonValidatorAttemptedIndex) into a single map, and to merge (ValidatorIndex, ValidatorAttemptedIndex) into a single map. This way, you would have a total of three maps:
- ValidatorIndex  map[bls.PublicKey]*RemoteNode
- NonValidatorIndex map[uint64]*RemoteNode
- AllRemoteNodes map[uint64]*RemoteNode

If my suggestion to merge the maps works, then you can simplify the indexer massively by exposing separate Add/Remove/Get functions for the two node types (Validator & NonValidator). It'll eliminate the need for the callback function below, the extra complexity in RemoveRemoteNode to iterate through all of the indices, and the RemoteNodeIndex templated struct itself. The code overall becomes easier to read and simpler to understand, even if does not look as fancy. It's < 3 hours of work to apply this code review feedback.

I know you personally may not like this suggested solution and may not want to apply it on first glance. Especially given you've already built something that works. That said, please give this simplification an honest consideration. And leave this comment on the PR for other reviewers to iterate on for simplicity. I'm sure we'll arrive at something extremely simple and easy to read through once other reviewers have weighed in.

```suggestion
func (rni *RemoteNodeIndexer) AddValidator(node *RemoteNode) {
	rni.ValidatorIndex[node.PublicKey] = node
	rni.AllRemoteNodes[node.PeerID] = node
}

func (rni *RemoteNodeIndexer) RemoveValidator(node *RemoteNode) {
	delete(rni.ValidatorIndex, node.HandshakeMetadata.validatorPublicKey)
	delete(rni.AllRemoteNodes, node.PeerID)
}

func (rni *RemoteNodeIndexer) GetValidator(pk *bls.PublicKey) *RemoteNode {
	node, _ := rni.ValidatorIndex[node.HandshakeMetadata.validatorPublicKey]
	return node
}

func (rni *RemoteNodeIndexer) GetAllValidators() []*RemoteNode {
	return collections.MapValues(rni.ValidatorIndex)
}

func (rni *RemoteNodeIndexer) AddNonValidator(node *RemoteNode) {
	rni.NonValidatorIndex[node.PeerID] = node
	rni.AllRemoteNodes[node.PeerID] = node
}

func (rni *RemoteNodeIndexer) RemoveNonValidator(node *RemoteNode) {
	...
}

func (rni *RemoteNodeIndexer) GetNonValidator(peerID uint64) *RemoteNode {
	...
}

func (rni *RemoteNodeIndexer) GetAllNonValidators() []*RemoteNode {
	...
}
```

lib/pos_remote_node_indexer.go
```diff
@@ -0,0 +1,231 @@
...
+)
+
+// RemoteNodeIndexer is a structure that holds information about all remote nodes and their indices.
+type RemoteNodeIndexer struct {
+	// AllRemoteNodes is a map storing all remote nodes by their IDs.
+	AllRemoteNodes map[RemoteNodeId]*RemoteNode
+
+	// Indices for various types of remote nodes.
+	ValidatorIndex             *RemoteNodeIndex[bls.PublicKey]
+	ValidatorAttemptedIndex    *RemoteNodeIndex[bls.PublicKey]
```
tholonious, line 24: Just for my understanding, why do we need separate maps for connected validators and attempted validators?

We store the validator's connected status in the RemoteNode itself, so it's easy for us to look up the validator by public key and then check its status no? If so, then it's an O(1) operation to perform such a lookup. We don't need to separate ValidatorIndex and ValidatorAttemptedIndex, and can instead have a single maps for them.

lib/pos_remote_node_indexer.go
```diff
@@ -0,0 +1,231 @@
...
+	NonValidatorOutboundIndex  *RemoteNodeIndex[RemoteNodeId]
+	NonValidatorInboundIndex   *RemoteNodeIndex[RemoteNodeId]
+	NonValidatorAttemptedIndex *RemoteNodeIndex[RemoteNodeId]
```
tholonious, line 27: Same comment as above. Why do we need three separate NonValidatorOutboundIndex, NonValidatorInboundIndex, and NonValidatorAttemptedIndex maps? 

The outbound/inbound flag and the attempted/connected status are tracked within the RemoteNode and the underlying peer. So we should just be able to look up the node by id and check it's status when needed. It's an O(1) operation to do this. Unless I'm missing something fundamental, IMO these three maps can be merged into a single one.

lib/pos_remote_node_indexer.go
```diff
@@ -0,0 +1,231 @@
...
+type RNIndexId string
+
+// Constants for different types of remote node indices.
+const (
+	RNIndexId_Validator              RNIndexId = "VALIDATOR"
+	RNIndexId_ValidatorAttempted     RNIndexId = "VALIDATOR_ATTEMPTED"
+	RNIndexId_NonValidator_Outbound  RNIndexId = "NONVALIDATOR_OUTBOUND"
+	RNIndexId_NonValidator_Inbound   RNIndexId = "NONVALIDATOR_INBOUND"
+	RNIndexId_NonValidator_Attempted RNIndexId = "NONVALIDATOR_ATTEMPTED"
+)
```
tholonious, line 15: IMO, there's no need for these identifiers. When an external caller calls the `GetValidatorIndex()` function, they know that the index returned will be the validator index. They don't additionally need a flag inside the index that identifies its type.

Removing this will make your code and the maps indexer more concise

lib/pos_remote_node_indexer.go
```diff
@@ -0,0 +1,231 @@
...
+
+// RemoteNodeIndexer is a structure that holds information about all remote nodes and their indices.
+type RemoteNodeIndexer struct {
+	// AllRemoteNodes is a map storing all remote nodes by their IDs.
+	AllRemoteNodes map[RemoteNodeId]*RemoteNode
```
tholonious, line 20: Any reason you use RemoteNodeId and not PeerID as the key for this map? If a RemoteNodeId is a tuple of (PeerID, AttemptID), and if PeerID is unique for each node, then PeerID alone is sufficient and simpler to use as a unique identifier for each node.

tholonious (changes requested):

PR #875, opened 2023-12-16
feature/pos-syncing-and-steady-state

tholonious:


tholonious:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/deso-protocol/core/875?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* `feature/proof-of-stake`
  * **PR #875** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/875?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
    * **PR #888** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/888?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/875?utm_source=stack-comment).

cmd/config.go
```diff
@@ -176,6 +179,10 @@ func (config *Config) Print() {
 		glog.Infof("Postgres URI: %s", config.PostgresURI)
 	}
 
+	if config.PosValidatorSeedHex != "" {
```
diamondhands0, line 182: Hmm seed hex not mnemonic? I guess because (mnemonic)->seedHex is a one-way operation, so everyone will have the latter but NOT everyone will have the former right?

lib/pos_blockchain.go
```diff
@@ -1128,12 +1440,24 @@ func (bc *Blockchain) shouldReorg(blockNode *BlockNode, currentView uint64) bool
...
+func (bc *Blockchain) addTipBlockToBestChain(blockNode *BlockNode) {
 	bc.bestChain = append(bc.bestChain, blockNode)
 	bc.bestChainMap[*blockNode.Hash] = blockNode
 }
 
+// removeTipBlockFromBestChain removes the current tip from the best chain. It
+// naively removes the tip regardless of the tip's status (committed or not).
+// This function is a general purpose helper function that bundles mutations to
+// the bestChain slice and bestChainMap map.
+func (bc *Blockchain) removeTipBlockFromBestChain() *BlockNode {
```
diamondhands0, line 1453: So I think we're modifying the bestHeaderChain independent of the bestChain, even though we want them to be in-sync. I think it's fine, just feels a bit scary to have two things that need to remain consistent being modified independently.

lib/server.go
```diff
@@ -252,33 +254,24 @@ func (srv *Server) BroadcastTransaction(txn *MsgDeSoTxn) ([]*MempoolTx, error) {
 		return nil, errors.Wrapf(err, "BroadcastTransaction: ")
 	}
```
diamondhands0, line 255: Ugh it deleted this long comment I wrote and now I have to type it again... I was saying that this could break frontends because they have the following flow:
1. Backend constructs txn and sends it to frontend to sign
2. Frontend signs and sends to /submit-transaction
3. /submit-transaction calls BroadcastTransaction (this thing) and returns
4. Then the frontend refreshes under the assumption the txn is in the mempool and its effect will show up after the refresh.

If we delete this thing then it could cause frontends to miss the txn's effect on refresh. A few ways to fix this:
1. Have frontend wait for inclusion in a block, which will take a bit of time. Not a big deal. We could wait for a post but maybe not a diamond or things like that.
2. Alternatively we could be smarter about how we render things in the mempool in the frontend.

I don't think we have to worry about this for now. I'm just mentioning for when you encounter it later.

lib/server.go
```diff
@@ -490,6 +484,36 @@ func NewServer(
...
+		return nil, errors.New("NewServer: Problem getting latest block from chain"), true
+	}
+	err = _posMempool.Init(
+		_params,
+		currentGlobalParamsEntry,
+		currentUtxoView,
+		uint64(_chain.blockTip().Height),
+		_mempoolDumpDir,
+		false,
+		1024*1024*1024*3, // Max mempool Size = 3GB; TODO make this a param
```
diamondhands0, line 508: I've been thinking about this. Better to make it a param or flag? I think flag to start would be good so we can tune it. Eventually for Revolution it needs to be more static, but that can wait...

lib/server.go
```diff
@@ -817,6 +847,16 @@ func (srv *Server) _handleHeaderBundle(pp *Peer, msg *MsgDeSoHeaderBundle) {
 		len(msg.Headers), srv.blockchain.chainState(), pp,
 		srv.blockchain.headerTip().Header.Height, printHeight)))
 
+	// If the node is running a Fast-HotStuff validator and the consensus is running,
+	// in the steady-state, then it means that we don't sync the header chain separately.
+	if srv.fastHotStuffConsensus != nil && srv.fastHotStuffConsensus.IsRunning() {
```
diamondhands0, line 852: I will come back to this, but we used to use this in the steady-state to keep the node in sync as follows:
1. We call GetHeaders on a peer
2. They send us a header bundle
3. Based on that we may or may not fetch a block

If we're not doing the above, then we must have a different way of fetchin ga block when the consensus is running. Is this because blocks are "pushed" to us? I'll keep reading.

lib/server.go
```diff
@@ -1728,16 +1790,36 @@ func (srv *Server) _addNewTxn(
...
+	if blockHeight < srv.blockchain.params.ForkHeights.ProofOfStake2ConsensusCutoverBlockHeight {
+		_, err := srv.mempool.ProcessTransaction(
+			txn, true /*allowUnconnectedTxn*/, rateLimit, peerID, verifySignatures)
+		if err != nil {
+			return nil, errors.Wrapf(err, "Server._addNewTxn: Problem adding transaction to mempool: ")
+		}
+
+		// At this point, we know the transaction has been run through the mempool.
+		// Now wait for an update of the ReadOnlyUtxoView so we don't break anything.
+		srv.mempool.BlockUntilReadOnlyViewRegenerated()
```
diamondhands0, line 1810: I think there could be a problem if you hold the the ChainLock.RLock() while you wait for this thing. Maybe not. It does take like a whole second or two, though, and I think kindof bad to hold the RLock for that long maybe idk. Think about it.

lib/server.go
```diff
@@ -2026,25 +2163,52 @@ func (srv *Server) _handleGetTransactions(pp *Peer, msg *MsgDeSoGetTransactions)
...
+		// it's an invalid txn.
+		if err != nil {
+			return nil, errors.Wrapf(err, "Server.ProcessSingleTxnWithChainLock: Problem adding transaction to PoW mempool: ")
+		}
+	}
+
+	// Regardless of the consensus protocol we're running (PoW or PoS), we use the PoS mempool's to house all
+	// mempool txns. If a txn can't make it into the PoS mempool, which uses a looser unspent balance check for
+	// the the transactor, then it must be invalid.
+	if err := srv.posMempool.AddTransaction(NewMempoolTransaction(txn, uint64(time.Now().UnixMicro())), true); err != nil {
```
diamondhands0, line 2203: This makes me a little nervous as there might be edge cases where it's rejected that we haven't thought of, but I can't think of any so probably OK.

lib/server.go
```diff
@@ -2338,11 +2500,11 @@ func (srv *Server) _startConsensus() {
 		}
 
 		select {
-		// case consensusEvent := <-srv.fastHotStuffEventLoop.GetEvents():
-		// 	{
-		// 		glog.Infof("Server._startConsensus: Received consensus event for block height: %v", consensusEvent.TipBlockHeight)
-		// 		srv._handleFastHostStuffConsensusEvent(consensusEvent)
-		// 	}
+		case consensusEvent := <-srv.fastHotStuffConsensus.fastHotStuffEventLoop.GetEvents():
```
diamondhands0, line 2503: Huh. That's all we had to add eh? Pretty elegant...

lib/pos_epoch_complete_hook.go
```diff
@@ -140,12 +140,28 @@ func (bav *UtxoView) runEpochCompleteSnapshotGeneration(epochNumber uint64) erro
 	return nil
 }
 
+// simulateNextEpochEntry simulates the block range for the next epoch given the current epoch's final
+// block height and epoch number. The view and timestamp for the simulated epoch are left empty since they can't
+// be easily simulated, so DO NOT USE CreatedAtBlockTimestampNanoSecs or InitialView from the returned EpochEntry.
+func (bav *UtxoView) simulateNextEpochEntry(currentEpochNumber uint64, currentEpochFinalBlockHeight uint64) (*EpochEntry, error) {
```
diamondhands0, line 146: Mention why we need this. I saw it but nice to have it there.

lib/server.go
tholonious: Hmm, good call. We really only need to hold the `ChainLock.RLock()` while reading the blockTip's height. I've updated the code to reflect that

lib/server.go
tholonious: Yup. That's all we need :)

cmd/config.go
tholonious: AFAIK, there isn't a standard derivation path to convert a BIP39 seed phrase to BLS private key. So we use the standard seedHex -> bls.PrivKey conversion provided by the FlowCrypto package.

lib/pos_epoch_complete_hook.go
tholonious: Done.

lib/server.go
tholonious: Yes, we want to make this a flag that's easily tunable via ENV variables. The comment to make it a param was just ambiguous. I've fixed it

lib/server.go
tholonious: Agreed. If a txn is accepted by the PoW mempool, then it must be accepted by the PoS mempool too. The PoS mempool has a strictly relaxed set of txn acceptance rules compared to its PoW counterpart. I can't think of any edge cases either, so this *should* be safe

lib/server.go
tholonious: Yes! We have a different and simpler way of fetching blocks once the new consensus is running:
1. Blocks are pushed to us without us having to request them in the new consensus
2. Whenever we see an orphan block with an unknown ancestor, we request the missing ancestor a peer until we have a fully connected chain

lib/pos_blockchain.go
tholonious: Hmm, I'm not sure this is much of an issue. I did notice in ProcessBlockPoW that we cann processHeaderPoW separately, which mutates the bestHeaderChain separately so I figured this would be safe.

I'll noodle on this to see if we can better couple mutations to the bestChain and bestHeaderChain here while keeping the reorgs simple. LMK if you have a strong opinion on this one

lib/server.go
```diff
@@ -490,6 +484,36 @@ func NewServer(
...
+		return nil, errors.New("NewServer: Problem getting latest block from chain"), true
+	}
+	err = _posMempool.Init(
+		_params,
+		currentGlobalParamsEntry,
+		currentUtxoView,
+		uint64(_chain.blockTip().Height),
+		_mempoolDumpDir,
+		false,
+		1024*1024*1024*3, // Max mempool Size = 3GB; TODO make this a flag
```
diamondhands0, line 508: Start this out as a configurable flag on the node.

lib/pos_consensus.go
```diff
@@ -0,0 +1,685 @@
...
+	if err != nil {
+		// If we get an error here, it means something went wrong with the block processing algorithm.
+		// Nothing we can do to recover here.
+		return errors.Errorf("FastHotStuffConsensus.HandleBlock: Error processing block as new tip: %v", err)
+	}
+
+	// If there are missing block hashes, then we need to fetch the missing blocks from the network
+	// and retry processing the block as a new tip. We'll request the blocks from the same peer.
+	if len(missingBlockHashes) > 0 {
+		pp.QueueMessage(&MsgDeSoGetBlocks{
```
diamondhands0, line 440: This seems like it should totally work, but a thought I had is it may be inefficient in the sense that you will do a lot of round-trips if you get a far-ahead block. For example, if the peer has `A -> B -> C -> D -> E -> F` and you're at just `A` and the peer sends you `F`, then you will try to connect F, fail, then request E, then you'll get E, try to connect it fail and request D, and so on. I don't think we need to optimize this right now because nodes will generally be nearly-synced at this stage. But leave a comment that this behavior exists if it's correct.

One way to fix this is to send your tip inside the GetBlocks message, and then the node can auto-send you everything from your tip to its tip. But again I think this is probably a premature optimization we don't need to make right now. If you like that optimization then leave it as a TODO for fun.

diamondhands0: Btw just added context: The above suggestion is a simplified version of what GetHeaders does. The GetHeaders thing does a sophisticated "common ancestor" check between the two nodes by sending an exponentially-backed-off list of your headers, eg: `tip, tip-1, tip-2, tip-4, tip-8, tip-16, ..., genesis`. This covers cases where nodes are super divergent, which can happen in Bitcoin but can't happen in PoS so it doesn't make sense to port it over. Instead, sending the tip only, and then having the node send you its next N blocks from the tip automatically, seems like the "most right" thing to do with PoS. But again I think what you have works and just adding a TODO is good and will work great.

lib/pos_consensus.go
```diff
@@ -0,0 +1,685 @@
...
+	if err != nil {
+		return nil, errors.Errorf("error fetching validator lists for safe blocks: %v", err)
+	}
+
+	// If the block was processed successfully but was not applied as the new tip, we need up date the safe
+	// blocks in the FastHotStuffEventLoop. This is because the new block may be safe to extend even though
+	// it did not result in a new tip.
+	if !successfullyAppliedNewTip {
+		// Update the safe blocks to the FastHotStuffEventLoop
+		if err = cc.fastHotStuffEventLoop.UpdateSafeBlocks(safeBlocksWithValidators); err != nil {
```
diamondhands0, line 493: Hmmm... it feels like you don't need to store the safe blocks on the fc object. Rather, you can just compute them on the fly from the blockchain object every time. This would be like a *tiny* bit less efficient, but it would also be less redundant and less error-prone.

Is there a reason you want to store and update a list of safeblocks on the fc object rather than recomputing them on the fly every time?

diamondhands0: Ah I see... it looks like the fastHotStuffEventLoop object doesn't have a blockchain object, which is why you have to do this update manually. This makes sense, and I like the design because it makes it much easier to test the consensus code.

diamondhands0: Ok but I'm still confused about something. Don't you *always* need to call UpdateSafeBlocks, even if the new block was added to the tip? I'm not seeing where FastHotStuffEventLoop.safeBlocks gets updated in the case where we just attach a block to the tip, even though I think we should be adding the tip block to safeBlocks in that case.

diamondhands0: Sorry nevermind. It's a few lines later, and gets updated in FastHotStuffEventLoop.ProcessTipBlock().

lib/server.go
```diff
@@ -2000,9 +2118,30 @@ func (srv *Server) _handleBlock(pp *Peer, blk *MsgDeSoBlock) {
...
+	// This requires the block height check to use ProofOfStake2ConsensusCutoverBlockHeight-1
+	tipHeight := srv.blockchain.blockTip().Height
+	if tipHeight < srv.blockchain.params.ForkHeights.ProofOfStake2ConsensusCutoverBlockHeight-1 {
+		return
+	}
+
+	// If the PoS validator FastHotStuffConsensus is initialized but not yet running, then
+	// we can start the validator consensus, and transition to it in the steady-state.
+	if srv.fastHotStuffConsensus != nil && !srv.fastHotStuffConsensus.IsRunning() {
+		srv.fastHotStuffConsensus.Start()
```
diamondhands0, line 2143: Just a heads-up. By the time we get here, we're guaranteed that our tip will be within MaxTipAge of the "true" tip on the network, where MaxTipAge is 24h. Specifically, the order is: 1) sync all the headers, 2) download all the blocks for those headers, 3) the next thing that would trigger with your code is _handleBlock().

If we assume the above, then we could potentially be an hour or more behind at step (2) causing the first tip block we get to do a lot of catch-up. I think this is probably fine, since an hour of blocks is only 3600 little blocks. But it's something to keep in mind as your testing. And if it's slow or suboptimal, you can do the update to add your committed tip to GetBlocks() that I mentioned previously.

cmd/config.go
diamondhands0: Hmm... so BIP39 is actually supposed to go from `mnemonic -> seedHex`, not to be confused with bip32 which deals with derivation paths and things like that. You can open identity and hit "advanced" on the part where it shows you a seed phrase and then see how it converts it:
https://ibb.co/wwW1rKH

My thinking is it would probably be good to make it so that the flag is a mnemonic, and that the seed hex used is identical to that generated by identity, which is standard bip39 generation.

I could also see the argument for keeping it as raw seed hex. In that case we'll just have to tell people to copy it from the advanced section or something. Also it feels easier to lose a seed hex than a mnemonic but idk.

diamondhands0: There should be a function somewhere that's like MnemonicToSeed or SeedToMnemonic

lib/pos_consensus.go
```diff
@@ -0,0 +1,685 @@
...
+// HandleLocalVoteEvent is triggered when FastHotStuffEventLoop has signaled that it wants to
+// vote on the current tip. This functions validates the vote signal, then it constructs the
+// vote message here.
+//
+// Steps:
+// 1. Verify that the event is properly formed.
+// 2. Construct the vote message
+// 3. Process the vote in the consensus module
+// 4. Broadcast the vote msg to the network
+func (cc *FastHotStuffConsensus) HandleLocalVoteEvent(event *consensus.FastHotStuffEvent) error {
```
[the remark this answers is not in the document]
tholonious: Yes, this requires the connection management code to be complete. For now, I independently implemented the consensus and added TODO's at the exact lines where network messages would be broadcast. The network send calls will be plug-and-play once Piotr has finished the networking code.

cmd/config.go
tholonious: Oh, I like this. And agreed that a seed phrase is easier to manage for node operators, vs a more arbitrary seed hex. The one thing I'm not sure of is whether the seed hex generated via BIP39 matches the number of bytes that the bls private key expects. I'll try to update to this.

tholonious: Implemented in PR https://github.com/deso-protocol/core/pull/946

lib/server.go
tholonious: Implemented in PR https://github.com/deso-protocol/core/pull/949

lib/server.go
tholonious: You're right, this totally breaks the frontends. I'm seeing it already on regtest. cc: @lazynina since this is something we'll need to solve.

Past discussions on this landed on optimistic UI updates for less critical and idempotent txns (posts, comments, likes,...) and to wait for block confirmation for more sensitive txns (basic transfers, limit order,...). The state syncer was the natural place to do optimistic updates. As a v0 solution though, maybe it's fine to wait for one block conf for everything cc: @lazynina so you're aware

lib/pos_blockchain.go
tholonious: On second review, updating the bestChain in lockstep with the bestHeaderChain here is fine. It's turns out to be a no-op from the context of running processBlockPoS because the bestHeaderChain always ends up with the same final state as the bestChain. But the peace of mind of updating them in lockstep here is a nice to have. I've addressed your feedback in https://github.com/deso-protocol/core/pull/962

lib/pos_consensus.go
tholonious: Added the comment here: https://github.com/deso-protocol/core/pull/963

lib/server.go
tholonious: The only caveat is that depending on how far back the timed out node is it, the syncing may be inefficient. It's the same inefficiency you pointed out in https://github.com/deso-protocol/core/pull/875#discussion_r1460183510

I think that's an optimization we can defer for the future depending on testing looks on devnet

cmd/config.go
```diff
@@ -42,9 +43,17 @@ type Config struct {
 	DisableEncoderMigrations  bool
 	HypersyncMaxQueueSize     uint32
 
+	// PoS Validator
+	PosValidatorSeed                       string
```
diamondhands0, line 47: Something I was telling Zordon is we should make it so these arguments accept EITHER a "seed hex" OR a "seed phrase." Not sure if you want to do that here, but I'm mentioning as it came in handy elsewhere. I think he made it so that if it starts with 0x, he reads it as a seed hex, otherwise he reads it as a seed phrase.

cmd/run.go
```diff
@@ -82,12 +82,28 @@ func SetupRunFlags(cmd *cobra.Command) {
 		- hypersync: Will sync by downloading historical state, and will NOT
 		  download historical blocks. Can only be set if HyperSync is true.`)
 
+	// PoS Validator
+	cmd.PersistentFlags().String("pos-validator-seed", "", "A BIP39 seed phrase used to generate the private key of the "+
+		"Proof of Stake validator. Setting this flag automatically makes the node run as a Proof of Stake Validator.")
+	cmd.PersistentFlags().Uint64("pos-block-production-interval-milliseconds", 1500, "The interval in milliseconds "+
```
diamondhands0, line 88: Hmm... is it better for these to be flags or for them to be ParamUpdater? I'm thinking about how we would change these values later, for example if we want to reduce the block production interval to 1s or even faster if we think the network can handle it.

Right now, there's not really a good upgrade path here. You would need everyone to upgrade their flags at the exact same time or else the network would break. I think originally we said we'd put these in ParamUpdater for this reason, since it can safely change these values.

diamondhands0: Separate question is whether we should start this number low and increase it OR start it high and decrease it. I thought about it, and it feels like starting LOW is actually better. If we start it low, we'll encounter issues in devnet and know what our "limits" are. However, if we start it high we may be running at too generous of a buffer and always be too afraid to reduce it.

This is also something that seems easier to test with ParamUpdater, since you can play with values quickly while the network is live and see what happens, rather than having to reboot the whole network.

cmd/run.go
```diff
@@ -82,12 +82,28 @@ func SetupRunFlags(cmd *cobra.Command) {
...
+		"that the Proof of Stake block proposer waits before timing out and producing a block. The default value is "+
+		"30 seconds, or 30000 milliseconds.")
+
 	// Mempool
 	cmd.PersistentFlags().Uint64("mempool-backup-time-millis", 30000,
 		"The frequency in milliseconds with which the mempool will persist its state to disk. "+
 			"The default value is 30 seconds, or 30,000 milliseconds.")
 	cmd.PersistentFlags().Uint64("max-mempool-pos-size-bytes", 3000000000,
 		"The maximum size of the PoS mempool in bytes. The default value is 3GB.")
+	cmd.PersistentFlags().Uint64("mempool-fee-estimator-num-mempool-blocks", 1,
```
diamondhands0, line 100: These mempool params seem appropriate as flags. In my view, any consensus variables that require coordination among node operators should be in ParamUpdater probably. Anything that individual nodes can update independently should be flags.

lib/server.go
```diff
@@ -1859,6 +2006,17 @@ func (srv *Server) _handleBlock(pp *Peer, blk *MsgDeSoBlock) {
 	glog.Infof(CLog(Cyan, fmt.Sprintf("Server._handleBlock: Received block ( %v / %v ) from Peer %v",
 		blk.Header.Height, srv.blockchain.headerTip().Height, pp)))
 
+	// If the node is running a Fast-HotStuff validator and the consensus is running,
+	// in the steady-state, then we handle the block according to the consensus rules.
+	if srv.fastHotStuffConsensus != nil && srv.fastHotStuffConsensus.IsRunning() {
+		srv.fastHotStuffConsensus.HandleBlock(pp, blk)
```
diamondhands0, line 2012: I notice you're ignoring the error here. What it looks like is happening right now is the following:
0. When we boot up the node before the PoS cut-over height, we will set its PoS params, which will cause us to enter this loop (note that IsRunning() doesn't have a block height check).
1. Before the PoS cut-over height, we are calling fasthotStuffConsensus.HandleBlock() -> processBlockPoS -> tryConnectBlockAsNewTip (or whatever it's called), which is erroring because we're not at the cut-over height yet.
2. The error bubbles all the way up to here, where it's ignored.
3. Eventually, we pass the cut-over height and this stops erroring.

Instead, I recommend the following:
1. Gate this on the cut-over height so we don't enter it until we're supposed to.
2. Catch the error here, since it probably means something is messed up.

lib/server.go
```diff
@@ -2010,9 +2168,30 @@ func (srv *Server) _handleBlock(pp *Peer, blk *MsgDeSoBlock) {
...
+	// Exit early if the chain isn't SyncStateFullyCurrent.
+	if srv.blockchain.chainState() != SyncStateFullyCurrent {
+		return
+	}
+
+	// Exit early if the current tip height is below the PoS cutover height. We are ready to enable
+	// the PoS validator FastHotStuffConsensus once we reach the final block of the PoW protocol.
+	// This requires the block height check to use ProofOfStake2ConsensusCutoverBlockHeight-1
+	tipHeight := uint64(srv.blockchain.blockTip().Height)
+	if tipHeight < srv.blockchain.GetFinalPoWBlockHeight() {
```
diamondhands0, line 2186: Why not `!srv.blockchain.IsPoSBlockHeight` here?

lib/pos_blockchain.go
```diff
@@ -1405,7 +1779,68 @@ func (bc *Blockchain) getMaxSequentialBlockHeightAfter(startingHeight uint64) ui
...
+	}
+
+	aggregatedSignature, signersList, err := BuildQuorumCertificateAsProofOfStakeCutoverValidator(finalPoWBlock.Header.Height, finalPoWBlock.Hash)
+	if err != nil {
+		return nil, err
+	}
+
+	qc := &QuorumCertificate{
+		BlockHash:      finalPoWBlock.Hash,
+		ProposedInView: finalPoWBlock.Header.Height,
```
diamondhands0, line 1795: I see we set ProposedInView here, which is good. But again I'm not seeing how ew get past the error in getLineageFromCommittedTip().

lib/blockchain.go
```diff
@@ -1907,20 +1912,21 @@ func (bc *Blockchain) ProcessBlock(desoBlock *MsgDeSoBlock, verifySignatures boo
...
-	// If the block's height is after the PoS cut-over fork height, then we use the PoS block processing logic. Otherwise, fall back
-	// to the PoW logic.
-	if desoBlock.Header.Height >= uint64(bc.params.ForkHeights.ProofOfStake2ConsensusCutoverBlockHeight) {
-		// TODO: call bc.processBlockPoS(desoBlock, verifySignatures) instead
-		isMainChain, isOrphan, err := bc.processBlockPoW(desoBlock, verifySignatures)
-		return isMainChain, isOrphan, nil, err
+	// If the block's height is after the PoS cut-over fork height, then we use the PoS block processing logic.
+	// Otherwise, fall back to the PoW logic.
+	if bc.IsPoSBlockHeight(desoBlock.Header.Height) {
+		return bc.processBlockPoS(desoBlock, 1, verifySignatures)
```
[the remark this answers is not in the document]
diamondhands0: Did some more digging. It looks like you define a MsgDeSoHeader.GetView() that patches the view with the height for HeaderVersion1/. BUT the code in getLineageFromTip() still looks wrong because it's calling block.Header.ProposedInview. Will keep reading...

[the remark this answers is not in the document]
diamondhands0: Mystery solved. Before the line that checks ProposedInView in getCommittedTip(), there is this line, which breaks out of the for loop:
```
if currentBlock.Hash.IsEqual(highestCommittedBlock.Hash) {
    break
```

This results in an empty ancestors list being returned, which is what we expect.

diamondhands0:
In pos_snapshot_entries.go, please verify that this assumption is still true:
```
func (bav *UtxoView) ComputeSnapshotEpochNumberForEpoch(epochNumber uint64) (uint64, error) {
	if epochNumber < SnapshotLookbackNumEpochs {
		// We want to return 0 in this case and not error. We start snapshotting with our StateSetup block height,
		// so we should have the correct number of snapshots and not hit this case once we hit the ConsensusCutover
		// block height. This case will only be hit immediately following the StateSetup block height. We run one
		// OnEpochCompleteHook right away on the StateSetup block height which will increment our CurrentEpochNumber
		// from zero (the starting default) to one. Then we wait one epoch and run our second OnEpochCompleteHook to
		// increment our CurrentEpochNumber from one to two. At this point, we will have the correct number of
		// snapshots and no longer hit this edge case.
		//
		// The problem is what about snapshot values we need to use in that first block where CurrentBlockHeight =
		// StateSetup block height and then the first epoch after that? The only snapshot values that we use relate
		// to our new PoS txn types. We pull the snapshot GlobalParamsEntry to retrieve the StakeLockupEpochDuration
		// and the ValidatorJailEpochDuration. Both of these impact the new PoS txn types which are unlocked after
		// the StateSetup block height. The ValidatorJailEpochDuration value doesn't really matter since no validators
		// will be jailed until the ConsensusCutover block height. For the StakeLockupEpochDuration (and all other
		// snapshot GlobalParamsEntry values), if there is no snapshot value, we return an empty GlobalParamsEntry with
		// just our defaults, which is what we intend. There's one other small edge case here which is if we update the
		// StakeLockupEpochDuration parameter within that first block (which would be weird and should not happen),
		// then that value will take immediate effect in the first epoch with no lagged snapshot wait period.
		return 0, nil
	}
	return SafeUint64().Sub(epochNumber, SnapshotLookbackNumEpochs)
}
```

diamondhands0:
AHA! Now I'm pretty sure I've got you good. In ConnectBlock(), the first time you call RunEpochCompleteHook() at height = cutover1, you call it with blockHeader.ProposedInView, which is ZERO because it's a PoW block. This trickles all the way down through runEpochCompleteEpochRollover to computeNextEpochEntry where this happens:
```
	nextEpochEntry := &EpochEntry{
		EpochNumber:                     currentEpochNumber + 1,
		InitialBlockHeight:              currentEpochFinalBlockHeight + 1,
		InitialView:                     currentEpochFinalView + 1,
		FinalBlockHeight:                nextEpochFinalBlockHeight,
		CreatedAtBlockTimestampNanoSecs: nextEpochBlockTimestampNanoSecs,
	}
```

Can you see what's wrong here? `currentEpochFinalView` is ZERO when it should (I think) be blockHeight. If you call GetView() instead of Header.ProposedInView, I *think* it should fix everything. As it's written now, I think you will set `InitialView = 0 + 1` until you get to cutover2, at which point I'm not really sure what happens yet...

diamondhands0:
At height = cutover2, you are called processBlockPoS on a PoW block that does not have `block.Header.ProposerVotingPublicKey` nor `block.Header.ProposerVotePartialSignature` set (I think). At minimum, I can't see how these fields would be set for the first call of processBlockPoS. If this is the case, I'm confused about how we're managing to pass through `processBlockPoS > validateLeaderAndQC > hasValidProposerPartialSignaturePoS` since the latter has this check at the top:
```
	votingPublicKey := block.Header.ProposerVotingPublicKey
	proposerPartialSig := block.Header.ProposerVotePartialSignature
	// If the proposer partial sig is nil, we can't validate it. That's an error.
	if proposerPartialSig.IsEmpty() {
		return false, nil
	}
```

I'm going to keep reading...

lib/server.go
tholonious: Sounds good. I'll catch and log the error here

lib/server.go
tholonious: We need to start the FastHotStuffConsensus once the chain has connected the final PoW block, which has a height of `ProofOfStake2ConsensusCutoverBlockHeight - 1`. The final PoW block is the genesis block for the PoS chain, on top of which the PoS consensus votes or times out on

cmd/run.go
tholonious: Sounds good. Will move these to ParamUpdater too

lib/server.go
tholonious: Resolved in PR https://github.com/deso-protocol/backend/pull/538

tholonious:
> AHA! Now I'm pretty sure I've got you good. In ConnectBlock(), the first time you call RunEpochCompleteHook() at height = cutover1, you call it with blockHeader.ProposedInView, which is ZERO because it's a PoW block. This trickles all the way down through runEpochCompleteEpochRollover to computeNextEpochEntry where this happens:
> 
> ```
> 	nextEpochEntry := &EpochEntry{
> 		EpochNumber:                     currentEpochNumber + 1,
> 		InitialBlockHeight:              currentEpochFinalBlockHeight + 1,
> 		InitialView:                     currentEpochFinalView + 1,
> 		FinalBlockHeight:                nextEpochFinalBlockHeight,
> 		CreatedAtBlockTimestampNanoSecs: nextEpochBlockTimestampNanoSecs,
> 	}
> ```
> 
> Can you see what's wrong here? `currentEpochFinalView` is ZERO when it should (I think) be blockHeight. If you call GetView() instead of Header.ProposedInView, I _think_ it should fix everything. As it's written now, I think you will set `InitialView = 0 + 1` until you get to cutover2, at which point I'm not really sure what happens yet...

Nice catch. I think this one may have become an issue in the edge case where the PoS cutover takes place in the middle on an epoch. That epoch would have had an InitialView value of 0 which would offset off the initial block proposer in an unexpected way after the cutover. It's safest to pass in the GetView() value as you suggest.

lib/miner.go
```diff
@@ -190,6 +198,10 @@ func (desoMiner *DeSoMiner) MineAndProcessSingleBlock(threadIndex uint32, mempoo
 		return nil, fmt.Errorf("DeSoMiner._startThread: _mineSingleBlock returned nil; should only happen if we're stopping")
 	}
 
+	if blockToMine.Header.Height >= uint64(desoMiner.params.ForkHeights.ProofOfStake2ConsensusCutoverBlockHeight) {
+		return nil, fmt.Errorf("DeSoMiner._startThread: _mineSingleBlock returned a block that is past the Proof of Stake Cutover")
```
AeonSw4n, line 202: Should we replace this with `IsPoSBlockHeight` for consistency?

lib/pos_blockchain.go
tholonious: Yeah, this line is safe. I updated it to set `ProposedInView: finalPoWBlock.Header.GetView()` so it's clearer too

lib/miner.go
tholonious: Resolved in PR https://github.com/deso-protocol/core/pull/1009

lib/server.go
tholonious: Clarified the comment further in PR https://github.com/deso-protocol/core/pull/1010

lib/pos_consensus.go
```diff
@@ -0,0 +1,843 @@
...
+	// change.
+	cc.blockchain.ChainLock.RLock()
+	defer cc.blockchain.ChainLock.RUnlock()
+
+	// Fetch the current tip of the chain
+	tipBlock := cc.blockchain.BlockTip()
+	tipHeight := tipBlock.Header.Height
+
+	// If the chain is not at the final PoW block height or higher, then we cannot start the PoS consensus.
+	if cc.params.IsPoWBlockHeight(tipHeight) && !cc.params.IsFinalPoWBlockHeight(tipHeight) {
```
diamondhands0, line 69: Can you simplify this as simply:
```
if tipHeight < cc.params.GetFinalPoWBlockHeight()
```
like you have in server.go?

lib/pos_constants.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/deso-protocol/core/collections/bitset"
+	"github.com/deso-protocol/core/consensus"
+	"github.com/holiman/uint256"
+	"github.com/pkg/errors"
+)
+
+// The Proof of Work -> Proof of Stake cutover requires a synthetic QC to protect against a timeout
+// during the exact point of the cutover. This synthetic QC is built and signed locally by every node
+// using a known and consistent private key.
+const proofOfStakeCutoverValidatorBLSPrivateKeyHex = "0x0570b78ce822f902b203ee075a7e2147d6b9a420a9409c038154589de64eec96"
```
diamondhands0, line 14: I would explain a bit more of the "why" here. I think we're doing this because it's the easiest way to special-case our QC validation logic at the cutover height. But just explain it so we know/remember.

consensus/types.go
```diff
@@ -51,7 +67,7 @@ type BlockHashValue = [32]byte
 type FastHotStuffEventLoop interface {
 	GetEvents() chan *FastHotStuffEvent
 
-	Init(time.Duration, time.Duration, BlockWithValidatorList, []BlockWithValidatorList) error
+	Init(time.Duration, time.Duration, QuorumCertificate, BlockWithValidatorList, []BlockWithValidatorList) error
 	GetCurrentView() uint64
 	AdvanceViewOnTimeout() (uint64, error)
 	ProcessTipBlock(BlockWithValidatorList, []BlockWithValidatorList) error
```
diamondhands0, line 73: Every time I hit "Go to Definition" for one of these, it takes me here rather than to the implementation. Even hitting "Show Usages" doesn't take me there. Is there an easy way to jump directly to the implementation that I'm not aware of?

lib/pos_consensus.go
```diff
@@ -0,0 +1,843 @@
...
+	if err != nil {
+		return errors.Wrapf(err, "Error producing unsigned block for proposal at height %d", event.TipBlockHeight+1)
+	}
+
+	// Sign the block
+	blockHash, err := unsignedBlock.Header.Hash()
+	if err != nil {
+		return errors.Errorf("Error hashing block: %v", err)
+	}
+	unsignedBlock.Header.ProposerVotePartialSignature, err = cc.signer.SignBlockProposal(unsignedBlock.Header.ProposedInView, blockHash)
```
diamondhands0, line 264: I don't think there's any way for this to be incorrect but just to be safe I would replace all instances of Header.ProposedInview with GetView(), including this instance here.

lib/pos_block_producer.go
```diff
@@ -75,7 +83,7 @@ func (pbp *PosBlockProducer) CreateUnsignedTimeoutBlock(latestBlockView *UtxoVie
 func (pbp *PosBlockProducer) createBlockTemplate(latestBlockView *UtxoView, newBlockHeight uint64, view uint64,
 	proposerRandomSeedSignature *bls.Signature) (BlockTemplate, error) {
 	// First get the block without the header.
-	currentTimestamp := time.Now().UnixNano()
+	currentTimestamp := pbp.timeSource.AdjustedTime().UnixNano()
```
diamondhands0, line 86: We shouldn't rely on a timeSource anymore. For context, Bitcoin has a really weird way of dealing with time that we sortof emulated before, but that we're not going to be porting over to PoS.

For PoS, we want to simply set the block timestamp to:
`currentTimestamp = MIN(time.UnixNano(), parentBlock.TstampNanos + 1)`

Critically, the nodes on the network should allow for some amount of drift, say a few minutes, but should reject blocks outside of that strict time range. Bitcoin does this rejection but its window is 2 hours or so, and it does not enforce monotonically increasing timestamps, which causes complications (it would be better if it did).

I remember we decided that monotonically increasing the timestamp would prevent attacks, but now that I'm thinking about it I'm not 100% sure we need it. It prevents attacks in Proof of Work because the timestamp is used by Proof of Work to adjust the difficulty, but I don't think we rely explicitly on the timestamp for anything in consensus itself, and so I think just having it be `currentTimestamp = time.UnixNano()` would probably work. In this case, a validator can fudge the time up to our strict time window for one block, but then the next validator would correct it.

The above being said, I recommend keeping monotonically increasing timestamps unless we have a good reason not to have them. Among other things, they lead to more consistent behavior for things that rely on the timestamp, such as lockups. In the specific example of lockups, you could have a weird case where a lockup is unlockable for one block but not unlockable for the next, which is kinda dumb (though not fatal).

So I would just leave a comment explaining as much as you can about the relevant context above.

Also just as a history lesson, I believe Bitcoin Core sets its timestamp based on the median of what its peers tell it the time is or something like that. I don't know if it still does this, but I think that's how Satoshi had it. It prevents your clock from drifting too hard from everyone else's clock, so that you're always suggesting blocks that are within their window.

diamondhands0:
I couldn't comment on this line but in pos_block_producer.go you have a line where you're setting `	blockRewardOutput.AmountNanos = math.MaxUint64`. You fix it later but could you just set that to zero by default or not? Consensus would reject it if it were wrong regardless, it just makes me a little more comfortable.

lib/pos_block_producer.go
```diff
@@ -169,6 +182,7 @@ func (pbp *PosBlockProducer) getBlockTransactions(
 		if err != nil {
 			return nil, nil, 0, errors.Wrapf(err, "Error getting transaction size: ")
 		}
+
 		// Skip over transactions that are too big.
 		if currentBlockSize+uint64(len(txnBytes)) > maxBlockSizeBytes {
```
diamondhands0, line 187: Make sure that the mempool cannot accept a txn above this size or it will sit in the mempool forever being skipped by the block producer.

diamondhands0:
In block_view.go there is a line `effectiveFeeU256 := failingTransactionRate.Mul(failingTransactionRate, failingTransactionFee)`. Change it to `effectiveFeeU256 := uint256.NewInt().Mul(failingTransactionRate, failingTransactionFee)`

Always use NewInt when assigning to a new variable.

diamondhands0:
Also in block_view.go:3928, I think you could overflow after the multiplication.
```
	effectiveFeeU256 := failingTransactionRate.Mul(failingTransactionRate, failingTransactionFee)
        // Add a check for overflow here. 
	effectiveFeeU256.Div(effectiveFeeU256, basisPointsAsUint256)
```

I think for multiplication, you check that the resulting value is bigger than each of the two multiplied values.

diamondhands0:
Also in block_view.go:_connectFailingTransaction on line 3938, I think you're comparing a *raw txn fee* to a *fee rate*:
```
	if effectiveFee < gp.MinimumNetworkFeeNanosPerKB {
		effectiveFee = gp.MinimumNetworkFeeNanosPerKB
	}
```

To fix this, I think you need to divide effectiveFee by BytesPerKB, and maybe add one in the comparison to cover truncation.

lib/blockchain.go
```diff
@@ -1144,7 +1154,7 @@ func (bc *Blockchain) isTipCurrent(tip *BlockNode) bool {
 	minChainWorkBytes, _ := hex.DecodeString(bc.params.MinChainWorkHex)
 
 	// Not current if the cumulative work is below the threshold.
-	if tip.CumWork.Cmp(BytesToBigint(minChainWorkBytes)) < 0 {
+	if bc.IsPoWBlockHeight(uint64(tip.Height)) && tip.CumWork.Cmp(BytesToBigint(minChainWorkBytes)) < 0 {
 		//glog.V(2).Infof("Blockchain.isTipCurrent: Tip not current because "+
 		//"CumWork (%v) is less than minChainWorkBytes (%v)",
 		//tip.CumWork, BytesToBigint(minChainWorkBytes))
```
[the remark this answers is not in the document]
diamondhands0: I think I mentioned this earlier. Can't remember what the decision was though.

lib/pos_mempool.go
```diff
@@ -241,11 +270,70 @@ func (mp *PosMempool) Start() error {
...
+						continue
+					}
+					// If the transaction failed to connect, we connect the transaction as a failed txn
+					// directly on newView.
+					if mp.latestBlockHeight+1 >= uint64(mp.params.ForkHeights.ProofOfStake2ConsensusCutoverBlockHeight) {
+						// Try to connect as failing txn directly to newView
+						_, _, _, err = newView._connectFailingTransaction(
+							txn.GetTxn(), uint32(mp.latestBlockHeight), false)
+						if err != nil {
+							glog.Errorf(
```
diamondhands0, line 319: I think if this fails, then newView is compromised and you have to recompute it before you go to the next iteration. Maybe what I would do to keep it light is set `copiedView, err = newView.CopyUtxoView()` and run this check on copiedView to guarantee newView is always clean. And then if it passes you do `newView = copiedView` like above.

lib/pos_mempool.go
```diff
@@ -663,7 +840,9 @@ func (mp *PosMempool) GetAugmentedUniversalView() (*UtxoView, error) {
 	if !mp.IsRunning() {
 		return nil, errors.Wrapf(MempoolErrorNotRunning, "PosMempool.GetAugmentedUniversalView: ")
 	}
-	newView, err := mp.readOnlyLatestBlockView.CopyUtxoView()
+	mp.augmentedLatestBlockViewMutex.RLock()
```
diamondhands0, line 843: You don't need to hold the lock when you're making a copy. Rewrite this as follows:
```
var readOnlyPointer *UtxoView
mp.augmentedLatestBlockViewMutex.RLock()
readOnlyPointer = mp.augmentedLatestBlockViewMutex
mp.augmentedLatestBlockViewMutex.RUnlock()

newView, err := readOnlyPointer.CopyUtxoView()
```

Note that we can do this because, unlike with the legacy mempool, we are *never modifying* augmentedLatestBlockView ever. This seems like a small change, but I think it will dramatically increase efficiency because it means we're almost never holding the lock. This change is also the reason why I recommend we decrease the wait time between recomputations of the augmented view.

lib/pos_mempool.go
```diff
@@ -672,6 +851,15 @@ func (mp *PosMempool) GetAugmentedUniversalView() (*UtxoView, error) {
 func (mp *PosMempool) GetAugmentedUtxoViewForPublicKey(pk []byte, optionalTx *MsgDeSoTxn) (*UtxoView, error) {
 	return mp.GetAugmentedUniversalView()
 }
+func (mp *PosMempool) BlockUntilReadOnlyViewRegenerated() {
+	oldSeqNum := atomic.LoadInt64(&mp.augmentedLatestBlockViewSequenceNumber)
+	newSeqNum := oldSeqNum
+	for newSeqNum == oldSeqNum {
+		// Check fairly often. Not too often.
+		time.Sleep(25 * time.Millisecond)
```
diamondhands0, line 859: For symmetry, I'd make this `augmentedBlockViewRefreshIntervalMillis / 10` so that it checks faster if you update that variable.

lib/pos_mempool.go
```diff
@@ -129,6 +140,14 @@ type PosMempool struct {
 	// feeEstimator is used to estimate the fee required for a transaction to be included in the next block
 	// based off the current state of the mempool and the most n recent blocks.
 	feeEstimator *PoSFeeEstimator
+
+	// augmentedBlockViewRefreshIntervalMillis is the frequency with which the augmentedLatestBlockView is updated.
+	augmentedBlockViewRefreshIntervalMillis uint64
```
diamondhands0, line 145: I'd like to start this value at 10ms. I don't think it will cause locking issues here the way it does with legacy mempool because here we hold the write-lock only to make a pointer-copy. In theory, that means we really could set this value to zero and probably be fine. But 10ms seems like a good middle-ground that users hopefully won't really notice.

lib/pos_mempool.go
```diff
@@ -118,6 +121,14 @@ type PosMempool struct {
 	// PosMempool only needs read-access to the block view. It isn't necessary to copy the block view before passing it
 	// to the mempool.
 	readOnlyLatestBlockView *UtxoView
+	// augmentedLatestBlockView is a copy of the latest block view with all the transactions in the mempool applied to
+	// it. This allows the backend to display the current state of the blockchain including the mempool.
+	// The augmentedLatestBlockView is updated every 100 milliseconds to reflect the latest state of the mempool.
+	augmentedLatestBlockView *UtxoView
```
diamondhands0, line 127: I'd rename this to readOnlyAugmentedLatestBlockView since we actually never modify it. Instead, we just always throw away the old one and let it get garbage-collected.

lib/pos_mempool.go
```diff
@@ -241,11 +270,70 @@ func (mp *PosMempool) Start() error {
...
+		for {
+			select {
+			case <-time.After(time.Duration(mp.augmentedBlockViewRefreshIntervalMillis) * time.Millisecond):
+				// If we're not within 10 blocks of the PoS cutover, we don't need to update the
+				// augmentedLatestBlockView.
+				if mp.latestBlockHeight+10 < uint64(mp.params.ForkHeights.ProofOfStake2ConsensusCutoverBlockHeight) {
+					continue
+				}
+				// Update the augmentedLatestBlockView with the latest block view.
+				mp.RLock()
```
diamondhands0, line 290: I don't think you need to hold the lock during the copy. I would rewrite this as:
```
mp.RLock()
readOnlyPointer = mp.readOnlyLatestBlockView
mp.RUnblock()

newView, err := readOnlyPointer.CopyUtxoView()
```

This works because we never modify the value of readOnlyLatestView. Instead, we just change the pointer. So you only need to guard the pointer access, not the copy.

I've always used locks for these cases, but a pointer-copy is so chill I thought *maybe* Go made it so that it's guaranteed to be atomic. And I think on most hardware the pointer-copy IS atomic, meaning you don't even *need* a lock around it. But it looks like it's not a guarantee so best to keep the guard you have there.

This was my research:
https://chat.openai.com/share/f8aade6f-7822-484a-9057-f656f7429669
https://www.reddit.com/r/golang/comments/yb1ijr/race_or_not_race/

lib/pos_mempool.go
```diff
@@ -118,6 +121,14 @@ type PosMempool struct {
 	// PosMempool only needs read-access to the block view. It isn't necessary to copy the block view before passing it
 	// to the mempool.
 	readOnlyLatestBlockView *UtxoView
```
diamondhands0, line 123: A minor note to help you understand my other comments: This field is *truly* readOnly now, meaning that we don't ever modify the underlying view after we set it. Instead, we just change the *pointer* to point to a new view. This means that the code for updating this value, as well as the augmentedLatestBlockView value, is *much* more efficient than it was with the legacy mempool. The reason is that the legacy mempool used to require a view-copy to be inside of a lock, whereas we only need a pointer-copy to be inside of a lock. Much better than before! Which means we can have a lower time between each update. I recommend 10ms.

lib/server.go
```diff
@@ -1738,16 +1842,36 @@ func (srv *Server) _addNewTxn(
...
+	// addition into the PoW mempool succeeded above.
+	mempoolTxn := NewMempoolTransaction(txn, uint64(time.Now().UnixMicro()))
+	if err := srv.posMempool.AddTransaction(mempoolTxn, true /*verifySignatures*/); err != nil {
+		return nil, errors.Wrapf(err, "Server._addNewTxn: problem adding txn to pos mempool")
+	}
 
-	return newlyAcceptedTxns, nil
+	// At this point, we know the transaction has been run through the mempool.
+	// Now wait for an update of the ReadOnlyUtxoView so we don't break anything.
+	srv.GetMempool().BlockUntilReadOnlyViewRegenerated()
```
diamondhands0, line 1872: I don't think you want BlockUntilReadOnlyViewRegenerated here. I think _addTxn is called when we initialize the mempool, which means you'll be blasting it with txns. Instead, you want to do this wait in BroadcastTransaction() only, which is where it was before anyway. You can just look at the diff and make sure that the position of BlockUntilReadOnlyViewRegenerated hasn't changed, and make sure it makes sense why it hasn't changed.

lib/pos_mempool.go
diamondhands0: Update: Resolved this myself.

lib/pos_mempool.go
diamondhands0: Update: Resolved this myself.

lib/pos_mempool.go
diamondhands0: Update: Resolved this myself.

lib/pos_mempool.go
diamondhands0: Update: Resolved this myself.

lib/pos_mempool.go
diamondhands0: Update: Resolved this myself.

consensus/types.go
tholonious: The event loop is referenced as the interface type. So, no there isn't a a built-in way to go to the implementation directly in VS Code. You'd have to Ctrl+Shift+F to search for it.

tholonious: The interface is useful so we can test the consensus using a mock event loop type. Marking this as resolved since it's not a code issue, and is instead an IDE limitation

lib/server.go
tholonious: Done

lib/pos_consensus.go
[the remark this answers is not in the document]
tholonious: Marking this as resolved since there's no action item for this feature PR on it.

lib/pos_block_producer.go
tholonious: We settled on having strictly increasing block timestamps with up to a 10 minute forward buffer. It should not be possible for a block proposer to roll back time with a previously proposed block. This would break the integrity of lockups, among what I'm sure will be other edge cases built on future assumptions of strictly increasing timestamps in other use cases.

I'll update this to get rid of the timeSource, and to introduce strictly increasing timestamps.

lib/pos_consensus.go
tholonious: The GetView() is a convenience when accepting incoming PoW blocks so that we can use block height as a pseudo "view". IMO, it's more appropriate to use the raw ProposedInView field here when constructing and validating blocks. This way, there's no way we can misread and conflate the block height as the view.

lib/pos_block_producer.go
tholonious: Addressed in PR https://github.com/deso-protocol/core/pull/1016

AeonSw4n (approved):

cmd/config.go
tholonious: I don't have a strong opinion on this. An optional check on the 0x prefix seems fine to add later on when needed. This uses the standard BIP39 seed phrase -> seed conversion, so it should just work out of the box to use a hex seed encoding

lib/blockchain.go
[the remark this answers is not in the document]
tholonious: Tracking this separately as a TODO for node syncing optimizations

PR #878, opened 2023-12-18
Feature/pos fee estimation v2

lazynina:


lib/blockchain.go
```diff
@@ -4899,12 +4899,22 @@ func (bc *Blockchain) CreateMaxSpend(
...
 		}
-
 		feeAmountNanos := uint64(0)
 		prevFeeAmountNanos := uint64(0)
 		for feeAmountNanos == 0 || feeAmountNanos != prevFeeAmountNanos {
 			prevFeeAmountNanos = feeAmountNanos
-			feeAmountNanos = _computeMaxTxV1Fee(txn, minFeeRateNanosPerKB)
+			if bc.BlockTip().Height >= bc.params.ForkHeights.ProofOfStake2ConsensusCutoverBlockHeight &&
+				!isInterfaceValueNil(mempool) &&
+				mempool.GetFeeEstimator() != nil {
```
tholonious, line 4908: When would the estimator be nil if the mempool isn't nil? If the mempool is properly initialized, I think it's fine to expect all of its contents to be properly initialized too

lib/blockchain.go
```diff
@@ -5028,12 +5038,24 @@ func (bc *Blockchain) AddInputsAndChangeToTransactionWithSubsidy(
...
-		if txArg.TxnMeta.GetTxnType() != TxnTypeBlockReward && minFeeRateNanosPerKB != 0 {
-			prevFeeAmountNanos := uint64(0)
-			for feeAmountNanos == 0 || feeAmountNanos != prevFeeAmountNanos {
-				prevFeeAmountNanos = feeAmountNanos
-				feeAmountNanos = _computeMaxTxV1Fee(txArg, minFeeRateNanosPerKB)
-				txArg.TxnFeeNanos = feeAmountNanos
+		if txArg.TxnMeta.GetTxnType() != TxnTypeBlockReward {
+			if blockHeight >= bc.params.ForkHeights.ProofOfStake2ConsensusCutoverBlockHeight &&
+				!isInterfaceValueNil(mempool) &&
+				mempool.GetFeeEstimator() != nil {
```
tholonious, line 5044: Ditto on when the estimator would be nil if the mempool isn't nil

lib/pos_mempool.go
```diff
@@ -169,6 +174,11 @@ func (mp *PosMempool) Start() error {
 
 	// Create the transaction register, the ledger, and the nonce tracker,
 	mp.txnRegister = NewTransactionRegister(mp.globalParams)
+	mp.feeEstimator = NewPoSFeeEstimator()
```
tholonious, line 177: You can set `mp.feeEstimator = NewPoSFeeEstimator()` in the `NewPosMempool` constructor above. This would guarantee that the estimator is always are non-nil once the mempool has been constructed

lib/pos_mempool.go
```diff
@@ -169,6 +174,11 @@ func (mp *PosMempool) Start() error {
...
+	// TODO: parameterize num blocks. Also, how to pass in blocks.
+	if err := mp.feeEstimator.Init(mp.txnRegister, []*MsgDeSoBlock{}, 1, mp.globalParams); err != nil {
```
tholonious, line 179: You can create a separate `PosMempool.Init(...)` function that can take in all of the initialization params for the mempool and the initialization params for the estimator. It can be separate from the `PosMempool.Start()` which can be scoped to just starting internal event loop for the mempool.

Suggestion:

```
func (mp *PosMempool) Init(
	params *DeSoParams, 
	globalParams *GlobalParamsEntry, 
	readOnlyLatestBlockView *UtxoView,
	latestBlockHeight uint64,
	dir string,
	inMemoryOnly bool,
	maxMempoolPosSizeBytes uint64,
	mempoolBackupIntervalMillis uint64, 
	blocks []*MsgDeSoBlock, // The last n blocks at the tip of the blockchain
)
```

tholonious (changes requested):

lib/blockchain.go
lazynina: A couple things here:
1. The mempool passed into this function is actually an interface, so it could be either a DeSoMempool or a PosMempool. PosMempool returns a fee estimator when `GetFeeEstimator` is called, DeSoMempool returns nil.
2. There are some tests that pass in a nil value for the mempool to these functions. This is annoying more than anything else.

lazynina:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/deso-protocol/core/878?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* `feature/proof-of-stake`
  * **PR #878** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/878?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
    * **PR #881** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/881?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/878?utm_source=stack-comment).

lib/pos_mempool.go
```diff
@@ -156,22 +162,35 @@ func NewPosMempool(params *DeSoParams, globalParams *GlobalParamsEntry, readOnly
...
 		mempoolBackupIntervalMillis: mempoolBackupIntervalMillis,
+		txnRegister:                 NewTransactionRegister(globalParams),
+		feeEstimator:                NewPoSFeeEstimator(),
+		ledger:                      NewBalanceLedger(),
+		nonceTracker:                NewNonceTracker(),
+	}
+}
+
+func (mp *PosMempool) Init(feeEstimatorPastBlocks []*MsgDeSoBlock, feeEstimatorNumBlocks uint64) error {
+
```
tholonious, line 173: super nit
```suggestion
```

tholonious (approved):

lib/pos_fee_estimator.go
```diff
@@ -0,0 +1,506 @@
...
+		return errors.Wrap(err, "PoSFeeEstimator.removeBlockNoLock: error computing blockHash")
+	}
+	// Remove the block from the cached blocks.
+	for ii, cachedBlock := range posFeeEstimator.cachedBlocks {
+		cachedBlockHash, err := cachedBlock.Hash()
+		if err != nil {
+			return errors.Wrap(err, "PoSFeeEstimator.removeBlockNoLock: error computing cachedBlockHash")
+		}
+		if blockHash.IsEqual(cachedBlockHash) {
+			posFeeEstimator.cachedBlocks = append(posFeeEstimator.cachedBlocks[:ii],
```
diamondhands0, line 150: A long time ago I hit a bug related to reusing a slice in an append like this. Something about rewriting the slice in-place. I'm not sure if you have a bug here because of it, but the safer thing would be to make a copy of cachedBlocks like this:

```
newCachedBlocks = []
for ii, cachedBlock := range posFeeEstimator.cachedBlocks {
    if blockHash.IsEqual(cachedBlockHash) {
        continue
    }
    newCachedBlocks = append(newCachedBlocks, cachedBlock)
}
posFeeEstimator.cachedBlocks = newCachedBlocks
```

lib/pos_fee_estimator.go
```diff
@@ -0,0 +1,506 @@
...
+	}
+	return nil
+}
+
+// sortCachedBlocks sorts the cached blocks by height & tstamp just to be safe.
+func (posFeeEstimator *PoSFeeEstimator) sortCachedBlocks() {
+	posFeeEstimator.cachedBlocks = collections.SortStable(posFeeEstimator.cachedBlocks,
+		func(ii, jj *MsgDeSoBlock) bool {
+			if ii.Header.Height == jj.Header.Height {
+				return ii.Header.TstampNanoSecs < jj.Header.TstampNanoSecs
```
diamondhands0, line 178: Maybe add one more tie-breaker on block hash just so that the sort is 100% stable.

lib/pos_fee_estimator.go
```diff
@@ -0,0 +1,506 @@
...
+// the percentile position of the priorityPercentileBasisPoints param and then compute the fee bucket
+// range based on the fee rate per KB of the transaction at that position.
+func getPriorityFeeBucketFromTxns(
+	feeTimeOrderedTxns []*MempoolTx,
+	priorityPercentileBasisPoints uint64,
+	minimumNetworkFeeNanosPerKB *big.Float,
+	feeBucketGrowthRateBasisPoints *big.Float,
+) (uint64, uint64, error) {
+	percentilePosition := uint64(
+		len(feeTimeOrderedTxns)) - ((priorityPercentileBasisPoints * uint64(len(feeTimeOrderedTxns))) / 10000)
```
diamondhands0, line 496: So this percentile is based on number of txns, which means that a very large txn would probably be under-weighted in terms of its contribution. For example, if a block contains a single very large txn with very high fee, probably we should count that as more than one txn when computing percentile. This is OK for v0, but add a comment to mention this.

lib/pos_fee_estimator.go
```diff
@@ -0,0 +1,506 @@
...
+	if percentilePosition >= uint64(len(feeTimeOrderedTxns)) {
```
diamondhands0, line 497: This will happen if the priority is zero. In that case... probably what they want is `len(feeTimeOrderedTxns)-1`? In which case consider just returning that instead of erroring here.

lib/pos_fee_estimator.go
```diff
@@ -0,0 +1,506 @@
...
+
+	// Compute the congestion threshold. If our congestion factor is 100% (or 10,000 bps),
+	// then congestion threshold is simply max block size * numBlocks
+	// TODO: I don't know if I like this name really.
+	congestionThreshold := (congestionFactorBasisPoints * maxSizeOfNumBlocks) / MaxBasisPoints
+	// If the total size of the txns in the transaction register is less than the computed congestion threshold,
+	// we return one bucket lower than the Priority fee.
+	if totalTxnsSize <= congestionThreshold {
+		// Return one bucket lower than Priority fee
+		oneBucketLowerMinFee, _ := computeFeeTimeBucketRangeFromFeeNanosPerKB(
```
diamondhands0, line 467: I am a little bit worried that you'll run into a rounding error here because you're going: fee -> bucket exponent -> minFee -> subtract one -> back to bucket exponent. The reason is there could be an off by one in the first "bucket exponent -> minFee" step. Better would be to go: fee -> bucket exponent -> bucket exponent - 1 -> minFee.

lib/pos_fee_estimator.go
```diff
@@ -0,0 +1,506 @@
...
+			bucketMinFee-1,
+			txnRegister.minimumNetworkFeeNanosPerKB,
+			txnRegister.feeBucketGrowthRateBasisPoints,
+		)
+		return oneBucketLowerMinFee, nil
+	}
+	// If the total size of the txns in the mempool is greater than the computed congestion threshold
+	// but less than the max size of num blocks, we return the Priority fee.
+	if totalTxnsSize > congestionThreshold && totalTxnsSize <= maxSizeOfNumBlocks {
```
diamondhands0, line 476: Do we need this in-between case? I feel like we can just lower the fee if we're below the congestion threshold and raise it if we're above.

diamondhands0: Also you use this same code for the block fee estimator, which I think introduces a problem. In particular, the total size of the txns in the last N blocks will always be *slightly less* than maxSizeOfNumBlocks. So you'll never increase the fee according to the block fee estimator with the code as it is currently.

I think safer from a fee adjustment standpoint is just to get rid of the "static fee" case. It goes up if we're above the threshold and decreases if we're below it.

lib/pos_fee_estimator.go
```diff
@@ -0,0 +1,506 @@
...
+//  4. Compute the priority fee bucket for the transactions in the slice using the priorityPercentileBasisPoints param
+//     by calling getPriorityFeeBucketFromTxns on the slice.
+//  5. If the resulting priority fee bucket from step 4 is less than the global minimum network fee, return the global
+//     minimum network fee.
+//  6. Compute the congestion threshold as congestionFactorBasisPoints * maxSizeOfNumBlocks / 10000.
+//  7. If the total size of the transactions in the slice is less than the congestion threshold, return one bucket lower
+//     than the priority fee bucket from step 4.
+//  8. If the total size of the transactions in the size is greater than the congestion threshold and less
+//     than or equal to maxSizeOfNumBlocks, return the priority fee bucket from step 4.
+//  9. Otherwise, return one bucket higher than the priority fee bucket from step 4.
```
diamondhands0, line 409: I'd link to the doc we have analyzing this stuff here.

diamondhands0: I would give some examples of these parameters here. I had to noodle pretty hard to work out examples and I'm the guy who wrote it! I think the examples below helped me:
```
When run for mempool:
numBlocks = 1 (this is set at construction time)
congestionFactorBasisPoints = 900 (90%)
priorityPercentileBasisPoints = 10000 (100%)

With these numbers, I think we'd have the following behavior (and notice I'm getting rid of the case where we keep the fee the same, see comment below):
- We will find the lowest fee paid by anyone within the first block's worth ot txns because priorityPercentileBasisPoints=100% will return the lowest-fee txn.
- If the mempool has >90% of 1 block's worth of txns, we will pay one fee bucket HIGHER than this lowest-fee txn.
- If the mempool has <90% of 1 block's worth of txns, we will pay one fee bucket LOWER than this lowest-fee txn.
```

More examples:
```
When run for mempool:
numBlocks = 5 (this is set at construction time)
congestionFactorBasisPoints = 10000 (100%)
priorityPercentileBasisPoints = 2000 (20%)

With these numbers, I think we'd have the following behavior (and notice I'm getting rid of the case where we keep the fee the same, see comment below):
- We find the fee of the 20th percentile txn, which is enough to be included in the next block
- If the mempool has more than 5 blocks worth of txns, we pay one fee bucket higher than this guy, otherwise we pay one fee bucket lower than him.

This one is a bit weird, and I think the simple single-block thing previously would be better but lmk what you think of it.
```

Now to play with some block fee estimation:
```
numBlocks = 5 (this is set at construction time)
congestionFactorBasisPoints = 9000 (90%)
priorityPercentileBasisPoints = 2000 (20%)

Here, we would find the 20th percentile txn, enough to include in the next block. And if the last 5 blocks are more than 90% full we pay 1 bucket higher than this guy. Otherwise, we pay one fee bucket lower.

This I think makes senes for the block fee estimator. Lmk if you agree.
```

diamondhands0: To summarize, I think the params we'd probably want for mempool and block estimator respectively are:

Mempool:
```
numBlocks = 1
congestionFactorBasisPoints = 9000 (90%)
priorityPercentileBasisPoints = 9000 (90%)

This means we will take one block's worth of txns from the mempool, and we'll pay 1 fee bucket higher than the 90th percentile guy if there's more than 90% of a block's worth of txns in it.
```

For the block estimator I think our params would be something like tihs:
```
numBlocks = 60
congestionFactorBasisPoints = 9000 (90%)
priorityPercentileBasisPoints = 1/60 (since we want to always get included in the next block)

Here, we would take the last 60 blocks worth of txns. If we're more than 90% full, we pay one fee bucket higher than the "top block"'s txn. Otherwise, we pay one fee bucket lower than that txn. I think that makes sense.
```

lib/pos_fee_estimator.go
```diff
@@ -0,0 +1,506 @@
...
+// and past blocks using the congestionFactorBasisPoints, priorityPercentileBasisPoints, and
+// maxBlockSize params.
+func (posFeeEstimator *PoSFeeEstimator) EstimateFeeRateNanosPerKB(
+	congestionFactorBasisPoints uint64,
+	priorityPercentileBasisPoints uint64,
+	maxBlockSize uint64,
+) (uint64, error) {
+	posFeeEstimator.rwLock.RLock()
+	defer posFeeEstimator.rwLock.RUnlock()
+	pastBlockFeeRate, err := posFeeEstimator.estimateFeeRateNanosPerKBGivenTransactionRegister(
```
diamondhands0, line 194: I think we'll probably want to use a different congestion factor and percentile for block vs mempool. See my comment talking about the different scenarios.

PR #888, opened 2023-12-26
Create Entry Points For PoS Consensus In Server

tholonious:

PR #889, opened 2023-12-27
Wire up PoS Mempool Into Server

tholonious:


tholonious:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/deso-protocol/core/889?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* `feature/proof-of-stake`
  * **PR #875** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/875?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #888** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/888?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #889** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/889?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/889?utm_source=stack-comment).

lib/peer.go
```diff
@@ -183,14 +184,16 @@ func (pp *Peer) HandleGetTransactionsMsg(getTxnMsg *MsgDeSoGetTransactions) {
...
 
-	mempoolTxs := []*MempoolTx{}
-	txnMap := pp.srv.mempool.readOnlyUniversalTransactionMap
+	mempoolTxs := []*MempoolTransaction{}
+
+	// We fetch the request txns from the PoS mempool because the PoS mempool contains
+	// all txns that have been added to the PoW and PoS mempools.
 	for _, txHash := range getTxnMsg.HashList {
-		mempoolTx, exists := txnMap[*txHash]
+		mempoolTx := pp.srv.posMempool.GetTransaction(txHash)
```
lazynina, line 192: We could use `pp.srv.GetMempool().GetTransaction(txHash)` and finish the `GetTransaction` implementation in the legacy mempool since both the legacy mempool and pos mempool implement the interface. My only concern about using the PoSMempool is if we start ejecting transactions from it when it fills up. `GetMempool` returns the legacy mempool if we're before the ProofOfStake2ConsensusCutoverBlockHeight block height.

Here's what the `GetTransaction` implementation could look like for the legacy mempool:
```
func (mp *DeSoMempool) GetTransaction(txnHash *BlockHash) *MempoolTransaction {
	mempoolTx, exists := mp.readOnlyUniversalTransactionMap[*txnHash]
	if !exists {
		return nil
	}
	return NewMempoolTransaction(mempoolTx.Tx, uint64(mempoolTx.Added.UnixMicro()))
}
```

lib/pos_mempool.go
```diff
@@ -277,6 +277,88 @@ func (mp *PosMempool) IsRunning() bool {
...
+
+		// Get the transaction from the register. If the txn doesn't exist in the register,
+		// then there's nothing left to do.
+		txn := mp.txnRegister.GetTransaction(txnHash)
+		if txn == nil {
+			continue
+		}
+
+		mp.removeTransactionNoLock(txn, true)
+	}
```
lazynina, line 315: I think we also want to call `mp.feeEstimator.AddBlock` with the new block somewhere in here? Or are we managing the past blocks of the fee estimator elsewhere? Also, I need to trace how block connected events are fired in re-orgs. For example, let's say we currently have a chain such as:

A -> B -> C* -> D*, where * denotes uncommitted
and now we reorg to 
A -> B -> E* -> F*

Do we have an OnBlockDisconnected event for C* and D* and an OnBlockConnected event for E* and F*? If so, everything should work as expected. But if that's not the case, then we need to figure out how to add E* to the fee estimator's past blocks cache and remove C* and D* from the past blocks cache.

lib/pos_mempool.go
```diff
@@ -277,6 +277,88 @@ func (mp *PosMempool) IsRunning() bool {
...
+
+	if blockEvent == nil || blockEvent.Block == nil || len(blockEvent.Block.Txns) == 0 {
+		return
+	}
+
+	if !mp.IsRunning() {
+		return
+	}
+
+	// Remove all transactions in the block from the mempool.
```
lazynina, line 336:
```suggestion
	// Add all transactions in the block to the mempool.
```

lib/pos_mempool.go
```diff
@@ -277,6 +277,88 @@ func (mp *PosMempool) IsRunning() bool {
...
+
+		// Add the transaction to the mempool and then prune if needed.
+		if err := mp.addTransactionNoLock(mempoolTx, true); err != nil {
+			glog.Errorf("PosMempool.AddTransaction: Problem adding transaction to mempool: %v", err)
+		}
+	}
+
+	if err := mp.pruneNoLock(); err != nil {
+		glog.Errorf("PosMempool.AddTransaction: Problem pruning mempool: %v", err)
+	}
```
lazynina, line 359: We'll want to call `RemoveBlock` on `mp.feeEstimator` somewhere in here as well.

lib/peer.go
lazynina: Given that it's okay if we only respond with a subset, my recommendation above is not necessary. but wanted you to think about it. I generally like how backward compatible the PoSMempool is so it's kinda nice to use it everywhere once a node has updated their code.

lib/pos_mempool.go
tholonious: Great call-out. We do need to call `mp.feeEstimator.AddBlock(...)` here, which sorts blocks by block height underneath to internally maintain its sorted list of the last n cached blocks for fee estimation. It's agnostic to the order in which blocks are dinnected, so it's safe as-is.

I have it as a separate task to make sure that the OnBlockConnected and OnBlockDisconnected events will be called from ProcessBlockPoS. It can be a separate PR.

lib/pos_mempool.go
```diff
@@ -277,6 +277,94 @@ func (mp *PosMempool) IsRunning() bool {
...
+		txnHash := txn.Hash()
+
+		// This should never happen. We perform a nil check on the txn hash to avoid a panic.
+		if txnHash == nil {
+			continue
+		}
+
+		// Get the transaction from the register. If the txn doesn't exist in the register,
+		// then there's nothing left to do.
+		txn := mp.txnRegister.GetTransaction(txnHash)
```
lazynina, line 305: can we use a different variable name so it doesn't shadow the `txn` defined on line 295 when we start the for-loop?

lib/pos_mempool.go
```diff
@@ -277,6 +277,94 @@ func (mp *PosMempool) IsRunning() bool {
...
+	// Add the block to the fee estimator. This is a best effort operation. If we fail to add the block
+	// to the fee estimator, we log an error and continue.
+	if err := mp.feeEstimator.AddBlock(block); err != nil {
+		glog.Errorf("PosMempool.OnBlockConnected: Problem adding block to fee estimator: %v", err)
+	}
+}
+
+// OnBlockDisconnected is an event handler provided by the PoS mempool to handle the blockchain
+// event where a block is disconnected from the tip of the blockchain. The mempool updates its
+// internal state based on the new block that has been connected.
```
lazynina, line 322:
```suggestion
// internal state based on the block that has been disconnected.
```

lib/pos_mempool.go
lazynina, line 305: Ooooh - we use `mp.txnRegister.GetTransaction` instead of `mp.GetTransaction` here since `mp.GetTransaction` acquires the lock on the mempool (which would lead to a deadlock) right?

lib/pos_mempool.go
```diff
@@ -277,6 +277,94 @@ func (mp *PosMempool) IsRunning() bool {
...
+
+		// This should never happen. We perform a nil check on the txn hash to avoid a panic.
+		if txnHash == nil {
+			continue
+		}
+
+		// Add all transactions in the block to the mempool.
+
+		// Construct the MempoolTx from the MsgDeSoTxn.
+		mempoolTx, err := NewMempoolTx(txn, block.Header.TstampNanoSecs/1000, mp.latestBlockHeight)
```
lazynina, line 346: Interesting, there's no way for us to know exactly when this txn was actually added. This seems like the best approach since we know it was added to some mempool in the network before this timestamp.

lib/server.go
```diff
@@ -220,6 +219,9 @@ func (srv *Server) GetBlockchain() *Blockchain {
 
 // TODO: The hallmark of a messy non-law-of-demeter-following interface...
 func (srv *Server) GetMempool() Mempool {
+	srv.blockchain.ChainLock.RLock()
+	defer srv.blockchain.ChainLock.RUnlock()
```
lazynina, line 223: Out of curiosity, why do we need to acquire this lock here? Is it just to make sure the tip height doesn't shift underneath us?

lib/server.go
```diff
@@ -242,7 +244,7 @@ func (srv *Server) GetMiner() *DeSoMiner {
 	return srv.miner
 }
 
-func (srv *Server) BroadcastTransaction(txn *MsgDeSoTxn) ([]*MempoolTx, error) {
+func (srv *Server) BroadcastTransaction(txn *MsgDeSoTxn) ([]*MsgDeSoTxn, error) {
```
lazynina, line 247: Just a note for myself - changing the return value here is perfectly fine since we don't actually use the first return param in backend or core right now.

lib/server.go
```diff
@@ -252,33 +254,24 @@ func (srv *Server) BroadcastTransaction(txn *MsgDeSoTxn) ([]*MempoolTx, error) {
 		return nil, errors.Wrapf(err, "BroadcastTransaction: ")
 	}
 
-	// At this point, we know the transaction has been run through the mempool.
-	// Now wait for an update of the ReadOnlyUtxoView so we don't break anything.
-	srv.mempool.BlockUntilReadOnlyViewRegenerated()
-
```
lazynina, line 258: I think we want to keep this for PoW for now.

lib/server.go
```diff
@@ -490,6 +483,36 @@ func NewServer(
...
+	}
+	err = _posMempool.Init(
+		_params,
+		currentGlobalParamsEntry,
+		currentUtxoView,
+		uint64(_chain.blockTip().Height),
+		_mempoolDumpDir,
+		false,
+		1024*1024*1024*3, // Max mempool Size = 3GB; TODO make this a param
+		60*1000,          // Mempool dumper frequency = 60 seconds; TODO make this a param
```
lazynina, line 508: Maybe move these to `constants.go` for now?

lib/server.go
```diff
@@ -490,6 +483,36 @@ func NewServer(
...
+
+	// Initialize the PoS mempool. We need to initialize a best-effort UtxoView based on the current
+	// known state of the chain. This will all be overwritten as we process blocks later on.
+	currentUtxoView, err := _chain.GetUncommittedTipView()
+	if err != nil {
+		return nil, errors.Wrapf(err, "NewServer: Problem initializing latest UtxoView"), true
+	}
+	currentGlobalParamsEntry := currentUtxoView.GetCurrentGlobalParamsEntry()
+	latestBlockHash := _chain.blockTip().Hash
+	latestBlock := _chain.GetBlock(latestBlockHash)
```
lazynina, line 496: I think @diamondhands0 actually wants us to load up the last 60 blocks. We can configure the # of blocks via some param, but we're going to need to make it resilient to hypersync not syncing full blocks. So if we can only get the last 10 blocks instead of 60, that's okay too. I'm okay punting these changes to another PR.

lib/server.go
```diff
@@ -1684,14 +1708,14 @@ func (srv *Server) _handleDonePeer(pp *Peer) {
 }
 
 func (srv *Server) _relayTransactions() {
-	glog.V(1).Infof("Server._relayTransactions: Waiting for mempool readOnlyView to regenerate")
-	srv.mempool.BlockUntilReadOnlyViewRegenerated()
```
lazynina, line 1688: Maybe we want this for PoW still for now?

lib/server.go
```diff
@@ -2058,25 +2127,52 @@ func (srv *Server) _handleGetTransactions(pp *Peer, msg *MsgDeSoGetTransactions)
...
+	defer srv.blockchain.ChainLock.RUnlock()
+
 	// Note we set rateLimit=false because we have a global minimum txn fee that should
 	// prevent spam on its own.
-	return srv.mempool.ProcessTransaction(
-		txn, true /*allowUnconnectedTxn*/, false, /*rateLimit*/
-		pp.ID, true /*verifySignatures*/)
+
+	// Only attempt to add the transaction to the PoW mempool if we're on the
+	// PoW protocol. If we're on the PoW protocol, then we use the PoW mempool's,
```
lazynina, line 2143:
```suggestion
	// PoW protocol. If we're on the PoW protocol, then we use the PoW mempool's
```

lib/server.go
```diff
@@ -2058,25 +2127,52 @@ func (srv *Server) _handleGetTransactions(pp *Peer, msg *MsgDeSoGetTransactions)
...
+	// txn validity checks to signal whether the txn has been added or not. The PoW
+	// mempool has stricter txn validity checks than the PoW mempool, so this works
```
lazynina, line 2145:
```suggestion
	// mempool has stricter txn validity checks than the PoS mempool, so this works
```

lib/server.go
```diff
@@ -2058,25 +2127,52 @@ func (srv *Server) _handleGetTransactions(pp *Peer, msg *MsgDeSoGetTransactions)
...
+	// out conveniently, as it allows us to always add a txn to the PoS mempool.
+	blockHeight := srv.blockchain.blockTip().Height
+	if blockHeight < srv.blockchain.params.ForkHeights.BalanceModelBlockHeight {
```
lazynina, line 2148: Wrong fork height.

lib/server.go
```diff
@@ -2058,25 +2127,52 @@ func (srv *Server) _handleGetTransactions(pp *Peer, msg *MsgDeSoGetTransactions)
...
+		// it's an invalid txn.
+		if err != nil {
+			return nil, errors.Wrapf(err, "Server.ProcessSingleTxnWithChainLock: Problem adding transaction to PoW mempool: ")
+		}
+	}
+
+	// Regardless of the consensus protocol we're running (PoW or PoS), we use the PoS mempool's to house all
+	// mempool txns. If a txn can't make it into the PoS mempool, which uses a looser unspent balance check for
+	// the the transactor, then it must be invalid.
+	if err := srv.posMempool.AddTransaction(NewMempoolTransaction(txn, uint64(time.Now().UnixMicro())), true); err != nil {
```
lazynina, line 2167: Okay this isn't the right place for this comment, but wanted to get it out while it's in my head. Currently, we call `LoadTxnsFromDB` on the legacy mempool in `NewDeSoMempool`. When we start up a node for the first time with the PoSMempool, do we want to load these transation into the PoSMempool as well? I guess it's not a huge deal since those transactions should make it into a block relatively quickly, but we may want the two to be in lockstep.

lib/server.go
```diff
@@ -2153,6 +2249,10 @@ func (srv *Server) StartStatsdReporter() {
 				mempoolTotal := len(srv.mempool.readOnlyUniversalTransactionList)
 				srv.statsdClient.Gauge("MEMPOOL.COUNT", float64(mempoolTotal), tags, 1)
 
+				// Report PoS Mempool size
+				posMempoolTotal := srv.posMempool.txnRegister.Count()
```
lazynina, line 2253: You're my hero.

lib/server.go
```diff
@@ -2549,9 +2649,12 @@ func (srv *Server) Stop() {
 		if !srv.mempool.stopped {
 			srv.mempool.Stop()
 		}
+		srv.posMempool.Stop()
 		glog.Infof(CLog(Yellow, "Server.Stop: Closed Mempool"))
 	}
 
+	srv.posMempool.Stop()
```
lazynina, line 2656: Do we need both these stop calls?

lazynina (changes requested):

lib/pos_mempool.go
tholonious: Correct. The usage of `mp.txnRegister.GetTransaction()` here matches how we do txn removals elsewhere in this file

lib/pos_mempool.go
tholonious: No need. This solution ends up working nicely. Assume we have txn A from block 1 and txn B from block 2, and block 2 has a higher timestamp than block 1. If we disconnect both blocks 1 and 2, then txn A will have precedence over txn B. And all txns in the mempool that didn't end up in blocks 1 or 2 must have not been proposed in the blocks because they either have lower fee buckets or higher timestamps.

lib/server.go
tholonious: Yikes, good catch

lib/server.go
tholonious: Good question. I didn't have a great solution for it. My thought is that it becomes a self-balancing thing once 2-3 blocks have mined. When a node first upgrades its code to the latest code and initializes the two mempools, the PoW mempool will hold all of the persisted txns while the PoS mempool will be empty. Then as txns are received organically, and as txns are mined into blocks, the two mempools will end up with the same contents. Both mempools will also dump their contents to disk so that they will maintain parity on future node restarts.

lib/server.go
tholonious: Sounds good. Good callout. I'll add a todo for myself for this. For now, I just want this wired up with the simplest params so we can regtest. We can parametrize in another PR>

lib/server.go
tholonious: Eh, this will all be dead code in a few PRs. You think it's worth creating consts elsewhere in the code for them?

lib/server.go
tholonious: Nope. Good catch

lib/server.go
tholonious: Correct. I had to move this call into the `srv._addNewTxn(...)` function where it's only called if we're on PoW

lib/pos_mempool.go
lazynina: I think this is fine. There may be an extreme edge case where transaction reach certain nodes before a block is produced by another node, but I think it all washes out in the end. I tried to type it out multiple times and haven't been able to concretely define a case where this approach would break, so I think we're okay here.

lib/server.go
lazynina: nah screw it.

lib/server.go
lazynina: yeah I like this better than trying to have some condition for the situation like:
if the PosMempool hasn't been dumped before, then load it with the transactions from the PowMempool. So I think we're okay to ignore this comment.

lazynina (approved):

PR #892, opened 2023-12-29
Implement ProcessHeaderPoS

tholonious:


tholonious:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/deso-protocol/core/892?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* `feature/proof-of-stake`
  * **PR #875** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/875?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #892** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/892?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/892?utm_source=stack-comment).

lib/pos_blockchain.go
```diff
@@ -34,12 +47,91 @@ import (
...
+		if _, innerErr := bc.storeValidateFailedHeaderInBlockIndex(header); innerErr != nil {
+			return false, false, errors.Wrapf(innerErr, "processHeaderPoS: Problem adding validate failed header to block index")
+		}
+		return false, false, nil
+	}
+
+	// Verify that the block header is properly formed.
+	if err := bc.isProperlyFormedBlockHeaderPoS(header); err != nil {
+		// If the block header is not properly formed, we store it as HeaderValidateFailed and exit early.
+		if _, innerErr := bc.storeValidateFailedHeaderInBlockIndex(header); innerErr != nil {
```
lazynina, line 91: If we continue hitting these nested if _, immersed := bc.storeValidateFailedHeaderInBlockIndex, we should write a wrappedStoreValidateFailedHeaderInBlockIndex

lib/pos_blockchain.go
```diff
@@ -34,12 +47,91 @@ import (
...
+		return false, false, nil
+	}
+
+	// If the block header is properly formed, we store it as HeaderValidated.
+	headerBlockNode, err := bc.storeValidatedHeaderInBlockIndex(header)
+	if err != nil {
+		return false, false, errors.Wrapf(err, "processHeaderPoS: Problem adding header to block index: ")
+	}
+
+	// If the block header is an orphan, or its view is less than the current header chain's tip,
```
lazynina, line 103: What does orphan status have to do with the comment here? Looks like we basically just throw away orphan headers for now based on the logic above

lazynina (changes requested):

lib/pos_blockchain.go
tholonious: Typo left over from a previous iteration. The comment is removed now.

lib/pos_blockchain.go
tholonious: It's just these two uses. Since the two usages are side by side like this, I'm partial between leaving it as-is vs adding the helper function. The code is most concise as-is.

lib/pos_blockchain.go
```diff
@@ -26,20 +47,129 @@ import (
...
+	// ------------------------------------ Base Cases ----------------------------------- //
+
+	// The header is already validated. Exit early.
+	if exists && blockNode.IsHeaderValidated() {
+		return blockNode, false, nil
+	}
+
+	// The header has already failed validations. Exit early.
+	if exists && blockNode.IsHeaderValidateFailed() {
+		return nil, false, errors.New("validateAndIndexHeaderPoS: Header already failed validation")
```
lazynina, line 107: Should we return the block node here? It doesn't really matter based on how it's currently being used, but this is how `validateAndIndexBlockPoS` behaves.

lib/pos_blockchain.go
```diff
@@ -12,8 +12,29 @@ import (
...
+func (bc *Blockchain) ProcessHeaderPoS(header *MsgDeSoHeader) (_isMainChain bool, _isOrphan bool, _err error) {
+	// Grab the chain lock
+	bc.ChainLock.Lock()
+	defer bc.ChainLock.Unlock()
+
+	if header == nil {
+		return false, false, fmt.Errorf("ProcessHeaderPoS: Header is nil")
+	}
+
+	headerHash, err := header.Hash()
```
lazynina, line 25: Why not push the header hashing into `processHeaderPoS`? This way you don't have to worry about the caller passing through any funny business.

lazynina (approved):

lib/pos_blockchain.go
tholonious: There's no need, IMO. Currently, the block node is returned only if the header is validated and can be applied as the tip. It's a simpler spec than if we returned it in some cases where it's also invalid (but not all cases)

PR #879, opened 2023-12-19
PoS Remote Node Indexer and Manager

AeonSw4n:

PR #861, opened 2023-12-07
PoS ConnectionController

AeonSw4n:

PR #860, opened 2023-12-07
PoS HandshakeController

AeonSw4n:

PR #859, opened 2023-12-07
PoS RemoteNodeIndexerManager

AeonSw4n:

PR #899, opened 2023-12-29
feature/pos-networking-and-syncing

tholonious:


AeonSw4n:
* `feature/proof-of-stake`
  * **PR #899** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/899?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
    * **PR #857** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/857?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #879** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/879?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
        * **PR #860** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/860?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
          * **PR #861** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/861?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #858** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/858?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
        * **PR #859** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/859?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/899?utm_source=stack-comment).

lib/server.go
```diff
@@ -457,10 +467,9 @@ func NewServer(
 	timesource := chainlib.NewMedianTime()
 
 	// Create a new connection manager but note that it won't be initialized until Start().
-	_incomingMessages := make(chan *ServerMessage, (_targetOutboundPeers+_maxInboundPeers)*3)
+	_incomingMessages := make(chan *ServerMessage, 100+(_targetOutboundPeers+_maxInboundPeers)*3)
```
diamondhands0, line 470: Why is this a hardcoded value of 100? I assume you want to use the number of validators from GlobalParams?

lib/server.go
```diff
@@ -235,7 +245,6 @@ func (srv *Server) GetBlockProducer() *DeSoBlockProducer {
 	return srv.blockProducer
 }
```
diamondhands0, line 247: Why delete this comment? It's still correct.

lib/server.go
```diff
@@ -770,11 +793,13 @@ func (srv *Server) GetSnapshot(pp *Peer) {
 	}
 	// If operationQueueSemaphore is full, we are already storing too many chunks in memory. Block the thread while
```
diamondhands0, line 794: Update comment, as this is not blocking the thread anymore.

lib/server.go
```diff
@@ -770,11 +793,13 @@ func (srv *Server) GetSnapshot(pp *Peer) {
...
-	}, false)
+	go func() {
+		srv.snapshot.operationQueueSemaphore <- struct{}{}
+		// Now send a message to the peer to fetch the snapshot chunk.
+		pp.AddDeSoMessage(&MsgDeSoGetSnapshot{
+			SnapshotStartKey: lastReceivedKey,
+		}, false)
+	}()
 
 	glog.V(2).Infof("Server.GetSnapshot: Sending a GetSnapshot message to peer (%v) "+
```
diamondhands0, line 804: Move this inside the func or else our logs will be confusing.

lib/remote_node.go
```diff
@@ -0,0 +1,744 @@
...
+	rn.setStatusAttempted()
+	return nil
+}
+
+// AttachInboundConnection creates an inbound peer once a successful inbound connection has been established.
+func (rn *RemoteNode) AttachInboundConnection(conn net.Conn, na *wire.NetAddress) error {
+	rn.mtx.Lock()
+	defer rn.mtx.Unlock()
+
+	if !rn.IsNotConnected() {
```
diamondhands0, line 305: This is a confusing if statement. Add a comment explaining it.

lib/constants.go
```diff
@@ -492,6 +492,30 @@ func GetEncoderMigrationHeightsList(forkHeights *ForkHeights) (
 	return migrationHeightsList
 }
 
+type ProtocolVersionType uint64
+
+const (
+	ProtocolVersion0 ProtocolVersionType = 0
+	ProtocolVersion1 ProtocolVersionType = 1
+	ProtocolVersion2 ProtocolVersionType = 2
```
diamondhands0, line 500: Comment these

lib/remote_node.go
```diff
@@ -0,0 +1,744 @@
...
+	vMeta.advertisedProtocolVersion = NewProtocolVersionType(verMsg.Version)
+	// Make sure the latest supported protocol version is ProtocolVersion2.
+	if vMeta.advertisedProtocolVersion.After(ProtocolVersion2) {
+		return fmt.Errorf("RemoteNode.HandleVersionMessage: Requesting disconnect for id: (%v) "+
+			"protocol version too high. Peer version: %v, max version: %v", rn.id, verMsg.Version, ProtocolVersion2)
+	}
+
+	// Decide on the protocol version to use for this connection.
+	negotiatedVersion := rn.params.ProtocolVersion
+	if verMsg.Version < rn.params.ProtocolVersion.ToUint64() {
```
diamondhands0, line 483: I think this check is problematic. If you include it, then you will be partitioning the network between upgraded nodes and non-upgraded nodes. This is likely not what we want. Instead, we want PoS nodes to be able to talk to un-upgraded nodes until at least the first PoS cutoff, and likely the second PoS cutoff as well. After the second cutoff, then it makes more sense to refuse connections to/from un-upgraded nodes.

lib/remote_node.go
```diff
@@ -0,0 +1,744 @@
...
+
+	// Set the latest time by which we should receive a verack message from the peer.
+	verackTimeExpected := time.Now().Add(rn.params.VerackNegotiationTimeout)
+	rn.verackTimeExpected = &verackTimeExpected
+	if err := rn.sendVerack(); err != nil {
+		return errors.Wrapf(err, "RemoteNode.HandleVersionMessage: Problem sending verack message to peer (id= %d)", rn.id)
+	}
+
+	// Update the timeSource now that we've gotten a version message from the peer.
+	rn.cmgr.AddTimeSample(rn.peer.Address(), timeConnected)
```
diamondhands0, line 547: It's fine to leave this here for legacy reasons but we don't want to rely on this timeSource anymore. Leave a comment explaining this and make sure we're not using the timeSource anywhere.

lib/remote_node.go
```diff
@@ -0,0 +1,744 @@
...
+		return fmt.Errorf("RemoteNode.HandleVerackMessage: Requesting disconnect for id: (%v) "+
+			"verack timeout. Time expected: %v, now: %v", rn.id, rn.verackTimeExpected.UnixMicro(), time.Now().UnixMicro())
+	}
+
+	var err error
+	vMeta := rn.handshakeMetadata
+	switch vMeta.negotiatedProtocolVersion {
+	case ProtocolVersion0, ProtocolVersion1:
+		err = rn.validateVerackPoW(vrkMsg)
+	case ProtocolVersion2:
```
diamondhands0, line 619: I think this will still work, you just need to be more lenient with regard to version negotiation. In particular, you need to allow negotiated version to be ProtocolVersion1 even if you're on ProtocolVersion2.

lib/remote_node.go
```diff
@@ -0,0 +1,744 @@
...
+	if err != nil {
+		return errors.Wrapf(err, "RemoteNode.validateVerackPoS: Requesting disconnect for id: (%v) "+
+			"verack signature verification failed with error", rn.id)
+	}
+	if !ok {
+		return fmt.Errorf("RemoteNode.validateVerackPoS: Requesting disconnect for id: (%v) "+
+			"verack signature verification failed", rn.id)
+	}
+
+	if rn.validatorPublicKey != nil && rn.validatorPublicKey.Serialize() != vrkMsg.PublicKey.Serialize() {
```
diamondhands0, line 704: You can't use != on two byte arrays. Must use bytes.Equal. Surprised this is working.

lib/remote_node.go
```diff
@@ -0,0 +1,744 @@
...
+		return fmt.Errorf("RemoteNode.validateVerackPoS: Requesting disconnect for id: (%v) "+
+			"verack public key mismatch; message: %v; expected: %v", rn.id, vrkMsg.PublicKey, rn.validatorPublicKey)
+	}
+
+	// If we get here then the verack message is valid. Set the validator public key on the peer.
+	vMeta.validatorPublicKey = vrkMsg.PublicKey
```
diamondhands0, line 710: This is a weird thing to do. Didn't you already check that rn.validatorPublicKey == vrkMsg.PublicKey explicitly above?

lib/remote_node_manager.go
```diff
@@ -0,0 +1,349 @@
...
+
+	// keystore is a reference to the node's BLS private key storage.
+	keystore *BLSKeystore
+
+	// configs
+	minTxFeeRateNanosPerKB uint64
+	nodeServices           ServiceFlag
+
+	// Used to set remote node ids. Must be incremented atomically.
+	remoteNodeIndex uint64
```
diamondhands0, line 37: Call this something without "index" in it like remoteNodeNextId or something.

lib/remote_node_manager.go
```diff
@@ -0,0 +1,349 @@
...
+	// Only remove from the validator index if the fetched remote node is the same as the one we are trying to remove.
+	// Otherwise, we could have a fun edge-case where a duplicated validator connection ends up removing an
+	// existing validator connection from the index.
+	fetchedRn, ok := indexer.GetValidatorIndex().Get(rn.GetValidatorPublicKey().Serialize())
+	if ok && fetchedRn.GetId() == rn.GetId() {
+		indexer.GetValidatorIndex().Remove(rn.GetValidatorPublicKey().Serialize())
+	}
+}
+
+func (manager *RemoteNodeManager) SendMessage(rn *RemoteNode, desoMessage DeSoMessage) error {
```
diamondhands0, line 130: Why don't we just let them call rn.SendMessage()?

lib/network_manager.go
```diff
@@ -0,0 +1,787 @@
...
+	return nm.activeValidatorsMap.Clone()
+}
+
+// refreshValidatorIndex re-indexes validators based on the activeValidatorsMap. It is called periodically by the
+// validator connector.
+func (nm *NetworkManager) refreshValidatorIndex(activeValidatorsMap *collections.ConcurrentMap[bls.SerializedPublicKey, consensus.Validator]) {
+	// De-index inactive validators. We skip any checks regarding RemoteNodes connection status, nor do we verify whether
+	// de-indexing the validator would result in an excess number of outbound/inbound connections. Any excess connections
+	// will be cleaned up by the NonValidator connector.
+	validatorRemoteNodeMap := nm.rnManager.GetValidatorIndex().ToMap()
```
diamondhands0, line 459: Feels like there's a race conditoin here. If activeValidatorsMap is modified, eg during an epoch transitoin, at the same time as refreshValidatorIndex is called, then you could wind up with a weird set of validators in the remoteNodeManager indexes.

lib/network_manager.go
```diff
@@ -0,0 +1,787 @@
...
+		// if the RemoteNode advertised support for the SFValidator service flag during handshake, and provided us
+		// with a public key, and a corresponding proof of possession signature.
+		pk := rn.GetValidatorPublicKey()
+		if pk == nil {
+			continue
+		}
+		// It is possible that through unlikely concurrence, and malevolence, two non-validators happen to have the same
+		// public key, which goes undetected during handshake. To prevent this from affecting the indexing of the validator
+		// set, we check that the non-validator's public key is not already present in the validator index.
+		if _, ok := nm.rnManager.GetValidatorIndex().Get(pk.Serialize()); ok {
```
diamondhands0, line 481: This will actually be common. It looks like every validator creates an outbound connection to every other validator, and so you will frequently have one outbount connectoin AND one inbound connectoin for each validator. Disconnecting here seems like it will screw up the outbound validator connections for all peers.

lib/network_manager.go
```diff
@@ -0,0 +1,787 @@
...
+		publicKey, err := pk.Deserialize()
+		if err != nil {
+			continue
+		}
+
+		// For now, we only dial the first domain in the validator's domain list.
+		if len(validator.GetDomains()) == 0 {
+			continue
+		}
+		address := string(validator.GetDomains()[0])
```
diamondhands0, line 525: Should probably randomize across their domains or else everyone will pick the same one.

lib/network_manager.go
diamondhands0: Nevermind I see that you clone the map with an RLock before this function. Confusing that it's a ConcurrentMap since you don't need it to be here. Leave a comment explaining the race condition.

lib/network_manager.go
```diff
@@ -0,0 +1,787 @@
...
+	// among other things.
+	limitOneInboundRemoteNodePerIP bool
+
+	startGroup sync.WaitGroup
+	exitChan   chan struct{}
+	exitGroup  sync.WaitGroup
+}
+
+func NewNetworkManager(params *DeSoParams, cmgr *ConnectionManager, rnManager *RemoteNodeManager,
+	blsKeystore *BLSKeystore, addrMgr *addrmgr.AddrManager, connectIps []string,
```
diamondhands0, line 81: Can you name this NewNetworkManager?

lib/network_manager.go
```diff
@@ -0,0 +1,787 @@
...
+	return &NetworkManager{
+		params:                                params,
+		cmgr:                                  cmgr,
+		blsKeystore:                           blsKeystore,
+		handshake:                             NewHandshakeController(rnManager),
+		rnManager:                             rnManager,
+		AddrMgr:                               addrMgr,
+		connectIps:                            connectIps,
+		persistentIpToRemoteNodeIdsMap:        collections.NewConcurrentMap[string, RemoteNodeId](),
+		activeValidatorsMap:                   collections.NewConcurrentMap[bls.SerializedPublicKey, consensus.Validator](),
```
diamondhands0, line 94: So where would this be set initially, and where would it be updated? It feels like we want to set it in Server but only after we've gotten our tip up-to-date. Then in the steady-state it can be set after every epoch update.

tholonious: @diamondhands0 - the activeValidatorsMap is set in FastHotStuffConsensus when transitioning to the steady state and when updating the tip block. Both are cheap and result in minimal additional overhead since we have to fetch the validator set to verify every block

AeonSw4n (approved):

PR #900, opened 2024-01-03
Z/hypersync db settings fix

superzordon:
Fix badger db options for node restarts when hypersyncing.

### DIFF
--- a/cmd/node.go
+++ b/cmd/node.go
@@ -157,17 +157,19 @@ func (node *Node) Start(exitChannels ...*chan struct{}) {
 	// Check to see if this node has already been initialized with performance or default options.
 	// If so, we should continue to use those options.
 	// If not, it should be based on the sync type.
-	performanceOptions, err := lib.DbInitializedWithPerformanceOptions(node.Config.DataDirectory)
+	//performanceOptions, err := lib.DbInitializedWithPerformanceOptions(node.Config.DataDirectory)
+	//
+	//// If the db options haven't yet been saved, we should base the options on the sync type.
+	//if os.IsNotExist(err) {
+	//	performanceOptions = !node.Config.HyperSync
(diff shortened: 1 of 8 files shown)


lib/constants.go
```diff
@@ -1348,3 +1348,6 @@ const MaxAssociationTypeByteLength int = 64
 const MaxAssociationValueByteLength int = 256
 const AssociationTypeReservedPrefix = "DESO"
 const AssociationNullTerminator = byte(0)
+
+// THe name of the txt file that contains whether the current Badger DB is using performance or default options.
```
lazynina, line 1352:
```suggestion
// The name of the txt file that contains whether the current Badger DB is using performance or default options.
```

lazynina (approved):

cmd/node.go
```diff
@@ -151,13 +151,30 @@ func (node *Node) Start(exitChannels ...*chan struct{}) {
...
-		opts = lib.DefaultBadgerOptions(dbDir)
-	} else {
+
+	// Check to see if this node has already been initialized with performance or default options.
+	// If so, we should continue to use those options.
+	// If not, it should be based on the sync type.
+	performanceOptions, err := lib.DbInitializedWithPerformanceOptions(node.Config.DataDirectory)
+
+	// If the db options haven't yet been saved, we should base the options on the sync type.
+	if os.IsNotExist(err) {
```
diamondhands0, line 163: Whenever we encounter an annoying bug like this I basically like to leave the post-mortem in a long comment next to the change. Here, I would explain that the reason why you need to save the options to disk is because Badger breaks if you (a) shutdown (b) without closing (c) and then re-open with *different* options.

The other reason why this is good is it gives us the confidence to delete hacks like this in the future, rather than being scared that we'll break something if we do.

PR #902, opened 2024-01-07
Vested Linear Lockups in LockedBalanceEntry

poolcoke:


lazynina:
* `feature/proof-of-stake`
  * **PR #902** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/902?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
    * **PR #905** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/905?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/902?utm_source=stack-comment).

lib/block_view_lockups.go
```diff
@@ -9,42 +9,129 @@ import (
...
+// LockedBalanceEntryKey is a very crucial struct in the design of lockups.
+//
+// Consider a naive utilization of LockedBalanceEntryMapKey in the context of two subsequent vested unlocks
+// WITHOUT flushing to db in-between:
+//
+//		(1st Unlock, blockTimestampNanoSecs=x+100)
+//			(1) We read from disk the LockedBalanceEntry with UnlockTimestampNanoSecs=x
+//	     	(2) We cache the entry found in-memory (i.e. in the UtxoView)
+//	     	(3) We read from the in-memory cache the LockedBalanceEntry with UnlockTimestampNanoSecs=x
+//			(4) We update the in-memory entry to have UnlockTimestampNanoSecs=x+100
```
diamondhands0, line 59: (3) and (4) are wrong. What we normally do is mark the entry with UnlockTimestampNanoSex=x as isDeleted=true. Otherwise, we would not know to delete it from disk on flush. Overall, I am very confused by this comment and the edge-case it's trying to describe. And I thought we were trying to avoid all this complexity by just giving each entry a unique ID.

lib/block_view_lockups.go
```diff
@@ -9,42 +9,129 @@ import (
...
+//		(2nd Unlock, blockTimestampNanoSecs=x+150)
+//			(1) We read from disk the LockedBalanceEntry with UnlockTimestampNanoSecs=x
```
diamondhands0, line 61: If we did what I mentioned above, we would NOT read from disk here. Instead, we would see the entry in-memory with isDeleted=true.

lib/block_view_lockups.go
```diff
@@ -427,14 +710,51 @@ func (bav *UtxoView) GetLocalYieldCurvePoints(profilePKID *PKID, lockupDuration
...
+	}
+
+	return bav.PKIDToLockupYieldCurvePointKeyToLockupYieldCurvePoints[*profilePKID], nil
+}
+
 //
 // TYPES: CoinLockupMetadata
 //
 
 type CoinLockupMetadata struct {
```
diamondhands0, line 752: Include a lot of comments explaining how each param is to be used.

lib/block_view_lockups.go
```diff
@@ -739,16 +1075,36 @@ func (bav *UtxoView) _connectCoinLockup(
 			errors.Wrap(RuleErrorCoinLockupInvalidLockupDuration, "_connectCoinLockup")
 	}
 
-	// Compute the lockup duration in nanoseconds.
-	lockupDurationNanoSeconds := txMeta.UnlockTimestampNanoSecs - blockTimestampNanoSecs
+	// Validate the vesting end timestamp as expiring at a logically valid time.
+	if txMeta.VestingEndTimestampNanoSecs < txMeta.UnlockTimestampNanoSecs {
```
diamondhands0, line 1079: Isn't VestingEndTimestampNanoSecs zero in the case of a non-vesting lockup? It seems like this check would then fail for all non-vesting lockups. Not sure if I'm missing something.

lib/block_view_lockups.go
```diff
@@ -739,16 +1075,36 @@ func (bav *UtxoView) _connectCoinLockup(
...
+		return 0, 0, nil,
+			errors.Wrap(RuleErrorCoinLockupInvalidVestingEndTimestamp, "_connectCoinLockup")
+	}
 
-	// Determine the hodler PKID to use.
-	transactorPKIDEntry := bav.GetPKIDForPublicKey(txn.PublicKey)
-	if transactorPKIDEntry == nil || transactorPKIDEntry.isDeleted {
+	// In the vested case, validate that the underlying profile is the transactor.
+	if txMeta.VestingEndTimestampNanoSecs > txMeta.UnlockTimestampNanoSecs &&
```
diamondhands0, line 1085: Hmm... it looks like we're restricting vested lockups to only being creatable by the profile owner. Is this a necessary requirement? It kinda felt like any key should be able to send you locked coins of another person's supply. Example:
- N has minted 100 coins and sent them to A and B
- A wants to send B vested locked coins, which seems like it wouldn't be allowed here

I think this is also useful in the case where we want to isolate the minter of the coins from the distributor, which I think we probably want to do for security reasons.

lib/block_view_lockups.go
diamondhands0: Nevermind. I figure dout they're equal in the case of a "point" lockup. I would document this in the definition of the metadata, as I mentioned in a prior comment.

lib/block_view_lockups.go
```diff
@@ -818,109 +1174,541 @@ func (bav *UtxoView) _connectCoinLockup(
...
-	leftYieldCurvePoint, rightYieldCurvePoint, err :=
-		bav.GetLocalYieldCurvePoints(profilePKID, lockupDurationNanoSeconds)
-	if err != nil {
-		return 0, 0, nil,
-			errors.Wrap(err, "_connectCoinLockup failed to fetch yield curve points")
-	}
+	// If this is an unvested lockup, compute any accrued yield.
+	// In the vested lockup case, the yield earned is always zero.
+	yieldFromTxn := uint256.NewInt()
+	if txMeta.UnlockTimestampNanoSecs == txMeta.VestingEndTimestampNanoSecs {
```
diamondhands0, line 1180: Let's add one more safeguard here. Can you first look up if the profile has any yield curve points and add that as a check before we enter this logic? The reason is I just want profiles without yield settings to be super duper safe. So something like this:
```
if profileEnablesYield && txMeta.UnlockTimestampNanoSecs == txMeta.VestingEndTimestampNanoSecs {
```

diamondhands0: This will also protect DESO txns from entering this block at all, which is good.

lib/block_view_lockups.go
```diff
@@ -818,109 +1174,541 @@ func (bav *UtxoView) _connectCoinLockup(
...
+			return 0, 0, nil,
+				errors.Wrap(RuleErrorCoinLockupYieldCausesOverflow, "_connectCoinLockup: "+
+					"New DESO Locked Balance Entry Balance")
+		}
+
+		// (4) Set the new locked balance entry in the view
+		lockedBalanceEntry.BalanceBaseUnits = *newLockedBalanceEntryBalance
+		bav._setLockedBalanceEntry(lockedBalanceEntry)
+
+		// NOTE: While we could check for "global" overflow here, we let this occur on the unlock transaction instead.
```
diamondhands0, line 1287: Very interesting edge-case for yields. I have to think more about if we're covered here, but I think we're OK. We're definitely fine for coins that skip yield, which is good.

lib/block_view_lockups.go
```diff
@@ -818,109 +1174,541 @@ func (bav *UtxoView) _connectCoinLockup(
...
+		// Ensure in the case of DESO the resulting locked amount is less than 2**64.
+		if profilePKID.IsZeroPKID() && !newLockedBalanceEntryBalance.IsUint64() {
+			return 0, 0, nil,
+				errors.Wrap(RuleErrorCoinLockupYieldCausesOverflow, "_connectCoinLockup: "+
+					"New DESO Locked Balance Entry Balance")
+		}
+
+		// (4) Set the new locked balance entry in the view
+		lockedBalanceEntry.BalanceBaseUnits = *newLockedBalanceEntryBalance
+		bav._setLockedBalanceEntry(lockedBalanceEntry)
```
diamondhands0, line 1285: We agree that this works because we're NOT modifying the key, right?

lib/block_view_types.go
```diff
@@ -4077,10 +4077,47 @@ type GlobalParamsEntry struct {
...
+	//		April 1st 2024 -> May 1st 2024; Amount: 4 DESO
+	//		May 1st 2024 -> January 1st 2025; Amount: 5 DESO
+	//		January 1st 2025 -> February 1st 2025; Amount: 4 DESO
+	//		February 1st 2025 -> March 1st 2025; Amount: 3 DESO
+	//		March 1st 2025 -> April 1st 2025; Amount: 2 DESO
+	//		April 1st 2025 -> May 1st 2025; Amount: 1 DESO
+	//
+	// Notice that this is functionally identical but at any given point in time we hit exactly one vested
+	// locked balance entry. This consolidation on the lockup transaction operation could be computationally expensive.
+	// Hence, we limit this complexity with the MaximumVestedIntersectionsPerLockupTransactions (default: 100).
```
diamondhands0, line 4114: I'd make it 1,000. I don't think we're going to have issues with it.

lib/block_view_lockups.go
```diff
@@ -818,109 +1174,541 @@ func (bav *UtxoView) _connectCoinLockup(
...
+				HODLerPKID:                  hodlerPKID,
+				ProfilePKID:                 profilePKID,
+				UnlockTimestampNanoSecs:     txMeta.UnlockTimestampNanoSecs,
+				VestingEndTimestampNanoSecs: txMeta.VestingEndTimestampNanoSecs,
+				BalanceBaseUnits:            *lockupValue,
+			})
+		}
+
+		// (3b) Go through each existing locked balance entry and consolidate
+		if len(lockedBalanceEntries) > 0 {
```
diamondhands0, line 1334: Make this an if else. Less error-prone that way.

lib/block_view_lockups.go
```diff
@@ -818,109 +1174,541 @@ func (bav *UtxoView) _connectCoinLockup(
...
+					// Second case:
+					// proposed:                       <--------------------->
+					// existing:                                   <--------->
+					//
+					// The second case is fine to leave as it will be taken care of in the
+					// subsequent iteration. However, we make a special note to capture the
+					// remaining LockedBalanceEntry present in the first case.
+
+					// We check if there's another locked balance entry sometime in the future.
+					splitLockedBalanceEntry, remainingLockedBalanceEntry, err := SplitVestedLockedBalanceEntry(
```
diamondhands0, line 1457: This is super tricky. I think it works only because we've cut off everything on the left at this point, and so we're 100% sure that the existing entry wholly overlaps with the proposed one AND the proposed one doesn't overhang the existing entry. Only if you have both of those, THEN you can consolidate the left parts of the thing. I think it wouldn't hurt to make it clear that all of the previous chopping PLUS the fact that existing entries can't overlap with each other make this work.

lib/block_view_lockups.go
```diff
@@ -818,109 +1174,541 @@ func (bav *UtxoView) _connectCoinLockup(
...
+
+					// Set the now combined splitLockedBalanceEntry into the view.
+					bav._setLockedBalanceEntry(splitLockedBalanceEntry)
+
+					// Update the proposed locked balance entry with the remaining portion.
+					proposedLockedBalanceEntry = remainingLockedBalanceEntry
+
+					// (3b-iii) On the final iteration, the remaining proposedLockedBalanceEntry
+					//          is the only vesting schedule left.
+					if ii == len(lockedBalanceEntries)-1 {
```
diamondhands0, line 1485: I may be missing something but I *think* there's a bug here. I notice that you consolidate the proposed with the existing in this case, but you don't update the existing. So imagine a simple case like this:
```
proposed: <==================================>
existing: <============>
```

After this consolidation, I think you WANT to have:
```
proposed:               <=====================>
existing: <============>
```

And I think the timestamps are correct for the existing piece but NOT the balance. The balance of the existing seems like it will be whatever you had prior to consolidation, since you don't update the existing to be equal to the splitLockedBalanceEntry.

Lmk if I'm missing something here.

diamondhands0: I see... I think it doesn't matter because you won't use the existingBalanceEntry for anything after this point because they don't overlap. Interesting.

lib/block_view_lockups.go
```diff
@@ -818,109 +1174,541 @@ func (bav *UtxoView) _connectCoinLockup(
...
+								"_connectCoinLockup")
+					}
+
+					// Update the remaining entry.
+					proposedLockedBalanceEntry.BalanceBaseUnits = *combinedBalanceBaseUnits
+					bav._setLockedBalanceEntry(proposedLockedBalanceEntry)
+				}
+			}
+		}
+	}
```
diamondhands0, line 1512: I think this works, and it's really well-written. I really thought I'd be able to find a bug but... I couldn't so far. Great work with this. I'm still on the fence as to whether we want to do it this way vs the "capped unlock" way but will call you about it tomorrow to discuss. On the bright side, it looks like the "capped unlock" approach would be deleting a bunch of this code, except for the last part where you consolidate things with the same start+end time.

lib/block_view_lockups.go
```diff
@@ -818,109 +1174,541 @@ func (bav *UtxoView) _connectCoinLockup(
...
 
+func SplitVestedLockedBalanceEntry(
+	lockedBalanceEntry *LockedBalanceEntry,
+	startSplitTimestampNanoSecs int64,
+	endSplitTimestampNanoSecs int64,
+) (
+	_splitLockedBalanceEntry *LockedBalanceEntry,
+	_remainingLockedBalanceEntry *LockedBalanceEntry,
+	_err error,
+) {
```
diamondhands0, line 1534: Nit: One note I had while reading through this: It seems like a better API for this would be to pass it a single `splitTimestamp` and then it gives you back `leftBalanceEntry` and `rightBalanceEntry`. I also felt like that would make it a bit easier to read above, since "left" is something you can visualize, whereas "remaining" isn't. I don't think you need to redo it, but maybe leave a comment if you leave it as-is.

lib/block_view_lockups.go
```diff
@@ -818,109 +1174,541 @@ func (bav *UtxoView) _connectCoinLockup(
...
+			"((start timestamp - end timestamp) + 1) * lockedBalanceEntry.Balance overflow")
+	}
+	splitValue, err := SafeUint256().Div(numerator, denominator)
+	if err != nil {
+		return nil, errors.Wrap(err, "CalculateLockupSplitValue: "+
+			"(elapsedDuration * lockedBalanceEntry.BalanceBaseUnits) / "+
+			"(lockedBalanceEntry.UnlockTimestamp - lockedBalanceEntry.VestingEndTimestamp) has zero denominator")
+	}
+
+	return splitValue, nil
```
diamondhands0, line 1709: I think this works and is free of money printer. Good job!

lib/block_view_lockups.go
```diff
@@ -1404,6 +2196,10 @@ func (bav *UtxoView) _connectCoinLockupTransfer(
 	// Fetch PKIDs for the recipient, sender, and profile.
 	senderPKIDEntry := bav.GetPKIDForPublicKey(txn.PublicKey)
 	senderPKID := senderPKIDEntry.PKID
+	if txMeta.RecipientPublicKey.IsZeroPublicKey() {
```
diamondhands0, line 2199: Any particular reason we prohibit this for DESO? Just extra safety?

lib/block_view_lockups.go
```diff
@@ -1416,18 +2212,20 @@ func (bav *UtxoView) _connectCoinLockupTransfer(
...
 	if senderLockedBalanceEntry == nil || senderLockedBalanceEntry.isDeleted {
 		senderLockedBalanceEntry = &LockedBalanceEntry{
-			HODLerPKID:              senderPKID,
-			ProfilePKID:             profilePKID,
-			UnlockTimestampNanoSecs: txMeta.UnlockTimestampNanoSecs,
-			BalanceBaseUnits:        *uint256.NewInt(),
+			HODLerPKID:                  senderPKID,
+			ProfilePKID:                 profilePKID,
+			UnlockTimestampNanoSecs:     txMeta.UnlockTimestampNanoSecs,
+			VestingEndTimestampNanoSecs: txMeta.UnlockTimestampNanoSecs,
```
diamondhands0, line 2227: Nit: I would do this a bit differently. I would validate that txMeta.UnlockTimestampNanoSecs == txMeta.VestingTimestampNanoSecs at the top, and then use the proper values from txMeta here. The way you've done it now makes it very error-prone if we want to make vested lockups transferrable in the future.

lib/block_view_lockups.go
```diff
@@ -1695,6 +2506,53 @@ func (bav *UtxoView) _connectCoinUnlock(
...
+		// Depending on when the unlock occurs, we either DELETE or MODIFY the locked balance entry.
+		if blockTimestampNanoSecs >= unlockableLockedBalanceEntry.VestingEndTimestampNanoSecs {
+			bav._deleteLockedBalanceEntry(unlockableLockedBalanceEntry)
+		} else {
+			// DELETE the previous key.
+			bav._deleteLockedBalanceEntry(unlockableLockedBalanceEntry)
+
+			// Create and modify a copy to prevent pointer reuse.
+			modifiedLockedBalanceEntry := unlockableLockedBalanceEntry.Copy()
+			modifiedLockedBalanceEntry.UnlockTimestampNanoSecs = blockTimestampNanoSecs
```
diamondhands0, line 2540: I was noodling trying to figure out if this should be `blockTimestampNanoSecs+1` instead. I think it's right as-is but noodle on it.

lib/block_view_lockups.go
```diff
@@ -1754,6 +2612,60 @@ func (bav *UtxoView) _connectCoinUnlock(
...
+	// Compute the vested earnings.
+	vestedEarnings, err := SafeUint256().Div(numerator, denominator)
+	if err != nil {
+		return uint256.NewInt(),
+			errors.Wrap(err, "ComputeVestedEarnings failed to compute division "+
+				"((time elapsed * balance) / total time)")
+	}
+
+	// Sanity check that vestedEarnings < BalanceBaseUnits
+	if vestedEarnings.Gt(&lockedBalanceEntry.BalanceBaseUnits) ||
```
diamondhands0, line 2659: I like sanity checks like this.

lib/block_view_lockups.go
```diff
@@ -1754,6 +2612,60 @@ func (bav *UtxoView) _connectCoinUnlock(
 	return totalInput, totalOutput, utxoOpsForTxn, nil
 }
 
+func CalculateVestedEarnings(
```
diamondhands0, line 2615: Feels like you can consolidate this with CalculateLockupOverElapsedDuration, or maybe call it in this function to save doing math twice. Keep the sanity-check at the end if you do though!

lib/db_utils.go
```diff
@@ -573,12 +573,13 @@ type DBPrefixes struct {
...
 	// Retrieves LockedBalanceEntries that may or may not be claimable for unlock.
-	// LockedBalanceEntries can be retrieved by HodlerPKID and CreatorPKID are have their
-	// corresponding unlock timestamp appended to sort by timestamp.
-	// Prefix, <HodlerPKID [33]byte>, <ProfilePKID [33]byte>, <UnlockTimestampNanoSecs int64> -> <LockedBalanceEntry>
-	PrefixLockedBalanceEntryByHODLerPKIDProfilePKIDUnlockTimestampNanoSecs []byte `prefix_id:"[93]" is_state:"true"`
+	// A discriminator byte it placed before the UnlockTimestampNanoSecs to allow for separation
+	// among the vested and unvested locked balance entries without separate indexes.
+	// Prefix, <HodlerPKID [33]byte>, <ProfilePKID [33]byte>, <vested/unvested byte>,
+	// <UnlockTimestampNanoSecs [8]byte>, <VestingEndTimestampNanoSecs [8]byte> -> <LockedBalanceEntry>
+	PrefixLockedBalanceEntry []byte `prefix_id:"[93]" is_state:"true"`
```
diamondhands0, line 582: Kinda feels like you don't need to add the vested/unvested byte, since it kinda makes you go through two lookups for no reason (one for vested and then another for unvested). Alternatively, you could just look up everything that's unlockable and then split them by UnlockTimestamp != VestingEndTime or whatever. But there also doesn't seem to be any harm in it so I'd keep it this way if you like it.

lib/block_view_lockups.go
```diff
@@ -739,16 +1075,36 @@ func (bav *UtxoView) _connectCoinLockup(
...
+	if txMeta.VestingEndTimestampNanoSecs > txMeta.UnlockTimestampNanoSecs &&
+		!reflect.DeepEqual(txn.PublicKey, txMeta.ProfilePublicKey.ToBytes()) {
 		return 0, 0, nil,
-			errors.Wrap(RuleErrorCoinLockupInvalidTransactorPKID, "_connectCoinLockup")
+			errors.Wrapf(RuleErrorCoinLockupInvalidVestedTransactor, "_connectCoinLockup: Profile "+
+				"pub key: %v, signer public key: %v", PkToString(txn.PublicKey, bav.Params),
+				PkToString(txn.PublicKey, bav.Params))
 	}
-	hodlerPKID := transactorPKIDEntry.PKID
+
```
diamondhands0, line 1092: I want you to add a deep sanity-check as follows:
1. Near the top of this function, after you have everyone's pkids, snapshot the balances of the transactor, the recipient, and the creator.
2. Toward the bottom, pull the three balances again AND compute the delta from (sum(newLockedBlanaceEntries) - sum(prevLockedBalanceEntries))
3. Verify that sum(newBalances) + sum(lockedBalanceEntryDelta) <= sum(oldBalances)

diamondhands0: We do a similar check for NFTs and DAO Coin Limit Orders btw. Those are more hardcore than what I'm suggesting here but you can look at them.

lib/block_view_lockups.go
```diff
@@ -1404,6 +2196,10 @@ func (bav *UtxoView) _connectCoinLockupTransfer(
 	// Fetch PKIDs for the recipient, sender, and profile.
 	senderPKIDEntry := bav.GetPKIDForPublicKey(txn.PublicKey)
 	senderPKID := senderPKIDEntry.PKID
+	if txMeta.RecipientPublicKey.IsZeroPublicKey() {
+		return 0, 0, nil,
+			errors.Wrap(RuleErrorCoinLockupTransferToZeroPublicKey, "_connectCoinLockupTransfer")
+	}
```
diamondhands0, line 2202: Add a sanity check to this function as follows:
1. Snapshot the sender's balance at the beginning.
2. Toward the end, compute lockedBalanceDelta = sum(prevLockedBlanaceEntries) - sum(newLockedBalanceEntries).
3. Verify that newBalance + lockedBalanceDelta <= oldBalance

diamondhands0: Extra credit: Add a similar sanity-check to the lockup transfer function as well.

PR #913, opened 2024-01-16
Remove error prone pointer equality check from core.

poolcoke:


lib/blockchain.go
```diff
@@ -1548,7 +1548,7 @@ func _FindCommonAncestor(node1 *BlockNode, node2 *BlockNode) *BlockNode {
 	// reach the end of the lists. We only need to check node1 for nil
 	// since they're the same height and we are iterating both back
 	// in tandem.
-	for node1 != nil && node1 != node2 {
+	for node1 != nil && !node1.Hash.IsEqual(node2.Hash) {
```
AeonSw4n, line 1551: Can node2 be nil here?

poolcoke: It's a good question.

At the top of the function if node2 is nil it'll return nil, meaning calling _FindCommonAncestor with node2 being nil isn't possible.

That means the only risk of node2 being nil should come from this block of code later in _FindCommonAncestor:

`	if node1.Height > node2.Height {
		node1 = node1.Ancestor(node2.Height)
	} else if node1.Height < node2.Height {
		node2 = node2.Ancestor(node1.Height)
	}`
	
The Ancestor function just goes through the Parent chain of node2 until it hits the specified height. I guess if we have a block without a parent/grandparent/great-grandparent somewhere it could be issue... maybe someone with more PoS specific implementation knowledge could comment.

If node2 is ever nil it would've caused the node to panic (even before this change) as the calling code depends on node2 existing. We could add the check if we want to be extra sure... but then callers of _FindCommonAncestor need to know that it could possibly return nil (we don't do this now).

lazynina (approved):

tholonious (approved):

PR #918, opened 2024-01-18
Move Validator Set Fetch Helper Function to UtxoView

tholonious:
No functionality change. This just moves the helper function that maps block height -> validator set into the UtxoView struct so it can be reused there.

Code behavior and unit test results do not change.


tholonious:
* `feature/pos-syncing-and-steady-state`
  * **PR #918** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/918?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/918?utm_source=stack-comment).

lib/pos_consensus.go
```diff
@@ -602,61 +602,30 @@ func (cc *FastHotStuffConsensus) fetchValidatorListsForSafeBlocks(blocks []*MsgD
...
-			validatorSetAtBlock, err = utxoView.GetAllSnapshotValidatorSetEntriesByStakeAtEpochNumber(snapshotEpochNumber)
-			if err != nil {
-				return nil, errors.Errorf("Error fetching validator set for block: %v", err)
-			}
+		validatorSetAtBlock, ok := validatorSetByBlockHeight[block.Height+1]
+		if !ok {
+			return nil, errors.Errorf("Error fetching validator set for block: %v", block)
 		}
 
 		blocksWithValidatorLists[ii] = consensus.BlockWithValidatorList{
```
lazynina, line 631: Would it make sense to convert all the ValidatorEntries to ConsensusInterface before the loop? Minor optimization I guess, but if the number of safe blocks is large (m) and the number of validators is also somewhat large (n), we get a O(m*n) with having it in the loop.

lazynina (approved):
One minor comment about possible performance optimization (not introduced in this PR)

lib/pos_consensus.go
tholonious: Yes! Great idea

tholonious: It'll be O(m*n) even if we perform the conversion before the loop and want to do it in a simple way. I feel fine about leaving the conversion as-is since the number of safe blocks will be small (< 10).

PR #942, opened 2024-01-24
PoS NetworkManager Persistent Connector

AeonSw4n:


AeonSw4n:
* `feature/pos-networking-and-syncing`
  * **PR #942** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/942?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
    * **PR #943** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/943?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #944** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/944?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
        * **PR #945** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/945?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/942?utm_source=stack-comment).

lib/connection_controller.go
```diff
@@ -5,12 +5,23 @@ import (
...
 	"github.com/pkg/errors"
 	"net"
 	"strconv"
+	"sync"
+	"time"
 )
 
+type GetActiveValidatorsFunc func() *collections.ConcurrentMap[bls.SerializedPublicKey, *ValidatorEntry]
+
+var GetActiveValidatorImpl GetActiveValidatorsFunc = BasicGetActiveValidators
```
lazynina, line 19: Is this going to be used outside of test files? If it's only going to be used in test files, let's move it to one of the test files instead of sticking it in the connection controller file. I know we've used this pattern in the past for the encoder migration tests, but it's not my favorite as there's extra overhead to figure out what's actually happening.

lib/connection_controller.go
```diff
@@ -141,13 +203,34 @@ func (cc *ConnectionController) cleanupFailedOutboundConnection(connection Conne
...
+func (cc *ConnectionController) refreshConnectIps() {
+	// Connect to addresses passed via the --connect-ips flag. These addresses are persistent in the sense that if we
+	// disconnect from one, we will try to reconnect to the same one.
+	for _, connectIp := range cc.connectIps {
+		if _, ok := cc.persistentIpToRemoteNodeIdsMap[connectIp]; ok {
+			continue
+		}
+
+		glog.Infof("ConnectionController.initiatePersistentConnections: Connecting to connectIp: %v", connectIp)
+		id, err := cc.CreateNonValidatorPersistentOutboundConnection(connectIp)
```
lazynina, line 223: so connect-ips flag won't allow a connection to a validator node?

lazynina (approved):

lib/connection_controller.go
tholonious, line 19: How will GetActiveValidatorImpl be used? Seems like the intention is to globally reassign from somewhere else in the code?

lib/connection_controller.go
```diff
@@ -32,8 +43,12 @@ type ConnectionController struct {
 	// it's aware of at random and provides it to us.
 	AddrMgr *addrmgr.AddrManager
 
-	// When --connectips is set, we don't connect to anything from the addrmgr.
+	// When --connect-ips is set, we don't connect to anything from the addrmgr.
 	connectIps []string
+	// persistentIpToRemoteNodeIdsMap maps persistent IP addresses, like the --connect-ips, to the RemoteNodeIds of the
+	// corresponding RemoteNodes. This is used to ensure that we don't connect to the same persistent IP address twice.
+	// And that we can reconnect to the same persistent IP address if we disconnect from it.
+	persistentIpToRemoteNodeIdsMap map[string]RemoteNodeId
```
tholonious, line 51: Consider tracking this in the RemoteNodeManager. If we're going to index and dedupe RemoteNodex by some key, that seems to be the most relevant place to do it

lib/connection_controller.go
```diff
@@ -57,16 +77,45 @@ func NewConnectionController(params *DeSoParams, cmgr *ConnectionManager, handsh
...
 }
 
+func (cc *ConnectionController) startPersistentConnector() {
+	cc.startGroup.Done()
+	for {
+		select {
+		case <-cc.exitChan:
+			cc.exitGroup.Done()
+			return
+		case <-time.After(1 * time.Second):
```
tholonious, line 113: What was the connectIps refresh timer in the old code? It'll be good to set this cadence as a flag if possible too

lib/connection_controller.go
AeonSw4n: Good call! We figured out a better approach based on pushing the latest validator set to the NetworkManager (ConnectionController) directly from consensus. This approach resembles how Mempool receives the latest UtxoView, if you remember how we solved it there.

lib/connection_controller.go
AeonSw4n: Persistent remote nodes can still be validators. The ValidatorConnector routine may reindex the persistent remote node after completing the handshake. We default to a NonValidator connection so that we don't have to assume the public key of this RemoteNode (Validator RemoteNodes require an expected public key passed on initialization).

lib/connection_controller.go
AeonSw4n: Fixed in #945

AeonSw4n: It's in #945 btw.

lib/connection_controller.go
AeonSw4n: Hmm, I thought about it, and it seems like keeping it in the NetworkManager is a better choice, given the `connectIps` map is part of the NetworkManager. It's also more convenient to iterate through this map when it is in the NetworkManager.

AeonSw4n: I changed it to ConcurrentMap by the way.

lib/connection_controller.go
AeonSw4n: I'll update it in #945.

PR #943, opened 2024-01-24
PoS NetworkManager Validator Connector

AeonSw4n:


AeonSw4n:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/deso-protocol/core/943?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* `feature/pos-networking-and-syncing`
  * **PR #942** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/942?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #943** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/943?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
      * **PR #944** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/944?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
        * **PR #945** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/945?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/943?utm_source=stack-comment).

lib/connection_controller.go
```diff
@@ -231,6 +252,84 @@ func (cc *ConnectionController) refreshConnectIps() {
...
+	}
+
+	validators := activeValidatorsMap.Copy()
+	for pk, validator := range validators {
+		_, exists := cc.rnManager.GetValidatorIndex().Get(pk)
+		// If we're already connected to the validator, continue.
+		if exists {
+			continue
+		}
+		if cc.blsKeystore.GetSigner().GetPublicKey().Serialize() == pk {
```
lazynina, line 315: We should use bytes.Equal here.

lib/connection_controller.go
```diff
@@ -231,6 +252,84 @@ func (cc *ConnectionController) refreshConnectIps() {
...
+
+		publicKey, err := pk.Deserialize()
+		if err != nil {
+			continue
+		}
+
+		// For now, we only dial the first domain in the validator's domain list.
+		address := string(validator.Domains[0])
+		if err := cc.CreateValidatorConnection(address, publicKey); err != nil {
+			// TODO: Do we want to log an error here?
```
lazynina, line 327: Yes please. Let's glog this error.

lazynina (changes requested):

lib/connection_controller.go
```diff
@@ -116,6 +117,26 @@ func (cc *ConnectionController) startPersistentConnector() {
...
+// startValidatorConnector is responsible for ensuring that the node is connected to all active validators. It does
+// this in two steps. First, it looks through the already established connections and checks if any of these connections
+// are validators. If they are, it adds them to the validator index. It also checks if any of the existing validators
+// are no longer active and removes them from the validator index. Second, it checks if any of the active validators
+// are missing from the validator index. If they are, it attempts to connect to them.
+func (cc *ConnectionController) startValidatorConnector() {
+	cc.startGroup.Done()
+	for {
+		select {
+		case <-cc.exitChan:
```
tholonious, line 129: AFAIK, trying to read from the exitChan channel here shouldn't work. You already call `case <-cc.exitChan:` in the `startPersistentConnector` event loop above. Only one of the event loops will be able to read from the exitChan event....

lib/connection_controller.go
```diff
@@ -231,6 +252,84 @@ func (cc *ConnectionController) refreshConnectIps() {
...
+func (cc *ConnectionController) refreshValidatorIndex(activeValidatorsMap *collections.ConcurrentMap[bls.SerializedPublicKey, *ValidatorEntry]) {
+	// De-index inactive validators. We skip any checks regarding RemoteNodes connection status, nor do we verify whether
+	// de-indexing the validator would result in an excess number of outbound/inbound connections. Any excess connections
+	// will be cleaned up by the peer connector.
+	validatorRemoteNodeMap := cc.rnManager.GetValidatorIndex().Copy()
+	for pk, rn := range validatorRemoteNodeMap {
+		// If the validator is no longer active, de-index it.
+		if _, ok := activeValidatorsMap.Get(pk); !ok {
+			cc.rnManager.SetNonValidator(rn)
+			cc.rnManager.UnsetValidator(rn)
```
tholonious, line 270: Would it make sense to disconnect from these validators instead? Ex: what happens if have the connectIps flag as non-empty, a validator leaves the validator set: we would end up with an extra non-validator connection that we don't intend to maintain

lib/connection_controller.go
AeonSw4n: I think this one is okay to leave as `==`, both types are strings actually.

lib/connection_controller.go
AeonSw4n: It's possible this inactive validator will be disconnected later by the NonValidator connector, which makes decisions about expunging from the NonValidator index. As I was thinking about the organization of the NetworkManager, this design seemed to result in a cleaner separation of concerns.

As for the example you mentioned, the connectIps RemoteNodes (persistent RemoteNodes) get "special treatment" in that we'll never intentionally disconnect from them. Persistent peers are also not included in the count for targetOutbound/InboundRemoteNodes. The only way for us to drop persistent RemoteNodes will be them closing the connection with us.

PR #944, opened 2024-01-24
PoS NetworkManager NonValidator Connector

AeonSw4n:


integration_testing/connection_controller_routines_test.go
```diff
@@ -302,6 +302,49 @@ func TestConnectionControllerValidatorInboundDeduplication(t *testing.T) {
...
+	node1 := spawnNonValidatorNodeProtocol2(t, 18000, "node1")
+	node1.Config.TargetOutboundPeers = 0
+	node2 := spawnNonValidatorNodeProtocol2(t, 18001, "node2")
+	node3 := spawnNonValidatorNodeProtocol2(t, 18002, "node3")
+	node4 := spawnNonValidatorNodeProtocol2(t, 18003, "node4")
+	node5 := spawnNonValidatorNodeProtocol2(t, 18004, "node5")
+	node6 := spawnNonValidatorNodeProtocol2(t, 18005, "node6")
+
+	node2 = startNode(t, node2)
+	defer node2.Stop()
```
lazynina, line 320: Your `startNode` function already cleans these up at the end of the test. We can get rid of all the `node.Stop` calls here.

lib/connection_controller.go
```diff
@@ -330,6 +351,146 @@ func (cc *ConnectionController) connectValidators(activeValidatorsMap *collectio
...
+
+func (cc *ConnectionController) connectNonValidators() {
+	// Only connect to addresses from the addrmgr if we don't specify --connect-ips. These addresses are *not* persistent,
+	// meaning if we disconnect from one we'll try a different one.
+	// TODO: We used this condition in the old code to prevent the node from connecting to non-connect-ips nodes.
+	// 	I'm not sure whether this is still necessary. I suppose the concern here was that the connect-ips nodes
+	//	should be prioritized over the addrmgr nodes, especially during syncing. However, I think we can achieve the
+	// 	same result by defining another flag, like a boolean --sync-from-persistent-peers-only, which could indicate that
+	//	we disregard non-persistent non-connect-ips nodes during syncing, if the flag is set to true.
+	// FIXME: This about uncommenting the below condition.
```
lazynina, line 441: @diamondhands0 will have to make the call on this. I think there are two flags: `--add-ips` and `--connect-ips`. My understanding is that `--connect-ips` means ONLY connect to these peers and no other peers. I don't know if that'll work for PoS.

lib/connection_controller.go
```diff
@@ -330,6 +351,146 @@ func (cc *ConnectionController) connectValidators(activeValidatorsMap *collectio
...
+		remainingOutboundPeers = cc.targetNonValidatorOutboundRemoteNodes - numOutboundPeers
+	}
+	for ii := uint32(0); ii < remainingOutboundPeers; ii++ {
+		addr := cc.getRandomUnconnectedAddress()
+		if addr == nil {
+			break
+		}
+		cc.AddrMgr.Attempt(addr)
+		// FIXME: error handle
+		cc.rnManager.CreateNonValidatorOutboundConnection(addr)
```
lazynina, line 459: Let's at least put it into a variable and glog it for now please.

lib/connection_controller.go
```diff
@@ -330,6 +351,146 @@ func (cc *ConnectionController) connectValidators(activeValidatorsMap *collectio
...
+		// Bitcoin and we do it to prevent Sybil attacks.
+		if cc.cmgr.IsFromRedundantOutboundIPAddress(addr.NetAddress()) {
+			//glog.V(2).Infof("ConnectionManager.getRandomUnconnectedAddress: Not choosing address due to redundant group key %v:%v", addr.NetAddress().IP, addr.NetAddress().Port)
+			continue
+		}
+
+		return addr.NetAddress()
+	}
+
+	//glog.V(2).Infof("ConnectionManager.getRandomAddr: Returning nil")
```
lazynina, line 490: Can we remove all these commented out glog statements in this function please?

lib/connection_controller.go
```diff
@@ -137,6 +138,26 @@ func (cc *ConnectionController) startValidatorConnector() {
...
+// startNonValidatorConnector is responsible for ensuring that the node is connected to the target number of outbound
+// and inbound remote nodes. To do this, it periodically checks the number of outbound and inbound remote nodes, and
+// if the number is above the target number, it disconnects the excess remote nodes. If the number is below the target
+// number, it attempts to connect to new remote nodes.
+func (cc *ConnectionController) startNonValidatorConnector() {
+	cc.startGroup.Done()
+
+	for {
+		select {
+		case <-cc.exitChan:
```
tholonious, line 150: Same comment as in the ValidatorConnector PR. This thread won't be able to read from the existChan. Please make sure this is safe

lib/connection_controller.go
tholonious: @AeonSw4n - if I understand correctly, you want to be able to dial outbound connections to ips defined in the addrmgr even when the `--connect-ips` flag is defined. And you want to define the separate `--sync-from-persistent-peers-only` to prevent the node from syncing from the addrmgr peers.

What is the purpose of dialing an outbound non-validator connection from the addrmgr if you don't intend to sync from the peer?

lib/connection_controller.go
```diff
@@ -330,6 +351,146 @@ func (cc *ConnectionController) connectValidators(activeValidatorsMap *collectio
...
+		cc.rnManager.CreateNonValidatorOutboundConnection(addr)
+	}
+}
+
+func (cc *ConnectionController) getRandomUnconnectedAddress() *wire.NetAddress {
+	for tries := 0; tries < 100; tries++ {
+		addr := cc.AddrMgr.GetAddress()
+		if addr == nil {
+			//glog.V(2).Infof("ConnectionManager.getRandomUnconnectedAddress: addr from GetAddressWithExclusions was nil")
+			break
```
tholonious, line 468: Why `break` and not `continue`?

tholonious (changes requested):

integration_testing/connection_controller_routines_test.go
AeonSw4n: Lol forgot about it. Nice find!

lib/connection_controller.go
AeonSw4n: We discussed this off-GH with @tholonious and figured out that the `--sync-from-persistent-peers-only` flag is redundant. The `--connect-ips` flag is sufficient to maintain the syncing behavior of the old code.

lib/connection_controller.go
AeonSw4n: If the `GetAddress()` returns nil, the `AddrMgr` is empty. So we can break early.

PR #945, opened 2024-01-24
PoS NetworkManager RemoteNode Cleanup

AeonSw4n:


integration_testing/connection_controller_routines_test.go
```diff
@@ -29,9 +30,13 @@ func TestConnectionControllerInitiatePersistentConnections(t *testing.T) {
 	node5 := spawnValidatorNodeProtocol2(t, 18004, "node5", blsPriv5)
 
 	node2 = startNode(t, node2)
+	defer node2.Stop()
```
lazynina, line 33: delete these!! your startNode helper is already cleaning these up for you.

integration_testing/connection_controller_routines_test.go
```diff
@@ -66,23 +71,93 @@ func TestConnectionControllerInitiatePersistentConnections(t *testing.T) {
 		node5.Listeners[0].Addr().String(),
 	}
 	node6 = startNode(t, node6)
+	defer node6.Stop()
```
lazynina, line 74: delete.

integration_testing/connection_controller_routines_test.go
```diff
@@ -66,23 +71,93 @@ func TestConnectionControllerInitiatePersistentConnections(t *testing.T) {
...
+	node1 := spawnNonValidatorNodeProtocol2(t, 18000, "node1")
+	node2 := spawnNonValidatorNodeProtocol2(t, 18001, "node2")
+
+	node1.Config.ConnectIPs = []string{"127.0.0.1:18001"}
+	node2.Config.ConnectIPs = []string{"127.0.0.1:18000"}
+
+	node1 = startNode(t, node1)
+	node2 = startNode(t, node2)
+	defer node1.Stop()
+	defer node2.Stop()
```
lazynina, line 99:
```suggestion
```

integration_testing/connection_controller_routines_test.go
```diff
@@ -339,26 +414,121 @@ func TestConnectionControllerNonValidatorConnector(t *testing.T) {
...
+	waitForValidatorConnection(t, node1, node4)
+	waitForValidatorConnection(t, node1, node5)
+	waitForValidatorConnection(t, node1, node6)
+	waitForCountRemoteNodeIndexerHandshakeCompleted(t, node1, 6, 5, 0, 1)
+}
 
-	waitForCountRemoteNodeIndexer(t, node1, 2, 0, 1, 1)
-	waitForCountRemoteNodeIndexer(t, node2, 2, 0, 1, 1)
+func TestConnectionControllerNonValidatorConnectorAddressMgr(t *testing.T) {
+	require := require.New(t)
```
lazynina, line 516: let's stop doing this? shadowing the package name is bad practice. If you want to create a new `*Assertions`, please save to a different variable name so it does not shadow the package name `require`. Please apply this to all tests where you do this.

integration_testing/connection_controller_routines_test.go
```diff
@@ -339,26 +414,121 @@ func TestConnectionControllerNonValidatorConnector(t *testing.T) {
...
+	// Spawn a non-validator node1. Set node1's targetOutboundPeers to 2 and targetInboundPeers to 0. Then
+	// add two ip addresses to AddrMgr. Make sure that node1 creates outbound connections to these nodes.
+	node1 := spawnNonValidatorNodeProtocol2(t, 18000, "node1")
+	node1.Config.TargetOutboundPeers = 2
+	node1.Config.MaxInboundPeers = 0
+
+	node1 = startNode(t, node1)
+	defer node1.Stop()
```
lazynina, line 524:
```suggestion
```

integration_testing/connection_controller_test.go
```diff
@@ -382,6 +394,8 @@ func TestConnectionControllerProtocolDifference(t *testing.T) {
 }
 
 func TestConnectionControllerPersistentConnection(t *testing.T) {
+	t.Cleanup(ResetDisableNetworkManagerRoutines)
```
lazynina, line 397: good - using t.Cleanup is best practice for cleaning up after your tests. However, I may update `SetDisableNetworkManagerRoutines` to take in `*testing.T` and add the `t.Cleanup` call in there so you don't need to have `t.Cleanup` and then `SetDisableNetworkManagerRoutines` everywhere. You can simply have `SetDisableNetworkManagerRoutines(t)`

integration_testing/connection_controller_utils_test.go
```diff
@@ -129,6 +129,21 @@ func waitForCountRemoteNodeIndexer(t *testing.T, node1 *cmd.Node, allCount int,
...
+
+	userAgent := node1.Params.UserAgent
+	rnManager := node1.Server.GetConnectionController().GetRemoteNodeManager()
+	condition := func() bool {
+		if true != checkRemoteNodeIndexerCountHandshakeCompleted(rnManager, allCount, validatorCount,
+			nonValidatorOutboundCount, nonValidatorInboundCount) {
+			return false
+		}
+		return true
+	}
```
lazynina, line 143: This feels like a roundabout way to write this? 
```suggestion
	condition := func() bool {
		return checkRemoteNodeIndexerCountHandshakeCompleted(rnManager, allCount, validatorCount,
			nonValidatorOutboundCount, nonValidatorInboundCount)
	}
```

integration_testing/connection_controller_utils_test.go
```diff
@@ -232,3 +283,11 @@ func spawnValidatorNodeProtocol2(t *testing.T, port uint32, id string, blsPriv *
...
 	return node
 }
+
+func ResetDisableNetworkManagerRoutines() {
+	lib.DisableNetworkManagerRoutines = false
+}
+
+func SetDisableNetworkManagerRoutines() {
+	lib.DisableNetworkManagerRoutines = true
+}
```
lazynina, line 293:
```suggestion
func SetDisableNetworkManagerRoutines(t *testing.T) {
	t.Cleanup(func() { lib.DisableNetworkManagerRoutines = false})
	lib.DisableNetworkManagerRoutines = true
}
```

integration_testing/connection_controller_utils_test.go
```diff
@@ -232,3 +283,11 @@ func spawnValidatorNodeProtocol2(t *testing.T, port uint32, id string, blsPriv *
 	node.Params.ProtocolVersion = lib.ProtocolVersion2
 	return node
 }
+
+func ResetDisableNetworkManagerRoutines() {
+	lib.DisableNetworkManagerRoutines = false
+}
```
lazynina, line 289:
```suggestion
```

lib/connection_controller.go
```diff
@@ -415,16 +448,30 @@ func (cc *ConnectionController) refreshNonValidatorOutboundIndex() {
 func (cc *ConnectionController) refreshNonValidatorInboundIndex() {
 	// First let's check if we have an excess number of inbound remote nodes. If we do, we'll disconnect some of them.
 	numConnectedInboundRemoteNodes := uint32(cc.rnManager.GetNonValidatorInboundIndex().Count())
-	excessiveInboundRemoteNodes := uint32(0)
-	if numConnectedInboundRemoteNodes > cc.targetNonValidatorInboundRemoteNodes {
-		excessiveInboundRemoteNodes = numConnectedInboundRemoteNodes - cc.targetNonValidatorInboundRemoteNodes
+	if numConnectedInboundRemoteNodes <= cc.targetNonValidatorInboundRemoteNodes {
```
lazynina, line 451: ❤️

lazynina (changes requested):

lib/connection_controller.go
```diff
@@ -158,6 +166,21 @@ func (cc *ConnectionController) startNonValidatorConnector() {
 	}
 }
 
+func (cc *ConnectionController) startRemoteNodeCleanup() {
+	cc.startGroup.Done()
+
+	for {
+		select {
+		case <-cc.exitChan:
```
tholonious, line 174: Same comment in the other PRs. Does this work when multiple event loops are reading from the exitChan?

lib/constants.go
```diff
@@ -43,6 +43,10 @@ const (
 	MessagesToFetchPerInboxCall = 10000
 )
 
+// DisableNetworkManagerRoutines is used to disable the network manager routines.
+// This is useful for testing.
+var DisableNetworkManagerRoutines = false
```
tholonious, line 48: Please consider defining this flag as a field on the ConnectionController that you can set on each instance of the controller. Global flags typically lead to headaches down the line as the code gets more complicated

lib/peer.go
```diff
@@ -140,6 +141,8 @@ type Peer struct {
 	// SyncType indicates whether blocksync should not be requested for this peer. If set to true
 	// then we'll only hypersync from this peer.
 	syncType NodeSyncType
+
+	startGroup sync.WaitGroup
```
tholonious, line 145: Add a comment here on why we need this startGroup

integration_testing/connection_controller_routines_test.go
AeonSw4n: Are you sure? I think it's fineee. We've been doing this everywhere in our unit tests.

lib/constants.go
```diff
@@ -43,6 +43,10 @@ const (
...
+
```
tholonious, line 49: Global flags used testing are more likely to become tech debt in the long run that outweights the convenience in the short term. Is there a way you can still test without needing a global flag to disable the routines? Try defining a boolean constructor param on NetworkManager that disables the routines when flipped.

lib/remote_node_manager.go
```diff
@@ -51,12 +51,19 @@ func NewRemoteNodeManager(srv *Server, bc *Blockchain, cmgr *ConnectionManager,
 	}
 }
 
-func (manager *RemoteNodeManager) newRemoteNode(validatorPublicKey *bls.PublicKey) *RemoteNode {
+func (manager *RemoteNodeManager) DisconnectAll() {
+	for _, rn := range manager.GetAllRemoteNodes().GetAll() {
```
tholonious, line 55: Is this thread-safe? What happens if a new remote node is added after `GetAllRemoteNodes()` returns

lib/network.go
```diff
@@ -854,23 +854,6 @@ func (msg *MsgDeSoDisconnectedPeer) FromBytes(data []byte) error {
 	return fmt.Errorf("MsgDeSoDisconnectedPeer.FromBytes not implemented")
 }
 
-// MsgDeSoPeerHandshakeComplete is a control message that is used to internally signal when a peer has
```
tholonious, line 857: Why remove this control message type?

tholonious (changes requested):

lib/network.go
AeonSw4n: Yeah, this is a fun simplification. The `MsgDeSoPeerHandshakeComplete` message turned out to be redundant: it was emitted by the HandshakeManager, and then handled back at HandshakeManager. Not only was this an unnecessary redirection, it actually caused deadlocking! 

The thread that was enqueuing the `HandshakeComplete` message to the server's `incomingMessages` channel was the same thread that was dequeueing messages from the `incomingMessages`. So if the `incomingMessages` channel was full as we're enqueueing the `HandshakeComplete` message, we would have deadlocked, since no thread would be able to evict the `incomingMessages` channel.

The simplest solution was to just remove the `HandshakeComplete` message indirection, and replace it with a direct call to HandshakeManager's handler for the `HandshakeComplete` message.

lib/remote_node_manager.go
AeonSw4n: I guess technically it is possible that some new connection came through as we're stopping the Node. But right now NetworkManager is stopped at the very end of Server's stop, after ConnectionManager, meaning as we get to NetworkManager.Stop, the ConnectionManager already seized to serve new incoming connections. Which makes the `DisconnectAll` implicitly thread-safe. If you're wondering if we should mutex-gate this method, I tried it initially, but it actually causes us to deadlock, since Disconnect also calls on the `mtx` mutex. That said, I think it's fine to leave this method without the mutex, as it's only called by the NetworkManager on exit. I will document this method a little more in a future PR.

lib/constants.go
AeonSw4n: Okay, agreed. Just fixed it.

integration_testing/connection_controller_routines_test.go
AeonSw4n: I thought about it, and I agree with you. It's a bad practice. It's fixed now.

lib/connection_controller.go
```diff
@@ -48,7 +41,12 @@ type ConnectionController struct {
 	// persistentIpToRemoteNodeIdsMap maps persistent IP addresses, like the --connect-ips, to the RemoteNodeIds of the
 	// corresponding RemoteNodes. This is used to ensure that we don't connect to the same persistent IP address twice.
 	// And that we can reconnect to the same persistent IP address if we disconnect from it.
-	persistentIpToRemoteNodeIdsMap map[string]RemoteNodeId
+	persistentIpToRemoteNodeIdsMap *collections.ConcurrentMap[string, RemoteNodeId]
+
+	validatorMapLock sync.RWMutex
+	// validatorMap is a list of all validators that we are connected to. It will be updated periodically by the
+	// owner of the ConnectionController.
+	validatorMap *collections.ConcurrentMap[bls.SerializedPublicKey, consensus.Validator]
```
tholonious, line 49: Please clarify the naming. If I understand correctly, these are the validators the NetworkManager wants to connect to
```suggestion
	validatorsToConnectLock sync.RWMutex
	// validatorsToConnect is a map of all validators registered in consensus that we need to connect to.
	// It will be updated periodically by the owner of the ConnectionController.
	validatorsToConnect *collections.ConcurrentMap[bls.SerializedPublicKey, consensus.Validator]
```

lib/connection_controller.go
```diff
@@ -269,21 +296,34 @@ func (cc *ConnectionController) refreshConnectIps() {
...
-		cc.persistentIpToRemoteNodeIdsMap[connectIp] = id
+		cc.persistentIpToRemoteNodeIdsMap.Set(connectIp, id)
 	}
 }
 
 // ###########################
 // ## Validator Connections
 // ###########################
 
+func (cc *ConnectionController) SetValidatorMap(validatorMap *collections.ConcurrentMap[bls.SerializedPublicKey, consensus.Validator]) {
```
tholonious, line 307:
```suggestion
func (cc *ConnectionController) SetValidatorToConnect(validatorsToConnect *collections.ConcurrentMap[bls.SerializedPublicKey, consensus.Validator]) {
```

lib/connection_controller.go
```diff
@@ -269,21 +296,34 @@ func (cc *ConnectionController) refreshConnectIps() {
...
+	cc.validatorMapLock.Lock()
+	defer cc.validatorMapLock.Unlock()
+	cc.validatorMap = validatorMap.Clone()
+
+}
+
+func (cc *ConnectionController) getValidatorMap() *collections.ConcurrentMap[bls.SerializedPublicKey, consensus.Validator] {
```
tholonious, line 314:
```suggestion
func (cc *ConnectionController) getValidatorsToConnect() *collections.ConcurrentMap[bls.SerializedPublicKey, consensus.Validator] {
```

tholonious (approved):

PR #972, opened 2024-01-30
Convert MsgDeSoHeader TstampNanoSecs to int64

tholonious:


tholonious:
* `feature/proof-of-stake`
  * **PR #972** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/972?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/972?utm_source=stack-comment).

lib/postgres.go
```diff
@@ -1289,7 +1289,7 @@ func (postgres *Postgres) UpsertBlockTx(tx *pg.Tx, blockNode *BlockNode) error {
 
 		TxMerkleRoot: blockNode.Header.TransactionMerkleRoot,
 		Version:      blockNode.Header.Version,
-		Timestamp:    blockNode.Header.TstampNanoSecs,
+		Timestamp:    uint64(blockNode.Header.TstampNanoSecs),
```
tholonious, line 1292: This will be deprecated anyway, so I'm adding this change so the code complies.

lib/network.go
```diff
@@ -2194,7 +2194,7 @@ func (msg *MsgDeSoHeader) EncodeHeaderVersion2(preSignature bool) ([]byte, error
 
 	// TstampNanosSecs: this field can be encoded to take up to the full 64 bits now
 	// that MsgDeSoHeader version 2 does not need to be backwards compatible.
-	retBytes = append(retBytes, UintToBuf(msg.TstampNanoSecs)...)
+	retBytes = append(retBytes, IntToBuf(msg.TstampNanoSecs)...)
```
lazynina, line 2197: I don't think it's safe to just change this, but I'm not 100% sure. Basically I'd opt to continue encoding it with UintToBuf after casting to uint64. We should code defensively to make sure the value is never negative. I'm worried that this will mess with syncing since headers won't match.

lib/network.go
```diff
@@ -2309,7 +2309,7 @@ func DecodeHeaderVersion0(rr io.Reader) (*MsgDeSoHeader, error) {
 		if err != nil {
 			return nil, errors.Wrapf(err, "MsgDeSoHeader.FromBytes: Problem decoding TstampSecs")
 		}
-		retHeader.SetTstampSecs(uint64(binary.LittleEndian.Uint32(scratchBytes[:])))
+		retHeader.SetTstampSecs(int64(binary.LittleEndian.Uint32(scratchBytes[:])))
```
lazynina, line 2312: This seems fine to me since we're not changing how we encode it.

lib/network.go
```diff
@@ -2409,7 +2409,7 @@ func DecodeHeaderVersion2(rr io.Reader) (*MsgDeSoHeader, error) {
 	}
 
 	// TstampNanoSecs
-	retHeader.TstampNanoSecs, err = ReadUvarint(rr)
+	retHeader.TstampNanoSecs, err = ReadVarint(rr)
```
lazynina, line 2412: Yeah can we change this back? I know it's kinda unfortunate that we can't just wipe and move everything to Int64, but this makes me a bit nervous.

lib/pos_epoch.go
```diff
@@ -40,7 +40,7 @@ func (epochEntry *EpochEntry) RawEncodeWithoutMetadata(blockHeight uint64, skipM
 	data = append(data, UintToBuf(epochEntry.InitialBlockHeight)...)
 	data = append(data, UintToBuf(epochEntry.InitialView)...)
 	data = append(data, UintToBuf(epochEntry.FinalBlockHeight)...)
-	data = append(data, UintToBuf(epochEntry.CreatedAtBlockTimestampNanoSecs)...)
+	data = append(data, IntToBuf(epochEntry.CreatedAtBlockTimestampNanoSecs)...)
```
lazynina, line 43: This is okay since this is all new stuff.

lib/pos_fee_estimator.go
```diff
@@ -126,7 +126,7 @@ func addBlockToTransactionRegister(txnRegister *TransactionRegister, block *MsgD
 		if txn.TxnMeta.GetTxnType() == TxnTypeBlockReward {
 			continue
 		}
-		mtxn, err := NewMempoolTx(txn, block.Header.TstampNanoSecs, block.Header.Height)
+		mtxn, err := NewMempoolTx(txn, NanoSecondsToUint64MicroSeconds(block.Header.TstampNanoSecs), block.Header.Height)
```
lazynina, line 129: Should we use Int64 for the microsecond tracking of mempool txs? this could be addressed separately.

lib/postgres.go
lazynina: Yeah this is fine.

lazynina (changes requested):

lib/network.go
tholonious: Header version is newly added with PoS. This decoding is safe! Or are you concerned with something else

lazynina: Thanks for clarifying! I forgot that this was new, so yeah this is good then!

lazynina (approved):

lib/pos_fee_estimator.go
tholonious: I'm open to leaving this as-is, switching to int64 microseconds, or to pass in the `Time` struct type. Happy to defer to @AeonSw4n on how he wants to move forward for the Mempool

AeonSw4n: I think let's just keep it as is for simplicity. It shouldn't change how transactions are ordered within the TransactionRegister instance.

AeonSw4n (approved):

PR #1025, opened 2024-02-13
_connectCoinLockup transfer restrictions

poolcoke:


lazynina (approved):

lib/block_view_lockups.go
```diff
@@ -1303,6 +1303,36 @@ func (bav *UtxoView) _connectCoinLockup(
...
+			return 0, 0, nil,
+				errors.Wrap(RuleErrorCoinLockupRestrictedToProfileOwner, "_connectCoinLockup")
+		}
+
+		// Check if the lockup respects transfer limited to DAO members only.
+		// Here, a "DAO member" is anyone who holds either unlocked or locked DAO coins associated with the profile.
+		if transferRestrictionStatus == TransferRestrictionStatusDAOMembersOnly {
+			receiverBalanceEntry :=
+				bav._getBalanceEntryForHODLerPKIDAndCreatorPKID(hodlerPKID, profilePKID, true)
+			if receiverBalanceEntry.BalanceNanos.IsZero() && lockedBalanceEntry.BalanceBaseUnits.IsZero() {
```
diamondhands0, line 1330: I think you need to check nil or isDeleted here.

poolcoke: In DBGetBalanceEntryForHODLerAndCreatorPKIDsWithTxn which is called from the _getBalanceEntryForHODLerPKIDAndCreatorPKID function, we return an empty balance entry with zero nanos as the balance. I believe this is the only place in the code where we do this pattern, but it's not necessary to check if it's nil here as a reuslt.

lazynina (approved):

PR #1044, opened 2024-02-15
Mark PoS Blocks as STORED If Block Timestamp Drift Check Fails

tholonious:


tholonious:
* `feature/proof-of-stake`
  * **PR #1044** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/1044?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/1044?utm_source=stack-comment).

lib/pos_blockchain.go
```diff
@@ -646,6 +633,19 @@ func (bc *Blockchain) validateAndIndexBlockPoS(block *MsgDeSoBlock) (*BlockNode,
 		return blockNode, nil
 	}
 
+	// Base case - Check if the block has already been stored and fails the timestamp drift check.
+	// If it fails the check, then we leave it as stored and return early.
+	if exists && blockNode.IsStored() {
+		// If the block is too far in the future, we leave it as STORED and return early.
+		failsTimestampDriftCheck, err := bc.isBlockTimestampTooFarInFuturePoS(block.Header)
```
tholonious, line 640: When check the timestamp drift here as a base case to prevent recursive validation calls on the block's ancestors. If the check fails, then we can leave the block as STORED and exit early

lib/pos_blockchain.go
```diff
@@ -831,11 +835,41 @@ func (bc *Blockchain) isBlockTimestampValidRelativeToParentPoS(header *MsgDeSoHe
...
 }
 
+// isBlockTimestampTooFarInFuturePoS validates that the block's timestamp is not too far in the future based
+// on the configured block timestamp drift.
+//
+// We use the snapshotted global params to validate that the block's timestamp isn't too far ahead in the
+// future. We use the snapshotted global params specifically so that the drift timestamp check behaves
+// consistently even for orphan blocks that are 1 epoch in the future..
+func (bc *Blockchain) isBlockTimestampTooFarInFuturePoS(header *MsgDeSoHeader) (bool, error) {
+	utxoView, err := NewUtxoView(bc.db, bc.params, nil, bc.snapshot, nil)
```
lazynina, line 845: Let's leave a comment on why using `NewUtxoView` is okay here.

lib/pos_blockchain.go
```diff
@@ -831,11 +835,41 @@ func (bc *Blockchain) isBlockTimestampValidRelativeToParentPoS(header *MsgDeSoHe
...
+	if err != nil {
+		return false, errors.Wrap(err, "isBlockTimestampTooFarInFuturePoS: Problem initializing UtxoView")
+	}
+
+	simulatedEpochEntryForBlock, err := utxoView.SimulateAdjacentEpochEntryForBlockHeight(header.Height)
```
lazynina, line 850: Ah okay this won't work if we're blocksyncing and we've got a block that's more than 1 epoch in the past. I think a very simple solution to this would be to simply check if the header's timestamp is less than the current timestamp at the top of the function. This way we don't even need to create a view or compute the snapshot at epoch number. Basically, if the block is in the past, we can just exit early and say that the block is certainly not too far in the future.

lazynina (changes requested):

lib/pos_blockchain.go
tholonious: Great suggestion. Done

lazynina (approved):

PR #1062, opened 2024-02-21
Remove Peer Disconnects in Server._handleBlock

tholonious:
Remove Peer Disconnects in Server._handleBlock

Warning for unrequested blocks

Catch case for spam prevention check

### DIFF
--- a/lib/server.go
+++ b/lib/server.go
@@ -2102,14 +2102,12 @@ func (srv *Server) _handleBlock(pp *Peer, blk *MsgDeSoBlock) {
 		return
 	}
 
-	if pp != nil {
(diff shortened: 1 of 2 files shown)


lib/server.go
```diff
@@ -2102,14 +2102,12 @@ func (srv *Server) _handleBlock(pp *Peer, blk *MsgDeSoBlock) {
 		return
 	}
 
-	if pp != nil {
```
lazynina, line 2105: Why don't we need this `pp != nil` check anymore?

tholonious: The peer can't be nil. Other parts of this code and the FastHotStuffConsensus.HandleBlock function would break if it was nil

PR #1086, opened 2024-03-04
Atomic Transactions Support

poolcoke:


lib/block_view_atomic_txns.go
```diff
@@ -0,0 +1,474 @@
...
+	// key in the ExtraData map to ensure that start of the chain is not compromised.
+	//
+	// The AtomicTxnsChainLength key is crucial for pinning the start of the atomic transaction. It's
+	// arbitrary and redundant that we use the chains length, but it adds an extra sanity check when
+	// connecting the transaction to the blockchain. Without a key representing the starting transaction,
+	// a malicious entity could reorder the transactions while still preserving the validity of the hashes
+	// in the circularly linked list. The AtomicTxnsChainLength included in the first transaction ensures
+	// the transactions are atomically executed in the order specified.
+	// NOTE: As a measure to reduce potential mempool divergences, only one transaction can have an
+	// AtomicTxnsChainLength key.
```
diamondhands0, line 64: This is an excellent comment. Great job codifying all the logic into an easy-to-understand comment.

diamondhands0: So did we get rid of the ability to do an "unordered" list of txns? I think it's probably good to kill it for now since it could open up potential attack vectors.

lib/block_view_atomic_txns.go
```diff
@@ -0,0 +1,474 @@
...
+
+	// Convert the transaction to bytes but do NOT encode the transaction signature.
+	preSignature := true
+	txBytes, err := msgDuplicate.ToBytes(preSignature)
+	if err != nil {
+		return nil, errors.Wrap(err, "MsgDeSoTxn.AtomicHash: cannot convert modified transaction to bytes")
+	}
+
+	// Return the SHA256 double hash of the resulting bytes.
+	return Sha256DoubleHash(txBytes), nil
```
diamondhands0, line 178: This seems like it should work.

lib/block_view_atomic_txns.go
```diff
@@ -0,0 +1,474 @@
...
+	blockTimestampNanoSecs int64,
+	verifySignature bool,
+) (
+	_totalInput uint64,
+	_totalOutput uint64,
+	_utxoOps []*UtxoOperation,
+	_err error,
+) {
+	var utxoOpsForTxn []*UtxoOperation
+
```
AeonSw4n, line 198: Should we cap how many transactions an atomic txns wrapper can contain?

lib/block_view_atomic_txns.go
```diff
@@ -0,0 +1,476 @@
...
+func (msg *MsgDeSoTxn) AtomicHash() (*BlockHash, error) {
+	// Create a duplicate of the transaction to ensure we don't edit the existing transaction.
+	msgDuplicate, err := msg.Copy()
+	if err != nil {
+		return nil, errors.Wrap(err, "MsgDeSoTxn.AtomicHash: Cannot create duplicate transaction")
+	}
+
+	// Sanity check that the transaction includes the necessary extra data to be included in an atomic transaction.
+	if !msgDuplicate.IsAtomicTxnsInnerTxn() {
```
diamondhands0, line 162: @poolcoke do we want to do this check in AtomicHash()? Feels like we should probably remove it. I could see us wanting to compute the atomic hash for a txn BEFORE we've set its previous/next hashes in transaction construction. Wdyt?

lib/block_view_atomic_txns.go
diamondhands0: Nah I think max txn size should cover it.

lib/block_view_atomic_txns.go
```diff
@@ -0,0 +1,476 @@
...
+	var innerUtxoOps [][]*UtxoOperation
+	var totalInput, totalOutput, totalFees uint64
+	for _, innerTxn := range txMeta.Txns {
+		// NOTE: By recursively calling _connectNonAtomicTransaction, each inner transaction is checked that
+		// it is capable of paying for its own fees as well as having a valid signature.
+		innerTxnUtxoOps, txnInput, txnOutput, txnFees, err := bav._connectNonAtomicTransaction(
+			innerTxn, txHash, blockHeight, blockTimestampNanoSecs, verifySignatures, ignoreUtxos)
+		if err != nil {
+			return nil, 0, 0, 0,
+				errors.Wrap(err, "_connectAtomicTxnsWrapper")
```
diamondhands0, line 248: Usually error should be descriptive and end in ": " when using Wrap. See other examples.

lib/block_view_atomic_txns.go
```diff
@@ -0,0 +1,476 @@
...
+		totalInnerTxnFees, err = SafeUint64().Add(totalInnerTxnFees, innerTxn.TxnFeeNanos)
+		if err != nil {
+			return RuleErrorAtomicTxnsWrapperHasInternalFeeOverflow
+		}
+	}
+	if txn.TxnFeeNanos != totalInnerTxnFees {
+		return RuleErrorAtomicTxnsWrapperMustHaveEqualFeeToInternalTxns
+	}
+
+	// Technically, the txn.TxnNonce field could be
```
diamondhands0, line 342: Finish comment

lib/block_view_atomic_txns.go
```diff
@@ -0,0 +1,476 @@
...
+	if txn.TxnNonce.ExpirationBlockHeight != 0 || txn.TxnNonce.PartialID != 0 {
+		return RuleErrorAtomicTxnsWrapperMustHaveZeroedNonce
+	}
+
+	// Since the wrapper is free and modifiable by anyone, we check to ensure the
+	// associated ExtraData is empty to prevent free storage on the blockchain.
```
diamondhands0, line 348: I'm not sure we want this. If we're charging for the wrapper it's fine and could be useful?

lib/block_view_atomic_txns.go
```diff
@@ -0,0 +1,476 @@
...
+		}
+
+		// Validate the inner transaction as meant to be included in an atomic transaction.
+		if !innerTxn.IsAtomicTxnsInnerTxn() {
+			return RuleErrorAtomicTxnsHasNonAtomicInnerTxn
+		}
+
+		// Validate the starting point of the atomic transactions chain.
+		_, keyExists := innerTxn.ExtraData[AtomicTxnsChainLength]
+		if keyExists && ii == 0 {
```
diamondhands0, line 376: Shouldn't it be !keyExists?

lib/block_view_atomic_txns.go
```diff
@@ -0,0 +1,476 @@
...
+		if currentIndex == 0 {
+			return chainLength - 1
+		}
+		return currentIndex - 1
+	}
+
+	// Validate the chain sequence specified.
+	for ii, innerTxn := range txnMeta.Txns {
+		// Check the next transaction.
+		if !reflect.DeepEqual(
```
diamondhands0, line 417: Prefer bytes.Equal

lib/block_view.go
```diff
@@ -3496,17 +3507,88 @@ func (bav *UtxoView) ValidateDiamondsAndGetNumDeSoNanos(
...
+	}
+
+	// Check that we're not trying to commit a transaction meant to be part of a series of atomic transactions
+	// outside an atomic transactions wrapper.
+	if txn.IsAtomicTxnsInnerTxn() {
+		return nil, 0, 0, 0, RuleErrorAtomicTxnsRequiresWrapper
+	}
+
+	// By here, we should know the transaction to be non-atomic.
+	return bav._connectNonAtomicTransaction(
```
diamondhands0, line 3568: I feel like we should have a better name for this. Right now you actually do use it to connect an atomic txn later on so it's not a great name. Maybe like _connectSingleTxn? Idk noodle on it.

lazynina: I like `_connectSingleTxn`

lib/block_view_atomic_txns.go
```diff
@@ -0,0 +1,472 @@
...
+	if txnSizeBytes != 0 && bav.GlobalParamsEntry.MinimumNetworkFeeNanosPerKB != 0 {
+		// Make sure there isn't overflow in the fee.
+		if txn.TxnFeeNanos != ((txn.TxnFeeNanos * 1000) / 1000) {
+			return nil, 0, 0, 0, RuleErrorOverflowDetectedInFeeRateCalculation
+		}
+		// If the fee is less than the minimum network fee per KB, return an error.
+		if (txn.TxnFeeNanos*1000)/uint64(txnSizeBytes) < bav.GlobalParamsEntry.MinimumNetworkFeeNanosPerKB {
+			return nil, 0, 0, 0, RuleErrorTxnFeeBelowNetworkMinimum
+		}
+	}
```
lazynina, line 279: This is a cheaper validation to do before we connect the inner txns.

lib/block_view_atomic_txns.go
```diff
@@ -0,0 +1,472 @@
...
+		// Validate the starting point of the atomic transactions chain.
+		_, keyExists := innerTxn.ExtraData[AtomicTxnsChainLength]
+		if !keyExists && ii == 0 {
+			return RuleErrorAtomicTxnsMustStartWithChainLength
+		}
+		if keyExists && ii > 0 {
+			return RuleErrorAtomicTxnsHasMoreThanOneStartPoint
+		}
+
+		// The error check in AtomicHash() is almost redundant, but we must keep it in the event
```
lazynina, line 379: well we actually do this to be more efficient when checking the atomic hashes below as well! This allows us to avoid calling `AtomicHash` an extra time below.

lib/block_view_atomic_txns.go
```diff
@@ -0,0 +1,472 @@
...
+		}
+		atomicHashes = append(atomicHashes, innerTxnAtomicHash)
+	}
+
+	// Construct special helper functions for circular doubly linked list indexing.
+	nextIndex := func(currentIndex int, chainLength int) int {
+		// Check for the special case of an atomic chain of length 1.
+		if chainLength == 1 {
+			return currentIndex
+		}
```
lazynina, line 394: Isn't this unnecessary really? If `chainLength == 1` then any (integer + 1) % 1 will be 0.

lib/block_view_atomic_txns.go
```diff
@@ -0,0 +1,472 @@
...
+	// There exists three design options for txn.TxnFeeNanos rules in atomic transaction wrappers:
+	//	(1) Force txn.TxnFeeNanos to equal zero.
+	//	(2) Force txn.TxnFeeNanos to equal the sum of the internal transaction's txn.TxnFeeNanos fields.
+	//	(3) Ignore txn.TxnFeeNanos entirely.
+	//
+	// Because txn.TxnFeeNanos gets used in several places for non-connection logic (e.g. BMF),
+	// it's important to use design option (2) to be consistent across core. This check as a result
+	// becomes extremely important in _connectAtomicTxnsWrapper().
+	var totalInnerTxnFees uint64
+	var err error
```
lazynina, line 327: Should we add a validation that the number of inner txns is non-zero? seems like a decent sanity check to me.

lib/block_view_atomic_txns.go
```diff
@@ -0,0 +1,472 @@
...
+		// Check for the special case of an atomic chain of length 1.
+		if chainLength == 1 {
+			return currentIndex
+		}
+
+		// Check for the wrap around case.
+		if currentIndex == 0 {
+			return chainLength - 1
+		}
+		return currentIndex - 1
```
lazynina, line 407: These also feel a little complicated. I don't love having functions defined in functions if we can help it as well. 
See this link for some simpler version of these functions [here](https://go.dev/play/p/a7EbQAHqGoo)

Basically, next index is always `(currentIndex + 1) % chainLength` and prev index is always `(currentIndex - 1 + chainLength) % chainLength`. We should use this anywhere else we have these special helper functions defined.

lib/block_view_atomic_txns.go
```diff
@@ -0,0 +1,472 @@
...
+		}
+
+		// The error check in AtomicHash() is almost redundant, but we must keep it in the event
+		// that the byte buffer for the Sha256 hash fails to allocate. This should almost never
+		// occur, and there's more serious issues if it does.
+		innerTxnAtomicHash, err := innerTxn.AtomicHash()
+		if err != nil {
+			return errors.Wrap(err, "_verifyAtomicTxnsChain")
+		}
+		atomicHashes = append(atomicHashes, innerTxnAtomicHash)
```
lazynina, line 386: Since we're going to be converting `innerTxnAtomicHash` to bytes below, should we make `atomicHashes` a 2D slice of bytes instead of a slice of block hashes?

lib/block_view_atomic_txns_test.go
```diff
@@ -0,0 +1,388 @@
...
+	// Now we test that the transactions are truly dependent on each-other by reorganizing them.
+	//
+
+	// Reorganize the transactions.
+	initialTxn := atomicTxns[0]
+	atomicTxns[0] = atomicTxns[len(atomicTxns)-1]
+	atomicTxns[0] = initialTxn
+
+	// Initialize test chain, miner, and testMeta for failing use.
+	testMetaFail := _setUpMinerAndTestMetaForAtomicTransactionTests(t)
```
lazynina, line 221: Why do we need a new chain and miner here? We should only ever need one for a given test. Running multiple really messes up memory usage on our CI pods.

lib/block_view_atomic_txns_test.go
```diff
@@ -0,0 +1,388 @@
...
+		}
+	}
+}
+
+//----------------------------------------------------------
+// (Testing) Atomic Transactions Setup Helper Functions
+//----------------------------------------------------------
+
+// The goal of _generateDependentAtomicTransactions is to generate
+// a sequence of transactions who CANNOT be reordered meaning they
```
lazynina, line 249:
```suggestion
// a sequence of transactions which CANNOT be reordered meaning they
```

lib/block_view_atomic_txns_test.go
```diff
@@ -0,0 +1,388 @@
...
+// must be executed in the sequence returned. This mean transaction
```
lazynina, line 250:
```suggestion
// must be executed in the sequence returned. This means the transaction
```

lib/block_view_atomic_txns_test.go
```diff
@@ -0,0 +1,388 @@
...
+// How can we generate "dependent atomic transactions" algorithmically using
+// TestMeta initialized with _setUpUsersForAtomicTransactionTesting?
+//
+//	(1) Choose an arbitrary starter public key with DESO (m0PkBytes)
+//	(2) For ii in [0, numberOfTransactions):
+//		(2a) Generate a new public/private key pair (pub_ii, priv_ii)
+//		(2b) Have pub_(ii-1) do a max DESO transfer to pub_ii. Use m0PkBytes as pub_(-1).
+//	(3) Have pub_numberOfTransactions perform a max transfer back to m0PkBytes
+//
+// Notice that because pub_ii only has DESO at following the transaction with
```
lazynina, line 263:
```suggestion
// Notice that because pub_ii only has DESO following the transaction at
```

lib/block_view_atomic_txns_test.go
```diff
@@ -0,0 +1,388 @@
...
+	var receiverPublicKeysBase58 []string
+	var receiverPrivateKeysBase58 []string
+	var receiverBalancesNanos []uint64
+
+	// Get the initial balance of m0.
+	m0InitialBalanceNanos := _getBalance(
+		testMeta.t, testMeta.chain, testMeta.mempool, m0Pub)
+
+	// Generate the atomic transactions.
+	for ii := 0; ii < numberOfTransactions-1; ii++ {
```
lazynina, line 284: Why `numberOfTransactions - 1`? This means we'll have `numberOfTransactions - 1` in the list of innerTxns, which seems incorrect.

lib/block_view_types.go
```diff
@@ -682,7 +682,8 @@ const (
 	OperationTypeStakeDistributionRestake      OperationType = 49
 	OperationTypeStakeDistributionPayToBalance OperationType = 50
 	OperationTypeSetValidatorLastActiveAtEpoch OperationType = 51
-	// NEXT_TAG = 52
+	OperationTypeAtomicTxnsWrapper             OperationType = 52
```
lazynina, line 685: Please update `func (op OperationType) String() string {` which starts on line 689.

lib/block_view_types.go
```diff
@@ -682,7 +682,8 @@ const (
...
+	// NEXT_TAG = 53
```
lazynina, line 686: This also reminds me, please update `GetEncoderForTxType` to return some value for `TxnTypeAtomicTxnsWrapper`. Without this, state syncer will panic and crash on the first atomic txn it sees. It can simply be an empty struct for now - although I'd probably just make it the following struct:
```
type AtomicTxnsWrapper struct {
  InnerTxnsTransactionMetadata []TransactionMetadata 
}
```
This way we can compute txindex metadata properly for it and use it in state syncer. And then a simple update to `ComputeTransactionMetadata` that calls itself for each inner txn would get you want you want.

lib/blockchain.go
```diff
@@ -5747,3 +5747,126 @@ func (bc *Blockchain) CreateCoinUnlockTxn(
...
+		if chainLength == 1 {
+			return currentIndex
+		}
+
+		// Check for the wrap around case.
+		if currentIndex == 0 {
+			return chainLength - 1
+		}
+		return currentIndex - 1
+	}
```
lazynina, line 5820: Update these based on my suggestions from [this link](https://go.dev/play/p/a7EbQAHqGoo)

lib/blockchain.go
```diff
@@ -5747,3 +5747,126 @@ func (bc *Blockchain) CreateCoinUnlockTxn(
...
+
+		// Track the total fees paid.
+		totalFees, err = SafeUint64().Add(totalFees, txnDuplicate.TxnFeeNanos)
+		if err != nil {
+			return nil, 0, errors.Wrapf(err, "CreateAtomicTxnsWrapper: total fee overflow")
+		}
+	}
+
+	// Set the starting point for the atomic transactions.
+	chainedUnsignedTransactions[0].ExtraData[AtomicTxnsChainLength] = UintToBuf(uint64(len(unsignedTransactions)))
```
lazynina, line 5858: move me to line 5821

lib/constants.go
```diff
@@ -1560,6 +1560,11 @@ const (
 	DiamondLevelKey    = "DiamondLevel"
 	DiamondPostHashKey = "DiamondPostHash"
 
+	// Atomic Transaction Keys
+	AtomicTxnsChainLength    = "AtomicTxnsChainLength"
+	NextAtomicTxnPreHash     = "NextAtomicTxnPreHash"
+	PreviousAtomicTxnPreHash = "PreviousAtomicTxnPreHash"
+
```
lazynina, line 1567: If we can shorten these constants while still having them be meaningful, that would be best. We enlarge the size each transaction w/ each of these, so the fewer bytes the better.

lazynina (changes requested):

lib/block_view_atomic_txns.go
```diff
@@ -0,0 +1,472 @@
...
+		return nil, errors.Wrap(err, "MsgDeSoTxn.AtomicHash: Cannot create duplicate transaction")
+	}
+
+	// Delete the NextAtomicTxnPreHash and PreviousAtomicTxnPreHash from the ExtraData map.
+	delete(msgDuplicate.ExtraData, NextAtomicTxnPreHash)
+	delete(msgDuplicate.ExtraData, PreviousAtomicTxnPreHash)
+
+	// Convert the transaction to bytes but do NOT encode the transaction signature.
+	preSignature := true
+	txBytes, err := msgDuplicate.ToBytes(preSignature)
```
lazynina, line 166:
```suggestion
	txBytes, err := msgDuplicate.ToBytes(true)
```

lib/block_view_atomic_txns.go
```diff
@@ -0,0 +1,557 @@
...
+
+func (txnData *AtomicTxnsWrapperMetadata) New() DeSoTxnMetadata {
+	return &AtomicTxnsWrapperMetadata{}
+}
+
+//
+// HELPER FUNCTIONS: MsgDeSoTxn
+//
+
+// IsAtomicTxnsInnerTxn is used to determine if a MsgDeSoTxn is an inner tnx
```
lazynina, line 141:
```suggestion
// IsAtomicTxnsInnerTxn is used to determine if a MsgDeSoTxn is an inner txn
```

lib/block_view_types.go
```diff
@@ -1043,6 +1049,21 @@ type UtxoOperation struct {
...
 	LockedAtEpochNumber uint64
+
+	// AtomicTxnInnerUtxoOps maintains a 2D slice of all UtxoOps collected from transactions
+	// who were executed atomically. The 2D array allows us to easily disconnect transactions
+	// who are part of an atomic transaction as we hold each of their UtxoOps separately.
+	//
+	// NOTE: While it may seem erroneous to have a field within the UtxoOperation struct of
+	// type UtxoOperation, this is valid because the size of the pointer is always known at
+	// compile time. Hence, there's no circular dependency as is the case if we were to use
+	// [][]UtoOperation for this field instead. This could equivalently be a 2D array of
```
lazynina, line 1060:
```suggestion
	// [][]UtxoOperation for this field instead. This could equivalently be a 2D array of
```

lib/constants.go
```diff
@@ -1560,6 +1560,11 @@ const (
 	DiamondLevelKey    = "DiamondLevel"
 	DiamondPostHashKey = "DiamondPostHash"
 
+	// Atomic Transaction Keys
+	AtomicTxnsChainLength    = "AtmcChnLen"
+	NextAtomicTxnPreHash     = "NxtAtmcHsh"
+	PreviousAtomicTxnPreHash = "PrvAtmcHsh"
```
lazynina, line 1566: ❤️

lib/block_view_atomic_txns.go
```diff
@@ -0,0 +1,557 @@
...
+		}
+
+		// Read the transaction metadata into the txnBytes memory buffer.
+		if _, err = io.ReadFull(rr, txnMetadataBytes); err != nil {
+			return errors.Wrap(err,
+				"AtomicTxnsWrapperTxindexMetadata.RawDecodeWithoutMetadata: Problem reading bytes for transaction")
+		}
+
+		// Convert the txnBytes buffer to a TransactionMetadata struct.
+		if err = txindexMetadata.InnerTxnsTransactionMetadata[ii].RawDecodeWithoutMetadata(blockHeight, rr); err != nil {
```
lazynina, line 543: This looks wrong. You're reading past the bytes you want on line 537. I'll leave my suggested changes in another comment.

lib/block_view_atomic_txns.go
```diff
@@ -0,0 +1,557 @@
...
+
+func (txindexMetadata *AtomicTxnsWrapperTxindexMetadata) RawEncodeWithoutMetadata(
+	blockHeight uint64,
+	skipMetadata ...bool,
+) []byte {
+	var data []byte
+	data = append(data, UintToBuf(uint64(len(txindexMetadata.InnerTxnsTransactionMetadata)))...)
+	for _, innerMetadata := range txindexMetadata.InnerTxnsTransactionMetadata {
+		txnBytes := innerMetadata.RawEncodeWithoutMetadata(blockHeight, skipMetadata...)
+		data = append(data, UintToBuf(uint64(len(txnBytes)))...)
```
lazynina, line 493:
```suggestion
```
we don't need the number of bytes - the RawEncodeWithoutMetadata and RawDecodeWithoutMetadata functions will read the exact number of bytes required.

lib/block_view_atomic_txns.go
```diff
@@ -0,0 +1,557 @@
...
+				"AtomicTxnsWrapperTxindexMetadata.RawDecodeWithoutMetadata: "+
+					"Problem allocating bytes for transaction")
+		}
+
+		// Read the transaction metadata into the txnBytes memory buffer.
+		if _, err = io.ReadFull(rr, txnMetadataBytes); err != nil {
+			return errors.Wrap(err,
+				"AtomicTxnsWrapperTxindexMetadata.RawDecodeWithoutMetadata: Problem reading bytes for transaction")
+		}
+
```
lazynina, line 541:
```suggestion

```
we don't need any of this num bytes stuff!

lib/block_view_atomic_txns_test.go
```diff
@@ -0,0 +1,667 @@
...
+	// Connect the transactions to ensure they can actually be connected.
+	for _, txn := range atomicTxns {
+		// Connect the transaction.
+		txHash := txn.Hash()
+		_, _, _, _, err := utxoView.ConnectTransaction(
+			txn, txHash, blockHeight, 0, true, false)
+		if err != nil {
+			require.Contains(t, err.Error(), RuleErrorInsufficientBalance)
+		}
+	}
```
lazynina, line 405: can we add a call to `_executeAllTestRollbackAndFlush` to make sure all connect and disconnect stuff behaves as expected please?

lib/block_view_atomic_txns_test.go
lazynina: ah okay it's the max transfer at the end.

lazynina (changes requested):

lib/block_view_atomic_txns_test.go
```diff
@@ -0,0 +1,667 @@
...
+		// Connect the transaction.
+		txHash := txn.Hash()
+		_, _, _, fees, err := utxoView.ConnectTransaction(
+			txn, txHash, blockHeight, 0, true, false)
+		require.NoError(t, err)
+		totalFees += fees
+	}
+
+	// Flush the view to ensure everything is working properly.
+	require.NoError(t, utxoView.FlushToDb(uint64(blockHeight)))
```
lazynina, line 370: Let execute all test rollback and flush handle the flushing for you.
```suggestion
```

lib/block_view_atomic_txns_test.go
```diff
@@ -0,0 +1,667 @@
...
+
+	// Get the final balance for m0.
+	m0FinalBalanceNanos := _getBalance(t, testMeta.chain, testMeta.mempool, m0Pub)
```
lazynina, line 373:
```suggestion
	m0FinalBalanceNanos := _getBalanceWithView(t, testMeta.chain, utxoView, m0Pub)
```
here are my suggestions on getting test to pass w/ execute all test rollback and flush.

lib/block_view_atomic_txns_test.go
lazynina, line 405:
```suggestion
	}
	_executeAllTestRollbackAndFlush(testMeta)
```

lazynina (approved):
Please wait for @tholonious before merging

lib/block_view_atomic_txns.go
```diff
@@ -0,0 +1,532 @@
...
+	_utxoOps []*UtxoOperation,
+	_totalInput uint64,
+	_totalOutput uint64,
+	_fees uint64,
+	_err error,
+) {
+	var utxoOpsForTxn []*UtxoOperation
+
+	// Validate the connecting block height.
+	if blockHeight < bav.Params.ForkHeights.ProofOfStake1StateSetupBlockHeight {
```
tholonious, line 207: If atomic txns are gated by `ProofOfStake1StateSetupBlockHeight`, then they need to be supported by the old PoW mempool. @poolcoke, @lazynina - has that been regtested yet? NBD if not. We'll just need to remember to regtest it.

tholonious (approved):

PR #1089, opened 2024-03-06
PoS Hypersync Archival Mode

AeonSw4n:


AeonSw4n:
* **#1089** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/1089?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `feature/pos-hypersync-updates`

<h2></h2>

 Join @AeonSw4n and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

lib/pos_blockchain.go
```diff
@@ -537,6 +547,29 @@ func (bc *Blockchain) processOrphanBlockPoS(block *MsgDeSoBlock) error {
...
+	// If the blockNode doesn't exist, or the block is not committed, or it's already stored, then we're not dealing
+	// with an archival block. Archival blocks must have an existing blockNode, be committed, and not be stored.
+	if !exists || !blockNode.IsCommitted() || blockNode.IsStored() {
+		return false, nil
+	}
+
+	// If we get to this point, we're dealing with an archival block, so we'll attempt to store it.
+	_, err = bc.storeBlockInBlockIndex(block)
+	if err != nil {
+		return true, errors.Wrap(err, "checkAndStoreArchivalBlock: Problem storing block in block index")
```
tholonious, line 568: Shouldn't the success flag be false here?
```suggestion
		return false, errors.Wrap(err, "checkAndStoreArchivalBlock: Problem storing block in block index")
```

lib/pos_blockchain.go
```diff
@@ -537,6 +547,29 @@ func (bc *Blockchain) processOrphanBlockPoS(block *MsgDeSoBlock) error {
...
+	}
+	blockNode, exists := bc.blockIndexByHash[*blockHash]
+	// If the blockNode doesn't exist, or the block is not committed, or it's already stored, then we're not dealing
+	// with an archival block. Archival blocks must have an existing blockNode, be committed, and not be stored.
+	if !exists || !blockNode.IsCommitted() || blockNode.IsStored() {
+		return false, nil
+	}
+
+	// If we get to this point, we're dealing with an archival block, so we'll attempt to store it.
+	_, err = bc.storeBlockInBlockIndex(block)
```
lazynina, line 566: Can we add a comment that we expect the block node to already be marked as COMMITTED and VALIDATED?

lazynina (approved):

lib/pos_blockchain.go
```diff
@@ -537,6 +547,29 @@ func (bc *Blockchain) processOrphanBlockPoS(block *MsgDeSoBlock) error {
 	return errors.Wrap(err, "processBlockPoS: Problem adding block to block index: ")
 }
 
+// checkAndStoreArchivalBlock is a helper function that takes in a block and checks if it's an archival block.
+// If it is, it stores the block in the db and returns true. If it's not, it returns false, or false and an error.
+func (bc *Blockchain) checkAndStoreArchivalBlock(block *MsgDeSoBlock) (_success bool, _err error) {
```
tholonious, line 552: Just as a simplification, you make this work without the extra _success return value. The _err return value should be sufficient:
- if the block was successfully stored as an archival block, then _err is nil
- if the block doesn't pass the archival block conditions or there was an error, then _err is non-nil
```suggestion
func (bc *Blockchain) storeArchivalBlock(block *MsgDeSoBlock) (_err error) {
```

lib/pos_blockchain.go
```diff
@@ -537,6 +547,29 @@ func (bc *Blockchain) processOrphanBlockPoS(block *MsgDeSoBlock) error {
...
+	blockHash, err := block.Hash()
+	if err != nil {
+		return false, errors.Wrap(err, "checkAndStoreArchivalBlock: Problem hashing block")
+	}
+	blockNode, exists := bc.blockIndexByHash[*blockHash]
+	// If the blockNode doesn't exist, or the block is not committed, or it's already stored, then we're not dealing
+	// with an archival block. Archival blocks must have an existing blockNode, be committed, and not be stored.
+	if !exists || !blockNode.IsCommitted() || blockNode.IsStored() {
+		return false, nil
+	}
```
tholonious, line 563:
```suggestion
	if !exists || !blockNode.IsCommitted() {
		return false, errors.New("checkAndStoreArchivalBlock: block is not committed")
	}
	
	// If the block is already stored, then there's nothing left to do.
	if blockNode.isStored() {
		return true, nil
	}
```

tholonious (changes requested):

lib/pos_blockchain.go
AeonSw4n: I feel like returning an error here could be problematic. If a blocknode exists, and a block is not committed, I think we should continue with the remainder of `processBlockPoS`. Instead, if we error here, we would break out of that function, which seems undesirable.

PR #1095, opened 2024-03-07
Fix Mempool and Fee Estimator Past Blocks Refresh on Block Connect

tholonious:


lib/pos_fee_estimator.go
```diff
@@ -94,27 +104,35 @@ func (posFeeEstimator *PoSFeeEstimator) Init(
...
-		return errors.Wrap(err, "PoSFeeEstimator.addBlockNoLock: error adding block to pastBlocksTransactionRegister")
-	}
-	posFeeEstimator.cachedBlocks = append(posFeeEstimator.cachedBlocks, block)
-	// Sort the cached blocks by height & tstamp just to be safe.
-	posFeeEstimator.sortCachedBlocks()
-	if uint64(len(posFeeEstimator.cachedBlocks)) > posFeeEstimator.numPastBlocks {
-		// Remove the oldest block.
-		if err := posFeeEstimator.removeBlockNoLock(posFeeEstimator.cachedBlocks[0]); err != nil {
-			return errors.Wrap(err, "PoSFeeEstimator.addBlockNoLock: error removing block from PoSFeeEstimator")
+	for _, pastBlock := range newPastBlocks {
```
tholonious, line 126: Rather than trying to incrementally manage the `pastBlocksTransactionRegister` as new blocks come in, it ends up being less error prone but just as performant to nuke it and recreate a new one from scratch. The number of past blocks will be small (<= 50) so this is not an issue.

lib/pos_fee_estimator.go
```diff
@@ -10,6 +10,9 @@ import (
 )
 
 type PoSFeeEstimator struct {
+	// globalParams is the latest GlobalParams used by the PoSFeeEstimator. The fee estimation uses
+	// this whenever the cached blocks are refreshed
+	globalParams *GlobalParamsEntry
```
tholonious, line 15: Store the globalParams now, so we can move the feeEstimator's params over into GlobalParams in a later PR

lib/pos_fee_estimator.go
```diff
@@ -194,19 +210,44 @@ func (posFeeEstimator *PoSFeeEstimator) removeBlockNoLock(block *MsgDeSoBlock) e
...
+	sortedBlocks := posFeeEstimator.sortBlocksByBlockHeightAndTstamp(dedupedBlocks)
+	return posFeeEstimator.pruneBlocksToMaxNumPastBlocks(sortedBlocks)
+}
+
+// dedupeBlocksByBlockHeight deduplicates the blocks by block height. If multiple blocks have the same
+// height, it keeps the one with the highest view.
+func (posFeeEstimator *PoSFeeEstimator) dedupeBlocksByBlockHeight(blocks []*MsgDeSoBlock) []*MsgDeSoBlock {
+	blocksByBlockHeight := make(map[uint64]*MsgDeSoBlock)
+	for _, block := range blocks {
+		existingBlock, hasExitingBlock := blocksByBlockHeight[block.Header.Height]
```
lazynina, line 240:
```suggestion
		existingBlock, hasExistingBlock := blocksByBlockHeight[block.Header.Height]
```

lib/pos_fee_estimator.go
```diff
@@ -194,19 +210,44 @@ func (posFeeEstimator *PoSFeeEstimator) removeBlockNoLock(block *MsgDeSoBlock) e
...
+		if !hasExitingBlock || existingBlock.Header.GetView() < block.Header.GetView() {
```
lazynina, line 241:
```suggestion
		if !hasExistingBlock || existingBlock.Header.GetView() < block.Header.GetView() {
```

lib/pos_fee_estimator.go
```diff
@@ -194,19 +210,44 @@ func (posFeeEstimator *PoSFeeEstimator) removeBlockNoLock(block *MsgDeSoBlock) e
...
+			blocksByBlockHeight[block.Header.Height] = block
+		}
+	}
+	return collections.MapValues(blocksByBlockHeight)
```
lazynina, line 245: Should cached blocks be a map instead of a slice? Is there any downside to using a map instead? I guess the prune operation becomes less efficient, but we could probably keep track of the min block height to make that operation more efficient.

lib/pos_fee_estimator.go
```diff
@@ -194,19 +210,44 @@ func (posFeeEstimator *PoSFeeEstimator) removeBlockNoLock(block *MsgDeSoBlock) e
...
+}
+
+// sortBlocksByBlockHeightAndTstamp sorts the blocks by height & tstamp just to be safe.
+func (posFeeEstimator *PoSFeeEstimator) sortBlocksByBlockHeightAndTstamp(blocks []*MsgDeSoBlock) []*MsgDeSoBlock {
+	return collections.SortStable(blocks,
 		func(ii, jj *MsgDeSoBlock) bool {
 			if ii.Header.Height != jj.Header.Height {
```
lazynina, line 252: With these updates I think we should no longer have blocks at the same height?

lib/pos_fee_estimator.go
tholonious: There's no practical difference. The number of cached blocks will always be small < 50, so time complexity is no issue. I have no preference between keeping it as a map vs slice

lib/pos_fee_estimator.go
tholonious: Yes, you're right. We don't need the tie-breaking anymore. Maybe it's safer to keep it regardless? This is a pretty robust sorting function that will work if the the deduping functionality elsewhere in this file gets changed

lazynina (approved):

PR #1098, opened 2024-03-11
PoS MempoolTx Validated status

AeonSw4n:

PR #1090, opened 2024-03-06
PoS Mempool Simplified AddTransaction Validation

AeonSw4n:

PR #1100, opened 2024-03-11
PoS Mempool Updated Refresh

AeonSw4n:


AeonSw4n:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/deso-protocol/core/1100?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1101** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/1101?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1100** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/1100?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#1099** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/1099?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1098** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/1098?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1090** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/1090?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `feature/pos-mempool-updates-and-transaction-exchange`

<h2></h2>

 Join @AeonSw4n and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

lib/pos_mempool.go
```diff
@@ -151,6 +151,8 @@ type PosMempool struct {
 	// based off the current state of the mempool and the most n recent blocks.
 	feeEstimator *PoSFeeEstimator
 
+	maxValidationViewConnects uint64
```
lazynina, line 154: Please add a comment above this that explains what `maxValidationViewConnects` does.

lib/pos_mempool.go
```diff
@@ -716,56 +720,43 @@ func (mp *PosMempool) GetIterator() MempoolIterator {
...
-		mp.feeEstimator.numPastBlocks,
-		mp.augmentedBlockViewRefreshIntervalMillis,
-	)
+	var txns []*MsgDeSoTxn
+	var txHashes []*BlockHash
+	for _, txn := range mempoolTxns {
+		txns = append(txns, txn.Tx)
+		txHashes = append(txHashes, txn.Hash)
+	}
+	copyValidationView, err := validationView.CopyUtxoView()
```
lazynina, line 738: Tests are failing because the validationView is nil. When would the validation view be nil? We should have a check in here somewhere that the view is nil or patch where it allows the view to be nil.

lib/pos_mempool.go
```diff
@@ -834,9 +825,6 @@ func (mp *PosMempool) UpdateGlobalParams(globalParams *GlobalParamsEntry) {
 	}
 
 	mp.globalParams = globalParams
-	if err := mp.refreshNoLock(); err != nil {
```
lazynina, line 837: I think we need a refresh here. @tholonious - I think we were relying on a call to `UpdateGlobalParams` in the `OnBlockConnected` event to remove txns that were in a block from the mempool.

lazynina (changes requested):

lib/pos_mempool.go
AeonSw4n: I've noticed there's also a separate method in PosMempool called `OnBlockConnected` which is doing something similar to what you're describing. Maybe that method is sufficient?

Also, interestingly, I think Refresh will eventually remove the stale transactions anyway since they'll fail to connect if they were already included in the chain.

lib/pos_mempool.go
AeonSw4n: Good observation. Added the check. I'll also fix the tests in a later PR.

PR #1102, opened 2024-03-11
PoS Block Producer updates

AeonSw4n:


lib/pos_blockchain_test.go
```diff
@@ -1891,8 +1870,6 @@ func testProcessBlockPoS(t *testing.T, testMeta *TestMeta) {
 		var malformedOrphanBlock *MsgDeSoBlock
 		malformedOrphanBlock = _generateRealBlock(testMeta, 18, 18, 9273, testMeta.chain.BlockTip().Hash, false)
 		malformedOrphanBlock.Header.PrevBlockHash = randomHash
-		// Modify anything to make the block malformed, but make sure a hash can still be generated.
-		malformedOrphanBlock.Header.TxnConnectStatusByIndexHash = randomHash
```
lazynina, line 1895: We still need to modify something to make sure the block is malformed such that the test passes.

lib/pos_blockchain_test.go
```diff
@@ -1909,7 +1886,6 @@ func testProcessBlockPoS(t *testing.T, testMeta *TestMeta) {
 		require.True(t, malformedOrphanBlockInIndex.IsStored())
 
 		// If a block can't be hashed, we expect to get an error.
-		malformedOrphanBlock.Header.TxnConnectStatusByIndexHash = nil
```
lazynina, line 1912: Modify something else such that we can't hash the block please.

lib/state_change_syncer.go
```diff
@@ -750,23 +750,6 @@ func (stateChangeSyncer *StateChangeSyncer) SyncMempoolToStateSyncer(server *Ser
 			// If the transaction successfully connected, we update mempoolTxUtxoView to the copied view.
 			if err == nil {
 				mempoolTxUtxoView = copiedView
-			} else {
```
lazynina, line 753: If err is not nil, we should glog an error maybe?

lazynina (changes requested):

lib/state_change_syncer.go
AeonSw4n: Agreed. However, I think we should only glog here, similarly to the block producer, which skips over failing transactions now. But I'm not fully familiar with the use of this function, so I may be wrong here.

PR #1114, opened 2024-03-14
Review of POS mempool (remove _connectFailingTxn)

diamondhands0:


lib/block_view.go
```diff
@@ -3920,6 +3920,89 @@ func (bav *UtxoView) _connectTransaction(
...
+		if err != nil && ignoreFailing {
+			glog.V(2).Infof("ConnectTransactions: Ignoring failing txn %d: %v", ii, err)
+			updateValues(nil, 0, 0, 0, false)
+			continue
+		} else if err != nil {
+			return nil, nil, nil, nil, nil,
+				errors.Wrapf(err, "ConnectTransactions: Problem connecting txn %d on copy view", ii)
+		}
+
+		utxoOpsForTxn, totalInput, totalOutput, fee, err = bav.ConnectTransaction(
```
diamondhands0, line 3985: Connecting the txn twice is definitely going to slow us down. I almost wish there was a way to do this in such a way that we just update the pointer to point to the copiedView.

Also I'm not 100% sure we need this _connectTransactions primitive on the view vs doing this logic in the mempool one time but will keep reading.

lib/block_view.go
```diff
@@ -3920,6 +3920,89 @@ func (bav *UtxoView) _connectTransaction(
...
+			txn, txHashes[ii], blockHeight, blockTimestampNanoSecs, verifySignatures, ignoreUtxos)
+		if err != nil {
+			return nil, nil, nil, nil, nil,
+				errors.Wrapf(err, "ConnectTransactions: Problem connecting txn %d", ii)
+		}
+		updateValues(utxoOpsForTxn, totalInput, totalOutput, fee, true)
+
+		if totalConnectedTxns == 0 {
```
diamondhands0, line 3993: I don't see how we would ever get past this line since totalConnectedTxns isn't incremented until AFTER this line. Could this be a bug?

lib/block_view.go
```diff
@@ -3952,104 +4035,6 @@ func (bav *UtxoView) ValidateTransactionNonce(txn *MsgDeSoTxn, blockHeight uint6
 	return nil
 }
```
diamondhands0, line 4036: Love this deletion.

lib/block_view.go
```diff
@@ -3920,6 +3920,89 @@ func (bav *UtxoView) _connectTransaction(
 	return utxoOpsForTxn, totalInput, totalOutput, fees, nil
 }
 
+func (bav *UtxoView) ConnectTransactions(
```
diamondhands0, line 3923: I wish there was a comment here explaining why we need this. You obviously added it for a reason and it would be super helpful to have that reason here so that when I review I can internalize what drove the addition quickly instead of having to figure it out. poolcoke has recently internalized this advice and started writing extremely long comments on everything. I love this and would like to see more of it from everyone. See here as an example all throughout this file:
* https://github.com/deso-protocol/core/blob/3dc58f301235dfc663577c8182b1f85aa0bf0f95/lib/block_view_lockups.go#L114

lib/handshake_manager.go
```diff
@@ -1,168 +0,0 @@
-package lib
```
diamondhands0, line 1: Love this deletion. I assume the logic is now embedded in one of the other parts of the code but will keep reading.

lib/network.go
```diff
@@ -2797,29 +2770,12 @@ func (msg *MsgDeSoBlock) EncodeBlockVersion1(preSignature bool) ([]byte, error)
 	return data, nil
 }
```
diamondhands0, line 2772: All of these deletions are amazing. There was so much complexity driven by the changing block header that is now no longer needed. I really love it.

lib/network_manager.go
```diff
@@ -77,18 +98,25 @@ type NetworkManager struct {
...
-		rnManager:                             rnManager,
+		keystore:                              blsKeystore,
 		AddrMgr:                               addrMgr,
+		minTxFeeRateNanosPerKB:                minTxFeeRateNanosPerKB,
+		nodeServices:                          nodeServices,
+		AllRemoteNodes:                        collections.NewConcurrentMap[RemoteNodeId, *RemoteNode](),
+		ValidatorIndex:                        collections.NewConcurrentMap[bls.SerializedPublicKey, *RemoteNode](),
+		NonValidatorOutboundIndex:             collections.NewConcurrentMap[RemoteNodeId, *RemoteNode](),
+		NonValidatorInboundIndex:              collections.NewConcurrentMap[RemoteNodeId, *RemoteNode](),
+		usedNonces:                            lru.NewCache(1000),
```
diamondhands0, line 119: I assume this is to make it so that you can recognize when a validator is reusing a nonce? If yes I would bump it to a higher number like 100k since we could have a lot of validator connections coming in and a lot of nonces to track.

lib/network_manager.go
```diff
@@ -214,16 +238,79 @@ func (nm *NetworkManager) startRemoteNodeCleanup() {
 // ## Handlers (Peer, DeSoMessage)
 // ###########################
 
-// _handleVersionMessage is called when a new version message is received. It is a wrapper around the handshake's
-// handleVersionMessage function.
+// _handleVersionMessage is called when a new version message is received.
 func (nm *NetworkManager) _handleVersionMessage(origin *Peer, desoMsg DeSoMessage) {
-	nm.handshake.handleVersionMessage(origin, desoMsg)
+	if desoMsg.GetMsgType() != MsgTypeVersion {
```
diamondhands0, line 243: Ok cool I see this is what you did with the handshake manager. I love this consolidation.

lib/peer.go
```diff
@@ -180,7 +180,7 @@ func (pp *Peer) HandleGetTransactionsMsg(getTxnMsg *MsgDeSoGetTransactions) {
 		// If the transaction isn't in the pool, just continue without adding
 		// it. It is generally OK to respond with only a subset of the transactions
 		// that were requested.
-		if mempoolTx == nil {
+		if mempoolTx == nil || !mempoolTx.IsValidated() {
```
diamondhands0, line 183: Nice. I think this should prevent us from sending unvalidated txns on the first sync which is good.

lib/pos_block_producer.go
```diff
@@ -197,68 +191,34 @@ func (pbp *PosBlockProducer) getBlockTransactions(
...
 		if err != nil {
-			return nil, nil, 0, errors.Wrapf(err, "Error copying UtxoView: ")
+			return nil, 0, errors.Wrapf(err, "Error copying UtxoView: ")
 		}
 		_, _, _, fees, err := blockUtxoViewCopy._connectTransaction(
 			txn.GetTxn(), txn.Hash(), uint32(newBlockHeight), newBlockTimestampNanoSecs,
 			true, false)
 
 		// Check if the transaction connected.
-		if err == nil {
```
diamondhands0, line 207: Love this change. Just skipping txns that don't connect. Easy.

lib/pos_mempool.go
```diff
@@ -149,6 +154,11 @@ type PosMempool struct {
 	// based off the current state of the mempool and the most n recent blocks.
 	feeEstimator *PoSFeeEstimator
 
+	maxValidationViewConnects uint64
```
diamondhands0, line 157: Here again I wish you had super detailed comments that explained what these control and all of the design decisions around them.

lib/pos_mempool.go
```diff
@@ -256,7 +269,6 @@ func (mp *PosMempool) Start() error {
 	// Create the transaction register, the ledger, and the nonce tracker,
 	mp.txnRegister = NewTransactionRegister()
 	mp.txnRegister.Init(mp.globalParams)
-	mp.ledger = NewBalanceLedger()
 	mp.nonceTracker = NewNonceTracker()
```
diamondhands0, line 272: I was thinking maybe you don't need the nonce tracker anymore but I think you actually do need it to power replace-by-fee for txns that are beyond the 10k limit of our validation window. So good to keep it.

lib/pos_mempool.go
```diff
@@ -721,61 +726,45 @@ func (mp *PosMempool) GetIterator() MempoolIterator {
...
-// the main mempool. The temp mempool will have the most up-to-date readOnlyLatestBlockView, Height, and globalParams. Any
-// transaction that fails to add to the temp mempool will be removed from the main mempool.
-func (mp *PosMempool) Refresh() error {
-	mp.Lock()
-	defer mp.Unlock()
-
+// Any transaction that fails to add to the temp mempool will be removed from the main mempool.
+func (mp *PosMempool) validateTransactions() error {
+	mp.RLock()
 	if !mp.IsRunning() {
```
diamondhands0, line 732: Normally when you have more than one line in a Lock/RLock we like to break it up into a function like:

func getThings() {
    mp.RLock()
    defer mp.RLock()

    validationView := mp.readOnlyLatestBlockView
	mempoolTxns := mp.getTransactionsNoLock()

    return validationView, mempoolTxns
}

diamondhands0: This prevents disasters as people modify the code in the future.

lib/pos_mempool.go
```diff
@@ -721,61 +726,45 @@ func (mp *PosMempool) GetIterator() MempoolIterator {
...
+		txHashes = append(txHashes, txn.Hash)
+	}
+	copyValidationView, err := validationView.CopyUtxoView()
 	if err != nil {
-		return errors.Wrapf(err, "PosMempool.refreshNoLock: Problem initializing temp pool")
+		return errors.Wrapf(err, "PosMempool.refreshNoLock: Problem copying utxo view")
 	}
-	if err := tempPool.Start(); err != nil {
-		return errors.Wrapf(err, "PosMempool.refreshNoLock: Problem starting temp pool")
+	_, _, _, _, successFlags, err := copyValidationView.ConnectTransactionsWithLimit(txns, txHashes, uint32(mp.latestBlockHeight)+1, time.Now().UnixNano(),
```
diamondhands0, line 750: I think I get why you made this function a member of the view but since we only really use it here, I would just embed the logic here and try to get it down to only one Connect call. I think it's slightly better from a readability standpoint, but from an efficiency standpoint the double Connect is definitely going to be a bottleneck so it fixes that.

diamondhands0: To clarify: I'm suggesting you delete ConnectTransactionsWithLimit and embed the logic that it's running in here, but get rid of the second Connect.

lib/pos_mempool.go
```diff
@@ -721,61 +726,45 @@ func (mp *PosMempool) GetIterator() MempoolIterator {
...
-	// Add all transactions from the main mempool to the temp mempool. Skip signature verification.
 	var txnsToRemove []*MempoolTx
-	txns := mp.getTransactionsNoLock()
-	for _, txn := range txns {
-		mtxn := NewMempoolTransaction(txn.Tx, txn.Added)
-		err := tempPool.AddTransaction(mtxn, false)
-		if err == nil {
-			continue
+	for ii, successFlag := range successFlags {
+		if ii >= len(mempoolTxns) {
```
diamondhands0, line 758: Interesting. How would ii ever be more than mempoolTxns if you're passing len(mempoolTxns) number of txns into the big COnnect above? A comment here would be good.

lib/pos_mempool.go
```diff
@@ -721,61 +726,45 @@ func (mp *PosMempool) GetIterator() MempoolIterator {
...
+			break
+		}
+		if successFlag {
+			mp.Lock()
+			mp.updateTransactionValidatedStatus(mempoolTxns[ii].Hash, true)
```
diamondhands0, line 763: It looks like you have a double-locking bug here. You lock on the outside of updateTransactionValidationStatus and within the function as well.

lib/pos_mempool.go
```diff
@@ -527,32 +541,27 @@ func (mp *PosMempool) validateTransaction(txn *MsgDeSoTxn, verifySignature bool)
...
-	// Check transaction signature.
-	if _, err := mp.readOnlyLatestBlockView.VerifySignature(txn, uint32(mp.latestBlockHeight)); err != nil {
-		return errors.Wrapf(err, "PosMempool.AddTransaction: Signature validation failed")
+	txn := mp.txnRegister.GetTransaction(txnHash)
+	if txn == nil {
+		return
 	}
 
-	return nil
+	txn.SetValidated(validated)
```
diamondhands0, line 560: Since this is a write-lock, I'm worried this could cause a bottleneck. Just to max efficiency, I would try and do something more like this if you can:
```
// 1. With no lock, or just with RLock, get all the txns you want to mark as validated.
txnsToMarkAsValidated = // whatever
mp.Lock()
// 2. Inside the lock blitz through these txns and call txn.SetValidated() on all of them in one go
mp.Unlock()
```

diamondhands0: It almost feels like you shouldn't even need to lock the whole mempool to update a single txn like this but if you're going to do it I'd just try to minimize the time spent holding the lock a little bit more, as I mentioned above.

lib/remote_node_indexer.go
```diff
@@ -1,46 +0,0 @@
-package lib
```
diamondhands0, line 1: I could have sworn we deleted this in an earlier PR. Not sure why it's showing up here but hopefully I didn't mess anything up with my branch.

lib/server.go
```diff
@@ -1867,6 +1871,10 @@ func (srv *Server) _relayTransactions() {
 		// for which the minimum fee is below what the Peer will allow.
 		invMsg := &MsgDeSoInv{}
 		for _, newTxn := range txnList {
+			if !newTxn.IsValidated() {
```
diamondhands0, line 1874: I had to stare at this for a while and trace through things but I think it achieves everything we want:
0. We will not broadcast a txn to our peers until after it has passed validation, which is great.
1. We are waiting to bundle txns at regular intervals thanks to TransactionValidationRefreshIntervalMillis
2. If a user submits N txns to a single node in rapid succession, those should get bundled together IN ORDER thanks to the fact that the InvLis is an ordered list.

Do you see any reason why we wouldn't be achieving all of these properties with this approach? Also I would add all of this in a comment here to explain the full dynamics of what adding this line does.

lib/pos_mempool.go
```diff
@@ -277,14 +289,35 @@ func (mp *PosMempool) Start() error {
...
+				mp.exitGroup.Done()
+				return
+			}
+		}
+	}()
+}
+
 func (mp *PosMempool) startAugmentedViewRefreshRoutine() {
 	go func() {
+		mp.startGroup.Done()
```
diamondhands0, line 320: The logic in this function seems to implement almost 1:1 what you're doing in ConnectTransactions\[WithLimit\]. Since you have to do some flavor of `connect all these transactions to a view`, I would consider a change as follows:
Instead of having ConnectTransactions ON the view, move it to pos_mempool.go over here with a signature something like the following:
```
func ConnectTransactionsToViewWithLimit(oldView *UtxoView, txnsToConnect []*MempoolTxn, numTxnsLimit int, ignoreFailing bool, etc...)
```
Then have this function simply return a new view that has connected everything to a copy of oldView. And you can use that function here and in the place where you're currently using COnnectTransactions. The upshot of this approach is you avoid adding code to the view and, critically, you only do ONE connect per txn because you don't need to update the internal state of a view.

lib/pos_mempool.go
```diff
@@ -721,61 +726,45 @@ func (mp *PosMempool) GetIterator() MempoolIterator {
...
+		if ii >= len(mempoolTxns) {
+			break
+		}
+		if successFlag {
+			mp.Lock()
+			mp.updateTransactionValidatedStatus(mempoolTxns[ii].Hash, true)
+			mp.Unlock()
+		} else {
+			txnsToRemove = append(txnsToRemove, mempoolTxns[ii])
 		}
```
diamondhands0, line 767: Below this line you call removeTransactionNoLock(). Is that OK to do here given that you're not holding any kind of lock in the outer function? It feels like maybe you need to hold a mp.Lock() while you do it. Or call the RemoveTransaction function or something.

PR #1118, opened 2024-03-15
Diamondhands Hyeprsync Review

diamondhands0:


lib/blockchain.go
```diff
@@ -2769,6 +2769,7 @@ func (bc *Blockchain) processBlockPoW(desoBlock *MsgDeSoBlock, verifySignatures
 	return isMainChain, false, nil
 }
 
+// DEPRECATED: This function is no longer used in PoS.
```
diamondhands0, line 2772: Can you delete this since we no longer call it?

PR #1139, opened 2024-03-21
Make MaxBlockSizeBytesPoS part of GlobalParams

lazynina:


lazynina:
* **#1139** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/1139?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `feature/proof-of-stake`

<h2></h2>

lib/block_view.go
```diff
@@ -5016,3 +5033,10 @@ func mergeExtraData(oldMap map[string][]byte, newMap map[string][]byte) map[stri
 
 	return retMap
 }
+
+func (bav *UtxoView) GetMaxBlockSizeBytes(height uint64) uint64 {
```
tholonious, line 5037: IMO, this has a good chance of becoming error prone in the future. Imagine if someone instantiates a UtxoView at block height n, which uses the global params at height n, but passes a height of n+2 to this function.

Ideally, the PoS code just reads `bav.GetCurrentGlobalParamsEntry().MaxBlockSizeBytesPoS` directly where needed. Direct access is fine in this case

lib/block_view.go
```diff
@@ -3581,7 +3598,7 @@ func (bav *UtxoView) _connectTransaction(
 			err, "_connectTransaction: Problem serializing transaction: ")
 	}
 	txnSizeBytes := uint64(len(txnBytes))
-	if txnSizeBytes > bav.Params.MaxBlockSizeBytes/2 {
+	if txnSizeBytes > bav.GetMaxBlockSizeBytes(uint64(blockHeight))/2 {
```
tholonious, line 3601: See comment below
```suggestion
	if txnSizeBytes > bav.GetCurrentGlobalParamsEntry().MaxBlockSizeBytesPoS/2 {
```

tholonious (changes requested):

lib/block_view.go
lazynina: Thinking more about it - I think we want to use the value from a snapshot anyways. we really just need to distinguish between PoS block size and PoW block size.

lib/block_view.go
lazynina: This will break PoW. we need to support having two different values.

lib/blockchain.go
```diff
@@ -1509,7 +1509,9 @@ func (bc *Blockchain) _validateOrphanBlock(desoBlock *MsgDeSoBlock) error {
 	if err != nil {
 		return fmt.Errorf("_validateOrphanBlock: Could not serialize block")
 	}
-	if uint64(len(serializedBlock)) > bc.params.MaxBlockSizeBytes {
+	// TODO: seems like we don't use this function w/ PoS so it's safe to
+	// leave this as a direct access to MaxBlockSizeBytesPoW.
```
tholonious, line 1513: Yeah, this is fine as-is. Would it make sense to rename the above function to `_validateOrhpahBlockPoW`? NBD either way.

tholonious (approved):

lib/blockchain.go
lazynina: done!

lazynina:
### Merge activity

* **Mar 29, 3:19 PM EDT**: @lazynina started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/1139).

PR #1145, opened 2024-03-25
Fix minor issues to speed up block processing logic

diamondhands0:
* Fix a bug whereby block production would not start until the first call to SubmitBlock, which resulted in erratic miner behavior. Specifically, all calls to get-block-template would fail until the first call to submit-block. Normally in the steady-state this is not an issue, but during a reboot it can become a "chicken and egg" where miners need get-block-template to generate a valid block, but get-block-template won't return until they actually produce a block (which they can't do without a template).

* Reduce the default block size used by miners to 500kb, down from 2mb. This ends up giving better performance during periods of high congestion because smaller blocks are processed more quickly. With very large blocks, we were encountering an issue whereby a block would take a minute or two to fuly validate, causing erratic behavior as the block propagated through the network.

* Add a rest period during initial transaction download that allows for other processes to grab the ChainLock if needed. This was the least obviously-needed of the fixes, but it doesn't hurt to leave it in, and could prevent deadlock-related issues during periods of high congestion.

* Add more useful logging in various places.

### DIFF
--- a/lib/block_producer.go
+++ b/lib/block_producer.go
@@ -3,12 +3,13 @@ package lib
 import (
 	"encoding/hex"
 	"fmt"
-	"github.com/btcsuite/btcd/wire"
-	"github.com/tyler-smith/go-bip39"
 	"math"
 	"sync"
 	"sync/atomic"
 	"time"
+
+	"github.com/btcsuite/btcd/wire"
+	"github.com/tyler-smith/go-bip39"
 
 	"github.com/deso-protocol/go-deadlock"
 
@@ -582,6 +583,8 @@ func (desoBlockProducer *DeSoBlockProducer) SignBlock(blockFound *MsgDeSoBlock) 
 
 func (desoBlockProducer *DeSoBlockProducer) Start() {
 	// If we set up the max sync block height, we will not be running the block producer.
+	glog.V(1).Infof("DeSoBlockProducer.Start() called. MaxSyncBlockHeight: %v",
+		desoBlockProducer.chain.MaxSyncBlockHeight)
 	if desoBlockProducer.chain.MaxSyncBlockHeight > 0 {
 		glog.V(2).Infof("DeSoBlockProducer.Start() exiting because "+
 			"MaxSyncBlockHeight: %v is greater than 0", desoBlockProducer.chain.MaxSyncBlockHeight)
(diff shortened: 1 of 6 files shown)


lib/block_producer.go
```diff
@@ -582,23 +583,28 @@ func (desoBlockProducer *DeSoBlockProducer) SignBlock(blockFound *MsgDeSoBlock)
...
 	}
 
 	// Set the time to a nil value so we run on the first iteration of the loop.
-	var lastBlockUpdate time.Time
+	lastBlockUpdate := time.Now()
 	desoBlockProducer.producerWaitGroup.Add(1)
 
 	for {
-		if atomic.LoadInt32(&desoBlockProducer.exit) >= 0 {
+		if atomic.LoadInt32(&desoBlockProducer.exit) > 0 {
```
diamondhands0, line 599: This was causing a lot of the erratic behavior when nodes would restart. See main comment.

lib/constants.go
```diff
@@ -905,7 +906,7 @@ var DeSoMainnetParams = DeSoParams{
 	// We set this to be lower initially to avoid winding up with really big
 	// spam blocks in the event someone tries to abuse the initially low min
 	// fee rates.
-	MinerMaxBlockSizeBytes: 2000000,
+	MinerMaxBlockSizeBytes: 500000,
```
diamondhands0, line 909: Note that this does not reduce the maximum block size permitted by the network, only the default block size used by miners when generating block templates.

lib/server.go
```diff
@@ -2045,10 +2047,25 @@ func (srv *Server) _processTransactions(pp *Peer, transactions []*MsgDeSoTxn) []
...
 	transactionsToRelay := []*MempoolTx{}
-	for _, txn := range transactions {
+	for ii, txn := range transactions {
+		// Take some time to allow other threads to get the ChainLock if they need it
+		//
+		// TODO: It's not obvious how necessary this rest period is, and it's also not obvious if
+		// five seconds is the right amount. We added it during a mission-critical sprint
+		// to find and fix slow block production issue as one of several patches. It clearly doesn't
+		// hurt so we decided to leave it in for now.
+		if (ii+1)%1000 == 0 {
```
diamondhands0, line 2060: Early on, I suspected that the reason get-block-template wasn't working was because of a ChainLock starvation caused by downloading a lot of transactions. Ultimately, it wasn't that, but giving a breather for other processes to acquire the ChainLock seemed like a fine thing to keep, and doesn't slow down sync too much.

lib/block_producer.go
```diff
@@ -582,23 +583,28 @@ func (desoBlockProducer *DeSoBlockProducer) SignBlock(blockFound *MsgDeSoBlock)
...
 	// If we set up the max sync block height, we will not be running the block producer.
+	glog.V(1).Infof("DeSoBlockProducer.Start() called. MaxSyncBlockHeight: %v",
+		desoBlockProducer.chain.MaxSyncBlockHeight)
 	if desoBlockProducer.chain.MaxSyncBlockHeight > 0 {
 		glog.V(2).Infof("DeSoBlockProducer.Start() exiting because "+
 			"MaxSyncBlockHeight: %v is greater than 0", desoBlockProducer.chain.MaxSyncBlockHeight)
 		return
 	}
 
 	// Set the time to a nil value so we run on the first iteration of the loop.
```
lazynina, line 594: Let's update this comment since time is no longer nil.

lazynina (approved):

PR #1164, opened 2024-04-01
Add GetAugmentedUniversalViewWithAdditionalTransactions for Backend APIs.

poolcoke:


lazynina (approved):

lib/pos_mempool.go
```diff
@@ -60,6 +60,60 @@ type Mempool interface {
...
+// GetAugmentedUniversalViewWithAdditionalTransactions which will create a
+// view that has connected a set of transactions (specified by optionalTxns).
+//
+// NOTE: GetAugmentedUniversalViewWithAdditionalTransactions DOES NOT validate fees
+// as fees are computed in UtxoView.ConnectBlock and optionalTxns are not included
+// in any block that can be connected yet.
+func GetAugmentedUniversalViewWithAdditionalTransactions(
+	mempool Mempool,
+	optionalTxnsBlockHeight uint32,
+	optionalTxnsTimestampNanoSecs int64,
```
tholonious, line 84: Any reason you want the block height and timestamp to be parametrizable? Ideally, these are decided by the mempool based on the chain current height + 1 and the current timestamp. That way, a caller can't connect the txn in a way that doesn't reflect how the txns will be connect on-chain

tholonious (approved):

lib/pos_mempool.go
poolcoke: I addressed this concern by adding in a "GetMempoolTipBlockHeight() uint64" method to the Mempool interface and implementing it for both PoS Mempool and legacy mempool. Now you only must pass the mempool and optionalTxns to generate a UtxoView. 

I opted to make the interface method get the tip block height as this seemed the most universally useful function moving forward.

lazynina (approved):

PR #1177, opened 2024-04-04
Account for size of header in block producer

lazynina:


lib/pos_block_producer.go
```diff
@@ -111,10 +145,77 @@ func (pbp *PosBlockProducer) createBlockTemplate(latestBlockView *UtxoView, newB
...
+}
+
+func (pbp *PosBlockProducer) mockHeader(
+	latestBlockView *UtxoView,
+	newBlockHeight uint64,
+	view uint64,
+	proposerRandomSeedSignature *bls.Signature,
+	validatorsVoteQC *QuorumCertificate,
+	validatorTimeoutAggregateQC *TimeoutAggregateQuorumCertificate,
+) (*MsgDeSoHeader, error) {
```
tholonious, line 175: This is great. Only suggestion is to encapsulate all of the mock header size computation within this function. One risk with exposing the mocked header outside of this function is that you then have a mocked MsgDeSoHeader instance laying around that could be mistaken for the true header. So you may as well encapsulate all of its usage here and expose the mocked header size instead
```suggestion
// estimateHeaderSize computes the MsgDeSoHeader message size resulting from the given params. It's 
// used to estimate the header size for block size computations when the true block header has not been
// constructed yet
func (pbp *PosBlockProducer) estimateHeaderSize(
	latestBlockView *UtxoView,
	newBlockHeight uint64,
	view uint64,
	proposerRandomSeedSignature *bls.Signature,
	validatorsVoteQC *QuorumCertificate,
	validatorTimeoutAggregateQC *TimeoutAggregateQuorumCertificate,
) (uint64, error) {
```

tholonious (changes requested):

tholonious (approved):

PR #1185, opened 2024-04-05
Checkpoint syncing providers system

lazynina:


lazynina:
* **#1185** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/1185?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `feature/proof-of-stake`

<h2></h2>

cmd/config.go
```diff
@@ -177,6 +181,29 @@ func LoadConfig() *Config {
...
+	config.CheckpointSyncingProviders = viper.GetStringSlice("checkpoint-syncing-providers")
+	for _, provider := range config.CheckpointSyncingProviders {
+		if _, err := url.ParseRequestURI(provider); err != nil {
+			glog.Fatalf("Invalid checkpoint syncing provider URL: %v", provider)
+		}
+		// TODO: do we want to make a request to the checkpoint syncing provider to ensure it's valid?
+	}
+	// TODO: add default provider here based on network. However, if someone wants to sync w/o checkpoint
+	// syncing, they should be able to do so. How do we support this? another flag I guess.
+	if len(config.CheckpointSyncingProviders) == 0 && !config.Regtest {
```
diamondhands0, line 194: I think how you have it here is fine. The ideal way to do it would be to set a default flag value but we can't really do that because of testnet/mainnet needing different providers here.

cmd/run.go
```diff
@@ -226,6 +226,11 @@ func SetupRunFlags(cmd *cobra.Command) {
 		"from an empty string to a non-empty string (or from a non-empty string to the empty string) requires a resync.")
 	cmd.PersistentFlags().Uint("state-syncer-mempool-txn-sync-limit", 10000, "The maximum number of transactions to "+
 		"process in the mempool tx state syncer at a time.")
+
+	// PoS Checkpoint Syncing
+	// TODO: a default value is a little tricky here since we don't know what network we're using.
```
diamondhands0, line 231: I'd get rid of the TODO and just make the comment explain that we actually DO set a default value, but in code rather than explicitly. I would also embed the default in the message for the flag. Like this:

```
fmt.Sprintf("A comma-separated list of URLs that ... . If this flag is unset it will default to %v for mainnet and %v for testnet", DefaultMainnetBlah, DefaultTestnetBlah)
```

lib/blockchain.go
```diff
@@ -517,9 +538,98 @@ type Blockchain struct {
...
+
 	timer *Timer
 }
 
+func (bc *Blockchain) updateCheckpointBlockInfo() {
+	if len(bc.checkpointSyncingProviders) == 0 {
+		glog.V(2).Info("updateCheckpointBlockInfo: No checkpoint syncing providers set. Skipping update.")
+		return
+	}
+	var wg sync.WaitGroup
```
diamondhands0, line 558: Nit: I don't *think* you need the waitgroup here because the for-loop where you consume from the channel using `<-ch` will naturally force blocking until they all return.

cmd/run.go
diamondhands0: Hold on I'm going to see if I can edit this myself the way you did on my other PR...

lib/blockchain.go
```diff
@@ -517,9 +538,98 @@ type Blockchain struct {
...
+	glog.V(2).Infof("updateCheckpointBlockInfo: Setting checkpoint block info to: %v", highestHeightCheckpointBlockInfo)
+	bc.checkpointBlockInfo = highestHeightCheckpointBlockInfo
+}
+
+func getCheckpointBlockInfoFromProvider(provider string, ch chan<- *CheckpointBlockInfoAndError, wg *sync.WaitGroup) {
+	defer wg.Done()
+	url := fmt.Sprintf("%s%s", provider, RoutePathGetCommittedTipBlockInfo)
+	req, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		ch <- &CheckpointBlockInfoAndError{
```
diamondhands0, line 598: I don't love how easy it is to forget to add to the channel before you return here. I would break it into two functions so you can get compile-time checking on the return like this:

```
func getCheckpointBlockInfoFromProviderHelper(provider string) (*CheckpointBlockInfoAndError) {

}

func getCheckpointBlockInfoFromProvider(provider string, ch chan<- *CheckpointBlockInfoAndError, wg *sync.WaitGroup) {
```

cmd/run.go
```diff
@@ -226,6 +226,11 @@ func SetupRunFlags(cmd *cobra.Command) {
...
+	cmd.PersistentFlags().StringSlice("checkpoint-syncing-providers", []string{}, "A comma-separated list of URLs that "+
+		"support the committed tip block info endpoint to be used for checkpoint syncing.")
```
diamondhands0, line 233:
```suggestion
	// PoS Checkpoint Syncing
	// TODO: a default value is a little tricky here since we don't know what network we're using.
	cmd.PersistentFlags().StringSlice(fmt.Sprintf("checkpoint-syncing-providers", []string{}, "A comma-separated list of URLs that "+
		"supports the committed tip block info endpoint to be used for checkpoint syncing. "+
		"If unset, the field will default to %v on mainnet and %v on testnet", DefaultMainnetCheckpointProvider, DefaultTestnetCheckpointProvider))
```

lib/blockchain.go
```diff
@@ -517,9 +538,98 @@ type Blockchain struct {
...
+		return
+	}
+	var wg sync.WaitGroup
+	ch := make(chan *CheckpointBlockInfoAndError, len(bc.checkpointSyncingProviders))
+	for _, provider := range bc.checkpointSyncingProviders {
+		wg.Add(1)
+		go getCheckpointBlockInfoFromProvider(provider, ch, &wg)
+	}
+
+	wg.Wait()
```
diamondhands0, line 565:
```suggestion
	ch := make(chan *CheckpointBlockInfoAndError, len(bc.checkpointSyncingProviders))
	for _, provider := range bc.checkpointSyncingProviders {
		go getCheckpointBlockInfoFromProvider(provider, ch)
	}
```

lib/blockchain.go
```diff
@@ -517,9 +538,98 @@ type Blockchain struct {
...
+	if err = json.NewDecoder(resp.Body).Decode(responseData); err != nil {
+		ch <- &CheckpointBlockInfoAndError{
+			Error: errors.Wrapf(err, "getCheckpointBlockInfoFromProvider: Problem decoding response data"),
+		}
+		return
+	}
+	ch <- &CheckpointBlockInfoAndError{
+		CheckpointBlockInfo: responseData,
+	}
+}
```
diamondhands0, line 631:
```suggestion
func getCheckpointBlockInfoFromProvider(provider string, ch chan<- *CheckpointBlockInfoAndError) {
  ch <- getCheckpointBlockInfoFromProviderHelper(provider)
}

func getCheckpointBlockInfoFromProviderHelper(provider string) *CheckpointBlockInfoAndError {
	url := fmt.Sprintf("%s%s", provider, RoutePathGetCommittedTipBlockInfo)
	req, err := http.NewRequest("GET", url, nil)
	if err != nil {
		return &CheckpointBlockInfoAndError{
			Error: errors.Wrapf(err, "getCheckpointBlockInfoFromProvider: Problem creating HTTP request"),
		}
	}
	client := &http.Client{}
	resp, err := client.Do(req)
	if err != nil {
		return &CheckpointBlockInfoAndError{
			Error: errors.Wrapf(err, "getCheckpointBlockInfoFromProvider: Problem sending HTTP request"),
		}
	}
	if resp.StatusCode != 200 {
		return &CheckpointBlockInfoAndError{
			Error: fmt.Errorf(
				"getCheckpointBlockInfoFromProvider: Problem getting checkpoint block info from provider: %s",
				provider,
			),
		}
	}
	defer resp.Body.Close()
	responseData := &CheckpointBlockInfo{}
	if err = json.NewDecoder(resp.Body).Decode(responseData); err != nil {
		return &CheckpointBlockInfoAndError{
			Error: errors.Wrapf(err, "getCheckpointBlockInfoFromProvider: Problem decoding response data"),
		}
	}
	return &CheckpointBlockInfoAndError{
		CheckpointBlockInfo: responseData,
	}
}
```

lib/server.go
```diff
@@ -930,15 +931,73 @@ func (srv *Server) GetBlocks(pp *Peer, maxHeight int) {
...
+		return true, false
+	}
+	// For PoS blocks, we check if we've seen the checkpoint block.
+	// If we don't have a check point block info, we return true so that we verify signatures.
+	if srv.blockchain.checkpointBlockInfo == nil {
+		return true, false
+	}
+	var hasSeenCheckpointBlockHash bool
+	var checkpointBlockNode *BlockNode
+	if isHeaderChain {
```
diamondhands0, line 960: I think there's a small bug here. If we NEVER find the checkpoint header then I think this code does not disconnect from the peer properly. Instead, we need to add as a condition that the height has not progressed beyond the checkpoint as well.

lib/server.go
```diff
@@ -930,15 +931,73 @@ func (srv *Server) GetBlocks(pp *Peer, maxHeight int) {
...
+		checkpointBlockNode, hasSeenCheckpointBlockHash = srv.blockchain.bestHeaderChainMap[*srv.blockchain.checkpointBlockInfo.Hash]
+	} else {
+		checkpointBlockNode, hasSeenCheckpointBlockHash = srv.blockchain.bestChainMap[*srv.blockchain.checkpointBlockInfo.Hash]
+	}
+	// If we haven't seen the checkpoint block hash yet, we skip signature verification.
+	if !hasSeenCheckpointBlockHash {
```
diamondhands0, line 966:
```suggestion
	if !hasSeenCheckpointBlockHash {
	    	// If we're past the checkpoint height and we haven't seen the checkpoint block, we should
	// disconnect from the peer.
	if header.Height >= srv.blockchain.checkpointBlockInfo.Height {
	    return true, true
	}
```

lib/server.go
diamondhands0: I suggested the edit for this below

lib/server.go
```diff
@@ -985,12 +1043,20 @@ func (srv *Server) _handleHeaderBundle(pp *Peer, msg *MsgDeSoHeaderBundle) {
...
+				pp, srv.blockchain.chainState(), headerReceived.Height,
+				srv.blockchain.checkpointBlockInfo.String())
+			pp.Disconnect()
+			return
+		}
 
 		// Process the header, as we haven't seen it before, set verifySignatures to false
 		// if we're in the process of syncing.
-		_, isOrphan, err := srv.blockchain.ProcessHeader(headerReceived, headerHash, !srv.blockchain.isSyncing())
+		_, isOrphan, err := srv.blockchain.ProcessHeader(headerReceived, headerHash, verifySignatures)
```
diamondhands0, line 1059: Clean and efficient. I love it.

diamondhands0 (approved):
This is perfect. I made some minor suggestions in code but they're all nits.

lib/server.go
```diff
@@ -931,15 +932,80 @@ func (srv *Server) GetBlocks(pp *Peer, maxHeight int) {
...
+	}
+	// For PoS blocks, we check if we've seen the checkpoint block.
+	// If we don't have a check point block info, we return true so that we verify signatures.
+	checkpointBlockInfo := srv.blockchain.GetCheckpointBlockInfo()
+	if checkpointBlockInfo == nil {
+		return true, false
+	}
+	var hasSeenCheckpointBlockHash bool
+	var checkpointBlockNode *BlockNode
+	if isHeaderChain {
```
tholonious, line 962: There's an edge case here where the bestHeaderChain or bestChain contain the checkpointBlockNode, but the incoming header still has a lower height than the checkpoint's height.

I think you still need to check check for `header.Height > checkpointBlockInfo.Height` here. I imagine this could happen if we receive forking header bundles from the peer, so you may as well make the signature checking condition bullet proof

tholonious (changes requested):

tholonious (approved):

PR #1220, opened 2024-04-12
Fix PoW mempool and fees for Atomics

poolcoke:


lib/block_view.go
```diff
@@ -4420,21 +4420,53 @@ func (bav *UtxoView) ConnectBlock(
...
+				// For PoS, the maximum block reward is based on the maximum utility fee.
+				// Add the utility fees to the max utility fees. If any overflow
+				// occurs mark the block as invalid and return a rule error.
+				maxUtilityFee, err = SafeUint64().Add(maxUtilityFee, utilityFee)
+				if err != nil {
+					return nil, errors.Wrapf(RuleErrorPoSBlockRewardWithInvalidAmount,
+						"ConnectBlock: error computing maxUtilityFee: %v", err)
+				}
+			} else {
+				for _, innerTxn := range txn.TxnMeta.(*AtomicTxnsWrapperMetadata).Txns {
```
lazynina, line 4444: This is quite deeply nested, so cognitive complexity is pretty high in here. Can we write some sort of utility function that given an atomic transaction and a block reward recipient, iterates over it and return the total fees minus any fees paid by the block reward recipient? This function can return both BMF and old PoW fees. This way we can use the same function in each of the three places its used and not have such deeply nested for-loops and conditionals.

lazynina (changes requested):

lazynina (changes requested):
All the logic is correct, so it *works* in my eyes, but it's going to be a little annoying to maintain with the deeply nested logic as we have it. Any modifications you can make to simplify and de-duplicate looping over the inner txns to filter out block reward recipient fees would be very helpful.

lazynina:
* **#1224** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/1224?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1220** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/1220?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `feature/proof-of-stake`

<h2></h2>

 Join @poolcoke and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

lib/block_producer.go
```diff
@@ -452,17 +484,34 @@ func RecomputeBlockRewardWithBlockRewardOutputPublicKey(
...
+					glog.Errorf("DeSoMiner._startThread: Error parsing transactor public key: %v", err)
+					continue
+				}
+				if !transactorPublicKey.IsEqual(blockRewardOutputPublicKey) {
+					totalFees, err = SafeUint64().Add(totalFees, innerTxn.TxnFeeNanos)
+					if err != nil {
+						glog.Errorf("DeSoMiner._startThread: Error adding txn fee: %v", err)
+						continue
+					}
+				}
```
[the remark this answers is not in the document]
poolcoke: Yeah, I think this is better. We used `!transactorPublicKey.IsEqual(blockRewardOutputPublicKey)` before this change so I think it seemed natural to do but it was confusing while I was reading through it.

lib/block_producer.go
```diff
@@ -305,12 +305,43 @@ func (desoBlockProducer *DeSoBlockProducer) _getBlockTemplate(publicKey []byte)
...
+								"DeSoBlockProducer._getBlockTemplate: problem parsing transactor public key: ")
+					}
+
+					// Block the block reward output public key from getting free transaction fees within
+					// an atomic transaction.
+					if transactorPublicKey.IsEqual(blockRewardOutputPublicKey) {
+						if feeNanos < innerTxn.TxnFeeNanos {
+							return nil, nil, nil,
+								fmt.Errorf("DeSoBlockProducer._getBlockTemplate: " +
+									"feeNanos for atomic transaction underflowed during blow reward check; " +
```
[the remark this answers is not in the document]
poolcoke: Addressed with PR#1224.

lib/block_producer.go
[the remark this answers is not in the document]
poolcoke: Addressed with PR#1224.

lib/block_view.go
poolcoke: Addressed with PR#1224.

lazynina (approved):

tholonious (approved):

lib/block_view.go
```diff
@@ -4360,21 +4360,47 @@ func (bav *UtxoView) ConnectBlock(
...
-			// Add the fees from this txn to the total fees. If any overflow occurs
-			// mark the block as invalid and return a rule error. Note that block reward
-			// txns should count as having zero fees.
-			if totalFees > (math.MaxUint64 - currentFees) {
-				return nil, RuleErrorTxnOutputWithInvalidAmount
-			}
-			totalFees += currentFees
+			if txn.TxnMeta.GetTxnType() != TxnTypeAtomicTxnsWrapper {
+				// Compute the BMF given the current fees paid in the block.
+				_, utilityFee = computeBMF(currentFees)
```
diamondhands0, line 4365: Am I crazy or does this call to computeBMF need to be gated behind a block height?

diamondhands0 (changes requested):
See if you need to put _computeBMF behind a block height

diamondhands0 (approved):

lib/block_view.go
poolcoke: Gated on line 4449.

lazynina:
> See if you need to put _computeBMF behind a block height

As discussed offline, all downstream uses are confined to pos gated logic.

PR #1222, opened 2024-04-13
Simplify Mempool Transaction Validation Routine

tholonious:


tholonious:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/deso-protocol/core/1222?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1222** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/1222?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#1221** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/1221?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `feature/proof-of-stake`

<h2></h2>

 Join @tholonious and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

lib/pos_mempool.go
```diff
@@ -921,82 +910,73 @@ func (mp *PosMempool) validateTransactions() error {
...
-	mp.Lock()
-	for _, txn := range txnsToRemove {
-		if err := mp.removeTransactionNoLock(txn, true); err != nil {
-			glog.Errorf("PosMempool.validateTransactions: Problem removing transaction with hash (%v): %v", txn.Hash, err)
+		// Connect the transaction to a copy of the validation view. We can skip signatures on the transaction
+		// connect if the transaction has already previously been validated and been found to have a valid
+		// signature. This optimizes the connect by not repeating signature verification on a transaction
+		// more than once.
+		//
+		// TODO: Nina to confirm the signature validation assumption.
```
tholonious, line 943: @lazynina - I *think* this is safe. Provided the transaction has been validated once, it MUST have a valid signature

lazynina: The only problem is transactions signed by derived key. It's possible that a derived key has been de-authorized between runs of the transaction validation routine. But then the transaction would also fail validation because it wouldn't have sufficient spending limits. I need to think on this more, but I think it's okay to do.

tholonious: Sounds good @lazynina. Based on the cpu profiling, it looks like biggest cpu overhead in the mempool is repeat signature verification.

And it makes sense why that would be. This validation goroutine runs every 10ms and validates each txn multiple times in between the 1.5sec block time.

lazynina (approved):

lib/pos_mempool.go
lazynina: This seems fine. If the derived key was de-authorized, it'll get caught by the spending limits check and marked invalid. There should never be a need to check signatures a second time.

PR #1252, opened 2024-04-20
Fix PoS mempool dynamic fee market

diamondhands0:
A bunch of things were broken in our mempool fee estimation logic. All fixed now. Summary:
- Fixed a bug whereby we would never increase the fee above the minimum fee because of the placement of a `if bucketMinFee <= globalMinFeeRate` check.
- Fixed a bug where we were overwriting the mempool fee register in Start(), which was causing the mempool fee estimator to have no txns in it, and thus always return the minimum value. This meant we were not considering mempool congestion at all.
- In many places, we were confusing "fee bucket growth rate basis points" with "fee bucket multiplier". The former value would be something like `1000 (= 10%)` while the latter would map to `1.1 (= 10000 + 1000 / 10000)`. This caused fee-time ordering to be basically completely broken. All fixed now, and fixed tests.
  - Just to add a little more detail: The tests were very well-written and I think they exercise this logic very well. The reason why they were passing before, though, is because we were setting the value incorrectly in the Init() *and* passed it wrong as an argument, and the two sortof compensated for each other in the tests. But in production, we would go down a different path that wouldn't compensate properly, which is how I found the bug. Anyway it's all fixed now.
- Set optimized defaults for the mempool dynamic fee params and added a deep comment explaining why we chose these values where they are define. Also made sure we're using them consistently in all the relevant places. These params optimize heavily toward getting your txn into the next block, which is what we want. They cause reordering issues if you're sending txns at a rate much higher than 1 block per second, but this is correct behavior, and the comments include suggestions on how to mitigate these issues (eg by manually setting the fee or using an atomic txn):
  - MempoolCongestionFactorBasisPoints
  - MempoolPriorityPercentileBasisPoints
  - PastBlocksCongestionFactorBasisPoints
  - PastBlocksPriorityPercentileBasisPoints
- In `computeFeeTimeBucketRangeFromExponent`, there was a weird edge-case where we could have a fee bucket with start less than end. This can't happen in a real scenario, though, only when the bucket growth rate is like 1bp, which is ridiculously small. And I only found it because of the growth rate <> multiplier issue mentioned previously, which was causing a 10% growth rate to be threaded through as 1bp.

For reference, in case it's useful, the way I found all this stuff was I slowed the block time down to 1 block every 10s and made the NumPastBlocks for the block estimator 5 blocks using the params in constants.go (so that txns would accumulate in the mempool) and added logging of the fees. Then I wrote a script that blasted the mempool with txns and noticed that the fees weren't adjusting properly, which led me down the rabbit-hole to find all of these issues. After fixing all the issues I took some time to optimize all the params, and then used my script to exercise everything and make sure it's fully 100% adapting correctly. Specifically, I saw that the fee goes up correctly once the mempool has a full block's worth of txns accumulated in it, stays high for a few blocks because of the block estimator, and then starts to go down as more blocks come through. It all works really well.

### DIFF
--- a/lib/blockchain.go
+++ b/lib/blockchain.go
@@ -5010,9 +5010,15 @@ func (bc *Blockchain) CreateMaxSpend(
 				if bc.params.IsPoSBlockHeight(uint64(bc.BlockTip().Height)) {
 					maxBlockSizeBytes = utxoView.GetSoftMaxBlockSizeBytesPoS()
 				}
-				// TODO: replace MaxBasisPoints with variables configured by flags.
-				feeAmountNanos, err = mempool.EstimateFee(txn, minFeeRateNanosPerKB,
-					MaxBasisPoints, MaxBasisPoints, MaxBasisPoints, MaxBasisPoints, maxBlockSizeBytes)
+				feeAmountNanos, err = mempool.EstimateFee(
+					txn,
+					minFeeRateNanosPerKB,
+					// TODO: Make these flags or GlobalParams
+					bc.params.MempoolCongestionFactorBasisPoints,
+					bc.params.MempoolPriorityPercentileBasisPoints,
+					bc.params.PastBlocksCongestionFactorBasisPoints,
+					bc.params.PastBlocksPriorityPercentileBasisPoints,
+					maxBlockSizeBytes)
 				if err != nil {
 					return nil, 0, 0, 0, errors.Wrapf(err, "CreateMaxSpend: Problem estimating fee: ")
 				}
@@ -5147,9 +5153,15 @@ func (bc *Blockchain) AddInputsAndChangeToTransactionWithSubsidy(
 				if bc.params.IsPoSBlockHeight(uint64(bc.BlockTip().Height)) {
 					maxBlockSizeBytes = utxoView.GetSoftMaxBlockSizeBytesPoS()
 				}
-				// TODO: replace MaxBasisPoints with variables configured by flags.
-				newTxFee, err := mempool.EstimateFee(txArg, minFeeRateNanosPerKB, MaxBasisPoints,
-					MaxBasisPoints, MaxBasisPoints, MaxBasisPoints, maxBlockSizeBytes)
+				newTxFee, err := mempool.EstimateFee(
+					txArg,
+					minFeeRateNanosPerKB,
+					// TODO: Make these flags or GlobalParams
+					bc.params.MempoolCongestionFactorBasisPoints,
+					bc.params.MempoolPriorityPercentileBasisPoints,
+					bc.params.PastBlocksCongestionFactorBasisPoints,
+					bc.params.PastBlocksPriorityPercentileBasisPoints,
+					maxBlockSizeBytes)
 				UpdateTxnFee(txArg, newTxFee)
 				if err != nil {
 					return 0, 0, 0, 0, errors.Wrapf(err,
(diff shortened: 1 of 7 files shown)


lib/blockchain.go
```diff
@@ -5003,9 +5003,15 @@ func (bc *Blockchain) CreateMaxSpend(
...
-				feeAmountNanos, err = mempool.EstimateFee(txn, minFeeRateNanosPerKB,
-					MaxBasisPoints, MaxBasisPoints, MaxBasisPoints, MaxBasisPoints, maxBlockSizeBytes)
+				feeAmountNanos, err = mempool.EstimateFee(
+					txn,
+					minFeeRateNanosPerKB,
+					// TODO: Make these flags or GlobalParams
+					bc.params.MempoolCongestionFactorBasisPoints,
+					bc.params.MempoolPriorityPercentileBasisPoints,
+					bc.params.PastBlocksCongestionFactorBasisPoints,
+					bc.params.PastBlocksPriorityPercentileBasisPoints,
```
lazynina, line 5013: Making these global params is a forking change - just a heads up. Doing it with flags is chill though.

lib/blockchain.go
```diff
@@ -5856,7 +5868,15 @@ func (bc *Blockchain) CreateAtomicTxnsWrapper(
 		txn.ExtraData[NextAtomicTxnPreHash] = dummyAtomicHashBytes
 		txn.ExtraData[PreviousAtomicTxnPreHash] = dummyAtomicHashBytes
 		newFeeEstimate, err := mempool.EstimateFee(
-			txn, 0, MaxBasisPoints, MaxBasisPoints, MaxBasisPoints, MaxBasisPoints, maxBlockSizeBytes)
+			txn,
+			// TODO: Allow the caller to specify minFeeRateNanosPerKB
+			0,
```
lazynina, line 5873: I can update this in a separate PR. It's a relatively minor change.

lib/pos_fee_estimator.go
```diff
@@ -339,6 +339,20 @@ func (posFeeEstimator *PoSFeeEstimator) EstimateFee(
 	pastBlocksPriorityPercentileBasisPoints uint64,
 	maxBlockSize uint64,
 ) (uint64, error) {
+	// If minFeeRateNanosPerKB is specified, we use that to compute the txn fee. Otherwise,
+	// we'll use our fee estimation logic to compute the fee rate to use. This allows the developer
+	// to override the fee estimation logic if they need to for special circumstances.
+	//
+	// Note that it may make sense to change the name of this to overrideFeeRateNanosPerKB, but we'd
+	// have to change it in a lot of places so I don't think it's worth it.
+	if minFeeRateNanosPerKB != 0 {
```
lazynina, line 348: We're going to need to change a lot of logic in deso-js and our frontend repos to NOT specify a fee in order to get the optimal fee market suggestion with this change. I'd rather us introduce a NEW value to represent this "override" fee rate, but we're down to crunch time so it's a bit annoying at this point.

lazynina (approved):

lib/blockchain.go
tholonious: I'm moving these to GlobalParams. Marking this as resolved now that we have a resolution on what to do here

tholonious:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/deso-protocol/core/1252?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1252** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/1252?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#1253** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/1253?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `feature/proof-of-stake`

<h2></h2>

 Join @diamondhands0 and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

tholonious (approved):
LGTM. I moved the `overrideFeeRateNanosPerKB` change to https://github.com/deso-protocol/core/pull/1255. That way, we can merge this rest of this PR asap.

lazynina (approved):

PR #1275, opened 2024-04-27
Add buying/selling pkid to GetAllDAOCoinLimitOrders with tests

diamondhands0:
This PR contains a few SAFE modifications to support the launch of DeSo AMMs and CCV2:
- Make `_getDAOCoinLimitOrderEntry` public so we can call it on a UtxoView from backend.
- Add `buyingCoinPkid *PKID` and `sellingCoinPkid *PKID` to the call signature of `GetAllDAOCoinLimitOrdersForThisTransactor`. This allows you to get all the transactions for a particular transactor *by market*. Note that this change is SAFE because if you pass (nil, nil) for the new arguments then the behavior is *identical* to what it was previously. So it gives us the new functionality without any risk or any new indexes.
- Add new tests for `GetAllDAOCoinLimitOrdersForThisTransactor` to make sure the new arguments actually work and upgrade old tests.
- Add a public `IsDeleted` for `DAOCoinLimitOrderEntry`. This is needed to support filtering out unconfirmed orders in the backend, which is something new we added.


lib/db_utils.go
```diff
@@ -9822,10 +9822,23 @@ func DBGetAllDAOCoinLimitOrdersForThisDAOCoinPair(
...
+	sellingCoinPKID *PKID,
+) (
+	[]*DAOCoinLimitOrderEntry,
+	error,
+) {
+	// Get all DAO coin limit orders for this transactor. Potentially filter by the
+	// buying/selling coin pkids if provided
 	key := append([]byte{}, Prefixes.PrefixDAOCoinLimitOrderByTransactorPKID...)
 	key = append(key, transactorPKID[:]...)
+	if buyingCoinPKID != nil && sellingCoinPKID != nil {
```
lazynina, line 9838: So if we only specify sellingCoinPKID but not buyingCoinPKID, then this function returns all entries (and vice versa). We should probably clarify this in the comments describing this function.

lazynina (approved):

tholonious (approved):

lazynina:
* **#1275** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/1275?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `feature/proof-of-stake`

<h2></h2>

 Join @diamondhands0 and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

PR #1297, opened 2024-05-09
hypersync pos fixes

lazynina:


lazynina:
* **#1297** <a href="https://app.graphite.dev/github/pr/deso-protocol/core/1297?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `feature/proof-of-stake`

<h2></h2>

lib/snapshot.go
```diff
@@ -926,14 +906,14 @@ func (snap *Snapshot) DBSetAncestralRecordWithTxn(
...
-//	<prefix [1]byte, block height [8]byte, key []byte> -> <value []byte, existence_byte [1]byte>
+//	<mainDbPrefix [1]byte, prefix [1]byte, block height [8]byte, key []byte> -> <value []byte, existence_byte [1]byte>
 //
 // So we need to trim the first 9 bytes off of the ancestral_key to get the actual key.
 // And we need to trim the last 1 byte off of the ancestral_value to get the actual value.
 func (snap *Snapshot) AncestralRecordToDBEntry(ancestralEntry *DBEntry) *DBEntry {
 	var dbKey, dbVal []byte
 	// Trim the prefix and the block height from the ancestral record key.
-	dbKey = ancestralEntry.Key[9:]
+	dbKey = ancestralEntry.Key[10:]
```
diamondhands0, line 916: Oooh tricky one...

lazynina: hahah yeah this one took me a minute to find.

cmd/run.go
```diff
@@ -67,7 +67,7 @@ func SetupRunFlags(cmd *cobra.Command) {
 	cmd.PersistentFlags().Bool("force-checksum", false, "When true, the node will panic if the "+
 		"local state checksum differs from the network checksum reported by its peers.")
 	// Snapshot
-	cmd.PersistentFlags().Uint64("snapshot-block-height-period", 600000, "Set the snapshot epoch period. Snapshots are taken at block heights divisible by the period.")
+	cmd.PersistentFlags().Uint64("snapshot-block-height-period", lib.DefaultSnapshotEpochPeriodPoS, "Set the snapshot epoch period. Snapshots are taken at block heights divisible by the period.")
```
lazynina, line 70: Just defining as a constant to make it easier to re-use in other places. We'll eventually want to rip out this flag and enforce a consistent snapshot epoch length for PoS so that you can hypersync from any other hypersynced node.

lib/block_view.go
```diff
@@ -4568,17 +4568,6 @@ func (bav *UtxoView) ConnectBlock(
 	// after connecting all transactions in the block. These operations
 	// are always the last utxo operation in a given block.
 	var blockLevelUtxoOps []*UtxoOperation
-	if blockHeight >= uint64(bav.Params.ForkHeights.BalanceModelBlockHeight) &&
```
lazynina, line 4571: Deleting expired nonces significantly slows down the syncing process. Eventually we can do it on some different cadence or look at performance enhancements for this iteration.

lib/snapshot.go
```diff
@@ -2000,21 +2008,21 @@ func (status *SnapshotStatus) Initialize(snapshotDb *badger.DB, snapshotDbMutex
 
 func (status *SnapshotStatus) ToBytes() []byte {
 	var data []byte
-	data = append(data, UintToBuf(status.MainDBSemaphore)...)
-	data = append(data, UintToBuf(status.AncestralDBSemaphore)...)
+	data = append(data, UintToBuf(0)...) // Formerly status.MainDBSemaphore
```
lazynina, line 2011: I left this here for backward compatibility, but I guess we could simplify if we really want.

cmd/run.go
diamondhands0: Yep that's what I figured when I read it the first time.

lib/block_view.go
diamondhands0: Yep sounds fine to me. Maybe put a TODO here to delete nonces and explain the whole situation for posterity. It's always good to throw in gratuitous comments to capture the tribal knowledge. Something like the following:

```
TODO: To prevent the state from bloating, we should delete nonces periodically. We used to do that here but it was causing badger seeks to be slow due to a bug in badger whereby deleting keys slows down seeks. Eventually, we should go back to deleting txn nonces if we fix that badger bug or find a workaround.
```

lib/db_utils.go
```diff
@@ -1106,7 +1113,7 @@ func DBSetWithTxn(txn *badger.Txn, snap *Snapshot, key []byte, value []byte, eve
 		ancestralValue, getError = DBGetWithTxn(txn, snap, key)
 
 		// If there is some error with the DB read, other than non-existent key, we return.
-		if getError != nil && getError != badger.ErrKeyNotFound {
+		if getError != nil && !errors.Is(getError, badger.ErrKeyNotFound) {
```
diamondhands0, line 1116: Huh... do we need to do this in all places or were you just being extra defensive here?

lib/db_utils.go
```diff
@@ -1184,14 +1192,12 @@ func DBGetWithTxn(txn *badger.Txn, snap *Snapshot, key []byte) ([]byte, error) {
 		return nil, err
 	}
 
-	// If a flush takes place, we don't update cache. It will be updated in DBSetWithTxn.
+	// TODO: Do we want to update the database cache when performing GETs? I think it would be
+	// safer to ONLY update the cache when performing SETs. This way, we can avoid the possibility
+	// of the cache getting out of sync with the database when a badger view transaction is started
+	// before a badger update transaction begins.
```
diamondhands0, line 1198: This seems OK to me because we only get to this line if the item is NOT in the cache yet, which means we've never read it from the DB. If you look above on line 1179 you'll see it FIRST tries to look into the cache, and only gets here if it's NOT in the cache. If you think there's still an issue lmk.

lib/peer.go
```diff
@@ -501,16 +501,21 @@ func (pp *Peer) HandleGetSnapshot(msg *MsgDeSoGetSnapshot) {
 	// to the main DB or the ancestral records DB, and we don't want to slow down any of these updates.
 	// Because of that, we will detect whenever concurrent access takes place with the concurrencyFault
 	// variable. If concurrency is detected, we will re-queue the GetSnapshot message.
-	var concurrencyFault bool
+
+	// 05/09/2024: Leaving the above comment around for posterity. The concurrencyFault variable is no longer
+	// used because we performant all snapshot operations in a synchronous manner that blocks until the operation
```
diamondhands0, line 506: Yep makes sense. Good comment.

lib/pos_epoch_complete_hook.go
```diff
@@ -118,16 +118,6 @@ func (bav *UtxoView) runEpochCompleteStateTransition(blockHeight uint64, blockTi
 		return nil, errors.Wrapf(err, "runEpochCompleteStateTransition: problem rewarding snapshot stakes: ")
 	}
 
-	// Delete expired nonces
-	prevNonces := bav.GetTransactorNonceEntriesToDeleteAtBlockHeight(blockHeight)
```
diamondhands0, line 122: I'd add a TODO comment here that explains why we're not deleting nonces in this location.

lib/server.go
```diff
@@ -2417,6 +2428,9 @@ func (srv *Server) _handleBlockBundle(pp *Peer, bundle *MsgDeSoBlockBundle) {
 		// gracefully fail.
 		srv._handleBlock(pp, blk, ii == len(bundle.Blocks)-1 /*isLastBlock*/)
 		numLogBlocks := 1000
+		if srv.params.IsPoWBlockHeight(blk.Header.Height) {
+			numLogBlocks = 25
```
diamondhands0, line 2432: Do you want to leave this numLogBlocks at 25? Seems fine, just flagging in case you did it for debugging only.

lib/snapshot.go
```diff
@@ -394,10 +443,8 @@ func NewSnapshot(mainDb *badger.DB, mainDbDirectory string, snapshotBlockHeightP
 	//   Either way, it means our snapshot was compromised and we need to recompute it as described
 	//   in the previous bullet.
 	shouldRestart := false
-	if operationChannel.StateSemaphore > 0 || status.IsFlushing() {
+	if operationChannel.StateSemaphore > 0 {
```
diamondhands0, line 446: Sooo can StateSemaphore be > 0 now? I'm going to take a closer look at this after I work out just to see if we can eliminate this edge-case.

lib/snapshot.go
```diff
@@ -472,13 +530,15 @@ func (snap *Snapshot) Run() {
 
 		case SnapshotOperationChecksumAdd:
 			if err := snap.Checksum.AddOrRemoveBytesWithMigrations(operation.checksumKey, operation.checksumValue,
-				snap.Status.CurrentBlockHeight, snap.Migrations.migrationChecksums, true); err != nil {
+				snap.Status.CurrentBlockHeight, snap.Migrations.migrationChecksums,
+				snap.Migrations.migrationChecksumLock, true); err != nil {
```
diamondhands0, line 534: Interesting that we had to add this lock here. I'll take a closer look at it after gym.

lib/snapshot.go
```diff
@@ -598,6 +645,7 @@ func (snap *Snapshot) ProcessSnapshotChunk(mainDb *badger.DB, mainDbMutex *deadl
 	})
 }
 
+// TODO: should we still do checksum stuff before we hit the first snapshot height?
```
diamondhands0, line 648: Hmm... I think we do unfortunately. Since we don't want to "stop the world" and compute the checksum when we compute the snapshot height. Hopefully it's OK though?

lib/snapshot.go
```diff
@@ -717,20 +766,26 @@ func (snap *Snapshot) PrepareAncestralRecord(key string, value []byte, existed b
...
 	// flush, the order of operations is preserved; however, there could still be some
 	// snapshot worker threads running so we want to wait until they're done.
 	err := snap.Checksum.Wait()
 	if err != nil {
-		glog.Errorf("Snapshot.StartAncestralRecordsFlush: Error while waiting "+
+		glog.Errorf("Snapshot.FlushAncestralRecords: Error while waiting "+
 			"for checksum: (%v)", err)
-		snap.StartAncestralRecordsFlush(false)
+		// TODO: do we need this still?
+		snap.FlushAncestralRecordsWithTxn(txn)
```
diamondhands0, line 788: I'm not sure what this is doing. I'll take a closer look later.

lib/snapshot.go
```diff
@@ -756,96 +805,103 @@ func (snap *Snapshot) FlushAncestralRecords() {
...
 	snap.SnapshotDbMutex.Unlock()
 	if err != nil {
 		// If any error occurred, then we should redo this memory write. During the restart, we will re-write all
 		// entries. If the error happened during a partial write, e.g. we didn't write all records in recordsKeyList,
 		// we'll redo them in the next write of this ancestralCache. The only scenario where that wouldn't happen
 		// is if the node stopped suddenly. We can detect that via comparing semaphore counters on boot.
-		glog.Errorf("Snapshot.StartAncestralRecordsFlush: Problem flushing snapshot, error %v", err)
-		snap.StartAncestralRecordsFlush(false)
+		glog.Errorf("Snapshot.FlushAncestralRecords: Problem flushing snapshot, error %v", err)
+		// TODO: @diamondhands - do we need this still? Before this would queue up another operation
```
diamondhands0, line 820: I'll take a closer look.

lib/db_utils.go
lazynina: It's the new standard way of checking errors for a specific error type. I've been gradually fixing them all as I go.

lib/snapshot.go
lazynina: It's a recursive call so it's kinda weird.

lib/server.go
lazynina: What would be a better value for PoW? we're doing bundles of 250. Maybe 50 instead? I want to see something in the logs at least 😂

lib/db_utils.go
lazynina: I'm just not 100% sure one way or the other. I guess there's maybe a race condition that the value would be flushed while we're doing the get.

lib/snapshot.go
diamondhands0: @lazynina I just reviewed this and the instance below it. I think we should just make FlushAncestralRecords return an error in this case rather than doing a recursive call. It seems like the only reason this was done is because they didn't want to change the call signature to return an error maybe.

lib/snapshot.go
diamondhands0: See my comment here:
* https://github.com/deso-protocol/core/pull/1297/files#r1596274944

lazynina:
### Merge activity

* **May 10, 7:41 PM EDT**: @lazynina started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/deso-protocol/core/1297).

PR #1305, opened 2024-05-11
Fix checksum computation at the snapshot height

diamondhands0:
Making snapshot computation synchronous broke the checksum computation at the snapshot height because computing the snapshot was still asynchronous. This change waits for the checksum to be fully updated before it updates the snapshot to the new checksum value.

I also made the blocks requested+sent higher for testnet, which makes syncing much faster.

### DIFF
--- a/lib/server.go
+++ b/lib/server.go
@@ -2443,7 +2443,7 @@ func (srv *Server) _handleBlockBundle(pp *Peer, bundle *MsgDeSoBlockBundle) {
 		// gracefully fail.
 		srv._handleBlock(pp, blk, ii == len(bundle.Blocks)-1 /*isLastBlock*/)
 		numLogBlocks := 100
-		if srv.params.IsPoWBlockHeight(blk.Header.Height) ||
+		if srv.params.IsPoSBlockHeight(blk.Header.Height) ||
 			srv.params.NetworkType == NetworkType_TESTNET {
 			numLogBlocks = 1000
 		}
--- a/lib/server.go
+++ b/lib/server.go
@@ -824,7 +824,9 @@ func (srv *Server) GetBlocksToStore(pp *Peer) {
 		if blockNode.Status&StatusBlockStored == 0 {
 			maxBlocksInFlight := MaxBlocksInFlight
 			if pp.Params.ProtocolVersion >= ProtocolVersion2 &&
-				srv.params.IsPoSBlockHeight(uint64(blockNode.Height)) {
+				(srv.params.IsPoSBlockHeight(uint64(blockNode.Height)) ||
+					srv.params.NetworkType == NetworkType_TESTNET) {
+
 				maxBlocksInFlight = MaxBlocksInFlightPoS
 			}
 			numBlocksToFetch := maxBlocksInFlight - len(pp.requestedBlocks)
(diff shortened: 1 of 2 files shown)


lib/server.go
```diff
@@ -824,7 +824,9 @@ func (srv *Server) GetBlocksToStore(pp *Peer) {
 		if blockNode.Status&StatusBlockStored == 0 {
 			maxBlocksInFlight := MaxBlocksInFlight
 			if pp.Params.ProtocolVersion >= ProtocolVersion2 &&
-				srv.params.IsPoSBlockHeight(uint64(blockNode.Height)) {
+				(srv.params.IsPoSBlockHeight(uint64(blockNode.Height)) ||
+					srv.params.NetworkType == NetworkType_TESTNET) {
```
diamondhands0, line 828: Just for testnet, we use the bigger number of blocks. Works great.

lib/snapshot.go
```diff
@@ -547,7 +547,7 @@ func (snap *Snapshot) Run() {
 			if err != nil {
 				glog.Errorf("Snapshot.Run: Problem getting checksum bytes (%v)", err)
 			}
-			glog.V(2).Infof("Snapshot.Run: PrintText (%s) Current checksum (%v)", operation.printText, stateChecksum)
+			glog.V(0).Infof("Snapshot.Run: PrintText (%s) Current checksum (%v)", operation.printText, stateChecksum)
```
diamondhands0, line 550: Having this at a higher verbosity is nice so I left it like that.

lib/snapshot.go
```diff
@@ -664,10 +664,6 @@ func (snap *Snapshot) RemoveChecksumBytes(key []byte, value []byte) {
 // WaitForAllOperationsToFinish will busy-wait for the snapshot channel to process all
 // current operations. Spinlocks are undesired but it's the easiest solution in this case,
 func (snap *Snapshot) WaitForAllOperationsToFinish() {
-	// If we haven't hit the first snapshot block height yet, don't bother.
-	if !snap.shouldPerformSnapshotOperations() {
```
diamondhands0, line 668: This was a tricky one. We only call this function in two places right now, but the right thing to do is actually to always wait for the operation queue to empty, otherwise you will may have checksum operations that you haven't processed yet. This actually messed up the compute_db_checksum.go script, which was super annoying because it relied on this call. Super annoying to find but eventually I realized the values were different every time depending on what operations it had processed lol...

lib/snapshot.go
```diff
@@ -1036,12 +1032,27 @@ func (snap *Snapshot) snapshotProcessBlockNoLock(blockNode *BlockNode) {
...
+			if localErr := snap.Checksum.Wait(); localErr != nil {
+				glog.Errorf("Snapshot.SetSnapshotChunk: Problem waiting for the checksum: %v", localErr)
+				time.Sleep(1 * time.Second)
+				continue
+			}
+			// For good measure, let's also wait for the operation queue to empty. This will ensure
+			// that any pending checksum operations are processed before we proceed.
+			glog.V(1).Infof("Snapshot.SnapshotProcessBlock: Waiting for all operations to " +
+				"clear so we can update snapshot checksum...")
+			snap.WaitForAllOperationsToFinish()
```
diamondhands0, line 1047: This is the meat of the fix. Without this change, the print statement below prints a *different value* every time you hit the snapshot (I tested it). However, with the fix the value is always consistent. I believe this will fix our checksum issue, but I'm syncing a couple nodes right now to do a "final" test.

lib/server.go
```diff
@@ -2439,10 +2442,12 @@ func (srv *Server) _handleBlockBundle(pp *Peer, bundle *MsgDeSoBlockBundle) {
 		// though as we'll just connect all the blocks after the failed one and those blocks will also
 		// gracefully fail.
 		srv._handleBlock(pp, blk, ii == len(bundle.Blocks)-1 /*isLastBlock*/)
-		numLogBlocks := 1000
-		if srv.params.IsPoWBlockHeight(blk.Header.Height) {
-			numLogBlocks = 100
+		numLogBlocks := 100
+		if srv.params.IsPoWBlockHeight(blk.Header.Height) ||
```
lazynina, line 2446: I think you want to check if it's PoS
```suggestion
		if srv.params.IsPoSBlockHeight(blk.Header.Height) ||
```

lazynina (approved):

PR #1344, opened 2024-05-24
Z/pos bugfixes

superzordon:
Small bugfix for state change syncer encoder


lib/state_change_syncer.go
```diff
@@ -197,6 +197,11 @@ func (stateChangeEntry *StateChangeEntry) RawDecodeWithoutMetadata(blockHeight u
 	}
 	stateChangeEntry.BlockHeight = entryBlockHeight
 
+	// Don't decode the block if the encoder type is not a utxo operation.
+	if stateChangeEntry.EncoderType != EncoderTypeUtxoOperation && stateChangeEntry.EncoderType != EncoderTypeUtxoOperationBundle {
```
lazynina, line 201: Should we update the encode function as well so we don't encode a nil entry?

lazynina (approved):

PR #1391, opened 2024-07-22
accelerated regtest mode

lazynina:


lib/constants.go
```diff
@@ -911,9 +917,24 @@ func (params *DeSoParams) EnableRegtest() {
...
 	params.DefaultJailInactiveValidatorGracePeriodEpochs = 3
 
+	// Check for consensus transition every second.
+	params.FastHotStuffConsensusTransitionCheckDuration = 1 * time.Second
+
 	// In regtest, we start all the fork heights at zero. These can be adjusted
 	// for testing purposes to ensure that a transition does not cause issues.
 	params.ForkHeights = RegtestForkHeights
+	if isAcceleratedRegtest {
+		params.ForkHeights.ProofOfStake2ConsensusCutoverBlockHeight = 30
```
diamondhands0, line 927: So I actually had a manually-hacked version of this value, and whenever I set it to <40 I had issues so mine was always set to 40. If you've found that it works with 30 then leave it. But I think the epoch duration caused some issues under 40, and I suspect it was due to not having enough time to build the snapshot validator set or something (like you need two epochs or something).

lib/constants.go
```diff
@@ -911,9 +917,24 @@ func (params *DeSoParams) EnableRegtest() {
...
+		params.DefaultEpochDurationNumBlocks = uint64(5)
+		params.DefaultBlockProductionIntervalMillisecondsPoS = 1000 // 1s
+		params.DefaultTimeoutIntervalMillisecondsPoS = 2000         // 2s
+		newSeedBalance := &DeSoOutput{
```
diamondhands0, line 931: This is so smart. My hacked version of core never had this and it was so annoying.

lib/constants.go
```diff
@@ -911,9 +917,24 @@ func (params *DeSoParams) EnableRegtest() {
 	// Set the PoS default jail inactive validator grace period epochs to 3.
 	params.DefaultJailInactiveValidatorGracePeriodEpochs = 3
 
+	// Check for consensus transition every second.
+	params.FastHotStuffConsensusTransitionCheckDuration = 1 * time.Second
```
diamondhands0, line 921: I feel like I know exactly what this is doing because I've run the PoW->PoS cutover so many times but... where is it actually being used? I don't see where it's getting used here.

diamondhands0: Ok I see it's in `resetFastHotStuffConsensusTransitionCheckTime`. Looks good. Man, I *really* wish I had this when building the AMM haha. This is literally going to make all of my tests so much better.

diamondhands0 (approved):
This is my favorite of all the PRs. So needed!

lib/constants.go
lazynina: 30 works b/c our epoch lengths are 5 blocks! you may have been testing w/ 10 blocks, which I think I sent you when I was hacking this offline.

PR #1392, opened 2024-07-24
Ln/max tip age bug fix

lazynina:


consensus/event_loop.go
```diff
@@ -680,6 +681,16 @@ func (fe *fastHotStuffEventLoop) tryConstructVoteQCInCurrentView() *FastHotStuff
 	// Fetch the validator list at the tip.
 	validatorList := fe.tip.validatorList
 
+	for _, validator := range validatorList {
```
diamondhands0, line 684: I wrote a function called DomainsToString(). Let's use that here and elsewhere instead of iterating over GetDomains().

lib/constants.go
```diff
@@ -1110,7 +1110,7 @@ var DeSoMainnetParams = DeSoParams{
 	MinChainWorkHex: "000000000000000000000000000000000000000000000000006314f9a85a949b",
 
 	MaxTipAgePoW: 24 * time.Hour,
-	MaxTipAgePoS: time.Hour,
+	MaxTipAgePoS: 24 * time.Hour,
```
diamondhands0, line 1113: Ok so I have a question: If we're syncing from scratch, do we use MaxTipAgePoS to mark when we start checking block signatures? Because that would make sync take a lot longer unnecessarily. If this is NOT the case, then I like 24h as the time. If it IS the case, then maybe just do 3h or something more conservative until we can break that signature checking thing into its own flag...

lib/network_manager.go
```diff
@@ -177,6 +177,19 @@ func (nm *NetworkManager) Stop() {
 // ## NetworkManager Routines
 // ###########################
 
+func (nm *NetworkManager) printActiveValidators() {
```
diamondhands0, line 180: All of this logging was not actually that useful because it just prints the full validator set over and over again (doesn't distinguish between who you're connected to vs just "aware" of). I'd delete everything around printActiveValidators().

lib/network_manager.go
```diff
@@ -370,6 +395,16 @@ func (nm *NetworkManager) processInboundConnection(conn Connection) (*RemoteNode
 		return nil, fmt.Errorf("NetworkManager.handleInboundConnection: Connection is not an inboundConnection")
 	}
 
+	na, err := nm.ConvertIPStringToNetAddress(ic.connection.RemoteAddr().String())
+	if err != nil {
+		return nil, errors.Wrapf(err, "NetworkManager.handleInboundConnection: Problem calling "+
+			"ConvertIPStringToNetAddress for addr: (%s)", ic.connection.RemoteAddr().String())
+	}
+	//if !addrmgr.IsRoutable(na) {
```
diamondhands0, line 403: This change around excluding IsRoutable addresses was reverted so you can delete it. With the update to use external addresses it could actually be a good check, but it's not worth putting it through testing to figure that out so just delete.

You can also move the call to ConvertIPStringToNetAddress back to where it was.

lib/network_manager.go
```diff
@@ -976,7 +1010,8 @@ func (nm *NetworkManager) Disconnect(rn *RemoteNode, disconnectReason string) {
 	if rn == nil {
 		return
 	}
-	glog.V(2).Infof("NetworkManager.Disconnect: Disconnecting from remote node id=%v", rn.GetId())
+	glog.V(2).Infof("NetworkManager.Disconnect: Disconnecting from remote "+
```
diamondhands0, line 1013: These little extra loggings were kinda useful so I'd keep them.

lib/server.go
```diff
@@ -190,6 +190,34 @@ func (srv *Server) GetNetworkManager() *NetworkManager {
 	return srv.networkManager
 }
 
+func (srv *Server) AdminOverrideViewNumber(view uint64) error {
```
diamondhands0, line 193: This clearly worked right? If so then I'd keep it and just gate it behind a JWT check, which I think you probably did in backend (reviewing that next).

lib/server.go
```diff
@@ -1981,17 +2009,13 @@ func (srv *Server) _addNewTxn(pp *Peer, txn *MsgDeSoTxn, rateLimit bool) ([]*Msg
...
-		// We will error in two cases:
-		// - the chainState is not need blocks state
-		// - the chainState is need blocks state but the chain is not on PoS.
-		if chainState != SyncStateNeedBlocksss || !srv.blockchain.params.IsPoSBlockHeight(tipHeight) {
-			err := fmt.Errorf("Server._addNewTxnAndRelay: Cannot process txn "+
-				"from peer %v while syncing: %v %v", pp, srv.blockchain.chainState(), txn.Hash())
-			glog.Error(err)
-			return nil, err
-		}
+	if chainState != SyncStateFullyCurrent && !srv.blockchain.params.IsPoSBlockHeight(tipHeight) {
```
diamondhands0, line 2012: Did we change this to fix the SYNCING_HEADERS error on Diamond? I think it's fine to keep regardless just wondering.

lib/server.go
```diff
@@ -2856,6 +2880,9 @@ func (srv *Server) _handleFastHotStuffConsensusEvent(event *consensus.FastHotStu
 }
 
 func (srv *Server) _handleValidatorVote(pp *Peer, msg *MsgDeSoValidatorVote) {
+	if msg.GetMsgType() != MsgTypeValidatorVote {
```
diamondhands0, line 2883: I think these are good to have for defensiveness so I'd leave them.

cmd/config.go
```diff
@@ -86,6 +87,17 @@ type Config struct {
 	CheckpointSyncingProviders []string
 }
 
+// Viper doesn't work when you have environment variables. This is the
+// suggested workaround:
+// https://github.com/spf13/viper/issues/380
+func GetStringSliceWorkaround(flagName string) []string {
```
diamondhands0, line 93: I applied this thing pretty quickly, but I think we need to keep it generally so I'd leave it. If you don't have it then it literally can't process comma-separated values when you pass them as env vars.

diamondhands0 (approved):

consensus/event_loop.go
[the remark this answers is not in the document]
lazynina: here we don't have the `ValidatorEntry`, we only have the `Validator` interface, so I'll use `DomainsToString()`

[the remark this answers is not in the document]
lazynina: also renaming to `validatorItem` so it doesn't overlap w/ the type of name `validator` that we use in the consensus package.

lib/constants.go
lazynina: I'll lower to 3 hours - this does make us start checking signatures sooner.

lib/server.go
lazynina: Yes - this was the fix to allow us to allow transaction relay regardless of sync status.

lib/network_manager.go
```diff
@@ -190,8 +190,10 @@ func (nm *NetworkManager) startValidatorConnector() {
 			nm.exitGroup.Done()
 			return
 		case <-time.After(nm.peerConnectionRefreshInterval):
+			glog.V(2).Infof("NetworkManager.startValidatorConnector: Refreshing validator indices...")
```
diamondhands0, line 193: Sorry I should have been more clear. Let's also delete these glogs because they're too verbose. They just print every second.

lib/network_manager.go
```diff
@@ -404,6 +410,11 @@ func (nm *NetworkManager) processOutboundConnection(conn Connection) (*RemoteNod
 		return nil, fmt.Errorf("NetworkManager.handleOutboundConnection: Connection is not an outboundConnection")
 	}
 
+	if !addrmgr.IsRoutable(oc.address) {
```
diamondhands0, line 413: I thought we were going to delete this. Wasn't it commented out in the last revision? If we have it it could mess up nodes running within datacenters.

lib/network_manager.go
```diff
@@ -209,9 +209,11 @@ func (nm *NetworkManager) startNonValidatorConnector() {
 			nm.exitGroup.Done()
 			return
 		case <-time.After(nm.peerConnectionRefreshInterval):
+			glog.V(2).Infof("NetworkManager.startNonValidatorConnector: Refreshing NON validator indices...")
```
diamondhands0, line 212: Sorry delete this and the one below it as well. Don't want to have any logging in this select.

PR #1404, opened 2024-08-31
Supporting CreateNewCoin for Focus

diamondhands0:
- Small modifications to support getting all holders of a coin, including locked balance entries
- Adding keys to support associations related to coin restrictions

### DIFF
--- a/lib/block_view_balance_entry.go
+++ b/lib/block_view_balance_entry.go
@@ -134,7 +134,9 @@ func (bav *UtxoView) GetHoldings(pkid *PKID, fetchProfiles bool, isDAOCoin bool)
 	return entriesYouHold, profilesYouHold, nil
 }
 
-func (bav *UtxoView) GetHolders(pkid *PKID, fetchProfiles bool, isDAOCoin bool) (
+// We introduce fetchProfiles and fetchLockedBalances for efficiency. If set to false, we
+// save a lot of processing time by skipping these operations.
+func (bav *UtxoView) GetHolders(pkid *PKID, fetchProfiles bool, fetchLockedBalances bool, isDAOCoin bool) (
 	_unlockedBalanceEntrys []*BalanceEntry, _unlockedProfiles []*ProfileEntry,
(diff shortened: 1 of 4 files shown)


lib/block_view_balance_entry.go
```diff
@@ -165,17 +167,38 @@ func (bav *UtxoView) GetHolders(pkid *PKID, fetchProfiles bool, isDAOCoin bool)
 		}
 	}
 
+	// Fetch all the locked entries if we're dealing with a dao coin
```
lazynina, line 170: Please break this out into a separate function. Fetching all the holders for unlocked is already pretty slow since we have to get all the profiles as well. we can call both functions in backend if you need it, but it's a lot of unnecessary overhead for existing usages. This is especially confusing when this function is used to fetch creator coin holdings.

lib/constants.go
```diff
@@ -1822,7 +1822,11 @@ const (
 	// key represents who to pay the fee to and the value indicates how many basis points to charge. All fees are
 	// charged as "taker fees" and are paid to the pkid in the map in the *quote currency* of the market (typically
 	// DESO, USDC, or FOCUS).
-	TokenTradingFeesByPkidMapKey = "TokenTradingFeesByPkidMap"
+	TokenTradingFeesByPkidMapKey            = "TokenTradingFeesByPkidMap"
+	DisableTradingFeeUpdateKey              = "DisableTradingFeeUpdateKey"
+	CreatorRevsharePercentageBasisPointsKey = "CreatorRevsharePercentageBasisPoints"
+	DisableCreatorRevshareUpdateKey         = "DisableCreatorRevshareUpdateKey"
+	CoinCategoryExtraDataKey                = "CoinCategoryExtraDataKey"
```
lazynina, line 1829: Any way we can shorten these strings? It's just a lot of extra space in each transaction you'll be using.

lazynina (changes requested):

lib/constants.go
lazynina: It's fine here, but in the future as you define new ones, please start with shortened strings. We should force folks to pay for more than they need to when transacting.

lazynina (approved):

PR #1406, opened 2024-09-16
if snapshot initialization error is a checksum issue, ignore if force…

lazynina:
… checksum is set to false

### DIFF
--- a/lib/server.go
+++ b/lib/server.go
@@ -454,10 +454,19 @@ func NewServer(
 	// Setup snapshot
 	var _snapshot *Snapshot
 	shouldRestart := false
+	isChecksumIssue := false
 	archivalMode := false
 	if _hyperSync {
-		_snapshot, err, shouldRestart = NewSnapshot(_db, _snapshotBlockHeightPeriod, false, false, _params,
-			_disableEncoderMigrations, _hypersyncMaxQueueSize, eventManager)
+		_snapshot, err, shouldRestart, isChecksumIssue = NewSnapshot(
+			_db,
+			_snapshotBlockHeightPeriod,
+			false,
+			false,
+			_params,
+			_disableEncoderMigrations,
+			_hypersyncMaxQueueSize,
+			eventManager,
+		)
 		if err != nil {
 			panic(err)
 		}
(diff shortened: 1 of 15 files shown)


lib/server.go
```diff
@@ -709,9 +718,14 @@ func NewServer(
 		if stateChangeSyncer != nil {
 			stateChangeSyncer.Reset()
 		}
-		glog.Errorf(CLog(Red, "NewServer: Forcing a rollback to the last snapshot epoch because node was not closed "+
-			"properly last time"))
-		return nil, errors.Wrapf(err, "NewServer: Restart required"), true
+		if !_forceChecksum && isChecksumIssue {
```
diamondhands0, line 721: @lazynina I think you have this backwards. We want to return an error in the first part of the if, not the second. Would swap the two.

lib/snapshot.go
```diff
@@ -443,18 +444,29 @@ func NewSnapshot(
 	//   Either way, it means our snapshot was compromised and we need to recompute it as described
 	//   in the previous bullet.
 	shouldRestart := false
+	isChecksumIssue := false
+	// @diamondhands - I think the only operations tracked by the state semaphore are (1) checksum add, (2) checksum
+	// remove, (3) ProcessChunk, (4) ChecksumPrint, (5) Exit. So the only one operation that isn't checksum related
+	// is the ProcessChunk operation. However, I think we can get away with this being considered a "checksum" issue
+	// and not restarting the node (which causes a panic). If you have a node that didn't finish processing chunks,
+	// just start over again.
```
diamondhands0, line 452: I would add just for extra clarity:

However, I think we can get away with this being considered a "checksum" issue **because ProcessChunk only happens during initial sync**

lib/snapshot.go
```diff
@@ -443,18 +444,29 @@ func NewSnapshot(
...
 			metadata.SnapshotBlockHeight, metadata.CurrentEpochBlockHash)))
 		shouldRestart = true
+		isChecksumIssue = true
 	}
 
 	if !shouldRestart {
 		if err := migrations.StartMigrations(); err != nil {
+			// @diamondhands - If we hit this error, it means that the migration failed. Should we
+			// consider this a "checksum" issue or not? I believe the migrations are only necessary
+			// for checksums, but I'm unsure.
```
diamondhands0, line 466: Yep, migrations are only needed to guarantee that checksums match. They don't change anything about how the data the node is processing is handled. I'd add that to the comment and run with it.

diamondhands0 (approved):
I think I found a little bug, otherwise looks good to me.

PR #1410, opened 2024-09-20
check for no-op in update global params in pos mempool, update datadog dependencies

lazynina:


lib/pos_mempool.go
```diff
@@ -944,6 +945,15 @@ func (mp *PosMempool) UpdateLatestBlock(blockView *UtxoView, blockHeight uint64)
 // new minimum will be removed from the mempool. To safely handle this, this method re-creates the TransactionRegister
 // with the new global params and re-adds all transactions in the mempool to the new register.
 func (mp *PosMempool) UpdateGlobalParams(globalParams *GlobalParamsEntry) {
+	// If the global params haven't changed at all, then we don't need to do anything.
+	newGlobalParamBytes := globalParams.RawEncodeWithoutMetadata(mp.latestBlockHeight, true)
+	mp.RLock()
```
diamondhands0, line 950: I don't think you can do this because UpdateGlobalParams() is called after acquiring mp.Lock() in another function. Instead, maybe you can just get it directly, but you would need to check to make sure that we're **always** calling this with mp.Lock() held first.

diamondhands0: Ah wait sorry I just checked again and saw the outer function was acquiring a different lock. One was the mempool lock, while the other was the fee estimator lock. I also see that we acquire mp.Lock() right after so it's clearly fine. Disregard my above comment.

diamondhands0 (approved):

PR #1413, opened 2024-10-03
Ln/replace best chain with block index

lazynina:


lib/server.go
```diff
@@ -1013,14 +1109,20 @@ func (srv *Server) shouldVerifySignatures(header *MsgDeSoHeader, isHeaderChain b
 	if checkpointBlockInfo == nil {
 		return true, false
 	}
-	var hasSeenCheckpointBlockHash bool
-	var checkpointBlockNode *BlockNode
+	// TODO: @diamondhands - why can't we move this up in this function? It seems like we can avoid
```
lazynina, line 1106: Please check out this optimization

lib/blockchain.go
```diff
@@ -2115,24 +2190,26 @@ func (bc *Blockchain) processHeaderPoW(blockHeader *MsgDeSoHeader, headerHash *B
 			"before timestamp of previous block %v",
 			time.Unix(int64(blockHeader.GetTstampSecs()), 0),
 			time.Unix(int64(parentHeader.GetTstampSecs()), 0))
-		return false, false, HeaderErrorTimestampTooEarly
+		return nil, false, false, HeaderErrorTimestampTooEarly
 	}
 
+	// @diamondhands - this check is significantly slower because we need to go
```
lazynina, line 2196: @diamondhands0 - header processing for PoW blocks is significantly slower. Just a heads up. nice thing is that this is a fixed cost on each network. It took about 7m to sync all PoW headers (1.2 M headers) on testnet, so it's not *terrible*.

lib/server.go
```diff
@@ -767,6 +791,53 @@ func (srv *Server) _handleGetHeaders(pp *Peer, msg *MsgDeSoGetHeaders) {
...
+	}
+	nextNodeHeight := startNode.Header.Height + 1
+	nextNode, nextNodeExists, nextNodeError := srv.blockchain.GetBlockFromBestChainByHeight(nextNodeHeight, true)
+	if nextNodeError != nil {
+		return nil, fmt.Errorf("GetHeadersForLocatorAndStopHash: Error getting start node by height: %v", startNodeError)
+	}
+	if !nextNodeExists || nextNode == nil {
+		return nil, nil
+	}
+	for ii := uint32(0); ii < maxHeadersPerMsg; ii++ {
```
lazynina, line 823: @diamondhands0 - this is the part of the code that slows down header syncing as we individually fetch block nodes by height until we hit the stop hash or have enough headers.

Some alternatives:
1. Instead of using GetBlockFromBestChainByHeight, we could iterate over the height - hash prefix directly. Doing one single badger iteration is probably faster than a bunch of individual ones.
2. We could compare the stop node height to maxHeadersPerMsg + startNode.Height
    - if stop node height  is greater, then we can find the "true" stop node and iterate back backwards. This will be significantly faster since we'll have the block hash.
    - if stop node height is less, then we just iterate backwards from the stop node.

lib/server.go
```diff
@@ -862,7 +956,17 @@ func (srv *Server) GetBlocksToStore(pp *Peer) {
 	}
 
 	// Go through the block nodes in the blockchain and download the blocks if they're not stored.
-	for _, blockNode := range srv.blockchain.bestChain {
+	for ii := uint32(srv.blockchain.lowestBlockNotStored); ii <= srv.blockchain.blockTip().Height; ii++ {
```
lazynina, line 959: @diamondhands0 - we'll probably want to rewrite this so that we iterate in reverse. This allows us to more quickly look up blocks since we have both the block hash and block height.

lib/blockchain.go
```diff
@@ -1950,16 +2006,29 @@ func CheckTransactionSanity(txn *MsgDeSoTxn, blockHeight uint32, params *DeSoPar
 	return nil
 }
 
-func GetReorgBlocks(tip *BlockNode, newNode *BlockNode) (_commonAncestor *BlockNode, _detachNodes []*BlockNode, _attachNodes []*BlockNode) {
+func (bc *Blockchain) GetReorgBlocks(tip *BlockNode, newNode *BlockNode) (
```
diamondhands0, line 2009: I feel like this shouldn't ever be called for txindex, though, FWIW so not sure we need to optimize it.

lib/blockchain.go
diamondhands0: I think that's fine. We could do something to skip this check, but I'm worried it would expose us to attack so better to just leave it for now, and optimize it later if it just hits us when syncing a node from scratch.

diamondhands0: Another way to optimize it would be to keep a map from height to "node in the best chain at that height" in the blockindex. But no need to add that complexity now. It would need to be pruned and stuff as the LRU cache fills. Gross.

lib/blockchain.go
```diff
@@ -121,9 +122,19 @@ func (nn *BlockNode) IsValidateFailed() bool {
...
 
 // IsCommitted returns true if a BlockNode has passed all validations, and it has been committed to
-// the Blockchain according to the Fast HotStuff commit rule.
+// the Blockchain according to the Fast HotStuff commit rule. This function previously considered
+// as PoW blocks to be committed regardless of status, but since we no longer keep the entire best
+// chain in-memory, we needed to update all PoW blocks in the best chain to be Committed such that
+// we could distinguish them from forks that occurred in PoW and still determine if a node is in the
+// best chain or not.
 func (nn *BlockNode) IsCommitted() bool {
-	return nn.Status&StatusBlockCommitted != 0 || !blockNodeProofOfStakeCutoverMigrationTriggered(nn.Height)
```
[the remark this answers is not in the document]
diamondhands0: Ok I see. It's if a node has been running for a long time and had a reorg at some point. Then you have no way of knowing which PoW block was actually the best one. Got it.

lib/blockchain.go
```diff
@@ -1674,39 +1622,137 @@ func (bc *Blockchain) Snapshot() *Snapshot {
...
 }
 
 func (bc *Blockchain) BlockTip() *BlockNode {
 	return bc.blockTip()
 }
 
+// TODO: this won't work for now. Need to figure out how to handle this.
 func (bc *Blockchain) BestChain() []*BlockNode {
-	return bc.bestChain
+	panic("BestChain not supported.")
```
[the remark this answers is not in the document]
diamondhands0: Another thing you could do is add an argument like "minHeight" or something to this, and then the things that need it can get however many blocks in the past they need. This is what the hot feed needs and I see you kinda do it in the db_utils version of GetBestChain.

lib/server.go
diamondhands0: This seems fine to me. Probably just had it later because performance wasn't an issue before.

lib/block_producer.go
```diff
@@ -367,8 +367,8 @@ func (desoBlockProducer *DeSoBlockProducer) _getBlockTemplate(publicKey []byte)
 	blockRet.Header.TransactionMerkleRoot = merkleRoot
 
 	// Compute the next difficulty target given the current tip.
-	diffTarget, err := CalcNextDifficultyTarget(
-		lastNode, CurrentHeaderVersion, desoBlockProducer.params)
+	diffTarget, err := desoBlockProducer.chain.CalcNextDifficultyTarget(
```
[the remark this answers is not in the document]
diamondhands0: Yep this is fine.

lib/db_utils.go
```diff
@@ -5308,7 +5360,7 @@ func InitDbWithDeSoGenesisBlock(params *DeSoParams, handle *badger.DB,
 		diffTarget,
 		BytesToBigint(ExpectedWorkForBlockHash(diffTarget)[:]), // CumWork
 		genesisBlock.Header, // Header
-		StatusHeaderValidated|StatusBlockProcessed|StatusBlockStored|StatusBlockValidated, // Status
+		StatusHeaderValidated|StatusBlockProcessed|StatusBlockStored|StatusBlockValidated|StatusBlockCommitted, // Status
```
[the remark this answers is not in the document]
diamondhands0: Yep that makes sense.

lib/db_utils.go
```diff
@@ -5544,9 +5608,205 @@ func GetBlockIndex(handle *badger.DB, bitcoinNodes bool, params *DeSoParams) (
 	return blockIndex, nil
 }
 
+// LoadBlockIndexFromHeight loads the block index from the database starting at
+// a given height. This is only used by initChain, where we want to load all blocks
+// that could be descendents of the current tip block in the database.
+func (bi *BlockIndex) LoadBlockIndexFromHeight(height uint32, params *DeSoParams) error {
```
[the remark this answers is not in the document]
diamondhands0: It seemed good to me.

lib/db_utils.go
```diff
@@ -5544,9 +5608,205 @@ func GetBlockIndex(handle *badger.DB, bitcoinNodes bool, params *DeSoParams) (
...
+		return nil
+	})
+}
+
+// RunBlockIndexMigration runs a migration to populate the hash to height index from the height hash to
+// block node index. We can't use the encoder migrations to handle this situation since it's a new index
+// and not a modification of the existing entry type stored. This migration simply iterates over the keys in the
+// height hash to block node index, extracts the height and hash, and puts a key of the hash and a value of the
+// height in the hash to height prefix.
+func RunBlockIndexMigration(handle *badger.DB, snapshot *Snapshot, eventManager *EventManager, params *DeSoParams) error {
```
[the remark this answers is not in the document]
diamondhands0: This one also looked good.

lib/db_utils.go
```diff
@@ -5544,9 +5622,205 @@ func GetBlockIndex(handle *badger.DB, bitcoinNodes bool, params *DeSoParams) (
...
+			err = PutHeightHashToNodeInfoBatch(handle, snapshot, blockNodeBatch, false /*bitcoinNodes*/, eventManager)
+			if err != nil {
+				return errors.Wrap(err, "RunBlockIndexMigration: Problem putting block node batch")
+			}
+		}
+		return nil
+	})
+}
+
+// TODO: refactor to actually get the whole best chain if that's
```
diamondhands0, line 5818: You may want to consider adding a numBlocks param to this.

lib/server.go
diamondhands0: I like the second option. Is it difficult to execute? It also isn't a huge issue if header sync is slow and you want to just leave it for now.

lib/server.go
diamondhands0: Yeah, that makes sense. I'd only do it if it's very slow right now, though. If it's bearable (eg only adds 20m or something like that to an initial mainnet sync) then I think it's fine to leave it.

lib/types.go
```diff
@@ -238,7 +237,7 @@ func (bh *BlockHash) IsEqual(target *BlockHash) bool {
 		return false
 	}
 
-	return reflect.DeepEqual(bh[:], target[:])
+	return bytes.Equal(bh[:], target[:])
```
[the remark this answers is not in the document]
diamondhands0: Huh. Interesting find!

diamondhands0 (approved):

lib/blockchain.go
lazynina: yeah exactly. it's gross 😂 we could add a new prefix to the DB to do it, but it does not make sense to do it in memory b/c eventually we'll need to move it to badger.

lib/server.go
lazynina: oh I already implementated #2 - it's much faster.

lib/blockchain.go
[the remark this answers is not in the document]
lazynina: yep you figured it out. Sorry I thought I dropped a comment about this somewhere.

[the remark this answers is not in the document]
lazynina: oh yeah the comment above this function explains it:
```
// This function previously considered
// as PoW blocks to be committed regardless of status, but since we no longer keep the entire best
// chain in-memory, we needed to update all PoW blocks in the best chain to be Committed such that
// we could distinguish them from forks that occurred in PoW and still determine if a node is in the
// best chain or not.
```

lib/blockchain.go
```diff
@@ -978,7 +1162,7 @@ func NewBlockchain(
 
 	timer := &Timer{}
 	timer.Initialize()
-
+	blockViewCache, _ := collections.NewLruCache[BlockHash, *BlockViewAndUtxoOps](100) // TODO: parameterize
```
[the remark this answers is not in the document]
lazynina: 100 is the value we've been using before. This PR does not change the size of the block view cache. This is a cache used to cache the utxo view built at a certain block. It is used to improve performance of process block pos. Do not confuse it with the new `BlockIndex` struct.

lib/blockchain.go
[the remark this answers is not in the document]
lazynina: I actually don't see this function used in the hot feed routine. We need to replace the usages of this in the rosetta repo though.

lib/blockchain.go
lazynina: This is still needed for some old PoW tests. so I'm just going to leave it unoptimized since we won't be hitting PoW reorgs in practice.

PR #1420, opened 2024-10-14
free operation queue semaphore in process snapshot chunk and when hitting error in handleSnapshot that prevents us from calling process snapshot chunk

lazynina:


lib/server.go
```diff
@@ -1441,13 +1441,16 @@ func (srv *Server) _handleSnapshot(pp *Peer, msg *MsgDeSoSnapshotData) {
 		msg.SnapshotChunk[0].Key, msg.SnapshotChunk[len(msg.SnapshotChunk)-1].Key, len(msg.SnapshotChunk),
 		msg.SnapshotMetadata, msg.SnapshotChunk[0].IsEmpty(), pp)))
 	// Free up a slot in the operationQueueSemaphore, now that a chunk has been processed.
-	srv.snapshot.FreeOperationQueueSemaphore()
+	// srv.snapshot.FreeOperationQueueSemaphore()
```
diamondhands0, line 1444: I left some comments in this file as FIXMEs because I couldn't comment on the lines that were unchanged in the PR. See my edited PR here:

https://github.com/deso-protocol/core/pull/1423/files

diamondhands0: I also tried a weird alternative with defer that may be less error-prone. Check it out up to you if you want to change it.

diamondhands0 (approved):

PR #1423, opened 2024-10-24
Diamondhands/review free operation queue semaphore in process chunk

diamondhands0:


lib/blockchain.go
```diff
@@ -1255,6 +1255,8 @@ func (bc *Blockchain) LatestLocator(tip *BlockNode) []*BlockHash {
 }
 
 func (bc *Blockchain) HeaderLocatorWithNodeHash(blockHash *BlockHash) ([]*BlockHash, error) {
+	// We can acquire the ChainLock because the only place this is called currently is from
```
diamondhands0, line 1258: Add a comment why we need it here and below.

lib/pos_blockchain.go
```diff
@@ -310,7 +310,7 @@ func (bc *Blockchain) processBlockPoS(block *MsgDeSoBlock, currentView uint64, v
 
 	// We expect the utxoView for the parent block to be valid because we check that all ancestor blocks have
 	// been validated.
-	parentUtxoViewAndUtxoOps, err := bc.getUtxoViewAndUtxoOpsAtBlockHash(*block.Header.PrevBlockHash)
```
diamondhands0, line 313: Idk why this is here ignore.

PR #1442, opened 2025-01-30
State syncer mempool syncing and test suite updates

lazynina:


lib/db_utils.go
```diff
@@ -1204,26 +1215,39 @@ func DBDeleteWithTxn(txn *badger.Txn, snap *Snapshot, key []byte, eventManager *
...
 	// If snapshot was provided, we will need to load the current value of the record
 	// so that we can later write it in the ancestral record. We first lookup cache.
-	if isState {
-		// We check if we've already read this key and stored it in the cache.
-		// Otherwise, we fetch the current value of this record from the DB.
-		ancestralValue, getError = DBGetWithTxn(txn, snap, key)
-		// If the key doesn't exist then there is no point in deleting this entry.
-		if getError == badger.ErrKeyNotFound {
-			return nil
+	if isState || (isCoreState && eventManager != nil && eventManager.isMempoolManager) {
```
lazynina, line 1222:
```suggestion
	if isState || (eventManager != nil && eventManager.isMempoolManager && isCoreStateKey(key)) {
```
same thing as above.

lib/db_utils.go
```diff
@@ -1204,26 +1215,39 @@ func DBDeleteWithTxn(txn *badger.Txn, snap *Snapshot, key []byte, eventManager *
 	var getError error
 	isState := snap != nil && snap.isState(key)
 
+	isCoreState := isCoreStateKey(key)
+
```
lazynina, line 1219:
```suggestion
```

lib/db_utils.go
```diff
@@ -1204,26 +1215,39 @@ func DBDeleteWithTxn(txn *badger.Txn, snap *Snapshot, key []byte, eventManager *
...
 				"with key: %v", key)
 		}
 	}
 
 	err := txn.Delete(key)
-	if err != nil {
+	if err != nil && err == badger.ErrKeyNotFound && eventManager != nil && eventManager.isMempoolManager {
+		// If the key doesn't exist then there is no point in deleting this entry.
+		return nil
+	} else if err != nil {
```
lazynina, line 1250:
```suggestion
	}
	if err != nil {
```
doesn't need to be an else if here since we return above.

lib/db_utils.go
```diff
@@ -1204,26 +1215,39 @@ func DBDeleteWithTxn(txn *badger.Txn, snap *Snapshot, key []byte, eventManager *
...
 			return errors.Wrapf(getError, "DBDeleteWithTxn: problem checking for DB record "+
 				"with key: %v", key)
 		}
 	}
 
 	err := txn.Delete(key)
-	if err != nil {
+	if err != nil && err == badger.ErrKeyNotFound && eventManager != nil && eventManager.isMempoolManager {
+		// If the key doesn't exist then there is no point in deleting this entry.
+		return nil
```
lazynina, line 1249: Was the previous code causing issues for you? just curious. I don't see any issue w/ this change, but it does change the intent a little bit.

lib/event_manager.go
```diff
@@ -59,7 +62,10 @@ type EventManager struct {
 	blockCommittedHandlers       []BlockEventFunc
 	blockAcceptedHandlers        []BlockEventFunc
 	snapshotCompletedHandlers    []SnapshotCompletedEventFunc
-	isMempoolManager             bool
+	// A transaction used by the state syncer mempool routine to reference the state of the badger db
+	// prior to flushing mempool transactions. This represents the last committed view of the db.
+	lastCommittedViewTxn *badger.Txn
```
lazynina, line 67: ❤️ magnifique

lib/server.go
```diff
@@ -165,7 +165,7 @@ type Server struct {
 	// It can be used to find computational bottlenecks.
 	timer *Timer
 
-	stateChangeSyncer *StateChangeSyncer
+	StateChangeSyncer *StateChangeSyncer
```
lazynina, line 168: Why does this need to be public? just curious.

lib/state_change_syncer.go
```diff
@@ -174,8 +175,11 @@ func (stateChangeEntry *StateChangeEntry) RawDecodeWithoutMetadata(blockHeight u
 	ancestralRecord := stateChangeEntry.EncoderType.New()
 	if exist, err := DecodeFromBytes(ancestralRecord, rr); exist && err == nil {
 		stateChangeEntry.AncestralRecord = ancestralRecord
+		stateChangeEntry.AncestralRecordBytes = EncodeToBytes(blockHeight, ancestralRecord)
 	} else if err != nil {
 		return errors.Wrapf(err, "StateChangeEntry.RawDecodeWithoutMetadata: error decoding ancestral record")
+	} else {
+		stateChangeEntry.AncestralRecordBytes = EncodeToBytes(blockHeight, nil)
```
lazynina, line 182: so I think I get what's going on here? for the `else` clause, we just want to encode an empty struct which signifies that the entry didn't exist before. and in the `if exists && err == nil` case, we get the bytes while we can. Can we add a small comment around this section just to clarify?

lib/state_change_syncer.go
```diff
@@ -455,6 +460,16 @@ func (stateChangeSyncer *StateChangeSyncer) _handleStateSyncerOperation(event *S
 		encoderType = keyEncoder.GetEncoderType()
 		stateChangeEntry.Encoder = keyEncoder
 		stateChangeEntry.EncoderBytes = nil
+
+		if stateChangeEntry.AncestralRecordBytes != nil && len(stateChangeEntry.AncestralRecordBytes) > 0 {
+			// Decode the ancestral record.
+			ancestralRecord, err := DecodeStateKey(stateChangeEntry.KeyBytes, stateChangeEntry.AncestralRecordBytes)
+			if err != nil {
+				glog.Fatalf("Server._handleStateSyncerOperation: Error decoding ancestral record: %v", err)
```
lazynina, line 468: Fatal is going to kill the node if we hit this error. Are we sure that's what we want?

lib/state_change_syncer.go
```diff
@@ -455,6 +460,16 @@ func (stateChangeSyncer *StateChangeSyncer) _handleStateSyncerOperation(event *S
...
+			}
+			stateChangeEntry.AncestralRecord = ancestralRecord
+			stateChangeEntry.AncestralRecordBytes = nil
```
lazynina, line 471: why do we set `AncestralRecordBytes` to nil?

lib/state_change_syncer.go
```diff
@@ -899,77 +911,45 @@ func (stateChangeSyncer *StateChangeSyncer) SyncMempoolToStateSyncer(server *Ser
 	startTime = time.Now()
 	glog.V(2).Infof("Mempool synced len after flush: %d", len(stateChangeSyncer.MempoolSyncedKeyValueMap))
 
-	//Check to see if every txn hash in our cached txns is in the first n txns in the mempool.
```
lazynina, line 902: Love that we're getting rid of this cache in here.

lib/state_change_syncer.go
```diff
@@ -986,6 +966,47 @@ func (stateChangeSyncer *StateChangeSyncer) SyncMempoolToStateSyncer(server *Ser
...
+	mempoolTxEventManager.OnStateSyncerFlushed(stateChangeSyncer._handleStateSyncerFlush)
+
+	mempoolTxEventManager.isMempoolManager = true
+	mempoolTxUtxoView.EventManager = &mempoolTxEventManager
+
+	// Kill the snapshot so that it doesn't affect the original snapshot.
+	mempoolTxUtxoView.Snapshot = nil
+
+	server.blockchain.ChainLock.RLock()
+	mempoolTxUtxoView.TipHash = server.blockchain.bestChain[len(server.blockchain.bestChain)-1].Hash
```
lazynina, line 986: let's use the `BlockTip` function on blockchain. This will reduce changes I need to make when we get rid of the best chain in memory.
```suggestion
	mempoolTxUtxoView.TipHash = server.blockchain.BlockTip().Hash
```

lib/state_change_syncer.go
```diff
@@ -986,6 +966,47 @@ func (stateChangeSyncer *StateChangeSyncer) SyncMempoolToStateSyncer(server *Ser
...
+	// more than once in the mempool transactions.
+	txn2 := server.blockchain.db.NewTransaction(true)
+	defer txn2.Discard()
+
+	// Create a read-only view of the badger DB prior to the mempool flush. This view will be used to get the ancestral
+	// records of entries that are being modified in the mempool.
+	mempoolTxEventManager.lastCommittedViewTxn = server.blockchain.db.NewTransaction(false)
+	defer mempoolTxEventManager.lastCommittedViewTxn.Discard()
+
+	err = mempoolTxUtxoView.FlushToDbWithTxn(txn, uint64(server.blockchain.bestChain[len(server.blockchain.bestChain)-1].Height))
```
lazynina, line 1001:
```suggestion
	err = mempoolTxUtxoView.FlushToDbWithTxn(txn, uint64(server.blockchain.BlockTip().Height))
```

lazynina: addressed in #1447

lib/db_utils.go
lazynina: addressed in #1447

lib/db_utils.go
lazynina: addressed in #1447

lib/db_utils.go
lazynina: addressed in #1447

lib/db_utils.go
```diff
@@ -1108,12 +1108,23 @@ func DBSetWithTxn(txn *badger.Txn, snap *Snapshot, key []byte, value []byte, eve
 	var ancestralValue []byte
 	var getError error
 
+	isCoreState := isCoreStateKey(key)
```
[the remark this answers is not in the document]
lazynina: addressed in #1447

lib/db_utils.go
```diff
@@ -1108,12 +1108,23 @@ func DBSetWithTxn(txn *badger.Txn, snap *Snapshot, key []byte, value []byte, eve
...
+
 	// If snapshot was provided, we will need to load the current value of the record
 	// so that we can later write it in the ancestral record. We first lookup cache.
-	if isState {
-		// We check if we've already read this key and stored it in the cache.
-		// Otherwise, we fetch the current value of this record from the DB.
-		ancestralValue, getError = DBGetWithTxn(txn, snap, key)
+	if isState || (isCoreState && eventManager != nil && eventManager.isMempoolManager) {
```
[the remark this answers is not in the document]
lazynina: addressed in #1447 (comment/documentation still should be addressed).

lib/block_view_flush.go
```diff
@@ -948,9 +949,12 @@ func (bav *UtxoView) _flushDAOCoinBalanceEntriesToDbWithTxn(txn *badger.Txn, blo
 				balanceKey, computedBalanceKey)
 		}
 
+		// Either a zero entry or an isDeleted annotation results in a deletion.
+		isDeleted := balanceEntry.BalanceNanos.Eq(uint256.NewInt(0)) || balanceEntry.isDeleted
```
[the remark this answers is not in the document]
lazynina: addressed in #1447

cmd/config.go
```diff
@@ -79,6 +80,7 @@ type Config struct {
 	LogDBSummarySnapshots bool
 	DatadogProfiler       bool
 	TimeEvents            bool
+	NoLogToStdErr         bool
```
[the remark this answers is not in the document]
superzordon: Correct - I found when running the test suite it was nice to be able to turn off logs for mining blocks, etc. I've made it such that these logs are done by default, but it was nice to be able to disable them, made the tests much more readable.

lib/db_utils.go
superzordon: This is specifically logic just for the mempool manager, which is only active during the mempool state sync routine. I am struggling to remember the exact circumstances here, but in some instances when deleting in the state syncer mempool routine, we need these extra checks here.

lib/server.go
superzordon: I believe we access it in some of the testing routines we run in the state consumer testing suite.

lib/db_utils.go
lazynina: works for me!

lib/state_change_syncer.go
superzordon: This function is unfortunately not in a clean place for errors to get passed up the chain and handled. That said, at the top of the chain here we would probably be throwing a fatal one if this were to trip (it never has in all the time we've been running this), since this does mean that our entire state chain is somehow messed up. This also should have no impact on non state-sync nodes.

lib/state_change_syncer.go
superzordon: Exactly - will do

lib/state_change_syncer.go
lazynina: Makes sense. Sounds good to me!

lib/state_change_syncer.go
superzordon: The encoder/decoder here is built to favor the bytes when present instead of the ancestral record. In this instance, we'll be modifying the `AncestralRecord` property down the line, so we want the encoder to encode that `AncestralRecord` field to bytes, rather than just copying the `AncestralRecordBytes`. I've added a comment to explain this in more detail.

lazynina (commented):
Approved

lib/block_view_flush.go
```diff
@@ -948,9 +948,12 @@ func (bav *UtxoView) _flushDAOCoinBalanceEntriesToDbWithTxn(txn *badger.Txn, blo
 				balanceKey, computedBalanceKey)
 		}
 
+		// Either a zero entry or an isDeleted annotation results in a deletion.
+		isDeleted := balanceEntry.BalanceNanos.IsZero() || balanceEntry.isDeleted
```
diamondhands0, line 952: Ok, so I assume this is doing the following:

1. Before, if a balance went from non-zero to zero, we would let it linger in badger.
2. This change makes it so that we hard delete balances that hit zero.

Does that sound right? If yes, feel free to resolve this comment.

lib/db_utils.go
```diff
@@ -1110,10 +1110,19 @@ func DBSetWithTxn(txn *badger.Txn, snap *Snapshot, key []byte, value []byte, eve
...
-		// Otherwise, we fetch the current value of this record from the DB.
-		ancestralValue, getError = DBGetWithTxn(txn, snap, key)
+	if isState || (eventManager != nil && eventManager.isMempoolManager && isCoreStateKey(key)) {
+
+		// When we are syncing state from the mempool, we need to read the last committed view txn.
+		// This is because we will be querying the badger DB, and during the flush loop, every entry that is
+		// updated will first be deleted. In order to counteract this, we reference a badger transaction that was
+		// initiated before the flush loop started.
+		if eventManager != nil && eventManager.isMempoolManager && eventManager.lastCommittedViewTxn != nil {
+			ancestralValue, getError = DBGetWithTxn(eventManager.lastCommittedViewTxn, nil, key)
```
diamondhands0, line 1120: I think this is low-risk because it's only run when you have the mempool manager there. So looks fine.

lib/db_utils.go
```diff
@@ -1206,23 +1215,35 @@ func DBDeleteWithTxn(txn *badger.Txn, snap *Snapshot, key []byte, eventManager *
...
 		// If there is some error with the DB read, other than non-existent key, we return.
-		if getError != nil {
+		if getError != nil && getError != badger.ErrKeyNotFound {
 			return errors.Wrapf(getError, "DBDeleteWithTxn: problem checking for DB record "+
 				"with key: %v", key)
 		}
 	}
 
 	err := txn.Delete(key)
+	if err != nil && err == badger.ErrKeyNotFound && eventManager != nil && eventManager.isMempoolManager {
```
diamondhands0, line 1243: I don't love an if with a lot of conditions on it. I like something more like the below. In writing it like this I noticed there may be a case where you have ErrKeyNotFound but no event manager. If this is doing what you expect then it's fine to leave it.

```
if err != nil {
  if err == badger.ErrKeyNotFound {
    if eventManager != nil && eventManager.isMempoolManager {
      // In this case your logic returns nil
      return nil
    } else {
      // What happens here?
      // It looks like you'll return an error in this case. Is this what you want?
    }
  }  else {
    // In this case there was a problem deleting the record
    return errors.Wrapf(err, "DBDeleteWithTxn: Problem deleting record "+
			"from DB with key: %v", key)
  }
}
```

lib/state_change_syncer.go
```diff
@@ -174,8 +175,12 @@ func (stateChangeEntry *StateChangeEntry) RawDecodeWithoutMetadata(blockHeight u
 	ancestralRecord := stateChangeEntry.EncoderType.New()
 	if exist, err := DecodeFromBytes(ancestralRecord, rr); exist && err == nil {
 		stateChangeEntry.AncestralRecord = ancestralRecord
+		stateChangeEntry.AncestralRecordBytes = EncodeToBytes(blockHeight, ancestralRecord)
```
diamondhands0, line 178: I don't think this is problematic, but it could slow things down a lot. Encode/Decode is one of the biggest consumers of sync time I think. Consider skipping the encode/decode if you can. But it's also fine to leave it if it's hard to get rid of it.

diamondhands0 (approved):

lib/block_view_flush.go
superzordon: This is actually a state-syncer specific bug. In particular, this addresses the issues we were seeing where max sells wouldn't disappear from postgres/graphql. Basically - when a balance entry hits zero, we have core logic that removes the entry entirely from badger. However, when we do this, we don't actually set `isDeleted` in core for some reason. In order for state syncer to know that this entry is going to be deleted, we need the above logic so that it can accurately mirror what's going on in badger.

lib/state_change_syncer.go
superzordon: This is a good call-out. There are instances in the state consumer where we do use the raw bytes on the decoded state change entry ancestral record, and we haven't noticed any noticeable slowdowns as a result of the encoders specifically, but in the future if we do find that these become a bottleneck, we could consider modifying our consumer logic to only decode when we need to, e.g. a subsequent encode call on `stateChangeEntry.AncestralRecordBytes`.

PR #1463, opened 2025-06-24
enforce limit on get nfts for user

lazynina:


superzordon (changes requested):
Overall looks good, one code cleanliness question

